PR #1, opened 2015-02-04
moved maas playbooks out of os-ansible-deployment repo

miguelgrinberg:
Work in progress -- DO NOT MERGE

PR #2, opened 2015-02-17
Move MaaS to rpc-extras

andymcc:
Set MaaS up as a galaxy-role
Move all the bits to the rpc-extras repo
Adjust the rackspace-monitoring-cli pip package to be downloaded from
pypi

### DIFF
--- a/etc/openstack_deploy/user_maas_settings.yml
+++ b/etc/openstack_deploy/user_maas_settings.yml
@@ -1,0 +1,94 @@
+---
+# Copyright 2014, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+## MaaS Options
+# Set maas_auth_method to 'token' to use maas_auth_token/maas_api_url
+# instead of maas_username/maas_api_key
+maas_auth_method: password
+maas_auth_url: "{{ rackspace_cloud_auth_url }}"
+maas_username: "{{ rackspace_cloud_username }}"
+maas_api_key: "{{ rackspace_cloud_api_key }}"
+maas_auth_token: some_token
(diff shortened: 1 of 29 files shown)


miguelgrinberg:
@andymcc so I had a mostly done move of the maas playbooks (see PR #1), but did not do much testing. If you have verified this playbook I'm happy to discard mine and merge this. Let me know!

andymcc:
Ugh sorry @miguelgrinberg massive fail by me. I'll take a look at your patch and we can figure it out from there - should really have checked before hand.

miguelgrinberg:
@andymcc No problem, you probably went further into it than I did, I did not do a conversion to galaxy style roles (I have only done that for the horizon plugin playbook so far).

I don't have a good way to test maas, but if you have tested this and it works we can go ahead and merge it whenever the galaxy changes make it into os-ansible-deployment.

andymcc:
@miguelgrinberg I'll do a bit more testing - I ran through an initial test and I "believe" (tm) that it worked. One side concern, and perhaps you would have an idea on this - I'm not sure how to make this "easily" usable. For example, this will still require the dynamic_inventory, and libraries (keystone specifically) from the os-ansible-deployment repo, so in testing I had to literally clone the rpc-extras repo, copy over the user_maas_variables from the etc dir and then copy the contents of playbooks directory to the playbooks directory in the os-ansible-deployment dir (i.e. a really inefficient and manual way to get this going). I'm not sure the best way to get that to work more sensibly, but I figured that could be step 2 after we know the code works.

klevenstein:
This has a doc impact, right? In that we'll have to change the URL for downloading the maas playbooks? Will this be implemented for v10.1.2?

miguelgrinberg:
@andymcc for the dynamic inventory I have a trick, the `dynamic_inventory.py` script in this repo just locates the inventory script for os-ansible-deployment and defers everything to it. That is how the horizon plugin playbook can use the horizon vars. The `vars/common.yml` has the path to the os-ansible-deployment source tree.

For the libraries the same trick can be used, you can write a keystone library in this repo that just invokes the original.

BTW, the structure you see in master is not galaxy friendly, but I have made the conversion, I have that in my fork: https://github.com/miguelgrinberg/rpc-extras/tree/galaxy. I plan to drop this on master once we go the galaxy way on the stackforge repo.

miguelgrinberg:
@klevenstein Not for 10.1.2 AFAIK, but yes, you'll have some (mostly minor) changes for maas and horizon-extensions. I think we are doing this at the same time we are applying the galaxy changes on os-ansible-deployment, at least that is the last I heard.

klevenstein:
Okay, thanks. Just wanted to make sure.

andymcc:
Closing this PR out - since the repo changed to public and I made some changes to ensure it all works! #3 is the new source of truth.

PR #3, opened 2015-02-20
Move MaaS to rpc-extras

andymcc:
Set MaaS up as a galaxy-role
Move all the bits to the rpc-extras repo
Adjust the rackspace-monitoring-cli pip package to be downloaded from
pypi
Utilise the existing os-ansible-deployment repo's keystone library
Utilise the existing os-ansible-deployment repo's dynamic_inventory

### DIFF
--- a/etc/openstack_deploy/user_maas_settings.yml
+++ b/etc/openstack_deploy/user_maas_settings.yml
@@ -1,0 +1,69 @@
+---
+# Copyright 2014, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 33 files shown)


miguelgrinberg:
@andymcc Hey, so this structure is close, but not identical to what I did to convert the horizon-extensions stuff that to ansible galaxy style. Do you want to take a look at #4 and give me your feedback? One of us will have to make some minor changes to adapt to the other's structure.

andymcc:
@miguelgrinberg I've run into a few issues - I'm trying to fully test and am almost there, here are the issues i've run into so far:
I can't import the rpc_repo var from vars/all.yml and then use it directly beneath so I moved it into the user_maas_variables which is essentially a place where all the maas vars are copied (additionally the new galaxy roles patch creates an openstack-ansible binary which will chain together the user_x.yml variable files so that happens automatically). I see you've solved this by adding a global_vars which seems pretty sensible!

Second issue is the library import - this will work if you only call the module on localhost, it seems ansible will copy up the module to the remote node, but it will copy the "link file" and not the module that it links to -which causes failures. As a result I'm creating a task that will copy the module over.

Once I've got this fully working (last blocker is the keystone user tasks which I didn't even add! and now that I've added isn't working as intended) I'll compare fully to yours and we can make a call.

PR #4, opened 2015-02-20
ansible galaxy restructure

miguelgrinberg:


andymcc:
@miguelgrinberg I've completed #3 in terms of functionality (and its working/setting up etc correctly now!) - a couple things I think we should do:

* Move the playbooks into the root of the repo, so rpc-extras/playbooks, this mirrors os-ansible-deployment now.
* Stick the playbook files in here, and the roles, inventory, modules in here also.

I think the main difference is the location of the rpc_repo_path var (I think yours won't work just because the var has moved to group_vars/all/vars.yml and it should just be group_vars/all.yml) - but I think I prefer your method, although the downside is that it will require the user to set that var if its not in /opt/os-ansible-deployment and the var is in group_vars whereas all other vars that they should ever need to set are in the user_x.yml files, though structurally I prefer the group_vars file.

Other than that and the directory structure, is there anything else we're not sync'd up on?

miguelgrinberg:
@andymcc so what I did is model the structure to match stuff you find at the ansible galaxy site. The nice thing about those is that you don't have to add those sym links to roles, vars, etc. since everything is at the root.

If you prefer to go with a rpc-extras/playbooks, and put the sym links in there that's fine, I can adapt to that. So what we can do is merge yours first, then I reorg the horizon playbook/roles to match.

PR #5, opened 2015-02-23
Move rpc-support roles to rpc-extras

andymcc:
rpc_support roles were removed from os-ansible-deployment.
Add a role for this, with tags.
Convert to a galaxy role.
Set variables appropriately.


miguelgrinberg:
@andymcc is this ready to go? I just merged my horizon stuff, and as part of that added a `site.xml` playbook that includes all others. I put maas and horizon-extensions in it, we should get this one added there as well.

andymcc:
I'll merge this in and adjust the site.xml. Thanks @miguelgrinberg

Issue #10, opened 2015-02-24
Add pccommon to deployment


toxick:
Please add pccommon.sh to the utility and neutron_agents containers.

  https://github.com/rsoprivatecloud/pubscripts/blob/master/pccommon.sh

There.

andymcc:
@toxick I added a PR for this - can you confirm that works for you? I.e. is having it in /root acceptable or would you prefer it gets placed somewhere else? I set the perms to 0700.

toxick:
Hi andy - thanks for that.  Looks like that should work just fine.

andymcc:
This merged - forgot to close the issue!

b3rnard0:
@andymcc My KPIs have suffered greatly.... :-)

PR #12, opened 2015-02-25
Get pccommon.sh on neutron_agent & utility hosts

andymcc:
Get the pccommon.sh script from rsoprivatecloud repository and set its
perms to be executable inside /root/ - this should be done on all
neutron_agent and utility hosts.

Fixes-Issue: #10


mattt416:
@toxick confirmed that this meets support's needs, so :+1: from me.

stevelle:
+1

Issue #22, opened 2015-03-02
Errors when running setup-maas.yml


mattt416:
I ran into the following two issues when deploying maas with the separated rpc-extras repo:

1. `maas_keystone_password` is not defined in etc/openstack_deploy/user_maas_settings.yml (should it exist there, or does the deployer need to add it to /etc/openstack_deploy/user_secrets.yml?)
2. `maas_fqdn_extension` does not default to '' (we either need to add a |default('') to entity_name variable in playbooks/roles/rpc_maas/tasks/main.yml or set this in etc/openstack_deploy/user_maas_settings.yml)

PR #23, opened 2015-03-03
Fix issue #22

mattt416:
This commit addresses the two issues outlined in
https://github.com/rcbops/rpc-extras/issues/22.  We now create a
user_extras_secrets.yml which should be copied to
/etc/openstack_deploy and can have pw-token-gen.py run against it.
Additionally, we rename user_maas_settings.yml to
user_extras_variables.yml to clarify that this file can be used for
other roles in rpc-extras, while also keeping the naming consistent
with user_variables.yml in os-ansible-deployment.


miguelgrinberg:
LGTM

stevelle:
:+1:

PR #28, opened 2015-03-05
Create external address setting for maas checks

### LINKED ISSUE #27: MaaS remote checks need a separate address setting
In configurations where the {{ external_vip_address }} is only available via VPN, the MaaS remote checks fail because we're also using the same setting, which MaaS can not access. We need a maas specific setting for the external address, which can default to being external_vip_address in the user config. Then this can be overridden for these installations.

mirated from https://bugs.launchpad.net/openstack-ansible/+bug/1402134 as the fix post-Juno should not go into os-ansible-deployment

stevelle:
New setting maas_external_ip_address defined with default value of
external_vip_address and existing remote maas checked migrated to
use the new setting.

Fixes #27


miguelgrinberg:
I don't understand maas enough to know why this is needed, but the change in itself is pretty safe.

:+1:

d34dh0r53:
Looks good to me

sigmavirus24:
Looks good to me! :+1: It might be worth noting (somewhere) that `{maas_,}external_ip_address` *can* be the same.

PR #29, opened 2015-03-06
Fix regression introduced in adcffa0

### LINKED ISSUE #30: Regression introduced in adcffa0
In commit adcffa0, we added |default('') to maas_fqdn_extension since
maas_fqdn_extension was not defaulted in any user variable files and
would cause an ansible run failure due to the variable not existing.
Upon doing this, I noticed that maas_fqdn_extension was always being
set to '', even when it was defined in a user variables file. Oddly,
I could print the contents of maas_fqdn_extension right above the
change made and it was visible, so it is unclear why adding the
|defauilt('') was causing this issue.

mattt416:
In commit adcffa0, we added |default('') to maas_fqdn_extension since
maas_fqdn_extension was not defaulted in any user variable files and
would cause an ansible run failure due to the variable not existing.
Upon doing this, I noticed that maas_fqdn_extension was _always_ being
set to '', even when it was defined in a user variables file.  Oddly,
I could print the contents of maas_fqdn_extension right above the
change made and it was visible, so it is unclear why adding the
|defauilt('') was causing this issue.

This change removes any |default('') reference on maas_fqdn_extension
and sets a default value in rpc_maas/defaults/main.yml instead. This is
the better solution and should have been done in the first place.

Closes #30

### DIFF
--- a/playbooks/roles/rpc_maas/defaults/main.yml
+++ b/playbooks/roles/rpc_maas/defaults/main.yml
@@ -64,7 +64,8 @@ maas_monitoring_zones:
   - mzhkg
 
 # Set the fqdn extension if required, sample only
-# maas_fqdn_extenstion: .example.com
+# maas_fqdn_extension: .example.com
+maas_fqdn_extension:
 
 # Set the following to skip creating alarms for this device for disk utilisation monitoring
 # maas_excluded_devices: ['xvde']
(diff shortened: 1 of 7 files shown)


miguelgrinberg:
Don't understand why this didn't work, but I agree that this is overall a better way, so :+1: from me.

Issue #9, opened 2015-02-24
 maas_local check ansible_fqdn is an issue


BjoernT:
Some checks fail in maas_local due to playbook changes :

nova_compute check for instance:

9.0.1
   check_details: file=nova_service_check.py,args=--host,args={{ ansible_hostname }},args={{ internal_vip_address }}
9.0.5
  check_details: file=nova_service_check.py,args=--host,args={{ ansible_fqdn }},args={{ internal_vip_address }}

This change caused all checks to go fatal because the nova service name is installed using the short hostname, the actual maas check is looking for a FQDN name. So we either revert this change or change the maas checks to look for short hostnames first and then the FQDN hostname.

b3rnard0:
We are also tracking this here: https://trello.com/c/SwJqrVzY/460-rpc-extras-maas-local-check-ansible-fqdn-is-an-issue

miguelgrinberg:
@BjoernT let me see if I understand. This isn't a problem introduced here in rpc-extras, same issue exists in the 9.0.5 version of the integrated maas plays?

BjoernT:
Yes it does exist in older releases too.
But I can't file it as os-openstack-ansible since it was generalized, the Rackspace specifics got removed.

miguelgrinberg:
Yes, that's fine, just wanted to make sure.

mattt416:
It looks like the check for neutron_linuxbridge_agent_check is also affected by this.

miguelgrinberg:
@mattt416 so it looks like you made the change from `ansible_hostname` to `ansible_fqdn`. Here is the commit: https://github.com/stackforge/os-ansible-deployment/commit/d7216eccb6e7733f1079e09d99b04ca58634da19.

Unfortunately I can't find issue #452, that probably refers to the old ansible-lxc-rpc repo, and the issues there were removed for some reason.

I can't find any information regarding the problem that made you change this, I really don't understand why `ansible_hostname` was not giving you the correct thing in the first place.

I can revert these back to `ansible_hostname`, but what am I breaking if I do that?

mattt416:
@miguelgrinberg Correct -- I appear to have broken this.  :)  The original issue was that we were using `ansible_hostname` everywhere, which was problematic if someone had deployed nodes w/ a proper FQDN.  The change made was fine except for `nova_compute_check` and `neutron_linuxbridge_agent_check` which use the FQDN:

```
root@aio1_utility_container-0cdd7735:~# nova service-list
[...]

PR #31, opened 2015-03-06
Dynamically determine hostname for MaaS checks

mattt416:
Currently, we assume hosts returned from a `nova service-list` and
`neutron agent-list` are FQDNs which is an incorrect assumption. The
original change was made as we ran these MaaS-related playbooks/roles
on hosts w/ incorrectly configured hostnames (meaning a `hostname`
would return the FQDN), and therefore the hosts being returned from the
above two commands were FQDNs.

This change replaces references to ansible_fqdn to ansible_nodename,
which uses a python platform.node() to obtain the hostname. This should
allow us to properly configure MaaS irrespective of whether or not an
installation has correctly configured hostnames on the individual nodes.

Closes issue #9

### DIFF
--- a/playbooks/roles/rpc_maas/tasks/local.yml
+++ b/playbooks/roles/rpc_maas/tasks/local.yml
@@ -28,7 +28,7 @@
 - include: local_setup.yml
   vars:
     check_name: cinder_scheduler_check
-    check_details: file=cinder_service_check.py,args=--host,args={{ ansible_fqdn }},args={{ internal_vip_address }}
+    check_details: file=cinder_service_check.py,args=--host,args={{ ansible_nodename }},args={{ internal_vip_address }}
     check_period: "{{ maas_check_period }}"
     check_timeout: "{{ maas_check_timeout }}"
     alarms:
@@ -40,7 +40,7 @@
 - include: local_setup.yml
   vars:
     check_name: cinder_volume_check
-    check_details: file=cinder_service_check.py,args=--host,args={{ ansible_fqdn }},args={{ internal_vip_address }}
+    check_details: file=cinder_service_check.py,args=--host,args={{ ansible_nodename }},args={{ internal_vip_address }}
     check_period: "{{ maas_check_period }}"
     check_timeout: "{{ maas_check_timeout }}"
     alarms:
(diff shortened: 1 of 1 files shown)


Apsu:
Per https://docs.python.org/2/library/socket.html#socket.gethostname the socket.gethostname() call doesn't always return the FQDN; one should use socket.getfqdn() if they want to always get the FQDN. This implies that gethostname will sometimes return the FQDN, which isn't desired in the case of this issue.

Therefore I think the easiest solution is to assume the shortname is the same as the first segment of the FQDN, and split to retrieve it, like:

    python -c "import socket; print socket.gethostname().split('.')[0]"

odyssey4me:
@Apsu The point is to return the same hostname as returned by ansible in the fact `{{ ansible_hostname }}`, not to return either a short host name or FQDN.

Apsu:
@odyssey4me That's not what I understood to be the case from our IRC discussion. But if that is the case, then this is very relevant:

https://github.com/ansible/ansible/issues/9971

Which quotes ansible's fact code for hostnames as:

    self.facts['fqdn'] = socket.getfqdn()
    self.facts['hostname'] = platform.node().split('.')[0]
    self.facts['nodename'] = platform.node()

Meaning that `ansible_hostname` is, in fact, just a split hostname ( https://docs.python.org/2/library/platform.html#platform.node ) like I suggested. That said, to ensure exact equivalency, I will suggest this code instead:

    python -c "import platform; print platform.node().split('.')[0]"

Apsu:
So wait... we're trying to match what openstack services will consider to be the hostname, when the host may have an /etc/hosts with a shortname + FQDN, or just shortname, or just FQDN. Neither `ansible_hostname` nor `ansible_fqdn` will necessarily cover both cases, but `ansible_nodename` (`platform.node()`) most likely will. How about we switch to using that and re-test the MaaS bits?

mattt416:
@Apsu that's exactly it.  I didn't know about the existence of `ansible_nodename`, so I'll certainly try that.  Thanks!

mattt416:
@Apsu @odyssey4me I've updated the PR to use `ansible_nodename` as that seems to work fine.

Apsu:
:+1:

sigmavirus24:
:+1:

PR #37, opened 2015-03-12
Search for the rpc_repo_path setting in several config files (fixes #33)

### LINKED ISSUE #33: Dynamic inventory assumes /opt/os-ansible-deployment
Despite there being an override for `rpc_repo_path` in `etc/openstack_deploy/user_extras_variables.yml`, this is never read as dynamic_inventory.py pulls `rpc_repo_path` from `playbooks/group_vars/all.yml`.  This means the user cannot relocate os-ansible-deployment without having to modify rpc-extras itself.

Something like this will help, but then puts an assumption on `/etc/openstack_deployment`:

```
#!/usr/bin/env python
import os
import sys
import subprocess
import yaml

basedir = os.path.abspath(os.path.dirname(__file__))
group_vars_yaml = os.path.join(basedir, '../group_vars/all.yml')
user_vars_yaml = '/etc/openstack_deploy/user_extras_variables.yml'

try:
    rpc_repo_path = yaml.load(open(user_vars_yaml, 'r'))['rpc_repo_path']
except KeyError:
    rpc_repo_path = yaml.load(open(group_vars_yaml, 'r'))['rpc_repo_path']

rpc_inventory = os.path.join(rpc_repo_path,
                             'playbooks/inventory/dynamic_inventory.py')
subprocess.call([rpc_inventory] + sys.argv[1:])
```

Anyone have any suggestions as to how we should approach this?

miguelgrinberg:


miguelgrinberg:
@mattt416 Please review.

playbooks/inventory/dynamic_inventory.py
```diff
@@ -4,10 +4,34 @@
...
+    file_check = [
+        '/etc/openstack_deploy/user_extras_variables.yml',
+        os.path.join(os.environ.get('HOME'),
+                     'openstack_deploy/user_extras_variables.yml'),
+        os.path.join(basedir, '../group_vars/all.yml')
+    ]
+
+    for f in file_check:
+        if os.path.isfile(f):
+            cfg = yaml.load(open(f, 'r'))
```
cloudnull: NIT: This does not close the file upon load. We should either use "with" or a "try/finally" to ensure that the file descriptor is closed after its been read.

miguelgrinberg: Make sense, will add that.

mattt416:
LGTM!

PR #54, opened 2015-04-09
Add F5 pool monitoring scripts

git-harry:
The scripts f5-config.py and f5-monitor.sh are part of the juno branch
of os-ansible-deployment. The were removed in master because they are
Rackspace specific files.

This commit adds these scripts here so they are available for the kilo
release.

f5-monitor.sh was modified from the juno version, this was done because
keystone no longer supports XML. The script now parses the JSON
response for the auth token.

Closes-Bug: https://github.com/rcbops/rpc-extras/issues/46

### DIFF
--- a/scripts/f5-config.py
+++ b/scripts/f5-config.py
@@ -1,0 +1,457 @@
+#!/usr/bin/env python
+# Copyright 2014, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


mancdaz:
:+1: 

Tested locally against an haproxy fronted deploy, and this works fine (in that, parsing the json responses is functional).

This needs properly tested on an F5. We need confirmation that the carriage returns (\r) are required in the F5 environment. If not they can be stripped.

git-harry:
@miguelgrinberg @nrb - thanks for spotting the references to rpc_deploy. I've switched them out for openstack_deploy.

nrb:
+1

I can't see any more issues with this, and I don't have an F5 to test against.

miguelgrinberg:
I'm also :+1: on this, but I think it would be a good idea to do some testing, in particular the JSON parsing in bash to get the auth token.

claco:
Re: testing, this is going to really fall into QE as they have more active labs than we do. Jonathan in support is in the loop to help out with this, so with that, I'll do the green button.

PR #56, opened 2015-04-09
Improve os-ansible-deployment integration

odyssey4me:
This patch removes the requirement for rpc-extras to hold duplicate libraries
and the dynamic inventory from os-ansible-deployment, implementing a way to
rather make use of the inventory, libraries and roles within
os-ansible-deployment through the ansible configuration file.

This dramatically simplifies workflow and reduces complexity in making use of
rpc-extras.


playbooks/ansible.cfg
```diff
@@ -11,5 +11,17 @@ forks = 15
 # SSH timeout
 timeout = 120
 
+# Set the path to the folder in os-ansible-deployment which holds the dynamic
+# inventory script
+inventory = /opt/os-ansible-deployment/playbooks/inventory/
```
hughsaunders: could we use opt/stackforge for consistency with gate?

andymcc:
we'll need to remove the reference to the module_import role from the setup-maas and rpc-support plays.

Is there a way we could template out the ansible.cfg, e.g. if i set that repo_path var it will use that as the path, that way i could adjust it in my vars and it'd use that? Or leave it to the default which is /opt/os-ansible-deployment like in the script, and it won't change how this functions. Maybe that is too tricky/convoluted.

playbooks/ansible.cfg
odyssey4me: @hughsaunders I think that the user guides make use of /opt/os-ansible-deployment, which is why this convention was used throughout rpc-extras (you'll see it in the group vars too).

odyssey4me:
@andymcc The only way I can see the contents of ansible.cfg being templated is if we use some sort of bash script to do the needful setup. I'm not opposed to that and it may be useful to use something like this to allow jenkins to put things whereever it needs to. @hughsaunders Any suggestions/thoughts?

playbooks/ansible.cfg
andymcc: We just need to add the "hostfile" setting here (its above in that file) and make a note that if using <1.9 for ansible you need to set that instead of inventory, and for >1.9 set inventory value. (And set hostfile also) Otherwise this works pretty well and definitely LGTM

andymcc:
+2 LGTM lets mergerate this thing

README.md
```diff
@@ -50,9 +50,13 @@ desired plays.  IE:
 openstack-ansible setup-everything.yml
 ```
 
+**Note**: If you have cloned the os-ansible-deployment repository to anywhere
+other than `/opt/os-ansible-deployment` then you will need to edit the paths
+located in `rpc-extras/playbooks/ansible.cfg` to ensure that the dynamic
+inventory, libraries and roles are accessible.
+
```
mancdaz: this is probably better as a step in the 'basic setup' steps.  ie

3. edit playbooks/ansible.cfg and ensure the paths to the roles, playbooks and inventory are correct

mancdaz:
working for me - 1 small comment on the readme

odyssey4me:
@mancdaz updated the README with your suggestion

mancdaz:
:+1:

Issue #58, opened 2015-04-10
Invalid defaults set in roles/rpc_maas/defaults/main.yml


mattt416:
We currently have these defaults:

maas_auth_url: "{{ rackspace_cloud_auth_url }}"
maas_username: "{{ rackspace_cloud_username }}"
maas_api_key: "{{ rackspace_cloud_api_key }}"

However, os-ansible-deployment no longer sets default values for rackspace_cloud_auth_url, rackspace_cloud_username, and rackspace_cloud_api_key.  We will need to update these defaults to point to static dummy values.

PR #59, opened 2015-04-10
Clean up roles/rpc_maas/defaults/main.yml

mattt416:
Currently, roles/rpc_maas/defaults/main.yml has variables which point
to other rackspace-related variables that are no longer defaulted in
os-ansible-deployment's user_variables.yml file.  This change updates
roles/rpc_maas/defaults/main.yml to point to generic (static) values,
while also adding a new maas_tenant_id variable for the Rackspace Cloud
tenant ID (this avoids the need of having to override the maas_api_url
variable).  Additionally, we create some defaults for the
rackspace-related variables in user_extras_variables.yml (we have other
overrides in user_extras_variables.yml which area already pointing to
these).

On a side note, the reason for retaining the rackspace-related
variables is so that user_variables.yml overrides for configuring
glance CloudFiles backend can also point to these rackspace-related
variables.  This will avoid having to maintain two sets of Rackspace
details in your overrides files.

Closes issue #58

### DIFF
--- a/etc/openstack_deploy/user_extras_variables.yml
+++ b/etc/openstack_deploy/user_extras_variables.yml
@@ -13,17 +13,24 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+## Rackspace Cloud Details
+rackspace_cloud_auth_url: https://identity.api.rackspacecloud.com/v2.0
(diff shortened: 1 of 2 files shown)


mancdaz:
:+1: 

Someone else please review

andymcc:
This LGTM. merging.

PR #63, opened 2015-04-13
Add full combined procedure for os-ansible-deployment/rpc-extras

odyssey4me:
This patch includes the combined procedure to deploy rpc-extras with
os-ansible-deployment in such a way that all the preparation is done initially
before executing the deployment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,28 +28,35 @@ for Rackspace Private Clouds.
 
 Basic Setup:
 
-1. Complete an installation of
-[os-ansible-deployment](https://github.com/stackforge/os-ansible-deployment)
-as per the standard mechanism described there.
-2. Clone this repository.
-3. Recursively copy `rpc-extras/etc/openstack_deploy/*` to
+1. Clone 
+[os-ansible-deployment](https://github.com/stackforge/os-ansible-deployment).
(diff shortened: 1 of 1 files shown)


hughsaunders:
LGTM :+1:

README.md
```diff
@@ -28,28 +28,35 @@ for Rackspace Private Clouds.
...
 
-1. Complete an installation of
-[os-ansible-deployment](https://github.com/stackforge/os-ansible-deployment)
-as per the standard mechanism described there.
-2. Clone this repository.
-3. Recursively copy `rpc-extras/etc/openstack_deploy/*` to
+1. Clone 
+[os-ansible-deployment](https://github.com/stackforge/os-ansible-deployment).
+2. Clone [rpc-extras](https://github.com/rcbops/rpc-extras).
+3. From the root of the os-ansible-deployment clone, execute
```
stevelle: Just as a small point, it seems like 4 could well be done before 3. That is also the order they are executed by gate-check-commit.sh and unless we have a reason to deviate I'm not sure why we should.

odyssey4me:
@stevelle Re-ordered steps 3 & 4 to match the order in which the gate check implements it.

mancdaz:
:+1:

PR #68, opened 2015-04-14
Add Log Shipping rule

stevelle:
Add example definition needed by rsyslog client to ship logs to
logstash. Example defines logstash_port as needed by the log shipping
rule.

Add play to run all logging related tasks and related README change.


etc/openstack_deploy/user_extras_variables.yml
```diff
@@ -78,3 +78,22 @@ maas_filesystem_threshold: 90.0
...
 # elasticsearch_heap_size_mb: 1024
+
+# Ship all logs to the logstash container as well as the rsyslog_server.
+# Uncomment the following if the setup-logging play is going to be used.
+# logstash_port: 5544
+# rsyslog_client_user_defined_targets:
+#    - name: "rax-logging"
+#      proto: "tcp"
+#      port: "{{ logstash_port }}"
+#      hostname: "groups['rsyslog_all'][0]"
```
odyssey4me: something appears to be wrong here, this is processed too literally and I'm not entirely sure about the intent here

etc/openstack_deploy/user_extras_variables.yml
```diff
@@ -78,3 +78,22 @@ maas_filesystem_threshold: 90.0
 # For an AIO it's recommended to set the following to limit the expected RAM
 # usage of elasticsearch.
 # elasticsearch_heap_size_mb: 1024
+
+# Ship all logs to the logstash container as well as the rsyslog_server.
+# Uncomment the following if the setup-logging play is going to be used.
+# logstash_port: 5544
+# rsyslog_client_user_defined_targets:
+#    - name: "rax-logging"
```
odyssey4me: rpc-logging is perhaps better?

playbooks/site.yml
```diff
@@ -1,5 +1,4 @@
----
```
odyssey4me: I don't think this should be removed, and would think that the copyright gumpf should be here too.

etc/openstack_deploy/user_extras_variables.yml
stevelle: the hostname was a bit of a WIP.  iterating on that bit now.

etc/openstack_deploy/user_extras_variables.yml
stevelle: done

playbooks/site.yml
stevelle: done

etc/openstack_deploy/user_extras_variables.yml
```diff
@@ -78,3 +78,22 @@ maas_filesystem_threshold: 90.0
...
 # elasticsearch_heap_size_mb: 1024
+
+# Ship all logs to the logstash container as well as the rsyslog_server.
+# Uncomment the following if the setup-logging play is going to be used.
+# logstash_port: 5544
+# rsyslog_client_user_defined_targets:
+#    - name: "rpc-logging"
+#      proto: "tcp"
+#      port: "{{ logstash_port }}"
+#      hostname: "groups['logstash_all'][0]"
```
d34dh0r53: This needs to have double curly braces:
    hostname: "{{ groups['logstash_all'][0] }}"

d34dh0r53:
With the exception of the double curly problem I think this is good to go.

etc/openstack_deploy/user_extras_variables.yml
stevelle: done

Issue #69, opened 2015-04-15
rpc_maas role still references nova_spice_console


mattt416:
This host group is now called nova_console.

andymcc:
merged closing this out!

PR #70, opened 2015-04-15
Rename nova_spice_console to nova_console

mattt416:
In commit 5cce6b5 in os-a-d, we renamed nova_spice_console group to
nova_console.  This commit updates rpc_maas so we correctly deploy
a check/alarm against the actual group.

Closes issue #69


odyssey4me:
:+1: LGTM

andymcc:
+2

Issue #71, opened 2015-04-15
hardware_setup.yml and dell_setup.yml are exactly the same


andymcc:
We should get the dell monitors to use the hardware_setup.yml file instead of duplicating the file.

PR #72, opened 2015-04-15
Use the generic hardware_setup.yml for Dell MaaS

andymcc:
Since the hp checks were added with a generic hardware_setup.yml task
that is the exact same as the dell_setup.yml task, move the dell
monitoring to be setup using the generic hardware_setup.yml.

Avoids duplication of a file that is exactly the same.

Closes-Issue: #71


git-harry:
+1

odyssey4me:
:+1: LGTM

Issue #74, opened 2015-04-16
Fix keystone_service_user_name / keystone_auth_admin_password


mattt416:
playbooks/roles/rpc_maas/tasks/keystone_user.yml is currently failing as we resolved https://bugs.launchpad.net/openstack-ansible/+bug/1443955 upstream in os-a-d.  We need to update this role to use a token instead.

PR #75, opened 2015-04-16
Run keystone module w/ token

mattt416:
The fix for https://bugs.launchpad.net/openstack-ansible/+bug/1443955
breaks the tasks using the keystone module.  This commit updates the
rpc_maas role to use a token rather than the keystone service user when
interacting with keystone.  This is the pattern used in other os-a-d
roles.

Closes issue #74


odyssey4me:
:+1: LGTM

PR #76, opened 2015-04-16
Add haproxy playbook/role for RPC

odyssey4me:
This adds a role for RPC haproxy configurations, including:

* elasticsearch
* kibana

The playbook is intentionally not included in the general logging playbook or
in the site playbook as haproxy is used for development and testing only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@ OpenStack cloud.
 Plays:
 
 * `elasticsearch.yml` - deploys an elasticsearch host
+* `haproxy` - deploys haproxy configurations for elasticsearch and kibana
 * `horizon_extensions.yml` - rebrands the horizon dashboard for Rackspace,
 as well as adding a Rackspace tab and a Solutions tab, which provides
 Heat templates for commonly deployed applications.
--- a/playbooks/haproxy.yml
+++ b/playbooks/haproxy.yml
@@ -1,0 +1,24 @@
+---
+# Copyright 2014, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+- name: Configure HAProxy for RPC
+  hosts: haproxy_hosts
+  user: root
+  roles:
+    - { role: "haproxy_rpc" }
+  vars_files:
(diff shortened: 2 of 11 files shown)


d34dh0r53:
+2 works for me

nrb:
+1

miguelgrinberg:
A few comments:

- The `playbooks/roles/haproxy_rpc/CONTRIBUTING.rst` and `playbooks/roles/haproxy_rpc/LICENSE` are not related to this patch. Shouldn't these files be added to the root directory in a separate patch?
- Calling the playbook "haproxy" is a bit misleading, in my opinion. For horizon I used "horizon-extensions" as name, to make it clear that the playbook builds on top of horizon. How do "haproxy_extensions" or "haproxy_logging" sound?
- The name of the role can also be more specific, should probably be the same name as the playbook.
- In the `meta/main.yml` file, you put `haproxy-server` as a dependency. Does that work at all? The `haproxy-server` role is in the OSAD repo, I don't think there is visibility into that from here.

sigmavirus24:
@miguelgrinberg regarding your first point, other directories have CONTRIBUTING, README, and LICENSE files in them: https://github.com/rcbops/rpc-extras/tree/master/playbooks/roles/kibana (for example)

I think the rest of your points are very valid though, so I'm -1 on this for the time being.

miguelgrinberg:
@sigmavirus24 yeah, okay. The only one that doesn't have these files is the one I wrote :)

I agree, this is all role-centric, so it makes sense to have these files on each role. But copies in the root directory wouldn't hurt.

odyssey4me:
If there should be README, CONTRIBUTING and LICENSE files in the root of this repo I think it's fair to say that this can be done in a separate PR. These files have been included in the role as per the convention with the os-ansible-deployment roles - this is done with the idea that at some stage each role could end up in its own repository.

@miguelgrinberg Having the OSAD haproxy role as a dependancy means that it is run first before the haproxy_rpc role. This works and that is why the playbook is imply called haproxy, not haproxy_rpc. It will setup haproxy and negate the need to run the OSAD playbook. This dependancy works because of the configuration in ansible.cfg which merged in another PR.

I don't see why the name of the role is inappropriate. This is a role to apply haproxy configuration changes that are appropriate for RPC. The way it's been done allows extensions for other RPC requirements if need be, even though the only use right now pertains to the ELK stack.

miguelgrinberg:
@odyssey4me wasn't aware of the changes in ansible.cfg to locate the roles and inventory of OSAD, thanks for pointing that out.

My main point regarding the naming is that people may want to use haproxy without the extensions in this repo. This playbook is not about haproxy, but about enabling an pre-existing haproxy to work with the rpc-extras component. Honestly it is not a big deal, you clarified the main problem I had with the dependencies, so I don't oppose this change, now it's just name choices, so I'm :+1:

odyssey4me:
@miguelgrinberg If there's consensus on a better name, I'm all for it... but at this stage I don't think it's worth too much fuss considering that the role is only for development/testing convenience anyway.

miguelgrinberg:
@odyssey4me yup, agree 100%.

claco:
Sounds like everyone is ok with current status and a merge.

PR #77, opened 2015-04-16
Update Kibana from v3.1.0 to v3.1.2

odyssey4me:
This updates the kibana package downloaded to v3.1.2, implements a checksum validation and also implements a package download if the copy on disk is different to the one expected.

### DIFF
--- a/playbooks/roles/kibana/defaults/main.yml
+++ b/playbooks/roles/kibana/defaults/main.yml
@@ -14,9 +14,9 @@
 # limitations under the License.
 
 kibana_root: /opt/kibana
-kibana_version: "kibana-3.1.0"
+kibana_version: "kibana-3.1.2"
 kibana_url: "https://download.elasticsearch.org/kibana/kibana/{{ kibana_version }}.tar.gz"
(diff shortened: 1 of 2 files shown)


stevelle:
:+1: checked values

d34dh0r53:
+2 works for me

nrb:
+1

miguelgrinberg:
+1

sigmavirus24:
Should these be squashed into one commit?

stevelle:
yes they should.

odyssey4me:
I don't see why the commits should be squashed. In fact, I think it's better that they aren't so that each commit has a clear purpose.

git-harry:
I agree with keeping them separate. I think an argument could be made for doing them as separate pull requests but I think it is okay to merge as is because the commit history is the important thing.

nrb:
@odyssey4me's reasoning for separate commits make sense to me. If we're not mandated by the tool, I'd say leave them separate.

sigmavirus24:
:+1: Also, we won't be "backporting" this so it's fine.

PR #78, opened 2015-04-16
Implement rpc-release file drop

odyssey4me:
This patch adds the drop of /etc/rpc-release into the support role as it's
expected that this role will be used in all RPC deployments, whereas the other
roles may be considered optional.

### DIFF
--- a/playbooks/roles/rpc_support/defaults/main.yml
+++ b/playbooks/roles/rpc_support/defaults/main.yml
@@ -13,6 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# RPC release version
+rpc_release: master
+
 # Packages required for support
 support_util_packages:
   - vim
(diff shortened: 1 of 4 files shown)


nrb:
+1. (note, only read, not executed)

git-harry:
Why not have a separate role for this? Then there need be no assumption.

claco:
Iterate. This seems like a functional equiv to juno issue. We can improve on it all we want later, but I'd rather now expand our work as we need to get get things wrapped up for a branch/tag today.

sigmavirus24:
:+1: looks good to me.

miguelgrinberg:
+1 for current release, but I think we should automate obtaining the tag in the future.

miguelgrinberg:
This does not work for me, getting a 500 back from nova. I'm looking into it, will submit a patch once I know what's going on.

PR #81, opened 2015-04-20
Add kernel modules, tuning and support packages for hosts

odyssey4me:
This patch adds several packages requested by RPC support to all hosts and
containers, adds kernel module loading and sysctl tuning to all hosts.

The reference for all additions can be found in the description and comments
for https://bugs.launchpad.net/openstack-ansible/+bug/1400424

### DIFF
--- a/playbooks/roles/rpc_support/defaults/main.yml
+++ b/playbooks/roles/rpc_support/defaults/main.yml
@@ -18,13 +18,33 @@ rpc_release: master
 
 # Packages required for support
 support_util_packages:
-  - vim
-  - linux-crashdump
-  - iotop
+  - bridge-utils
+  - debootstrap
   - dstat
-  - tcpdump
-  - lsof
   - htop
+  - iotop
+  - iptables-persistent
+  - linux-crashdump
+  - lsof
+  - lvm2
+  - ntp
+  - ntpdate
+  - openssh-server
+  - sudo
+  - tcpdump
+  - vim
+  - vlan
+
+# kernel modules required on hosts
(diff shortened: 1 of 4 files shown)


hughsaunders:
LGTM :+1:  
One question.. why debootstrap?

odyssey4me:
@hughsaunders I have no idea, it was requested in the launchpad bug... ultimately I think the main purpose here is to provide the facility to do these things. This allows support to override if need be.

hughsaunders:
Yep, lets get this merged :+1:

mattt416:
There is a lot of duplication here w/ what's already defaulted in https://github.com/stackforge/os-ansible-deployment/blob/master/playbooks/roles/openstack_hosts/defaults/main.yml.

rackertom:
debootstrap shouldn't be needed since we're presumably not building images. But it may have been pulled in as a dependency for one of the other packages. I suspect the reason for installing these outside of ansible plays may be chicken-and-the-egg situation where some level of configuration is required before ansible is run, but ansible is clearly the best tool to perform the install.

Either way. :+1: from me.

hughsaunders:
@mattt416 I guess that duplication is so support can tune and add packages without adding them to OSAD

odyssey4me:
@nrb The PR is https://github.com/rcbops/rpc-openstack/pull/81

nrb:
@odyssey4me Thanks. Does anyone know of a reason that this shouldn't be backported to Kilo? @hughsaunders @mattt416 @rackertom

Issue #86, opened 2015-04-27
nova_spice_console group no longer exists.


andymcc:
nova_spice_console has changed to nova_console as a group name so the rpc_support play is failing.

PR #87, opened 2015-04-27
Adjust group name for nova_console group

andymcc:
The nova_spice_console group has changed to be called "nova_console"
this change adjusted the reference in the play so that the spice console
tasks can be run successfully.

Fixes-Issues: #86

### DIFF
--- a/playbooks/roles/rpc_support/tasks/main.yml
+++ b/playbooks/roles/rpc_support/tasks/main.yml
@@ -25,7 +25,7 @@
 
 - include: spice_console.yml
   when: >
-    inventory_hostname in groups['nova_spice_console']
+    inventory_hostname in groups['nova_console']
   tags:
     - nova_spice_console
 


mattt416:
This LGTM, though I wonder if we need to update https://github.com/rcbops/rpc-extras/blob/master/scripts/f5-config.py#L151 as that seems to use the `nova_spice_console`.  Anyone w/ more knowledge of this file know?

nrb:
This change looks good to me.  :+1:

Apsu:
@mattt416 To my knowledge, the data structure for pool members in that script is purely an organizational component in the F5 config, and unrelated to users/groups on the backend nodes in a pool. That said, it probably should still be updated to at least stay consistent with the codebase, if only for reference purposes.

nrb:
Patch proposed for the f5 config here: https://github.com/rcbops/rpc-extras/compare/master...nrb-patch-1?quick_pull=1

mattt416:
@nrb, you going to create a PR w/ that patch?

nrb:
@mattt416 Thought I already had, oops. Done now.

PR #89, opened 2015-04-29
F5 config refresh

miguelgrinberg:
Several F5 script changes in this PR:

1. Updated script from https://github.com/ckent99999/f5-config-generator/blob/master/f5-5-5.py
2. Changed rpc_* references to openstack_*
3. Use variables for partition and prefix names
4. Updated "nova_api_ec2" and "nova_spice_console" group names
5. Fixed monitor port for pkg_repo
6. Fixed horizon_ssl monitor rule
7. Expanded f5-monitor.sh to support a list of acceptable status codes instead of just one
8. Added swift monitoring to f5-monitor.sh
9. Coverted f5-monitor.sh EOLs from CRLF to LF

### DIFF
--- a/scripts/f5-config.py
+++ b/scripts/f5-config.py
@@ -5,7 +5,7 @@
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
 #
-#     http://www.apache.org/licenses/LICENSE-2.0
+#    http://www.apache.org/licenses/LICENSE-2.0
 #
 # Unless required by applicable law or agreed to in writing, software
 # distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


sigmavirus24:
@Mudpuppy12 would you mind reviewing this as well?

scripts/f5-config.py
```diff
@@ -14,55 +14,155 @@
...
 )
 
-MONITORS = (
-    'create ltm monitor mysql %(prefix_name)s_MON_GALERA { count 0 database'
-    ' information_schema debug yes defaults-from mysql destination *:*'
-    ' interval 30 time-until-up 0 timeout 91 username haproxy }'
-)
+#Persistance Profile:
+PERSISTANCE = [
+    r'create ltm persistence source-addr /' + PART + '/' + PREFIX_NAME + '_PROF_PERSIST_IP {'
```
sigmavirus24: Why are you using raw strings here? Those are most useful when trying to use `\`s as just `\` instead of having them used to escape something.

Especially below, you have `'\n'` as something other than a raw string and everything else just has `/`. Am I missing something?

scripts/f5-config.py
```diff
@@ -324,22 +515,18 @@ def args():
...
     parser.add_argument(
-        '-r',
-        '--route-domain',
-        help="Route domain modifier."
-             " Default: [ %(default)s ]",
+        '-S',
+        '--Superman',
+        help='Yes, its Superman ... strange visitor from another planet,'
+             'who came to Earth with powers and abilities far beyond those of mortal men!  '
+             'Superman ... who can change the course of mighty rivers, bend steel in his bare hands,'
```
sigmavirus24: Wait... what?

scripts/f5-config.py
```diff
@@ -349,58 +536,131 @@ def main():
...
+            % user_args,
+        ])
+    if user_args['Superman']:
+        print "       **************************       "
+        print "    .*##*:*####***:::**###*:######*.    "
+        print "   *##: .###*            *######:,##*   "
+        print " *##:  :####:             *####*.  :##: "
+        print "  *##,:########**********:,       :##:  "
+        print "   .#########################*,  *#*    "
+        print "     *#########################*##:     "
```
sigmavirus24: I'm sooooo confused

sigmavirus24:
Oh, I missed the fact that we had lifted this script from another repository. I'm going to assume, unless otherwise told, that the things I commented on are all a result of updating from that repository. Ignore me.

scripts/f5-config.py
miguelgrinberg: This was in the script that I started from (https://github.com/ckent99999/f5-config-generator/blob/master/f5-5-5.py). I only made the fixes listed above.

The script needs a good pep8 pass, by the way. But it would make the review even more difficult if I add that in the same patch.

scripts/f5-config.py
miguelgrinberg: yup  :)

sigmavirus24:
:+1:

scripts/f5-config.py
```diff
@@ -14,55 +14,155 @@
...
+    r' HTTP/1.1\r\nHost: rpc\r\n\r\n" }',
+    r'create ltm monitor http /' + PART + '/' + PREFIX_NAME + '_MON_HTTP_HORIZON { defaults-from http'
+    r' destination *:80 recv "302 Found" send "HEAD / HTTP/1.1\r\nHost:'
+    r' rpc\r\n\r\n" }',
+    r'create ltm monitor http /' + PART + '/' + PREFIX_NAME + '_MON_HTTP_NOVA_SPICE_CONSOLE {'
+    r' defaults-from http destination *:6082 recv "200 OK" send "HEAD /'
+    r' HTTP/1.1\r\nHost: rpc\r\n\r\n" }',
+    r'create ltm monitor https /' + PART + '/' + PREFIX_NAME + '_MON_HTTPS_HORIZON_SSL { defaults-from'
+    r' https destination *:443 recv "302 FOUND" send "HEAD / HTTP/1.1\r\nHost:'
+    r' rpc\r\n\r\n"}',
```
Mudpuppy12: Place a space after the " so it reads " }'. This will aid in parsing the output with the F5api.

miguelgrinberg: Done

Mudpuppy12:
I had an issue today when using ./f5-config.py -f data/openstack_inventory.json --ssl-public-ip 72.4.117.94 --ssl-domain-name eng-iad3-2.rpchost.com --print

### CREATE SSL PROFILES ###
create ltm profile client-ssl /ENG-IAD3-2/ENG-2_PROF_SSL_%(ssl_domain_name)s { cert /ENG-IAD3-2/%(ssl_domain_name)s.crt key /ENG-IAD3-2/%(ssl_domain_name)s.key defaults-from clientssl }
create ltm profile server-ssl /ENG-IAD3-2/ENG-2_PROF_SSL_SERVER { defaults-from /Common/serverssl }

It was not expanding the ssl_domain_name in the output.

miguelgrinberg:
@Mudpuppy12 good catch, I fixed the ssl_domain_name template generation, works fine now.

scripts/f5-config.py
hughsaunders: WUT

hughsaunders:
I attempted to test this against an aio inventory, but it failed because elastic search is missing. I'm guessing thats because elastic search is added via env.d. 

May be nice to check for missing groups? I had to use pdb to work out which group it was. 

```
pdb ./f5-config.py
> /home/hugh/git/rpc-extras/scripts/f5-config.py(17)<module>()
-> import argparse
(Pdb) c
Traceback (most recent call last):
  File "/usr/lib/python2.7/pdb.py", line 1314, in main
    pdb._runscript(mainpyfile)
  File "/usr/lib/python2.7/pdb.py", line 1233, in _runscript
    self.run(statement)
  File "/usr/lib/python2.7/bdb.py", line 400, in run
    exec cmd in globals, locals
  File "<string>", line 1, in <module>
  File "./f5-config.py", line 17, in <module>
    import argparse
  File "./f5-config.py", line 606, in main
    pool_parts = build_pool_parts(inventory=inventory_json)
  File "./f5-config.py", line 375, in build_pool_parts
    inventory, group_name=value['group'], host_dict=value
  File "./f5-config.py", line 353, in recursive_host_get
    if 'children' in inventory_group and inventory_group['children']:
TypeError: argument of type 'NoneType' is not iterable
Uncaught exception. Entering post mortem debugging
Running 'cont' or 'step' will restart the program
> /home/hugh/git/rpc-extras/scripts/f5-config.py(353)recursive_host_get()
-> if 'children' in inventory_group and inventory_group['children']:
(Pdb) p group_name
'elasticsearch'
```

scripts/f5-config.py
sigmavirus24: yussss

miguelgrinberg:
@andymcc sure, that's a good idea. Added the check for existing group. A warning with the group name is printed to stdout for any missing groups.

hughsaunders:
@miguelgrinberg thanks for the missing groups message/skip, much better. Runs fine for me now, but I don't have an actual F5 to test against. So untested LGTM from me.

miguelgrinberg:
There are some concerns on the format I used to define the irules that implement the X-Forwarded headers. The F5 uses a strange command shell called tmsh, which appears to be unable to define irules through a command, it wants users to enter the rule definition interactively in a vim-like editor.

So in the spirit of getting things moving I removed that from this patch, and will merge the everything else which has been under review for about a week and has now no objections. I will follow up with another PR for the proxy headers, pending some more investigation.

PR #92, opened 2015-05-05
Combine rpc-extras and rpc-maas, include os-a-d

prometheanfire:
This moves the maas plugins into rpc-extras along osad.
We still need to rename the repo though.
Pinning will be easy, we just update the submodule to the pin, then pin this repo.

MaaS is more fun, given that it is made to run only with how we deploy live environments, I can only test some/most of the play.  It works and the failures are where I expect given I am testing on a public cloud instance and not an actual install.  So I'm calling it good til we get better testing.


sigmavirus24:
:-1: You should strip all commits from me just for giggles.

sigmavirus24:
Seriously though, :+1: good work.

prometheanfire:
ya, wasn't sure people wanted history, so defaulted to the safe option :D

sigmavirus24:
History is fine until someone tries to blame something on me. =P

nrb:
Thanks for this. I'm going to do a few test deploys with this before giving a thumbs up.

prometheanfire:
ya, if anything it probably needs that readme you were working on I think

nrb:
So far I've been able to spin up instances and run RPC playbooks without issue (save for where I've left configuration values out).

I think we should update the README before merging this stuff in, so we have at least have some reference outside of git history.

prometheanfire:
https://github.com/nrb/rpc-extras/commit/572db1e5784a1acdf971cbbd47e521a0e7975957#commitcomment-10949800

the readme still needs some updates and probably an additional script for the logstash deploy

prometheanfire:
it's my understanding that osad sources it's config from it's repo, then afterwords copies to /etc is it the other way around?

nrb:
os-a-d should only source from /etc/

prometheanfire:
https://review.openstack.org/#/c/181496/ should fix the aio issue

scripts/deploy.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
+
+export RPCD_LOGSTASH=${RPCD_LOGSTASH:-"FALSE"}
+export RPCD_AIO=${RPCD_AIO:-"FALSE"}
+
+# Enable logstash if wanted
+if [ "${RPCD_AIO}" != "yes" ]; then
+  if [ "${RPCD_LOGSTASH}" == "yes" ]; then
+    mv /etc/openstack_deploy/user_logging_variables.yml.example /etc/openstack_deploy/user_logging_variables.yml || exit 1
```
nrb: This line doesn't account for retrying; for the long term solution we should probably make our copy/move operations idempotent.

prometheanfire: I agree, along these lines I want to import osad's script-libraries so we can use that with the ansible runs as well.

PR #93, opened 2015-05-10
Adding warning and critical thresholds for fileystem checks

BjoernT:
Currently we do only create warning alarms which do not trigger core.
I added configurable warning and critical thresholds. Critical thresholds will also set off core.
Additionally we are working in getting warnings into core as well for our devices


BjoernT:
This fixed issue #94

sigmavirus24:
LGTM. I haven't tested these out on a run though

prometheanfire:
ditto, :1: baring convergence failures

claco:
@prometheanfire @sigmavirus24 @BjoernT This needs backporting, or it is likely to miss the 10.1.6 release this week.

/cc  @b3rnard0

PR #96, opened 2015-05-11
initial beaver commit

d34dh0r53:
This commit add initial beaver with multiline support.  In this commit beaver is configured to pull logs directly from the destination directory of the rsyslog server, from there they are run through the multiline filter which is configured for galera multiline and openstack multiline.  Other logs are passed through as is.


odyssey4me:
Active work for this is being done here: https://github.com/odyssey4me/rpc-extras/tree/log_shipping

PR #97, opened 2015-05-11
[WIP] F5 proxy headers

miguelgrinberg:
Add the X-Forwarded headers for proxy identification. Still TBD if the single-line syntax used is appropriate for the F5's tmsh, as it appears tmsh pops up vim session for the rules to be entered interactively.

### DIFF
--- a/scripts/f5-config.py
+++ b/scripts/f5-config.py
@@ -332,6 +332,19 @@ POOL_PARTS = {
 }
 
 
+HTTP_REQUEST_IRULE = ('create ltm rule /' + PART + '/' + PREFIX_NAME + '_%(name)s '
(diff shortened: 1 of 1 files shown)


nrb:
Is this still a work in progress?

miguelgrinberg:
The patch as is works well for configuring the F5 through the REST API (with @Mudpuppy12's script). But this script is not compatible with the F5's tmsh, which apparently can only accept irule commands interactively. We believe a new release of the F5 software (11.6) might have some improvements to tmsh that will allow us to set these rules from a script, but that has not been evaluated yet.

nrb:
Setting to blocked and wishlist for now, since the F5 software update is out of our hands.

PR #99, opened 2015-05-13
Add apt-pinning variable file.

nrb:
This file is separated from the user_extras_variables.yml file in order
to make it more clear to administrators that there are pinned packages.

Also, this is being included with the packages commented out until we
confirm which ones rpc-extras would like to pin.


stevelle:
:+1:

nrb:
A note: we'll have to update our os-a-d submodule commit/tag to one that includes the role (https://review.openstack.org/#/c/181492/).

It's merged in kilo/master, but not yet tagged.

prometheanfire:
:+1: merging

PR #100, opened 2015-05-13
Add support for Corekube on RPC in the Solutions Tab

metral:


galstrom21:
@miguelgrinberg: I think there is a fair discussion around if we need to build a repo in rcbops that holds the heat template (the way the other solutions have been done). I just wanted to make sure we were not talking about pulling corekube into rpc as that would be odd.

PR #101, opened 2015-05-13
bash cleanup and aio with maas fixes

prometheanfire:


stevelle:
:+1:

README.md
```diff
@@ -40,18 +40,21 @@ below for more.
...
-    it will need to be coded by hand.
-3. Run ```scripts/deploy.sh``` within ```/opt/rpc-extras```
-  1. If building AIO, set ```RPCD_AIO=yes``` before running
-  2. If building with Logstash, set ```RPCD_LOGSTASH=yes``` before running
+       it will need to be coded by hand.
+3. Run `scripts/deploy.sh` within `/opt/rpc-extras`
+  1. If building AIO, set `RPCD_AIO=yes` before running
+  2. If building with Logstash, set `RPCD_LOGSTASH=yes` before running
+4. If you want MaaS working with AIO, do the following:
+  1. edit `/etc/openstack_deploy/user_extras_variables.yml` to add credentials
```
nrb: Should these instructions be put in the 'Edit configuration options' step? It would seem we'd want to do configuration before running the deploy.

prometheanfire: for AIO we can't do it in that order, since the aio deploy copies everything and gives no chance to edit the options

nrb: Cool, thanks.

nrb:
:+1:

Issue #94, opened 2015-05-10
Add support for critical fileystems thresholds


BjoernT:
Currently we do only create warning alarms which do not trigger core.
I added configurable warning and critical thresholds. Critical thresholds will also set off core.
Additionally we are working in getting warnings into core as well for our devices

claco:
@b3rnard0 @BjoernT Going to bump this to 10.1.7. Because the backport needs to be worked back into OSAD and rpc-maas repos for Juno, this is going to take a bit longer.

BjoernT:
@claco That's ok

BjoernT:
Juno will be tracked at https://bugs.launchpad.net/openstack-ansible/+bug/1455999 since it still belongs to osad. I guess we can remove the juno-backport-potential tag

PR #108, opened 2015-05-21
Update maas_filesystem_warning_threshold

mattt416:
Currently, maas_filesystem_warning_threshold is set to 90 in
user_extras_variables.yml but defaults to 80 in
playbooks/roles/rpc_mass/defaults/main.yml.  I believe this to be a
typo and am updating to match.

Closes issue #94


stevelle:
:+1: agree on the analysis and solution looks good.

cloudnull:
:+1:

Issue #109, opened 2015-05-21
Add alarm for rabbitmq_max_channels_per_conn


mattt416:
https://github.com/rcbops/rpc-maas/pull/206 adds metric rabbitmq_max_channels_per_conn, we need to add an alarm that checks if this value is > 10.

PR #110, opened 2015-05-21
Add alarm for rabbitmq_max_channels_per_conn

mattt416:
This commit adds a new alarm for metric rabbitmq_max_channels_per_conn,
which is a metric representing the largest (max) number of channels per
connection across all rabbitmq connections.  We have chosen a threshold
of 10 -- running a dev cluster periodically sees
rabbitmq_max_channels_per_conn > 1 and 10 is an 'arbitrary but likely
reasonable limit' which we can use to prevent channel ballooning.

Closes issue #109


mattt416:
This will need to merge _after_ https://github.com/rcbops/rpc-maas/pull/206 is merged.

cloudnull:
:+1:  but someone else will have to merge

andymcc:
LGTM - merging.

PR #112, opened 2015-05-27
Repo builder for rpc-extras

d34dh0r53:
This patch adds the capability for rpc-extras to build wheels and
store them on the local repo server. The repo-build play scans
through all of the YAML files in the tree and creates wheels for
any packages or git sources it finds and places the wheels in the
rpc-extras tree of the repo server. The repo-pip-setup play adds
the rpc-extras tree to the pip configuration of all hosts. Due to
the way repo-build scans empty YAML files are not allowed, hence
the removal of the haproxy main.yml file.


playbooks/group_vars/all.yml
```diff
@@ -14,3 +14,11 @@
 
 # This is the location where the rpc repository was checked out
 rpc_repo_path: /opt/os-ansible-deployment
+
+# This is the OSAD version you are deploying rpc-extras against
```
odyssey4me: Is this correct? I don't think so - in the build play we are building an entirely separate repository, not including these packages in the OSAD repository.

odyssey4me:
I think that any of the roles that do pip installs now need to ensure that they're dependant on the pip_lockdown role instead of just the pip_install role (replace one with the other). That should probably be included in this PR.

Also, it may be prudent to ensure that the repo server's sync service is restarted (or at least started) within this play because I found that it sometimes doesn't sync in time, resulting in failed pip installs.

playbooks/group_vars/all.yml
```diff
@@ -14,3 +14,11 @@
...
+rpc_release: 11.0.1
```
claco: @d34dh0r53 @nrb @stevelle @prometheanfire With extras being the focus of the install experience, with it's own release timing/version numbers, I wonder about this line.

We're using a submodule to point to a specific version of OSAD I thought?. So, is the description just wrong? In fact, whatever you put in that value has no effect on what is actually deployed from OSAD, and is only used for release version files and such?

playbooks/group_vars/all.yml
d34dh0r53: Yeah, that is correct, I will update that comment.

playbooks/group_vars/all.yml
d34dh0r53: Yeah, the description is wrong.  I'll update that comment.

README.md
```diff
@@ -53,20 +58,22 @@ os-ansible-deployment clone.
 6. Set the `rpc_repo_path` in
 `/etc/openstack_deploy/user_extras_variables.yml` to the path of the
 `os-ansible-deployment` repository clone directory.
-7. Set all other variables in
+7. Set the `rpc_release` in `playbooks/group_vars/all.yml` to match the version
```
odyssey4me: This is now incorrect.

claco: @d34dh0r53 Same issue as the other file...

README.md
```diff
@@ -83,6 +90,8 @@ openstack-ansible setup-everything.yml --tags rsyslog-client
 
 # Ansible Roles
 
+* `repo-build`
+* `repo-pip-setup`
```
odyssey4me: you're only adding plays, not roles - remove these

playbooks/group_vars/all.yml
```diff
@@ -14,3 +14,11 @@
 
 # This is the location where the rpc repository was checked out
 rpc_repo_path: /opt/os-ansible-deployment
+
+# RPC Extras version
+rpc_release: master
```
odyssey4me: Seeing as we have this here, perhaps we should pull it out of the support role?

odyssey4me:
Can we squash the commits to get rid of the beaver-related history?

PR #114, opened 2015-05-27
Add support for Corekube on RPC in the Solutions Tab

metral:


git-harry:
-1 if this is a Rackspace-managed template, the code needs to live with all the others in github.com/rcbops

galstrom21:
@git-harry: I think we should be looking how to model 3rd party provided solution templates with this one.. You would not want to create a repo, and manage access rights for a 'partner' vendor providing their stack through this medium. 

Lets use this opportunity to figure out the semantics for that type of scenario.

metral:
adding link to original PR (which is now closed) but adds more context to the merging of this PR: https://github.com/metral/rpc-extras/commit/914679aff20e779e46dc03824f4500235ecdc15b

git-harry:
@galstrom21 

If this is a third-party template it seems misleading to brand it as RPC.

The URL being added is for a file in a branch not a specific SHA which doesn't seem a particularly strong guarantee of what is being used. That could possibly be argued as acceptable when Rackspace controls the repo but for third-party repos that seems like a bad idea.

Maybe customers would like to be able to choose whether or not they want third-party templates or even what list of templates they want to use.

galstrom21:
@git-harry: Ok you have an issue with the repo name containing RPC.. that is an easy fix. @metral should be able to rename the repository to drop the RPC branding.

With regards to the sha/branch/head topic.. Everyone seems to have differing opinions on how that needs to be managed for a 3rd party repository. I would like to see the dev team have a discussion and come to a consensus on how yall want that managed, and then that standard can be applied here.

mattt416:
Hi @metral,

Sounds like we are still having some discussions here -- but could you rebase this change in the mean time?

Thanks!

--Matt

cloudnull:
:-1:  patchset needs to be rebased.

andymcc:
Perhaps this should be made into a variable so that a user can add whichever extensions they want to add, 3rd party or no - in a more flexible way. Its then not our responsibility if those are added  by the user and are not "acceptable".
I think the point is valid, if the 3rd party extension changed in a malicious way we have no control over that and are simply deploying it as default to any deployers.

metral:
rebase complete

galstrom21:
@andymcc: I agree there needs to be a way for a customer to add solution tab entries themselves.. 

But for this example, I think we should be deciding how we want to handle a 3rd party (read partner) solution. A partner solution will be different than a customer added solution tab entry.

andymcc:
@galstrom21 sure - imo we shouldn't advocate/support any solutions tab entries that we can't control, so that would mean its either in our repository or we submodule it (or some other method - using a specific SHA etc) so we can ensure we're happy with what is being pushed up, and it can't just change without us knowing about it.
I don't think it is sensible or good practice to just link to a 3rd party (of any type) repository that we have no control over without ensuring it is stable/tested - this opens us up to potential issues if changes are made to these repositories, that we don't control, that could adversely impact the deployers/users or us.

galstrom21:
@andymcc: I do not disagree on pointing against a sha (or whatever mechanism for stability).. my point was that I would like to see us define a standard and apply it here.

Since this issue started there have heard requests for:

* SHA
* Branch
* Tag
* Point to Head

Once a standard has been defined, it can be applied in this pull request.

b3rnard0:
@claco @mancdaz 

Need some managerial decisions on this issue.

nrb:
Adding blocked status until we have the discussion/decision

git-harry:
@claco @mancdaz what managerial decisions were made?

mancdaz:
@git-harry the discussions around ownership/support of these templates are still ongoing. I will try and get a better answer and come back here with an update

PR #127, opened 2015-05-30
OS-A-D integration rebased

stevelle:
Rework of the RPC deploy to consume OSAD as a sub-repository.
Introduces an RPC deploy script.

RPC Upgrade concerns excluded from this PR to break down the size of the reviews.


prometheanfire:
:+1:

README.md
```diff
@@ -44,46 +44,23 @@ hosts and containers in the deployment using the related plays mentioned
...
 
+1. From `/opt`, clone [rpc-extras](https://github.com/rcbops/rpc-extras)
+with the `--recursive` option to get all the submodules.
+2. Unless doing an AIO build, prepare the os-ansible-deployment configuration.
+  1. copy everything from `os-ansible-deployment/etc/openstack_deploy` 
+     into `/etc/openstack_deploy`
+  2. copy everything from `rpcd/etc/openstack_deploy` into
+     `/etc/openstack_deploy`
+  3. Edit configurations in `/etc/openstack_deploy` for example:
+    1. `openstack_user_variables.yml.example` and
```
odyssey4me: s/and/or/

README.md
```diff
@@ -44,46 +44,23 @@ hosts and containers in the deployment using the related plays mentioned
...
-* `kibana`
-* `logstash`
-* `rpc_maas`
-* `rpc_support`
+# Basic Setup:
 
+1. From `/opt`, clone [rpc-extras](https://github.com/rcbops/rpc-extras)
+with the `--recursive` option to get all the submodules.
+2. Unless doing an AIO build, prepare the os-ansible-deployment configuration.
+  1. copy everything from `os-ansible-deployment/etc/openstack_deploy`
```
odyssey4me: recursively copy

README.md
```diff
@@ -44,46 +44,23 @@ hosts and containers in the deployment using the related plays mentioned
...
-* `rpc_maas`
-* `rpc_support`
+# Basic Setup:
 
+1. From `/opt`, clone [rpc-extras](https://github.com/rcbops/rpc-extras)
+with the `--recursive` option to get all the submodules.
+2. Unless doing an AIO build, prepare the os-ansible-deployment configuration.
+  1. copy everything from `os-ansible-deployment/etc/openstack_deploy` 
+     into `/etc/openstack_deploy`
+  2. copy everything from `rpcd/etc/openstack_deploy` into
```
odyssey4me: recursively copy

rpcd/etc/openstack_deploy/user_extras_variables.yml
```diff
@@ -81,4 +81,4 @@ maas_filesystem_critical_threshold: 90.0
 
 # For an AIO it's recommended to set the following to limit the expected RAM
 # usage of elasticsearch.
-# elasticsearch_heap_size_mb: 1024
+elasticsearch_heap_size_mb: 1024
```
odyssey4me: This should only be set for an AIO - leave it commented generally.

rpcd/playbooks/ansible.cfg
```diff
@@ -12,19 +12,19 @@ timeout = 120
 
 # Set the path to the folder in os-ansible-deployment which holds the dynamic
 # inventory script - new config setting for ansible v1.9 and above
-inventory = /opt/os-ansible-deployment/playbooks/inventory/
+inventory = ../../os-ansible-deployment/playbooks/inventory/
```
odyssey4me: Why are we using ../.. here and /opt everywhere else? Rather use one convention and stick to it. I would recommend rather replacing '../..' with '/opt'

scripts/deploy.sh
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env bash
+
+set -e -u -x
+set -o pipefail
+source /opt/rpc-extras/os-ansible-deployment/scripts/scripts-library.sh
+
+export RPCD_LOGGING=${RPCD_LOGGING:-"FALSE"}
```
odyssey4me: To keep with the current conventions, s/FALSE/no/.

scripts/deploy.sh
```diff
@@ -0,0 +1,50 @@
...
+if [[ "${RPCD_AIO}" == "yes" ]]; then
+  if [[ ! -d /etc/openstack_deploy/ ]]; then
+    ./scripts/bootstrap-aio.sh
+    cp -R "${RPCD_DIR}"/etc/openstack_deploy/* /etc/openstack_deploy/
+  fi
+fi
+
+# bootstrap ansible only if not installed
+which openstack-ansible
+if [[ $? -ne 0 ]]; then
```
odyssey4me: could be simplified to 'if ${which openstack-ansible); then'

scripts/deploy.sh
```diff
@@ -0,0 +1,50 @@
...
+cd "${OSAD_DIR}"/playbooks/
+install_bits setup-hosts.yml
+if [[ "${RPCD_AIO}" == "yes" ]]; then
+  install_bits haproxy-install.yml
+fi
+install_bits setup-infrastructure.yml setup-openstack.yml
+
+# setup the rest
+cd "${RPCD_DIR}"/playbooks/
+install_bits horizon_extensions.yml rpc-support.yml
```
odyssey4me: The playbooks repo-build.yml repo-pip-setup.yml must run before anything else in RPC to ensure that all python wheels for RPC are built.

rpcd/etc/openstack_deploy/user_extras_variables.yml
git-harry: Maybe use scripts/deploy.sh to set this when RPCD_AIO=yes

scripts/deploy.sh
```diff
@@ -0,0 +1,50 @@
...
+  fi
+fi
+
+# bootstrap ansible only if not installed
+which openstack-ansible
+if [[ $? -ne 0 ]]; then
+  ./scripts/bootstrap-ansible.sh
+fi
+
+# only set up the passwords once
```
git-harry: What about putting the RPCD_LOGGING conditional stuff around this? Partly because it doesn't need to run if that's not equal to "yes" and partly to make the relationship.

scripts/deploy.sh
```diff
@@ -0,0 +1,50 @@
...
+set -o pipefail
+source /opt/rpc-extras/os-ansible-deployment/scripts/scripts-library.sh
+
+export RPCD_LOGGING=${RPCD_LOGGING:-"FALSE"}
+export RPCD_AIO=${RPCD_AIO:-"FALSE"}
+
+OSAD_DIR='/opt/rpc-extras/os-ansible-deployment'
+RPCD_DIR='/opt/rpc-extras/rpcd'
+
+# setup the things
```
git-harry: For the idempotent type bits that have been added for bootstrap-aio.sh and bootstrap-ansible.sh - would that not be better done upstream in those scripts instread of here?

scripts/deploy.sh
```diff
@@ -0,0 +1,50 @@
...
+# bootstrap ansible only if not installed
+which openstack-ansible
+if [[ $? -ne 0 ]]; then
+  ./scripts/bootstrap-ansible.sh
+fi
+
+# only set up the passwords once
+grep -E '^kibana_password:$' /etc/openstack_deploy/user_extras_secrets.yml
+if [[ $? -ne 0 ]]; then
+  ./scripts/pw-token-gen.py --file /etc/openstack_deploy/user_extras_secrets.yml
```
git-harry: Is that right? Does pw-token-gen.py only ever need running if kibana_password doesn't exist?

rpcd/etc/openstack_deploy/user_apt_packages.yml
```diff
@@ -0,0 +1,11 @@
+# This file will contain any apt packages that the deployer would like to pin to a particular version.
```
git-harry: I don't see the justification for breaking this out into a separate file. Unless this is part of some more generalised approach to separate user_extra_variables into multiple files it seems a bit random just to do it for this one var.

scripts/deploy.sh
```diff
@@ -0,0 +1,50 @@
...
+if [[ "${RPCD_AIO}" == "yes" ]]; then
+  install_bits haproxy-install.yml
+fi
+install_bits setup-infrastructure.yml setup-openstack.yml
+
+# setup the rest
+cd "${RPCD_DIR}"/playbooks/
+install_bits horizon_extensions.yml rpc-support.yml
+# maas doesn't work with aio directly
+if [[ "${RPCD_AIO}" != "yes" ]]; then
```
git-harry: What doesn't work by having maas as part of the aio?

rpcd/etc/openstack_deploy/user_apt_packages.yml
odyssey4me: :+1: Putting these into the user_variables file will likely give it more exposure. Putting it into a separate file is likely to leave it forgotten.

rpcd/etc/openstack_deploy/user_extras_variables.yml
odyssey4me: :+1:

scripts/deploy.sh
odyssey4me: pw-token-gen can be run against a file and it won't change anything already set, so there's no need to put this inside a conditional at all.

scripts/deploy.sh
mattt416: `which openstack-ansible || ./scripts/bootstrap-ansible.sh`

My AIO install just bailed prematurely as openstack-ansible isn't there and we have a `set -e` set.

scripts/deploy.sh
prometheanfire: It's specifically related to running an aio deploy on a cloud instance.  Because of how rackspace sets up our public instances, MaaS gets prepopulated with data about the instance.  This can conflict with what we try to do in the play.

scripts/deploy.sh
```diff
@@ -0,0 +1,50 @@
...
+  install_bits setup-maas.yml
+fi
+if [[ "${RPCD_LOGGING}" == "yes" ]]; then
+  install_bits setup-logging.yml
```
odyssey4me: After the logging playbook on an AIO, the `haproxy.yml` playbook must be run to configure the ELK LB's.

scripts/deploy.sh
odyssey4me: :+1:

rpcd/playbooks/ansible.cfg
mancdaz: maybe discover the absolute path of the basedir and have everything relative to that. Then there is no hard set requirement to clone it to any specific place, then you can pushd and popd around as you feel like. eg:

```
DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && cd ../ && pwd )
OSAD_DIR="${DIR}/os-ansible-deployment'"
RPCD_DIR="${DIR}/rpcd"
```

etc

scripts/deploy.sh
git-harry: Are you making the assumption that if it's an AIO it's on public cloud and if it's not it's on bare metal? Or are you saying MaaS will deploy if broken out onto multiple public cloud instances?

README.md
stevelle: done

README.md
stevelle: done

README.md
stevelle: done

rpcd/etc/openstack_deploy/user_apt_packages.yml
mattt416: openstack-ansible bails for me when this file is empty (as it is by default), not sure if this is an ansible bug or what?

rpcd/playbooks/ansible.cfg
stevelle: Pointing at RPCD_DIR instead of ../../

scripts/deploy.sh
stevelle: done

scripts/deploy.sh
stevelle: done

scripts/deploy.sh
mattt416: this needs to be run irrespective of kibana_password.

scripts/deploy.sh
stevelle: done

scripts/deploy.sh
stevelle: done

scripts/deploy.sh
stevelle: fixed by going with the simplified model

scripts/deploy.sh
stevelle: Fine idea to make bootstrap scripts idempotent but deferring this for now.
Those scripts are not idempotent, and these bits are insufficient to make them so.

scripts/deploy.sh
stevelle: for setting secrets, see below for change made which goes the other direction since generating secrets is idempotent.

rpcd/etc/openstack_deploy/user_apt_packages.yml
stevelle: I'd like to defer this also. I didn't make the decision to put this in it's own file in the first place and I don't want to argue the thinking behind that decision.

prometheanfire: is this even a file we touched? don't think this is a bug related to this pr

mattt416: It's created by this PR, see https://github.com/stevelle/rpc-extras/commit/e6eb0d1bc09d2d18f63ad4bcd7e95af42ca8ef9e

rpcd/etc/openstack_deploy/user_extras_variables.yml
stevelle: done

scripts/deploy.sh
odyssey4me: :+1:

rpcd/playbooks/ansible.cfg
odyssey4me: Why can't it reference '/opt/rpc-extras/os-ansible-deployment/playbooks/inventory/' ?

stevelle:
reworking the README somewhat to integrate the actual commands to be used, per comment by @git-harry

rpcd/playbooks/ansible.cfg
stevelle: It can as an absolute path, as I tried to say.  Not seeing any other good simple solutions coming along so going to just do it that way.

rpcd/etc/openstack_deploy/user_apt_packages.yml
stevelle: Rather than deal with unknown (ansible?) bugs giving an inconsistent experience (I'm not recreating this, and neither @nrb nor @prometheanfire seem to have encoutered it before) I'm going to just fold it into the user_extra_variables for right now and move on.

odyssey4me:
Looks good to me, but I haven't tested it yet.

prometheanfire:
:+1:

rpcd/etc/openstack_deploy/user_apt_packages.yml
nrb: It was a file I created to call out the functionality specifically.  You're right that the file being split out for one variable is probably a bit much. I dislike the idea of shoving everything into user_variables, especially when running from 'outside' osad, but in this case I'm not concerned enough to fight for it to stay separated.

And no, I did not encounter errors with this, though I don't remember if I ran with this file present.

nrb:
:+1:

stevelle:
who's got two thumbs and is merging?

claco:
^ That guy!

Issue #133, opened 2015-06-02
rabbitmq_status.py issues


mattt416:
I've started to see this type of failure on newly built clusters:

```
root@jenk-heat-158-node1:/usr/lib/rackspace-monitoring-agent/plugins# ./rabbitmq_status.py -H 172.29.236.126 -n jenk-heat-158-node1_rabbit_mq_container-fd99ca31
status error Detected RabbitMQ connections with multiple channels. Please check RabbitMQ and all Openstack consumers
root@jenk-heat-158-node1:/usr/lib/rackspace-monitoring-agent/plugins#
```

It looks like https://github.com/rcbops/rpc-maas/pull/139 introduced this change -- however the issue seems to involve us alarming when `channel['number'] > 1`, but `number` here is actually the channel number which doesn't make sense to alarm on (assuming I'm reading this correctly).

I believe what @cfarquhar requested was not implemented -- we should be checking /api/connections and not /api/channels, and perhaps we can alarm if channels > channel_max?  Some research will need to be done here as I'm not overly clear on the specifics of rabbitmq channels.

Mentioning @BjoernT for input.

--Matt

PR #134, opened 2015-06-02
Create rabbitmq_max_channels_per_conn metric

mattt416:
Currently, we alarm if connection['channels'] > 1.  This change creates
a new metric called rabbitmq_max_channels_per_conn which we can alarm
against using a MaaS alarm.  Additionally, this change updates
CONNECTIONS_URL so we only return the data from /api/connections that
we need (channels).

TODO: We will need to create an alarm on rpc-extras which ideally
      alarms if rabbitmq_max_channels_per_conn > 10.  Running a
      dev cluster periodically sees rabbitmq_max_channels_per_conn > 1
      and 10 is an 'arbitrary but likely reasonable limit' which we can
      use to prevent channel ballooning.

Closes issue #133

### DIFF
--- a/maas/rabbitmq_status.py
+++ b/maas/rabbitmq_status.py
@@ -24,7 +24,7 @@ from maas_common import (metric, metric_bool, status_ok, status_err,
 
 OVERVIEW_URL = "http://%s:%s/api/overview"
 NODES_URL = "http://%s:%s/api/nodes"
-CONNECTIONS_URL = "http://%s:%s/api/connections"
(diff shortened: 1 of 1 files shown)


mattt416:
This issue / PR were pulled over from https://github.com/rcbops/rpc-maas/pull/206, which had multiple :+1: comments.

mancdaz:
Based on on the approvals in https://github.com/rcbops/rpc-maas/pull/206, I am merging this in

PR #139, opened 2015-06-02
Disable warnings conditionally

mattt416:
requests.packages.urllib3.disable_warnings is only available on requests
2.4.0 or later. Right now, we package have a lower limit on 2.2.0 and
all versions before 2.4.0 will cause an AttributeError when calling
this.

(cherry picked from commit 4cb0513b1122d20f074f1e985227d8e373898e28)

### DIFF
--- a/maas/horizon_check.py
+++ b/maas/horizon_check.py
@@ -26,7 +26,8 @@ from lxml import html
 
 def check(args):
     # disable warning for insecure cert on horizon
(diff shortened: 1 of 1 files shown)


mattt416:
This is pulled over from open PR https://github.com/rcbops/rpc-maas/pull/210, since commits should no longer be going to rpc-maas.

stevelle:
:+1:

PR #140, opened 2015-06-02
Add 10.x to 11.x upgrade scripts.

nrb:
These scripts will allow for upgrading an existing 10.x RPC installation
to an 11.x-based one.

This is a squashed version of #131, with some changes from that PR integrated. I did this in order to get the process going without having to wake up @stevelle, not intending to snipe.


odyssey4me:
:+1: untested, but looks pretty straightforward

prometheanfire:
:+1: I tested this yesterday

Issue #144, opened 2015-06-03
ipaddr failing to install


mattt416:
Installing maas fails as it cannot install ipaddr python package.

PR #145, opened 2015-06-03
[WIP] Use virtualenv for maas plugins

mattt416:
This change implements the following two changes:

- maas plugins now use a .venv so there are no conflicts with packages
  installed on the hosts
- creates proper pip package variable names in rpc_maas role so that
  repo-builder.yml will include these packages when building repo
  packages

Closes issue #144


prometheanfire:
why can repo-builder not build it?

cloudnull:
@prometheanfire - the repo builder looks for lists with "*_pip_packages:" which is why the original package list was skipped.

@mattt416 - :+1: LGTM

mattt416:
Hold off on merging -- going to see if I can use a venv here.

mattt416:
@git-harry pointed out that this doesn't work, need to rethink or revert to original commit (not involving virtualenvs).

mattt416:
Actually this does appear to be working, but needs more testing.  Appreciate any feedback.

prometheanfire:
:+1: lgtm

PR #146, opened 2015-06-03

mattt416:
role itself.

Closes issue #144

(cherry picked from commit e2e7f7ea145bde4dfe14dd099de95004ad12d3aa)


mattt416:
Please hold off on merging this for a sec.

mattt416:
Closing, reworking this PR.

PR #147, opened 2015-06-03
Maas entity creation and no swift support

prometheanfire:


Apsu:
+1 LGTM

stevelle:
:+1:

claco:
@prometheanfire @stevelle Bug in Kilo? Needs backport?

prometheanfire:
yes, needs backport

mattt416:
Hi @prometheanfire ,

We made the conscious decision when putting these plays together not to create missing entities as these should be created at server provisioning and missing entities could mean other provisioning-related errors.  We felt it was best to bail and have the operator investigate.  Also, if you misconfigure some variables in ansible you could end up with a slew of newly created entities which are unneeded.

Additionally, commit https://github.com/rcbops/rpc-extras/commit/096b0a2a2bf679e326bbc9b75933c901ca3241c5 breaks my run (see https://github.com/rcbops/rpc-extras/issues/149).  It looks like the entity_count var gets reregistered even when the task is skipped resulting in there being no stdout field.  In this case it would probably be best to just re-issue the count task again with a `when` condition.

Personally I think we should revert this change -- however am open to hearing what others think. @git-harry, what are your thoughts?

--Matt

git-harry:
@mattt416 - I agree that was the reason we did not create the entities. They should alway exist and if they don't this is not the place to be creating them.

stevelle:
I would like to request adding comments explaining the above along with who creates them, when, and how to the ansible plays and roles. Nobody wants to be in a position where we break stuff or are afraid to work on it because of tribal knowledge.

Sadly, the model sounds like a tedious and error-prone user experience, and that was specifically what we were tasked with cleaning up. If the only concern with creating entities automatically is "risk of creating a bunch of extra stuff" in MaaS than I'd rather design a solution to that issue.

A fix for the entity_count being reregistered however doesn't need to wait

Issue #149, opened 2015-06-04
setup-maas.yml failing 


mattt416:
Commit https://github.com/rcbops/rpc-extras/commit/096b0a2a2bf679e326bbc9b75933c901ca3241c5 seems to have broken setup-maas.yml from running correctly.  I'm seeing:

```
TASK: [rpc_maas | At most one {{ inventory_hostname }}{{ maas_fqdn_extension }} entity exists] ***
fatal: [jenk-heat-297-node4] => error while evaluating conditional: entity_count.stdout|int != 1
fatal: [jenk-heat-297-node5] => error while evaluating conditional: entity_count.stdout|int != 1
fatal: [jenk-heat-297-node1] => error while evaluating conditional: entity_count.stdout|int != 1
fatal: [jenk-heat-297-node2] => error while evaluating conditional: entity_count.stdout|int != 1
```

prometheanfire:
I'm fine with reverting it, should be doc'd so testing knows they need to do entity creating at host provisioning.

mattt416:
I'm also seeing errors about galera_root_user var not existing, investigating now.

prometheanfire:
not sure I touched that...

mattt416:
Yep, identified cause, not related to your commit!

mattt416:
This issue should not affect kilo branch.

mattt416:
Only https://github.com/rcbops/rpc-extras/pull/151 needs backporting to kilo.

PR #151, opened 2015-06-04
Set defaults for galera vars

mattt416:
Commit eb95036a7ea0e98fc931dd19f416015c66f3620c in os-a-d relocated a
number of variables in playbooks/inventory/group_vars/all.yml to
etc/openstack_deploy/user_group_vars.yml.  However, in the process some
vars (namely galera_root_user) were not moved and this breaks execution
of the rpc_maas role as we include
playbooks/inventory/group_vars/all.yml in set-maas.yml playbook.

This change sets defaults for galera_root_user and removes inclusion of
playbooks/inventory/group_vars/all.yml since it is no longer required.

We will eventually want to use galera_client role here instead, however
the immediate issue identified is that already deployed hosts will run
into apt issues as they try to go from ubuntu mariadb packages to those
packaged by mariadb themself.

Closes issue #149

### DIFF
--- a/rpcd/playbooks/roles/rpc_maas/templates/client.my.cnf.j2
+++ b/rpcd/playbooks/roles/rpc_maas/templates/client.my.cnf.j2
@@ -2,5 +2,5 @@
 
 [client]
 host={{ container_address }}
-user={{ galera_root_user }}
+user={{ galera_root_user|default('root') }}
 password={{ galera_root_password }}
--- a/rpcd/playbooks/setup-maas.yml
+++ b/rpcd/playbooks/setup-maas.yml
@@ -19,5 +19,4 @@
   roles:
     - { role: "rpc_maas", tags: [ "maas-setup" ] }
   vars_files:
-    - "{{ rpc_repo_path }}/playbooks/inventory/group_vars/all.yml"
     - "{{ rpc_repo_path }}/playbooks/roles/os_cinder/defaults/main.yml"


andymcc:
LGTM - if somebody else reviews and approves lets get this merged.

rpcd/playbooks/roles/rpc_maas/templates/client.my.cnf.j2
```diff
@@ -2,5 +2,5 @@
 
 [client]
 host={{ container_address }}
-user={{ galera_root_user }}
-password={{ galera_root_password }}
+user={{ galera_root_user|default('root') }}
+password={{ galera_root_password|default('secrete') }}
```
cloudnull: I don't like that this sets a password that would/could be used in a production environment. I think it would be better to simply fail because the value is unset than it is to provide the deployer the ability to go to production with an insecure password.

cloudnull:
:-1: One inline comment, regarding the default galera vars but otherwise seems fine.

rpcd/playbooks/roles/rpc_maas/templates/client.my.cnf.j2
mattt416: Hi @cloudnull, we're not setting any password though -- if for whatever reason galera_root_password doesn't exist in the environment (a problem in itself) then all that will happen is we will end up with an unusable .my.cnf.

cloudnull: Not that I have a dog in this fight, however I think it would be better for ansible to complain about an unset variable than it would be to assume it can be set to "secrete".

mattt416:
Fixed @cloudnull, thanks!

andymcc:
LGTM

cloudnull:
LGTM - :+1:

PR #152, opened 2015-06-04
Revert "create entities if missing"

mattt416:
This change reverts 096b0a2 as we originally made a conscious decision to not create entities via rpc_maas.  These entities should be created at provisioning time and a lack of existence could indicate a problem with host provisioning.  Also, the rpc_maas role fails to run with this commit in place.   It looks like the entity_count var gets re-registered even when the task is skipped resulting in there being no stdout field. In this case it would probably be best to just re-issue the count task again without a when condition.

This reverts commit 096b0a2a2bf679e326bbc9b75933c901ca3241c5.

Closes issue #149

### DIFF
--- a/rpcd/playbooks/roles/rpc_maas/tasks/raxmon_agent_install.yml
+++ b/rpcd/playbooks/roles/rpc_maas/tasks/raxmon_agent_install.yml
@@ -42,15 +42,6 @@
   shell: "raxmon-entities-list | grep ' label={{ entity_name }} provider=Rackspace Monitoring ...>$' | wc -l"
   register: entity_count
 
-- name: Create missing entities
-  shell: "raxmon-entities-create --label={{ entity_name }}"
-  when: entity_count.stdout|int == 0
-
-- name: Entity {{ entity_name }} count
-  shell: "raxmon-entities-list | grep ' label={{ entity_name }} provider=Rackspace Monitoring ...>$' | wc -l"
-  register: entity_count
-  when: entity_count.stdout|int == 0
-
 - name: At most one {{ entity_name }} entity exists
   fail:
     msg: "Entity count of {{ entity_count.stdout }} != 1 for entity with the label {{ entity_name }}"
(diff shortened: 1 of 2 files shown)


stevelle:
Per comments in #147 I can be :+1: on this revert but created issue #155 to track the need to capture the reasoning for those decisions for future clarity.

mattt416:
Fair comments @stevelle -- thanks for raising.  I'd be more than happy to revisit if we want to create these missing entities but in light of the PR causing additional breakage I figured we could revisit this once rpc-extras is stable.  I'd be happy to get w/ you, @prometheanfire, and @git-harry to discuss.

prometheanfire:
we do need the allow for no swift and maas commit though, that's separate.

other then that, sure :D

git-harry:
@prometheanfire - based on the commit message, I think the purpose of this is to prevent the swift monitoring from being setup if swift is not deployed. The commit itself only appears to be preventing the remote checks/alarms from being created. If that is all correct I think it would be more appropriate for the guard to be in the main.yml task file.

stevelle:
Looking at it I can agree with @git-harry about moving the when to main.yml and will be an easy thumbs up on that patch.

Relatedly I find the roles/rpc_maas/tasks/main.yml file to be a little messy, currently. I would like to see the host tasks pulled out into their own include so I created a PR for this ( #168 ). It is not required for any of this work, but I will offer it as a quick refactor.

prometheanfire:
@git-harry that commit is needed for mass to run at all in a non-swift env.  If you have another fix to use instead I'm all ears, but don't revert that fix please (there's a reason it was in a separate commit).

stevelle:
treating this PR as now depending on #169 thx @git-harry

mattt416:
@prometheanfire I've removed the revert of 13ddcb5 as I think commit https://github.com/rcbops/rpc-openstack/pull/169/files is going to introduce initial issues. Once this PR goes through I'll look at https://github.com/rcbops/rpc-openstack/pull/169/files to see how we can ensure that `swift.yml` is only applied when we know swift is in the environment.

git-harry:
+1

mattt416:
Looks like commit 13ddcb5 doesn't even work (`defined x` should be `x is defined`), but rather than re-remove the commit from this PR I've created https://github.com/rcbops/rpc-openstack/pull/173.

andymcc:
+1 Merging.

PR #154, opened 2015-06-04

mattt416:
role itself.

Closes issue #144


prometheanfire:
:+1:

stevelle:
:+1:

PR #156, opened 2015-06-04

mattt416:
role itself.

Closes issue #144

(cherry picked from commit 2c9ca7512d2bfe318e55fa292af6d9a1abd96b63)


odyssey4me:
:+1:

cloudnull:
:+1:

PR #157, opened 2015-06-05
Deploy & Upgrade adjustments

odyssey4me:
This PR includes documentation and script fixes which adjust the deployment and upgrade processes. The upgrade process is now set to re-use the deployment script, and the deployment script will no longer create the ELK stack containers if they are not going to be used. The documentation also now includes an extra step to do the same for non-AIO builds.

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -54,13 +54,13 @@ if [[ "${DEPLOY_OSAD}" == "yes" ]]; then
     rm -f /etc/openstack_deploy/env.d/{elasticsearch,logstash,kibana}.yml
   fi
 
-  # setup the hosts and build the basic containers
-  install_bits setup-hosts.yml
-
   # setup the haproxy load balancer
   if [[ "${DEPLOY_HAPROXY}" == "yes" ]]; then
     install_bits haproxy-install.yml
   fi
+
+  # setup the hosts and build the basic containers
+  install_bits setup-hosts.yml
 
   # setup the infrastructure
   install_bits setup-infrastructure.yml
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -39,6 +39,12 @@ which openstack-ansible || ./scripts/bootstrap-ansible.sh
 # begin the openstack installation
 cd "${OSAD_DIR}"/playbooks/
 
+# ensure that the ELK containers aren't created if they're not
+# going to be used
+if [[ "${DEPLOY_ELK}" != "yes" ]]; then
+  rm -f /etc/openstack_deploy/env.d/{elasticsearch,logstash,kibana}.yml
+fi
+
 # setup the hosts and build the basic containers
 install_bits setup-hosts.yml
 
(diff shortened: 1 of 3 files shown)


mattt416:
:+1:

andymcc:
LGTM, @mattt416 can you confirm the 2 commits that came in after your previous comment are OK and then we can merge this?

scripts/deploy.sh
```diff
@@ -14,7 +15,7 @@ OSAD_DIR='/opt/rpc-extras/os-ansible-deployment'
 RPCD_DIR='/opt/rpc-extras/rpcd'
 
 # begin the bootstrap process
-cd "${OSAD_DIR}"
+cd ${OSAD_DIR}
```
prometheanfire: Why are we getting rid of quotes? they are important.

scripts/deploy.sh
```diff
@@ -37,24 +45,32 @@ which openstack-ansible || ./scripts/bootstrap-ansible.sh
...
-# setup the haproxy load balancer
-if [[ "${DEPLOY_HAPROXY}" == "yes" ]]; then
-  install_bits haproxy-install.yml
-fi
+  # setup the hosts and build the basic containers
+  install_bits setup-hosts.yml
+
+  # setup the haproxy load balancer
+  if [[ "${DEPLOY_HAPROXY}" == "yes" ]]; then
+    install_bits haproxy-install.yml
```
prometheanfire: haproxy can/should be run before setup-hosts (it's a totally stand alone play

scripts/deploy.sh
andymcc: In this case it doesn't really change anything - although the change itself seems unnecessary.

scripts/deploy.sh
andymcc: This doesn't change anything either - but the change seems unnecessary. Perhaps @odyssey4me has a reason for changing the ordering of haproxy?

In a chronological sense, you only need haproxy setup before running infra/os plays so that the db/keystone etc will work - so it does make sense to have it after host setup in some ways - there is no need for haproxy before the host setup.

prometheanfire: ya, probably will be fine, but running haproxy first is guaranteed to be fine (it's the order that osad upstream says to use, or at least kevin).

scripts/deploy.sh
prometheanfire: ya, looking at it, the quotes were only removed for the statically set local variables, so we are fine here

scripts/deploy.sh
andymcc: I'm against making changes for no reason, on principle - so if there's no reason I'd probably be against it. Although I can see that it would work.

scripts/deploy.sh
odyssey4me: they were removed as they add no real value, and it felt more consistent to me to have them used in this way (this is how it's done in the OSAD scripts)

scripts/deploy.sh
odyssey4me: my bad - I didn't realise that I changed the order and will revert that

prometheanfire:
:+1:

prometheanfire:
backport?

PR #164, opened 2015-06-08
[kilo] Deploy & Upgrade adjustments

odyssey4me:
This PR includes documentation and script fixes which adjust the deployment and upgrade processes. The upgrade process is now set to re-use the deployment script, and the deployment script will no longer create the ELK stack containers if they are not going to be used. The documentation also now includes an extra step to do the same for non-AIO builds.

This PR is a backport of https://github.com/rcbops/rpc-openstack/pull/157 to the kilo branch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # rpc-openstack
 Optional add-ons for Rackspace Private Cloud
 
-# os-ansible-deploy integration
+# os-ansible-deployment integration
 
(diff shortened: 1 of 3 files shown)


andymcc:
LGTM
NB incase anybody missed - its a backport - if another is happy with this feel free to merge!

mattt416:
:+1:

PR #169, opened 2015-06-08
do the checking for swift in the main rpc_mass task

prometheanfire:


stevelle:
filed #170 to leave a reminder for future possible cleanups.

:+1:

rpcd/playbooks/roles/rpc_maas/tasks/main.yml
```diff
@@ -60,6 +60,9 @@
 - include: swift.yml
   vars:
     external_vip_address: "{{ external_lb_vip_address }}"
+  when:
+    - defined groups['swift_proxy'][0]
+    - inventory_hostname == groups['swift_proxy'][0]
```
mattt416: This doesn't look right, but I'm not overly familiar w/ ansible.  I'd have expected this to be a simple `when: a and b`.  Can someone confirm?

mattt416:
I don't believe this is right as we'd be limiting the tasks in swift.yml from only being applied to node `groups['swift_proxy'][0]`.

git-harry:
@mattt416 I agree, that doesn't look like it will work. A condition of just "defined groups['swift_proxy'][0]" would probably do the trick but I'm not sure if that's the best way to confirm whether or not swift has been deployed.

I think the commit message is a bit lacking. I think a commit message should at the least contain details of what the current code does, why it needs changing and what is being changed.

mattt416:
Also, as per https://github.com/rcbops/rpc-openstack/pull/173, the syntax is `x is defined` and not `defined x`.  The latter causes ansible to puke with something along the lines of:

```
fatal: [jenk-heat-313-node5] => template error while templating string: expected token 'end of statement block', got 'groups'
```

mattt416:
I've looked at this a bit further, and adding a guard for swift in main.yml is actually pointless as each section in swift.yml is already guarded.  One would think that adding the guard would only apply swift.yml if swift is installed, but on a cluster w/ no swift deployed and `when: groups['swift_proxy']|length > 0` I still see all these tasks applied but skipped across the lot.

From http://docs.ansible.com/playbooks_conditionals.html#applying-when-to-roles-and-includes:

```
You will note a lot of ‘skipped’ output by default in Ansible when using this approach on systems that don’t match the criteria. Read up on the ‘group_by’ module in the About Modules docs for a more streamlined way to accomplish the same thing.
```

We can look at `group_by`, however this commit doesn't add anything further to what we are already doing.

git-harry:
Ah, apologies allI didn't appreciate it worked liked that. Well if that's the case it seems that https://github.com/rcbops/rpc-openstack/pull/173/files should be enough to address this.

prometheanfire:
:+1:

PR #171, opened 2015-06-09
allow for mass when not setting up swift

mattt416:
(cherry picked from commit 13ddcb59e4264965f632ff9a1ed6cc0e727af718)


mattt416:
Abandoning, this commit doesn't work.

Issue #172, opened 2015-06-09
Error in swift.yml


mattt416:
The following syntax is incorrect:

```
when:
  - defined groups['swift_proxy'][0]
  - inventory_hostname == groups['swift_proxy'][0]
```

It should read:

```
when:
  - groups['swift_proxy'][0] is defined
  - inventory_hostname == groups['swift_proxy'][0]
```

PR #173, opened 2015-06-09
Fix typo in swift.yml

mattt416:
The when statement added by 13ddcb5 to swift.yml is incorrect and
causes an ansible run to fail.

Closes issue #172


git-harry:
+1

prometheanfire:
:+1:

PR #176, opened 2015-06-09
RPC branding for Horizon

miguelgrinberg:
Backported from kilo release.


mattt416:
Is anyone deploying juno branch of rpc-openstack?  If so this one is a :+1: from me.

miguelgrinberg:
@mattt416 Not sure if it is going to be used, but they asked for it just in case (there's a Trello card in this sprint).

odyssey4me:
In that case :+1:

Issue #177, opened 2015-06-10
novaclient no longer supports client version 3


mattt416:
```
root@jenk-heat-316-node1:/usr/lib/rackspace-monitoring-agent/plugins# ./nova_api_local_check.py 172.29.237.104
status error Invalid client version '3'. must be one of:
root@jenk-heat-316-node1:/usr/lib/rackspace-monitoring-agent/plugins#
```

Flipping to client version 2 seems to work fine.

On a side note, the incomplete message returned by novaclient is being tracked here:

https://bugs.launchpad.net/python-novaclient/+bug/1463809

PR #178, opened 2015-06-10
Use novaclient version 2 instead of 3

mattt416:
The novaclient library has removed client version 3 (see
https://review.openstack.org/#/c/169378/).  This commit updates
get_nova_client() method in maas_common.py to use version 2 instead.

Closes issue #177


nrb:
:+1:

PR #179, opened 2015-06-10
Use novaclient version 2 instead of 3

mattt416:
The novaclient library has removed client version 3 (see
https://review.openstack.org/#/c/169378/).  This commit updates
get_nova_client() method in maas_common.py to use version 2 instead.

Closes issue #177

(cherry picked from commit f06a486e73c7a6d0a3061e0941212206e5e006f5)

### DIFF
--- a/maas/maas_common.py
+++ b/maas/maas_common.py
@@ -138,7 +138,7 @@ else:
                 'compute',
                 auth_ref['serviceCatalog'])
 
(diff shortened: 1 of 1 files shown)


mattt416:
Note that this isn't currently broken in the version of python-novaclient that we use in os-a-d kilo, but we may as well patch it to avoid hitting it somewhere down the line.

nrb:
:+1:

PR #185, opened 2015-06-12
Fix status error output

git-harry:
There is currently a print statement in the rabbitmq_status.py script
that prints out a line starting with 'status err '. This is a special
type of status line that causes a metric called legacy_state to be
created. This metric should not be created. The status_err function
should be used instead of using print directly.

This commit updates the line to use status_err. This will prevent the
metric legacy_state from being created. It will also cause the script
to exit with a return code of 1.


odyssey4me:
:+1:

prometheanfire:
:+1:

git-harry:
juno backport - https://github.com/rcbops/rpc-maas/pull/221

PR #187, opened 2015-06-23
Add user_variable.yml file to rpcd

git-harry:
This pull request is in response to https://bugs.launchpad.net/openstack-ansible/+bug/1457140

A commit is in os-ansible-deployment (master) to allow the requested variable to be set in the template.

This repo is for deploying an opinionated version of OpenStack using
os-ansible-deployment. There are instances where the defaults that a
Rackspace deployment will use differ from the upstream deployment
project. This commit creates a user_variables.yml file to use for
specifying these defaults and so this file should be used instead of
the one in os-ansible-deployment.

The user_variables.yml file in os-ansible-deployment currently sets two
variables that are not specified by default, by the associated roles.
These variables, ssl_protocol and ssl_cipher_suite, are specified in
this new file.

Support have requested that the nova configuration option
[cinder]/cross_az_attach be set to False and so is set here to override
the upstream default of True.

### DIFF
--- a/rpcd/etc/openstack_deploy/user_variables.yml
+++ b/rpcd/etc/openstack_deploy/user_variables.yml
@@ -1,0 +1,24 @@
+---
+# Copyright 2015, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


odyssey4me:
:+1:

nrb:
Has this been tested with an AIO? https://github.com/rcbops/rpc-openstack/blob/master/scripts/deploy.sh#L28 leads me to believe that the deploy script would overwrite what's already in place, which would be generated by osad's bootstrap-aio.sh command.

nrb:
Also, is there an issue associated with this? I couldn't find one, and would like to have one for release notes.

prometheanfire:
seems alright, but would like to know more about how tested this is.

git-harry:
@prometheanfire - there is no automatic testing of the AIO created by the deploy.sh script. You will need to test it manually as part of the review.

prometheanfire:
:+1:

PR #193, opened 2015-06-25
Add bash completion to openstack clients 

Alfano93:
Adds bash completion to: 
  
  Cinder
  Glance
  Heat
  Keystone
  Neutron
  Nova

All needed packages for bash completion to work have been added to the rpc_support playbooks. 

Closes-Bug:#1465723

### DIFF
--- a/rpcd/playbooks/roles/rpc_support/defaults/main.yml
+++ b/rpcd/playbooks/roles/rpc_support/defaults/main.yml
@@ -15,6 +15,7 @@
 
 # Packages required for support
 support_util_packages:
+  - bash-completion
   - bridge-utils
   - debootstrap
   - dstat
--- a/rpcd/playbooks/roles/rpc_support/tasks/autocompletion.yml
+++ b/rpcd/playbooks/roles/rpc_support/tasks/autocompletion.yml
@@ -1,0 +1,25 @@
+# Copyright 2015, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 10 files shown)


claco:
Could probably just squash these into 1. But :+1: otherwise. Cool!

BjoernT:
Please install it on the infra nodes as well, as we run most commands from these hosts.
Additionally the neutron completion should be installed inside the neutron agents container

nrb:
@BjoernT Is there any reason we shouldn't just remove the `when:` line on the include and install these everywhere?

odyssey4me:
I'm not fond of the idea of carrying the templates, forcing us to update them whenever a project releases an update. Are these files not already in the source we checkout, or located in the installed wheel? If there's a way that we can pull them from a known and consistent location then we should ideally try to do so.

BjoernT:
@odyssey4me Those are generic bash scripts and not likely going to change from Juno to Kilo etc, in fact not at all. The only upcoming change is once the openstack client takes over and all the other python clients die. I don't see that happening in the near future.

@nrb Not really but I prefer to prevent clutter when possible and not installing it everywhere.

prometheanfire:
generally in favour of this, but we need an issue to attach it to and to possibly track kilo backports on

nrb:
+1 to prometheanfire's comment on an issue. I think this originated from a LaunchPad request.

It's also likely this will get backported to Juno.

Alfano93:
Would it be safe to close this request, and resubmit after some more TLC has been given to this?

prometheanfire:
I think it's fine, if as was said this will not change much if at all.

we just need a tracking issue

git-harry:
@Alfano93 I'd like to look at getting this merged in. To do that the pull request could do with some attention. I'd like you to take a look at the following points:
- create an issue in the repo and reference this pull request against it
- rebase the commits
- squash the commits down to one commit
- remove the reference to the launchpad bug from the commit message
- all the templates don't look like they have any jinja code in them, if that's the case I think it would be better to use the copy module.

git-harry:
Replaced by https://github.com/rcbops/rpc-openstack/pull/470

Issue #195, opened 2015-06-26
how to determine if a wheel contains a commit


mancdaz:
An ask from support was "how do we know if a wheel that is installed inside a container contains a particular commit?", in order to know if a customer's environment contains a particular fix or not.  This is somewhat non-trivial because of the way we now build wheels outside the containers, and the wheel names do not directly indicate the SHA from which they were built (rather just the semver version)

prometheanfire:
@mancdaz backport to kilo or want me to?

prometheanfire:
merged into kilo, closing

PR #196, opened 2015-06-26
add script to check if a given commit is contained in a built wheel

mancdaz:
It's somewhat difficult to determine if a deployed wheel of an
openstack project (say neutron) contains a particular upstream commit,
because of the way we build wheels outside of the containers.

This commit adds a script to be run from the deployment host that does
the following things:

1) connects to all containers of type 'project' and determines the sha
from which the wheel was built
2) clones git repo locally (/tmp by default) to allow interrogation of
git history
3) checks to see if 'supplied commit sha' (the commit that you are
interested in) is contained in the wheel.

optional - supply a parent sha and use that instead of getting the
parent sha from the wheel in the containers.

Closes: #195

### DIFF
--- a/scripts/checksha.py
+++ b/scripts/checksha.py
@@ -1,0 +1,218 @@
+#!/usr/bin/env python
+# Copyright 2015, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# (c) 2015, Darren Birkett <darren.birkett@gmail.com>
+from __future__ import print_function
+
+import argparse
+import os
+import subprocess
+import sys
+
+from subprocess import CalledProcessError
+
+import ansible.inventory
+import ansible.runner
+
+
+def get_args():
+    """Get the arguments."""
+    parser = argparse.ArgumentParser(
+        prog='checksha',
+        description='Given a commit sha and an Openstack project name, '
+                    'find out if the wheel that is currently installed '
+                    'for that Openstack project in this deployment '
+                    'includes that commit.',
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+    parser.add_argument(
+        'project', type=str, default=None,
+        help='name of openstack project wheel to query')
+    parser.add_argument(
+        'commit', type=str, default=None,
+        help='SHA of commit that you are interested in')
+    parser.add_argument(
+        '-s', '--sha', type=str, default=None,
+        help='provide a parent sha rather than interrogating the wheels '
+             'that are deployed in the solution to get the sha')
+    parser.add_argument(
+        '-c', '--clonedir', type=str, default='/tmp',
+        help='where to clone the git repo for history interrogation')
+    parser.add_argument(
+        '-b', '--basedir', type=str,
+        default='/opt/rpc-openstack/os-ansible-deployment',
+        help='the base directory in which to recursively search for '
+             'the dynamic inventory file')
+    parser.add_argument(
+        '-i', '--inventory', type=str, default='dynamic_inventory.py',
+        help='the name of the dynamic inventory file to find'),
+    parser.add_argument(
+        '-u', '--cloneuri', type=str,
+        default='https://git.openstack.org/openstack',
+        help='base uri from which to clone the git repo. The project '
+             'name will be appended to make the full clone uri'),
+    parser.add_argument(
+        '-p', '--pattern', type=str, default=None,
+        help='host pattern to pass to the ansible command. By default '
+             '<PROJECT>_all will be used')
+
+    return parser.parse_args()
+
+
+def find_inventory(inv, path):
+    """Find the dynamic inventory file"""
+    if os.path.isfile(inv):
+        return inv
+    for root, dirs, files in os.walk(path):
+        if inv in files:
+            return os.path.join(root, inv)
+
+
+def clone_git_repo(project, clonedir, cloneuri):
+    """clone a git repo"""
+    if not os.path.exists('%s/%s' % (clonedir, project)):
+        print('cloning %s project to interrogate the history...'
+              'please be patient this may take a while' % project)
+        os.chdir(clonedir)
+        try:
+            subprocess.check_call(
+                'git clone %s/%s.git '
+                '--quiet' % (cloneuri, project),
+                shell=True, stderr=subprocess.STDOUT)
+        except CalledProcessError as e:
+            print('Something went wrong when trying to clone the repo')
+            sys.exit(e)
+    else:
+        print('%s/%s already exists...'
+              'fetching the latest commits' % (clonedir, project))
+        os.chdir('%s/%s' % (clonedir, project))
+        try:
+            subprocess.check_call(
+                'git fetch --all --quiet',
+                shell=True, stderr=subprocess.STDOUT)
+        except CalledProcessError as e:
+            print('Something went wrong when trying to update the repo')
+            sys.exit(e)
+
+
+def parse_results(results, project):
+    """parse results of the ansible run"""
+    shas = set()
+    output = dict()
+    for status in results.iterkeys():
+        if status == 'contacted':
+            for (hostname, result) in results['contacted'].iteritems():
+                if 'not installed' in result['stdout'] or not result['stdout']:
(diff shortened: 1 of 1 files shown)


scripts/checksha.py
```diff
@@ -0,0 +1,139 @@
...
+import ansible.inventory
+
+
+INVENTORY_FILENAME = 'dynamic_inventory.py'
+BASE_PATH = '/opt/stackforge'
+CLONE_DIR = '/tmp'
+
+
+def get_args():
+    '''Get the arguments.'''
```
prometheanfire: should be double quotes for docstrigs

scripts/checksha.py
```diff
@@ -0,0 +1,139 @@
...
+    except CalledProcessError as e:
+        if e.returncode == 1:
+            # not an ancestor
+            return False
+        else:
+            # command went wrong
+            sys.exit(e)
+
+
+def main(args):
```
prometheanfire: the args variable is shadowing the args variable you set below (very minor nit, but changing this name (or the one below) to something else would be nice.

scripts/checksha.py
```diff
@@ -0,0 +1,139 @@
...
+import subprocess
+import argparse
+from subprocess import CalledProcessError
+
+import ansible.runner
+import ansible.inventory
+
+
+INVENTORY_FILENAME = 'dynamic_inventory.py'
+BASE_PATH = '/opt/stackforge'
```
git-harry: stackforge?

scripts/checksha.py
```diff
@@ -0,0 +1,139 @@
...
+CLONE_DIR = '/tmp'
```
git-harry: Why not make the above three arguments with default values?

scripts/checksha.py
```diff
@@ -0,0 +1,139 @@
...
+            # command went wrong
+            sys.exit(e)
+
+
+def main(args):
+    project = args.project
+    commit = args.commit
+
+    # run pbr info on {project} on all hosts that might have it
+    dyn_inventory = find_inventory(INVENTORY_FILENAME, BASE_PATH)
```
git-harry: If find_inventory doesn't find the inventory file it will return None. It is probably worth failing the script at this point so that it is clear what is going on. I haven't tested the script yet but ansible.inventory.Inventory(None) returns and inventory object.

scripts/checksha.py
```diff
@@ -0,0 +1,139 @@
...
+    inventory = ansible.inventory.Inventory(dyn_inventory)
+    results = ansible.runner.Runner(
+        pattern='%s_all' % project,
```
git-harry: This seems a bit limiting (that it only works for openstack 'type:service' projects). If this is all that's required by support? I think the project help string in get_args should be more explicit if that is the case.

Alternatively, why to allow an ansible group and wheel to be supplied. That gives support the ability to check any wheel.

scripts/checksha.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+def parse_results(results, project):
+    '''parse and print results of the ansible run'''
+    shas = set()
+    hostnames = []
+    for (hostname, result) in results['contacted'].items():
+        shas.add(result['stdout'].split()[3])
+        hostnames.append(hostname)
+    # bail if multiple or no shas found
```
git-harry: Why exit if multiple shas? A heterogeneous deployment isn't necessary a bad thing. Why not just print a warning and carry on?

scripts/checksha.py
```diff
@@ -0,0 +1,139 @@
...
+    dyn_inventory = find_inventory(INVENTORY_FILENAME, BASE_PATH)
+    inventory = ansible.inventory.Inventory(dyn_inventory)
+    results = ansible.runner.Runner(
+        pattern='%s_all' % project,
+        inventory=inventory,
+        forks=10,
+        module_name='shell',
+        module_args='pbr info %s' % project).run()
+
+    if args.sha:
```
git-harry: If args.sha is supplied, the ansible stuff does not need to be run.

scripts/checksha.py
```diff
@@ -0,0 +1,139 @@
...
+
+def clone_git_repo(project, path=CLONE_DIR):
+    '''clone a git repo'''
+    os.chdir(path)
+    if not os.path.exists(project):
+        print 'cloning %s project to interrogate the history...' \
+              'please be patient this may take a while' % project
+        try:
+            subprocess.check_call(
+                'git clone https://github.com/openstack/%s.git '
```
git-harry: Should this be hardcoded? What about CDC?

scripts/checksha.py
mancdaz: it does only work on openstack service wheels since they are built with pbr. Non-pbr built wheels do not include git sha in pbr show/info output eg.:

```
# pbr freeze | egrep 'ipadd|nova'
nova==2015.1.1.dev30  # git sha 459f51e
python-novaclient==2.23.1.dev1  # git sha 0ae7a08
ipaddress==1.0.7
```
Accepted, it's quite specific in scope but this is the support ask.

scripts/checksha.py
mancdaz: Fixed, thanks

scripts/checksha.py
mancdaz: Fixed, thanks

scripts/checksha.py
mancdaz: Done, thanks

scripts/checksha.py
mancdaz: Fixed, thanks

scripts/checksha.py
```diff
@@ -0,0 +1,139 @@
...
+    '''Find the dynamic inventory file'''
+    for root, dirs, files in os.walk(path):
+        if inv in files:
+            return os.path.join(root, inv)
+
+
+def clone_git_repo(project, path=CLONE_DIR):
+    '''clone a git repo'''
+    os.chdir(path)
+    if not os.path.exists(project):
```
git-harry: Depending on how long this has been here it might not be up to date.

scripts/checksha.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+def find_inventory(inv, path):
+    '''Find the dynamic inventory file'''
+    for root, dirs, files in os.walk(path):
+        if inv in files:
+            return os.path.join(root, inv)
+
+
+def clone_git_repo(project, path=CLONE_DIR):
```
git-harry: Some of the projects are quite big. How often is this expected to get run? It seems quite expensive to be cloning from new all the time.

scripts/checksha.py
mancdaz: Fixed, thanks

scripts/checksha.py
mancdaz: need to be able to get an up to date clone from somewhere? Though on a separate note, if clonedir does exist we should probably do a fresh fetch

scripts/checksha.py
mancdaz: Fixed, thanks

scripts/checksha.py
prometheanfire: perhaps another option, like repo-uri-prefix or something?

mancdaz: done, thanks

scripts/checksha.py
```diff
@@ -0,0 +1,186 @@
+#!/usr/bin/env python
+# Copyright 2014, Rackspace US, Inc.
```
git-harry: 2015?

nrb:
Added labels to match the issue

d34dh0r53:
Hey @mancdaz this is failing flake8:
````
scripts/checksha.py:18:1: H306  imports not in alphabetical order (sys, os)

scripts/checksha.py:20:1: H306  imports not in alphabetical order (subprocess, argparse)

scripts/checksha.py:24:1: H306  imports not in alphabetical order (ansible.runner, ansible.inventory)

scripts/checksha.py:82:9: H233  Python 3.x incompatible use of print operator

scripts/checksha.py:91:13: H233  Python 3.x incompatible use of print operator

scripts/checksha.py:94:9: H233  Python 3.x incompatible use of print operator

scripts/checksha.py:102:13: H233  Python 3.x incompatible use of print operator

scripts/checksha.py:124:9: H233  Python 3.x incompatible use of print operator

scripts/checksha.py:127:9: H233  Python 3.x incompatible use of print operator

scripts/checksha.py:143:5: H233  Python 3.x incompatible use of print operator

scripts/checksha.py:174:9: H233  Python 3.x incompatible use of print operator

scripts/checksha.py:210:9: H233  Python 3.x incompatible use of print operator

scripts/checksha.py:212:9: H233  Python 3.x incompatible use of print operator
````

d34dh0r53:
Hiya @mancdaz can you rebase this please?

mancdaz:
@d34dh0r53 lols I keep rebasing, and then something else goes in and knocks it off again.  Done now, pending checks

mancdaz:
urgh github doesn't rebase, it merges.

prometheanfire:
well, :+1: when fixed

d34dh0r53:
:+1:

Issue #200, opened 2015-06-29
cinder_service_check.py not using --host option


git-harry:
The host argument is not used by the script leading to metrics being returned when they shouldn't.
Witthout host arg:
./cinder_service_check.py 172.29.236.3
status okay
metric cinder-scheduler_on_host_maas2-node3_cinder_scheduler_container-0377f639 uint32 1
metric cinder-scheduler_on_host_maas2-node1_cinder_scheduler_container-2706e6d1 uint32 1
metric cinder-scheduler_on_host_maas2-node2_cinder_scheduler_container-38784141 uint32 1
metric cinder-volume_on_host_maas2-node4 uint32 1

With host arg:
./cinder_service_check.py 172.29.236.3 --host maas2-node3_cinder_scheduler_container-0377f639
status okay
metric cinder-scheduler_status uint32 1
metric cinder-scheduler_on_host_maas2-node1_cinder_scheduler_container-2706e6d1 uint32 1
metric cinder-scheduler_on_host_maas2-node2_cinder_scheduler_container-38784141 uint32 1
metric cinder-volume_on_host_maas2-node4 uint32 1

Expected output with host arg:
./cinder_service_check.py 172.29.236.3 --host maas2-node3_cinder_scheduler_container-0377f639
status okay
metric cinder-scheduler_status uint32 1

nrb:
Does this need to go into Juno as well?

git-harry:
@nrb - It affects juno, if that is good enough reason for a backport then, yes, it does.

git-harry:
There look to be commits for this in master, kilo, juno.

PR #201, opened 2015-06-29
cinder_service_check.py not using --host option

git-harry:
The script cinder_service_check.py defines an optional host parameter
that is designed to filter the results, returned by the cinder API, so
that only services on the specified host are returned. The MaaS setup
role, that configures the checks that use this script, uses '--host'.
The script does not use the host argument if supplied.

This patch fixes the script so that the results returned are filtered
by host, where host includes any backends/pools on that host.

Closes-bug: https://github.com/rcbops/rpc-openstack/issues/200

### DIFF
--- a/maas/cinder_service_check.py
+++ b/maas/cinder_service_check.py
@@ -53,6 +53,12 @@ def check(auth_ref, args):
 
     services = r.json()['services']
 
+    # We need to match against a host of X and X@lvm (or whatever backend)
+    if args.host:
+        backend = ''.join((args.host, '@'))
+        services = [s for s in services if s['host'].startswith(backend) or
+                                           s['host'] == args.host]
+
     if len(services) == 0:
         status_err('No host(s) found in the service list')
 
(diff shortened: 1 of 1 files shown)


maas/cinder_service_check.py
```diff
@@ -56,6 +56,11 @@ def check(auth_ref, args):
     if len(services) == 0:
```
prometheanfire: I think the patch needs to go before this line, as services can be filtered down to 0 in the patch.

git-harry: Good spot, I'll update the patch.

mattt416:
I think this code is unnecessary.  We already filter on args.host, but the filter isn't right.  How about changing this from:

```
        # We need to match against a host of X and X@lvm (or whatever backend)
        if args.host and args.host in service['host']:
            name = '%s_status' % service['binary']
        else:
            name = '%s_on_host_%s' % (service['binary'], service['host'])
```
to:

```
        # We need to match against a host of X and X@lvm (or whatever backend)
        if args.host is None or (args.host and args.host in service['host']):
            name = '%s_on_host_%s' % (service['binary'], service['host'])
            metric_bool(name, service_is_up)
```

Thoughts?

git-harry:
@mattt416 - as per our conversation I have updated the patch to adjust that second section of filtering.

mattt416:
:+1: 

@prometheanfire can you re-review before we merge?  Thanks!

prometheanfire:
:+1:

Issue #206, opened 2015-06-30
Need some comments in README about where certain things live across branches


mancdaz:
Since things have moved around a lot between Juno and Kilo, it's worth
documenting where stuff lives, so we're not all having to remember it.

nrb:
Both Kilo/master merged. Attaching milestone for release note reference. Closing.

PR #207, opened 2015-06-30
Clarify where code lives for certain branches

mancdaz:
Since things have moved around a lot between Juno and Kilo, it's worth
documenting where stuff lives, so we're not all having to remember it.

Fixes: #206


mattt416:
Backport, LGTM.

nrb:
:+1: from me, this is a backport of my change.

PR #213, opened 2015-07-01
Prevent /boot from filling up by disabling unattended upgrades

### LINKED ISSUE #24: Prevent unattended upgrades

prometheanfire:
Simply disables the security branch of unattended upgrades (the only one enabled by default).

Closes #24

### DIFF
--- a/rpcd/playbooks/roles/rpc_support/tasks/disable_unattended_upgrades.yml
+++ b/rpcd/playbooks/roles/rpc_support/tasks/disable_unattended_upgrades.yml
@@ -1,0 +1,20 @@
+---
+# Copyright 2015, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


nrb:
Looks good, I will run a build with it just to be sure, though.

prometheanfire:
cool, I did test it here as well.

prometheanfire:
backport is in #214

rpcd/playbooks/roles/rpc_support/tasks/disable_unattended_upgrades.yml
```diff
@@ -0,0 +1,20 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+- name: Disable unattended upgrades
+  lineinfile:
+    dest: '/etc/apt/apt.conf.d/50unattended-upgrades'
+    regexp: '\-security\";$'
```
mattt416: Will this result in an ansible change on every run? Would it be easier to disable updates by flipping APT::Periodic::Unattended-Upgrade to "0" in /etc/apt/apt.conf.d/20auto-upgrades instead?

nrb: That's one way; the regexp could also be changed to look for `^\/\/?`at the beginning.

nrb:
Verified that re-running the play does not alter the file if the line is already commented, so I'm +1.

stevelle:
Verified, +1. Merging.

PR #214, opened 2015-07-01
Prevent /boot from filling up by disabling unattended upgrades

### LINKED ISSUE #24: Prevent unattended upgrades

prometheanfire:
Simply disables the security branch of unattended upgrades (the only one enabled by default).

Closes #24

### DIFF
--- a/rpcd/playbooks/roles/rpc_support/tasks/disable_unattended_upgrades.yml
+++ b/rpcd/playbooks/roles/rpc_support/tasks/disable_unattended_upgrades.yml
@@ -1,0 +1,20 @@
+---
+# Copyright 2015, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+- name: Disable unattended upgrades
+  lineinfile:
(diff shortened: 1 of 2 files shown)


rpcd/playbooks/roles/rpc_support/tasks/disable_unattended_upgrades.yml
```diff
@@ -0,0 +1,20 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+- name: Disable unattended upgrades
+  lineinfile:
+    dest: '/etc/apt/apt.conf.d/50unattended-upgrades'
+    regexp: '\-security\";$'
```
mattt416: This is going to match even after you disable it, resulting in this changing on each run.

prometheanfire: it might write the file every run, unless it checks if it already changed the line.  But the actual line won't change.

prometheanfire: this makes me think that it won't replace the line every time.

     lineinfile: dest=/etc/hosts regexp='^127\.0\.0\.1' line='127.0.0.1 localhost' owner=root group=root mode=0644

otherwise that would constantly get replaced as well.  That's from the examples for lineinfile

nrb: I'm spinning up my test AIO right now, so I'll run the role a few times just to verify this.

My suspicion is that the line will simply be replaced each time, not appended.

nrb:
I've confirmed this does not write multiple comment lines, and does not modify the mtimes if the comment already exists.

+1.

stevelle:
backport looks good, though missing msg in the commit.
:+1:  and merging.

PR #219, opened 2015-07-06
Add script to handle changing osad git tag

### LINKED ISSUE #218: Standardize osad tagging procedure.
When updating to new tags of os-ansible-deployment, it would be useful to have a standard tagging format, so we can quickly search for the commit should the need arise.

nrb:
This script will change the SHA used by the os-ansible-deployment
submodule to match a tag from the upstream repository. It will also
provide a commit message template that can be edited by the commiter.

Fixes #218

### DIFF
--- a/scripts/bump-osa-sha.sh
+++ b/scripts/bump-osa-sha.sh
@@ -1,0 +1,14 @@
+#!/bin/sh
+
+if (( $# < 1 )); then
+    echo "Please provide a tag value."
+    exit 1
+fi
+pushd openstack-ansible
+    git fetch
+    git checkout $1
+popd
+
+git add openstack-ansible
+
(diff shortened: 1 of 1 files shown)


odyssey4me:
Looks fine to me. Has this been tested? We had some shenanigans with sha's a few weeks ago.

nrb:
This is what I've been using to tag each time, so yeah, tested.

I think the shenanigans you were referring to were master and kilo pointing to the same tag? That was intentional, since I don't think we really want our master branch following master of OSAD.

odyssey4me:
LGTM :+1:

git-harry:
-1 this needs updating to openstack-ansible. There is a merge commit that should be removed and the branch rebased instead.

stevelle:
bump?

hughsaunders:
:-1: uses bashisms but specifies interpreter as sh

Issue #223, opened 2015-07-07
rabbit_mq user CVE fix breaks MaaS


andymcc:
Since we are removing the guest user (which maas uses by default) as part of the bug fix for: https://bugs.launchpad.net/openstack-ansible/+bug/1466982 MaaS will not work as intended anylonger and the rabbitmq monitors will start failing.

We should fix this moving forward - but this will require rabbitmq checks to be removed or manually edited, as it won't update existing checks monitors.

nrb:
Both PRs merged, closing.

PR #227, opened 2015-07-07
Add MaaS specific rabbitmq user

andymcc:
Since we are removing the guest/guest user from rabbitmq as part of
os-ad we need to create a maas specific user that has "administrator"
privileges, otherwise the monitors don't work.

This patch adds a maas_user to rabbit, with a password specified in
user_extras_secrets. This is then passed to the monitors for rabbit as
they are created.

This won't adjust existing MaaS checks, so existing checks will need to
be removed and re-added or manually adjusted to address this problem.

Fixes: #223

### DIFF
--- a/rpcd/playbooks/roles/rpc_maas/tasks/rabbitmq_user.yml
+++ b/rpcd/playbooks/roles/rpc_maas/tasks/rabbitmq_user.yml
@@ -13,7 +13,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-maas_keystone_password:
-rpc_support_holland_password:
-kibana_password:
-maas_rabbitmq_password:
+- name: Ensure MaaS rabbitmq user as administrator
+  rabbitmq_user:
+    user: "{{ maas_rabbitmq_user }}"
+    password: "{{ maas_rabbitmq_password }}"
+    vhost: "/"
+    configure_priv: ".*"
+    read_priv: ".*"
+    write_priv: ".*"
+    tags: "administrator"
+    state: "present"
+  tags:
+    - rabbitmq-user
--- a/rpcd/playbooks/roles/rpc_maas/defaults/main.yml
+++ b/rpcd/playbooks/roles/rpc_maas/defaults/main.yml
@@ -134,6 +134,12 @@ maas_scheme: http
 maas_keystone_user: maas
 
 #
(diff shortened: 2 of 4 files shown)


nrb:
:+1:

rpcd/playbooks/roles/rpc_maas/tasks/rabbitmq_user.yml
```diff
@@ -0,0 +1,27 @@
...
+
+- name: Ensure MaaS rabbitmq user as administrator
+  rabbitmq_user:
+    user: "{{ maas_rabbitmq_user }}"
+    password: "{{ maas_rabbitmq_password }}"
+    vhost: "/"
+    configure_priv: ".*"
+    read_priv: ".*"
+    write_priv: ".*"
+    tags: "administrator"
```
mattt416: There is a "monitoring" user which may be more appropriate.  See https://www.rabbitmq.com/management.html.

Anyone know if these read_priv/write_priv/config_priv privileges are necessary for what we're doing?

nrb: I think @sigmavirus24 and @cloudnull provided reviews on this while it was embargoed. I'm not 100% sure what the logic was for the administrator tag.

nrb: Reading that link makes the 'monitoring' tag seem like a good idea to me.

sigmavirus24: Hm, I wasn't aware of that tag. I, however, think this was to not break backwards compatibility with in a X.Y.* release. The reason for the CVE fix in os-ansible-deployment was that we are using the guest user which has admin privileges. That means that to not break monitoring's backwards compat we would want to have the new user be an administrator. I'd rather merge this now and then guarantee through extra testing that monitoring is the only tag we need, if that makes sense.

nrb: I think I concur with sigmavirus24 here - it's functional, and has parity with what we had prior. That said, if we want/need to change it, we can iterate.

sigmavirus24:
:+1: Merging.

PR #228, opened 2015-07-07
Add MaaS specific rabbitmq user

nrb:
Since we are removing the guest/guest user from rabbitmq as part of
os-ad we need to create a maas specific user that has "administrator"
privileges, otherwise the monitors don't work.

This patch adds a maas_user to rabbit, with a password specified in
user_extras_secrets. This is then passed to the monitors for rabbit as
they are created.

This won't adjust existing MaaS checks, so existing checks will need to
be removed and re-added or manually adjusted to address this problem.

Fixes: #223
(cherry picked from commit 6fc27ad7fd5a0b0191adc366faa91f8261f3c50e)

### DIFF
--- a/rpcd/playbooks/roles/rpc_maas/defaults/main.yml
+++ b/rpcd/playbooks/roles/rpc_maas/defaults/main.yml
@@ -134,6 +134,12 @@ maas_scheme: http
 maas_keystone_user: maas
 
 #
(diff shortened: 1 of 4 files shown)


nrb:
This is a kilo backport.

sigmavirus24:
Backport of a security issue so auto-merging

PR #231, opened 2015-07-07
Create remote check only if service exists.

### LINKED ISSUE #230: Standalone Swift - MaaS Remote Check Creation Failing
In standalone swift environments, MaaS remote check creation tasks fail. The `when:` criteria is trying to match on the [first host of each service host group](https://github.com/rcbops/rpc-openstack/blob/master/rpcd/playbooks/roles/rpc_maas/tasks/remote.yml#L35). In a standalone swift environment, most of these services won't exist. These tasks will fail.

byronmccollum:
Fixes #230.


byronmccollum:
The associated bug is prio-2. Can we bump the priority up to match? This is kinda important.

claco:
Bumped so they match.

git-harry:
@byronmccollum - you should rebase instead of merge. The second commit needs to be removed. This fix itself appear to do the trick.

git-harry:
+1

d34dh0r53:
:+1:

PR #232, opened 2015-07-07
Download of pccommon.sh should use raw URL

byronmccollum:
Download of pccommon.sh should use raw URL, or else you get the HTML source of the GitHub page.


odyssey4me:
:+1: LGTM

git-harry:
+1

Issue #209, opened 2015-06-30
f5-config.py Monitor for GALERA uses incorrect username haproxy


Mudpuppy12:
r'create ltm monitor mysql /' + PART + '/' + PREFIX_NAME + '_MON_GALERA { count 1 database'
    r' information_schema debug no defaults-from mysql destination *:*'
    r' interval 3 recv big5_chinese_ci recv-column 2 recv-row 0 send "select'
    r' * from CHARACTER_SETS;" time-until-up 0 timeout 10 username haproxy }',

The correct username should be monitoring, as the haproxy user no longer seems added to kilo releases

    r'create ltm monitor mysql /' + PART + '/' + PREFIX_NAME + '_MON_GALERA { count 1 database'
    r' information_schema debug no defaults-from mysql destination *:*'
    r' interval 3 recv big5_chinese_ci recv-column 2 recv-row 0 send "select'
    r' * from CHARACTER_SETS;" time-until-up 0 timeout 10 username monitoring }',

nrb:
Does this impact anything further back than Kilo?

mattt416:
I'm not familiar enough with this script or how the load balancers are operated.  Are we better off replacing reference to `haproxy` w/ `monitoring` or overriding galera_monitoring_user in rpc-openstack to use haproxy user?  The upside to doing the latter is it an upgrade from juno onwards will not create another monitoring user (called `monitoring`) when the `haproxy` user already exists.

Thoughts?

nrb:
From my perspective, the `galera_monitoring_user` override makes sense as a fix.

Mudpuppy12:
I am guessing the intent of this change was to remove all references to the haproxy. It doesn't make sense to have that username in the database, as monitoring is a better name for it. I am not sure how upgrades will handle it.

nrb:
I'll do a little bit more code spelunking and see if I can find anything related to this.

nrb:
So for upgrades, the f5-config.py script will need to be run anyway, since the repo server VIPs are not registered with the F5. They weren't present in Juno.

Given that, I *think* changing the user name during the F5 configuration part of an upgrade should be fine, but may want some input from netsec.

nrb:
Neither Juno nor Kilo's f5-config.py script are templated, so we might have to change that.

Juno: https://github.com/stackforge/os-ansible-deployment/blob/juno/scripts/f5-config.py#L32
[...]

PR #234, opened 2015-07-08
Update f5-config.py to use correct MySQL user

mattt416:
In kilo and up, os-a-d defaults 'galera_monitoring_user' to user
'monitoring'.  Previously, we created a static user called 'haproxy'.
This commit updates f5-config.py to replace reference to 'haproxy'
user with 'monitoring'.

Closes bug #209


mattt416:
It would be great to get some input from @BjoernT or NetSec.  The alternative here would be to override `galera_monitoring_user` with `haproxy`.

If we do proceed with this change, we will need to update osad's scripts/run-upgrade.sh to remove the defunct haproxy MySQL user.

nrb:
I am +1 on this.

sigmavirus24:
:+1: looks simple enough to me.

Issue #183, opened 2015-06-10
rpcd openstack_deploy copy needs '/*'


nrb:
In our deploy instructions, step 2ii should be `cp -R /opt/rpc-openstack/etc/openstack_deploy/* /etc/openstack_deploy`.

Otherwise, we get `/etc/openstack_deploy/openstack_deploy`.

mattt416:
Step 2ii also needs updating as we now include a user_variables.yml in rpc-openstack/rpcd/etc/openstack_deploy and just copying everything across would wipe out the user_variables.yml copied in from openstack-ansible-deployment.

mattt416:
We'll need to hold backport of this PR to kilo until https://github.com/rcbops/rpc-openstack/pull/243 has merged.

nrb:
Blocking for the 243 merge.

nrb:
Unblocking and closing since the changes merged.

PR #242, opened 2015-07-09
Fix incorrect steps in README.md

mattt416:
This change does the following:

- merges /etc/openstack_deploy/user_variables.yml with
- updates an incorrect cp which was resulting in
  /etc/openstack_deploy/openstack_deploy (we also selectively copy
  files to ensure we don't overwrite
  /etc/openstack_deploy/user_variables.yml)

Closes issue #183


nrb:
Bumping back to prio-1 since it affects #251.

I am +1 on this change.

sigmavirus24:
:+1:

Issue #203, opened 2015-06-29
Need to be able to specify default user_variables.yml for rpcd


git-harry:
rpcd needs some defaults that do not match those defined in osad.

The default for [cinder]/cross_az_attach in nova.conf is True for osad and support would like it set to False.
https://bugs.launchpad.net/openstack-ansible/+bug/1457140

nrb:
Are you then abandoning https://github.com/rcbops/rpc-openstack/pull/187 ?

git-harry:
@nrb - am I failing to github? You asked for an issue, so I created this. The other thing is the pull request.

nrb:
Nope, sorry, my mistake. You referenced it correctly, I was misreading.

d34dh0r53:
Can this be closed as #187 has merged?

nrb:
Looks like we still need it in Kilo.

prometheanfire:
kilo done, closing

b3rnard0:
@nrb Need doc impact information which Harry will provide.

PR #243, opened 2015-07-09
Add user_variable.yml file to rpcd

mattt416:
This repo is for deploying an opinionated version of OpenStack using
os-ansible-deployment. There are instances where the defaults that a
Rackspace deployment will use differ from the upstream deployment
project. This commit creates a user_variables.yml file to use for
specifying these defaults and so this file should be used instead of
the one in os-ansible-deployment.

The user_variables.yml file in os-ansible-deployment currently sets two
variables that are not specified by default, by the associated roles.
These variables, ssl_protocol and ssl_cipher_suite, are specified in
this new file.

Support have requested that the nova configuration option
[cinder]/cross_az_attach be set to False and so is set here to override
the upstream default of True.

The script update-yaml.py is added, for use when running an AIO, so
that the user_variables file added by this commit can be merged into
the one copied by bootstrap-aio.sh. Variables defined in the rpcd
user_variables.yml file will override those in the one supplied by
bootstrap-aio.sh.

Closes-bug: https://github.com/rcbops/rpc-openstack/issues/203
(cherry picked from commit 748a7259cc4f4e4193c4cc6eaf44a84f9057f8b9)

### DIFF
--- a/rpcd/etc/openstack_deploy/user_variables.yml
+++ b/rpcd/etc/openstack_deploy/user_variables.yml
@@ -1,0 +1,24 @@
+---
+# Copyright 2015, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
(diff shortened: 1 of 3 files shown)


claco:
Asking to ask. Since this is a behavior change people will have to learn, have we submitted some docs to the upgrade docs part of the guide for folks upgrading?

mattt416:
/cc @git-harry who authored original commit

(on a side note, I did amend this project's README.md to incorporate this change, that change will get backported to kilo branch once this merges)

nrb:
:+1:

odyssey4me:
:+1: LGTM backport

sigmavirus24:
@mattt416 please rebase this.

nrb:
I think this is made redundant by https://github.com/rcbops/rpc-openstack/pull/261

nrb:
Yeah, #261 implements this with a more up-to-date user_variables file. closing.

PR #248, opened 2015-07-15
SHA bump for OSAD to match kilo as of 15/07/2015

nrb:
This SHA bump is largely to grab some important fixes for the upcoming
11.0.0 release.

Of major concern is turning off MariaDB slow query logs
(https://bugs.launchpad.net/openstack-ansible/+bug/1463956)


claco:
:+1: for the right sha. :-)

sigmavirus24:
:+1:

Issue #258, opened 2015-07-16
OSAD default token is now Fernet


claco:
With the latest sha bump for OSAD, the default token type is now Fernet, which is a functional difference from 10.x, which is uuid+sql storage.

In order to keep the functionality the same from a support perspective, we should override the config defaults in RPC back to uuid+sql to match what we had in 10.x

odyssey4me:
I'm not entirely sure that enforcing UUID tokens as a default is doing anyone any favours. Currently the deployed environments have huge keystone databases which are causing operational problems. I really think that the decision to keep UUID tokens as a default should be thoroughly discussed with support.

I would rather suggest that moving to fernet tokens is a good new default, especially for new environments - and there should rather be a documentation note regarding how to set the token backend to UUID tokens if that is desired.

cloudnull:
:+1: to what @odyssey4me said. Fernet addresses some serious performance and security issues as such I think it should remain the default. IMHO.

JCallicoat:
Should we remove the user var and have fernet tokens be the default now for Liberty?

nrb:
If it's the default for Keystone in Liberty, I don't see why not.

I know there was a desire to keep things the same from Juno to Kilo, but I think it's been given a release cycle and we should catch up. Note that supporting fernet does NOT mean we have to support federation.

@d34dh0r53 will probably have to weigh in, though.

odyssey4me:
Confirmed - Fernet tokens have nothing to do with Federation. They have everything to do with saving yourself from wanting to burn everything... otherwise known as no more keystone token table truncates. Using fernet tokens is also a path to multi-region implementations (as the keystone DB is now much smaller and changing less often).

PR #260, opened 2015-07-17
use uuid tokens and sql persistence driver

mancdaz:
Upstream osad uses fernet tokens as the default. In rpc we want to
continue to use uuid for the time being, so this commit adds a var to
override the default, as well as setting the token persistence driver back to sql

Partially-Closes: https://github.com/rcbops/rpc-openstack/issues/258

### DIFF
--- a/rpcd/etc/openstack_deploy/user_variables.yml
+++ b/rpcd/etc/openstack_deploy/user_variables.yml
@@ -22,3 +22,7 @@ nova_cross_az_attach: False
 ssl_protocol: "ALL -SSLv2 -SSLv3"
 # Cipher suite string from https://hynek.me/articles/hardening-your-web-servers-ssl-ciphers/
 ssl_cipher_suite: "ECDH+AESGCM:DH+AESGCM:ECDH+AES256:DH+AES256:ECDH+AES128:DH+AES:ECDH+3DES:DH+3DES:RSA+AESGCM:RSA+AES:RSA+3DES:!aNULL:!MD5:!DSS"
+
+# Keystone overrides
+keystone_token_provider: "keystone.token.providers.uuid.Provider"
+keystone_token_driver: "keystone.token.persistence.backends.sql.Token"


cloudnull:
I think this is a bad idea in terms of future proofing an environment, scalability and supportability. I'm not sure why we'd choose to keep with SQL backed UUID tokens at this point.

claco:
I think the reason to stick with uuid+sql is that it's still the default in Kilo. Changing from the upstream defaults is an operational change for a 11.0 release where it's supposed to be a Kilo code refresh only,

Second, iirc, anyone who wants to use Federation/ADFS would have to change their config back to uuid [at least until keystone fixes the scoped things, which may or may not happen as a security patch]. Having the default change, to then have to change it again seems like a unnecessary thing to have to do for 11.0.

cloudnull:
In keystone the default is **uuid+sql** however the recommended in kilo going forward is **fernet**. That said, **fernet**, in its current form, would be an issue for Federation deployments however IMHO staying with **uuid+sql** because it is solving one problem, for a single use case, which will likely not be the norm while **fernet** will solve a lot of problems for most other deployment use cases. All that said, I understand the desire to change even if I don't agree with it.

mancdaz:
As has been pointed out, this is part of our commitment to keep the delta between juno, and this initial release of kilo (for rpc) as minimal as possible. Whilst fernet is going to be a better future choice, uuid+sql is what we have now, is what we have tested, is what support are comfortable and knowledgable about, and ultimately is a less risky choice at this time.

cloudnull:
While I understand the desire to minimize the delta I think this is the wrong course of action. I disagree that this is something that we've not tested. In truth QE has been deploying off the osad kilo branch and the rpc-openstack submodule which contains the bits that made fernet the default token provider. So rpc has been testing fernet for some time now.  Because of that, I'd also say there little to no risk here. While I'll I agree fernet is not something that support has experience with so there will need to be some documentation on how it works and how to do some of the functions manually. However,  I'd think the change would be welcomed as it solves much of the problems I know support has experienced using uuid+SQL.

odyssey4me:
:+1: LGTM as per requirements

cloudnull:
@mancdaz - It should be noted that UUID+SQL has never been tested in OSAD within the Kilo development time-frame if this is something that we intend to do, we should do it soon so QE can get  to running it through some paces.

mancdaz:
@cloudnull yes, that's the plan. QE are scheduled to do an entire run through with it, we just need to get this in so that they can.

Having said that, uuid is what has always been deployed in RPC, so I'd expect no regressions in behaviour.

nrb:
:+1:, merging with @odyssey4me's +1

PR #267, opened 2015-07-21
Disable Unattended Upgrades

### LINKED ISSUE #24: Prevent unattended upgrades

claco:
Unattended upgrades for the security branch is enabled by default.
After some period of time, the /boot partition fills up with new
kernels, which is unexpected and undesired.

This patch disables the security updates as part of the rpc_support
playbook by removing the configuration file.

Fixes #24


nrb:
:+1:

prometheanfire:
:+1:

PR #270, opened 2015-07-22
SHA bump for OSAD to match kilo as of 22/07/2015 This SHA bump is largely to grab some important fixes for the upcoming 11.0.0 release.

### LINKED ISSUE #269: Upgrades broken with RabbitMQ upgrade error
Filed upstream in OSAD: https://bugs.launchpad.net/openstack-ansible/+bug/1474992

We need to bump the OSAD sha to consume the fix for this.

claco:
Of major concern is fixing RabbitMQ upgrade failures.
(https://bugs.launchpad.net/openstack-ansible/+bug/1474992)

Fixes #269


nrb:
SHA matches head of OSAD's Kilo - I thought I spotted an 11.0.0rc3 reference somewhere in a variable file, so i'm checking for that now. If it's present, we should update that with this PR, too.

nrb:
Ok - it's in `rpcd/playbooks/group_vars/all.yml`, called `rpc_release`. For master, we can leave it. For kilo, we need to bump to rc8.

I am +1 on this PR as is.

odyssey4me:
While I think an appropriate update of rpc_release should be included every time the SHA is bumped, considering the timelines I'm happy to :+1: this for testing purposes.

sigmavirus24:
2 +1s already, so merging.

nrb:
@odyssey4me You're right, but in master it's always `master`. Same as OSAD

Issue #240, opened 2015-07-08
Add YAML lint/sanity checking to rpc-openstack


nrb:
Currently, rpc-openstack does not have any lint checking whatsoever. This resulted in a bug that halted QE work (#236). I have checked the below process and found that it would have caught this bug.

We can crib a lot of the logic for this from osad's `scripts/gate-check-lint.sh` script, though there will have to be some modifications. It's also possible that we could refactor osad's linting so that it could be parameterized, but that will require a spec to go through the community process.

To confirm that this can work my local machine, I did the following:

1. Install ansible and osad's `dev-requirements.txt` into a virtualenv
2. Symlinked `/opt/rpc-openstack` to my `rpc-openstack` checkout (since rpc-openstack has to use hard-coded paths for osad role lookups)
3. Created a local inventory file in `rpc-openstack/rpcd/playbooks/` with `echo -e '[all]\nlocalhost ansible_connection=local' | tee local_only_inventory`
4. Checked syntax with `ansible-playbook -i local_only_inventory --syntax-check *.yml --list-tasks` in `rpc-openstack/rpcd/playbooks`

Also, running `ansible-lint *.yml` in `rpc-openstack/rpcd/playbooks` produces a lot of errors; these could be cleaned up.

This definitely impacts QE, as it adds a layer of testing that they can use. It may impact docs, since they might want to at least note how lint checking will be done.

nrb:
@jpmontez @jacobwagner Do either of you have thoughts on this? Does QE already have some work on progress we could hook into?

nrb:
This is almost ready for merge - we just need #274's conflicts resolved and get it into master/kilo.

nrb:
(Closed by mistake)

PR #272, opened 2015-07-22
Add a linting script for rpcd.

nrb:
Linting for rpc-openstack should only cover the RPC-specific playbooks,
roles, scripts, and modules, since
os-ansible-deployment/openstack-ansible already does their own upstream
linting.

This change also uses relative paths in the ansible.cfg file so that
linting will work from Travis's container-based infrastructure. With the
hardcoded paths, we would have had to use sudo in the script, and add
configuration options to our .travis.yml. Also, their sudo-capable
infrastructure is considered legacy, so it's probably best avoided.

The relative paths mean we should be able to lint from whatever
directory we want, and if working within a virtualenv, run linting
without being root.

Addresses #240

### DIFF
--- a/scripts/linting.sh
+++ b/scripts/linting.sh
@@ -1,0 +1,60 @@
+#!/usr/bin/env bash
+# Copyright 2014, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+## Shell Opts ----------------------------------------------------------------
+set -euo pipefail
+
+trap "exit -1" ERR
+
+# Track whether linting failed; we don't want to bail on lint failures
+failed=0
+
+## Main ----------------------------------------------------------------------
+echo "Running Basic Ansible Lint Check"
+
+
+# Install the development requirements.
+if [[ -f "os-ansible-deployment/dev-requirements.txt" ]]; then
+  pip2 install -r os-ansible-deployment/dev-requirements.txt || pip install -r os-ansible-deployment/dev-requirements.txt
+else
+  pip2 install ansible-lint || pip install ansible-lint
+fi
+
+# Run hacking/flake8 check for all python files
+# Ignores the following rules due to how ansible modules work in general
+#     F403 'from ansible.module_utils.basic import *' used; unable to detect undefined names
+#     H303  No wildcard (*) import.
+# Excluding our upstream submodule, and our vendored f5 configuration script.
+flake8 $(grep -rln -e '^#!/usr/bin/env python' -e '^#!/bin/python' -e '^#!/usr/bin/python' * ) || failed=1
+
+# Perform our simple sanity checks.
+pushd rpcd/playbooks
+  # Put local inventory in a var so we're not polluting the file system too much
+  LOCAL_INVENTORY='[all]\nlocalhost ansible_connection=local'
+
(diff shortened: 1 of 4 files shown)


scripts/linting.sh
```diff
@@ -0,0 +1,59 @@
...
+  pip2 install -r os-ansible-deployment/dev-requirements.txt || pip install -r os-ansible-deployment/dev-requirements.txt
+else
+  pip2 install ansible-lint || pip install ansible-lint
+fi
+
+# Run hacking/flake8 check for all python files
+# Ignores the following rules due to how ansible modules work in general
+#     F403 'from ansible.module_utils.basic import *' used; unable to detect undefined names
+#     H303  No wildcard (*) import.
+flake8 --ignore=F403,H303 $(grep --exclude-dir=os-ansible-deployment -rln -e '^#!/usr/bin/env python' -e '^#!/bin/python' -e '^#!/usr/bin/python' * ) || failed=1
```
sigmavirus24: I'd really like us to exclude `scripts/f5-config.py` here. It's not really ours, we just vendor it, and it's full of lint problems that we'll just have to re-fix everytime we import it if we don't exclude it.

nrb: Ok, I'll look into that.

scripts/linting.sh
```diff
@@ -0,0 +1,60 @@
...
+# Track whether linting failed; we don't want to bail on lint failures
+failed=0
+
+## Main ----------------------------------------------------------------------
+echo "Running Basic Ansible Lint Check"
+
+
+# Install the development requirements.
+if [[ -f "os-ansible-deployment/dev-requirements.txt" ]]; then
+  pip2 install -r os-ansible-deployment/dev-requirements.txt || pip install -r os-ansible-deployment/dev-requirements.txt
```
sigmavirus24: IMO, this should be part of the travis config or a separate script that travis can run/anyone can run. It should also ideally happen in a virtual environment to prevent polution.

We should run with this for now, but we should follow upstream best practices.

scripts/linting.sh
```diff
@@ -0,0 +1,60 @@
...
+else
+  pip2 install ansible-lint || pip install ansible-lint
+fi
+
+# Run hacking/flake8 check for all python files
+# Ignores the following rules due to how ansible modules work in general
+#     F403 'from ansible.module_utils.basic import *' used; unable to detect undefined names
+#     H303  No wildcard (*) import.
+# Excluding our upstream submodule, and our vendored f5 configuration script.
+flake8 --ignore=F403,H303 $(grep --exclude-dir=os-ansible-deployment --exclude=f5-config.py -rln -e '^#!/usr/bin/env python' -e '^#!/bin/python' -e '^#!/usr/bin/python' * ) || failed=1
```
sigmavirus24: We can configure ignores and such in a tox.ini, setup.cfg, or .flake8 file. Obviously it isn't necessary though.

sigmavirus24:
:+1:

scripts/linting.sh
nrb: While I agree, I'm also trying to make sure we're not *too* beholden to Travis itself.

If it's a separate dependencies script, that's probably a decent compromise.

Also worth nothing this is taken from upstream OSAD, so this can be improved there, too.

scripts/linting.sh
nrb: Would the .flake8 file use the same settings listed here: https://flake8.readthedocs.org/en/2.3.0/config.html ?

scripts/linting.sh
sigmavirus24: True. Using tox would be *somewhat* appropriate here to take advantage of virtual environment management and such. If we use tox, we can then consolidate our flake8 configuration into one file (`tox.ini`)

scripts/linting.sh
sigmavirus24: Yes. Those are an example of the options you can set. Look at `flake8 -h` to get the full list of options.

scripts/linting.sh
nrb: I like that idea. There are already talks of doing some of that upstream.

prometheanfire:
:+1: 

needs another +1 because of new commit

sigmavirus24:
+1

Issue #276, opened 2015-07-23
Turn down ansible forks variable in deploy.sh


nrb:
By default, OSAD's scripts-library.sh uses [25 forks](https://github.com/stackforge/os-ansible-deployment/blob/master/scripts/scripts-library.sh#L23). However, Juno's run-playbooks.sh scripts [sets it to 10](https://github.com/stackforge/os-ansible-deployment/blob/juno/scripts/run-playbooks.sh#L30).

We should probably drop the number of forks down to 10 or 15, because we're seeing SSH failures on multi-node installs when the containers aren't quite ready.

nrb:
@robb-romans This affects docs; in the event that these patches don't merge, we should advise deployers to do the following:

When deploying, run `FORKS=15 scripts/deploy.sh`
When upgrades, run `FORKS=15 scripts/upgrade.sh`

sigmavirus24:
Would installing the ssh-retry ansible plugin help with this (as an alternative)?

robb-romans:
please review ^^^^

odyssey4me:
FYI please note that https://bugs.launchpad.net/openstack-ansible/+bug/1479812 has been reported upstream and the review for master has merged, with a review for kilo pending. While the rpc-openstack override will still work you may find that the upstream fix changes what you'll want to be doing. The reviews are: https://review.openstack.org/#/q/I31ad7353344f7994063127ecfce8f4733769234c,n,z

nrb:
Thanks @odyssey4me; I'm not totally sure we got the upstream changes, but I think setting this in the rpc-openstack scripts makes sense if for no other reason that to increase visibility for our deployers.

nrb:
A suggestion - we include @odyssey4me's patch grep pattern for the number of CPUs instead of a hardset 10. (https://review.openstack.org/#/c/209426/1/scripts/scripts-library.sh)

Thoughts, @jacobwagner?

nrb:
This was also merged kilo by mistake, but both @prometheanfire were fine with that. I'm closing this issue and leaving this comment to acknowledge that it happened.

nrb:
Re-opening in light of delegate_to issues found on https://bugs.launchpad.net/openstack-ansible/+bug/1479812/comments/5.

We should go with Jacob's original proposal of 10.

claco:
@nrb @jacobwagner What setting did we use today in the upgrade?

nrb:
I did not see what was set.

[...]

PR #277, opened 2015-07-23
Add FORKS env variable to deploy

jacobwagner:
Currently the FORKS variable is set by osads scripts-library
to 25. This can cause issues with multinode deployments and the
ssh service in containers not being up in time for plays.

This will allow the user of the deploy script to override the
FORKS ansible variable to suite their environment.

Closes issue #276

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -2,7 +2,6 @@
 
 set -e -u -x
 set -o pipefail
-source /opt/rpc-openstack/os-ansible-deployment/scripts/scripts-library.sh
 
 export ADMIN_PASSWORD=${ADMIN_PASSWORD:-"secrete"}
 export DEPLOY_AIO=${DEPLOY_AIO:-"no"}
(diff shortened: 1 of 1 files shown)


nrb:
This looks good to me, but needs a rebase due to https://github.com/rcbops/rpc-openstack/pull/279

sigmavirus24:
This looks good to me too. Are we opposed to doing manual merges in situations like this?

nrb:
I suppose not, but it's also not necessarily blocking anything. Since it's an environment variable, it can be set externally to these scripts.

nrb:
I think we should consider this for rc15, based on QE's experience of the lower fork setting being more successful. Any thoughts, @claco?

nrb:
Adding to rc15, since this has gotten us successful upgrades.

nrb:
I think we should consider the number of CPUs value, per https://review.openstack.org/#/c/209426/1/scripts/scripts-library.sh

jacobwagner:
rebased and pr updated

nrb:
:+1: from me.

ionosphere80:
Looks good, thanks!

prometheanfire:
:+1: lgtm

Issue #273, opened 2015-07-22
maas needs to be updated to use Keystone v3


sigmavirus24:
OSAD has updated its keystone library to perform all actions with Keystone v3. Some portions of Keystone v3 cannot be served via v2, e.g., endpoints. v2 endpoints can be returned via v2 but not vice-versa. Our MaaS scripts for keystone need to be updated to also use v3.

nrb:
So it sounds like MAaS scripts with 11.0 will fail against keystone without updating, correct?

sigmavirus24:
*Maybe*

nrb:
Sounds like we'll need to verify this, at the very least.

sigmavirus24:
I was able to confirm this today.

sigmavirus24:
So @xeregin and myself were able to confirm (yesterday) that pretty much every OpenStack check (except swift?) break due to the upgrade to Keystone v3. The entire monitoring dashboard for one of @xeregin's test deployments was red except for CDM checks, RabbitMQ, and Swift. Anything relying on Keystone will fail:

- All API local checks
- Most (if not all) service checks (Horizon service check may not be because we log in through the web page)

Our swift checks are mostly exempt because we use the swift cli tools to grab that (dispersion, recon) information.

sigmavirus24:
Sorry that I was thinking about [QUIC](http://tools.ietf.org/html/draft-tsvwg-quic-protocol-00).

Issue #199, opened 2015-06-27
MaaS plugins failing


git-harry:
The format of the token expiry string, returned by keystone, is different when using fernet tokens. This will break most/all of the OpenStack MaaS plugins. I've only tested this in master, but I think it will also affect kilo. I've logged a bug on launchpad [1] on the assumption this was not intentional.

./cinder_service_check.py --host 'maas2-node4' '172.29.236.3'
status error n.py", line 331, in is_token_expired\n    '%Y-%m-%dT%H:%M:%SZ')\n  File "/usr/lib/python2.7/_strptime.py", line 325, in _strptime\n    (data_string, format))\nValueError: time data '2015-06-28T05:01:12.403672Z' does not match format '%Y-%m-%dT%H:%M:%SZ'\n
Traceback (most recent call last):
  File "./cinder_service_check.py", line 88, in <module>
    main(args)
  File "./cinder_service_check.py", line 75, in main
    auth_ref = get_auth_ref()
  File "/usr/lib/rackspace-monitoring-agent/plugins/maas_common.py", line 341, in get_auth_ref
    if is_token_expired(auth_ref['token']):
  File "/usr/lib/rackspace-monitoring-agent/plugins/maas_common.py", line 331, in is_token_expired
    '%Y-%m-%dT%H:%M:%SZ')
  File "/usr/lib/python2.7/_strptime.py", line 325, in _strptime
    (data_string, format))
ValueError: time data '2015-06-28T05:01:12.403672Z' does not match format '%Y-%m-%dT%H:%M:%SZ'

[1] https://bugs.launchpad.net/keystone/+bug/1469458

prometheanfire:
This is designated for 11.1 iirc, so tagging kilo backport

nrb:
Bumping to prio-1 and r11.0.0rc9 per PR #286.

nrb:
Merged into master and kilo

PR #286, opened 2015-07-27
Update MaaS scripts to use Keystone v3

sigmavirus24:
Now that os-ansible-deployment uses Keystone v3 by default, we need to
upgrade MaaS to use v3 as well. v3 is necessary for Keystone
Federation/ADFS happening usptream which we will be consuming, so we
need to upgrade our monitoring scripts to adjust for that.

Co-Authored-By: Darren Birkett <darren.birkett@rackspace.co.uk>
Addresses #273
Addresses #199

Conflicts:
	maas/maas_common.py

To simplify dependency management for this cherry-pick, this also
includes changes in 936f23971e35b66f68e5be994f477d0632e6deb5 which were
never backported to Kilo.
(cherry picked from commit 182e5623405afffc44af42f2a5f906f4f7e8bb23)

### DIFF
--- a/maas/heat_api_local_check.py
+++ b/maas/heat_api_local_check.py
@@ -54,7 +54,7 @@ def check(args, tenant_id):
 
 def main(args):
     auth_ref = get_auth_ref()
-    tenant_id = auth_ref['token']['tenant']['id']
+    tenant_id = auth_ref['project']['id']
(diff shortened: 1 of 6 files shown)


nrb:
+1

prometheanfire:
:+1:

PR #288, opened 2015-07-27
Fix Galera Cluster Rejoin Failure on Upgrade

### LINKED ISSUE #284: Galera wsrep_sst_method=xtrabackup causing cluster failures when nodes attempt to rejoin
This is an osad bug that will need to be fixed for upgrades in rpc-openstack

https://bugs.launchpad.net/openstack-ansible/+bug/1478105

claco:
An outdated wsrep_sst_method setting will cause cluster members to fail
to rejoin with an error during upgrades. Using xtrabackup-v2 instead of
xtrabackup fixes this. Bumping os-ansible-deployment to consume the fix.

Fixes #284

(cherry picked from commit 03e6f7fb81b609f928dd730892ea2bb1a73d3be1)


nrb:
+1

sigmavirus24:
:+1:

PR #289, opened 2015-07-27
Fix yaml linting errors in rpcd

nrb:
This change cleans up the errors seen in the linting script. One line in
particular in playbooks/elasticsearch.yml looks like it was split to
maintain line length, but there isn't currently an Ansible yaml rule
for that, at least not one that is defined in ansible-lint's defaults.

Addresses #240


sigmavirus24:
:+1: :+1: :+1: :+1: 

Thank you for getting rid of the JSON syntax =P

prometheanfire:
:+1:

PR #290, opened 2015-07-27
Fix yaml linting errors in rpcd

### LINKED ISSUE #240: Add YAML lint/sanity checking to rpc-openstack

nrb:
This change cleans up the errors seen in the linting script. One line in
particular in playbooks/elasticsearch.yml looks like it was split to
maintain line length, but there isn't currently an Ansible yaml rule
for that, at least not one that is defined in ansible-lint's defaults.

Fixes #240
(cherry picked from commit 156002db2ca4fa06a9709b8789091fa72b38af90)

### DIFF
--- a/rpcd/playbooks/beaver.yml
+++ b/rpcd/playbooks/beaver.yml
@@ -21,71 +21,71 @@
     - { role: "beaver" }
   vars:
     beaver_log_monitors:
-      - { name: "cinder",
-          log_file: "/var/log/cinder/*",
-          tags: "cinder,openstack,oslofmt",
-          multiline_regex_before: '.*\sTRACE\s.*' }
-      - { name: "nova",
-          log_file: "/var/log/nova/*",
-          tags: "nova,openstack,oslofmt",
-          multiline_regex_before: '.*\sTRACE\s.*' }
-      - { name: "heat",
-          log_file: "/var/log/heat/*",
-          tags: "heat,openstack,oslofmt",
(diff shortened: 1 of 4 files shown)


sigmavirus24:
+1 backport

prometheanfire:
:+1:

PR #291, opened 2015-07-27
Add a linting script for rpcd.

### LINKED ISSUE #240: Add YAML lint/sanity checking to rpc-openstack

nrb:
Linting for rpc-openstack should only cover the RPC-specific playbooks,
roles, scripts, and modules, since
os-ansible-deployment/openstack-ansible already does their own upstream
linting.

This change also uses relative paths in the ansible.cfg file so that
linting will work from Travis's container-based infrastructure. With the
hardcoded paths, we would have had to use sudo in the script, and add
configuration options to our .travis.yml. Also, their sudo-capable
infrastructure is considered legacy, so it's probably best avoided.

The relative paths mean we should be able to lint from whatever
directory we want, and if working within a virtualenv, run linting
without being root.

Also, the flake8 ignores and excludes were moved into a tox.ini file to
reduce line noise.

Fixes #240 

(cherry picked from commit e5fbfdc697350bd19599c32445043dfb912f0b58)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,3 @@
+language: python
+
+script: ./scripts/linting.sh
--- a/rpcd/playbooks/ansible.cfg
+++ b/rpcd/playbooks/ansible.cfg
@@ -12,23 +12,23 @@ timeout = 120
 
 # Set the path to the folder in os-ansible-deployment which holds the dynamic
 # inventory script - new config setting for ansible v1.9 and above
-inventory = /opt/rpc-openstack/os-ansible-deployment/playbooks/inventory/
+inventory = ../../os-ansible-deployment/playbooks/inventory/
 
 # Set the path to the folder in os-ansible-deployment which holds the dynamic
 # inventory script - uncomment if using ansible below v1.9
-#hostfile = /opt/rpc-openstack/os-ansible-deployment/playbooks/inventory/
+#hostfile = ../../os-ansible-deployment/playbooks/inventory/
 
 # Set the path to the folder in os-ansible-deployment which holds the
 # libraries required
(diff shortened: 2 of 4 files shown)


prometheanfire:
:+1:

odyssey4me:
:+1: for a basic linting check - it'd be good to implement role implementations, etc and idempotency checks but I think it'd be better to do that in follow-up patches.

PR #297, opened 2015-07-29
set uri correctly for maas checks by default

### LINKED ISSUE #249: MaaS check fails on self-signed certificates
Hi,
I'm starting to see this issue more. All have taken place in customer datacenters. Self signed certificates are being used on the LB. When MaaS goes to check if Horizon is up, the check fails. I can see the error when I try to curl the vip ip: "curl -vvv "vip_ip".  However, if i do a --insecure curl connection (curl -vvnk "vip_ip"), I get back a 200 ok along with the website content. 

Https checks need to account for self signed certificates.

Best Regards,

Jeff Telesford
OpenStack Engineer
Rackspace Hosting

prometheanfire:
it seems that by default we are enabling ssl for horizon (self signed)
because of this we need to set the default uri for checks to be https
or maas will get confused (the check already defaults to port 443 (hard
coded)).

fixes #249

### DIFF
--- a/rpcd/playbooks/roles/rpc_maas/defaults/main.yml
+++ b/rpcd/playbooks/roles/rpc_maas/defaults/main.yml
@@ -121,7 +121,7 @@ maas_scheme: http
 # maas_keystone_scheme: https
 # maas_neutron_scheme: https
 # maas_nova_scheme: https
-# maas_horizon_scheme: https
+maas_horizon_scheme: https
 # maas_heat_api_scheme: https
 # maas_heat_cfn_scheme: https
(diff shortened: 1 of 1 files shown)


odyssey4me:
Yep, Horizon has always had SSL enabled by default with a self signed cert. I'm not sure why this hasn't been set this way from the start! :+1:

prometheanfire:
Just a resync with master, also looks like we already set the path in master at least.

nrb:
+1

Apsu:
:+1:

d34dh0r53:
LGTM

Issue #308, opened 2015-08-05
on upgrades, the galera restart fails due to the new cluster name in kilo


mancdaz:
In juno, the galera cluster name is ```rpc_galera_cluster```. In kilo it was changed to ```openstack_galera_cluster```.

On an upgrade, some config file changes are made, necessitating a restart of the galera nodes.  The galera play restarts them one at a time, in a rolling upgrade fashion.  Unfortunately this means that the first node to restart is now trying to connect to a cluster called ```openstack_galera_cluster```, and the other 2 are still running a cluster called ```rpc_galera_cluster```, so the restart times out and fails.

We need to override this back to rpc_galera_cluster in:

```
 rpcd/etc/openstack_deploy/user_variables.yml
```

nrb:
Master merged, needs kilo backport.

nrb:
Both issues merged, closing.

PR #309, opened 2015-08-05
keep galera_cluster_name the same as juno

mancdaz:
A juno deploy will have galera_cluster_name set to rpc_galera_cluster.
When upgrading to kilo, we need to keep this name the same (osad kilo
changes the default to openstack_galera_cluster) otherwise rolling
restarts of the cluster during upgrades will timeout and fail.

Closes: #308


claco:
:+1:

nrb:
+1

Issue #310, opened 2015-08-05
the rpc-openstack upgrade.sh script does not merge in the rpcd specific user_variables


mancdaz:
In ```scripts/deploy.sh``` we have this:

```
pushd ${RPCD_DIR}
        for i in $(find etc/openstack_deploy/ -type f -iname '*.yml'); do ../scripts/update-yaml.py /$i $i; done
popd
```

This does a nice job of merging in all the ```rpcd/etc/openstack_deploy/*.yml``` user variables files, meaning all our overrides work nicely.

In ```scripts/upgrade.sh```, we do not have this. This script actually calls ```deploy.sh```, but not until it's too late (i.e. after the osad ```run-upgrade.sh``` has run) so the rpcd specific variables do not get merged in. They therefore get ignored on upgrades.

We need to run the above three liner at the top of the ```rpc-openstack/scripts/upgrade.sh```. Something like:

```
diff --git a/scripts/upgrade.sh b/scripts/upgrade.sh
index 3ca67a8..6649436 100755
--- a/scripts/upgrade.sh
+++ b/scripts/upgrade.sh
@@ -18,6 +18,11 @@ set -eux pipefail

 BASE_DIR=$( cd "$( dirname ${0} )" && cd ../ && pwd )
 OSAD_DIR="$BASE_DIR/os-ansible-deployment"
+RPCD_DIR="$BASE_DIR/rpcd"
+
+pushd ${RPCD_DIR}
+    for i in $(find etc/openstack_deploy/ -type f -iname '*.yml'); do ../scripts/update-yaml.py /$i $i; done
+popd

 # Do the upgrade for os-ansible-deployment components
 cd ${OSAD_DIR}
```

mancdaz:
actually not sure this works.  We've not yet created the ```/etc/openstack_deploy/*``` files yet to update at this point.  Need to figure something out to ensure we're using the rpcd specific user variables on upgrade though.

nrb:
Yeah - we'd need to inject this in somehow; I'm not sure we have a good way to do that right now. Might this necessitate a hook in the OSAD upgrade scripts?

nrb:
On more thought, I have the following proposal:

Move [update-yaml.py](https://github.com/rcbops/rpc-openstack/blob/master/scripts/update-yaml.py) into OSAD, and modify the OSAD upgrade script to take the override file as a parameter.

[...]

PR #314, opened 2015-08-05
Ensure Kilo user_variables are present for upgrade

stevelle:
New user_variables introduced for Kilo will not be present at time of
upgrade due to the upstream (os-ansible-deployment) run-upgrade.sh
script's processing flow. Applying them to the pre-upgraded
user_variables ensures they are available as upstream upgrade runs
playbooks.

Related to #310


mancdaz:
:+1:

stevelle:
@sigmavirus24 pointed out an issue where existing overrides could be replaced by new values.
Updated patch.

nrb:
:+1: for me. Merging.

PR #319, opened 2015-08-06
Ensure Kilo user_variables are present for upgrade

claco:
New user_variables introduced for Kilo will not be present at time of
upgrade due to the upstream (os-ansible-deployment) run-upgrade.sh
script's processing flow. Applying them to the pre-upgraded
user_variables ensures they are available as upstream upgrade runs
playbooks. New values should have lower precedence than existing
overrides.

Related to #310

(cherry picked from commit 0a415f188e63675b207029eb5888a84881eda1a7)


rackertom:
:+1:

sigmavirus24:
:+1:

Issue #322, opened 2015-08-07
we need to ensure we are getting the correct keystone v3 api endpoint in maas_common


mancdaz:
in ```maas_common```, as part of the ```get_*_client``` methods, we call the ```get_endpoint_url_for_service``` to get the endpoint for a particular service type out of the json catalog.

Now that we're using keystone v3 api in the rest of the maas plugins, we need to ensure we grab the correct v3 keystone endpoint.  The problem is the identity part of the catalog looks like this:

```
    {
      "endpoints": [
        {
          "url": "http:\/\/172.29.236.3:35357\/v2.0",
          "interface": "admin",
          "region": "RegionOne",
          "region_id": "RegionOne",
          "id": "69d833769ef5490db0f0c52d7bc0ee30"
        },
        {
          "url": "http:\/\/172.29.236.3:35357\/v3",
          "interface": "admin",
          "region": "RegionOne",
          "region_id": "RegionOne",
          "id": "882843da8a864da3861977903d89dca8"
        },
        {
          "url": "http:\/\/172.29.236.3:5000\/v2.0",
          "interface": "internal",
          "region": "RegionOne",
          "region_id": "RegionOne",
          "id": "99d44a7b8e9f4c78888f8956110b40ea"
        },
        {
          "url": "http:\/\/162.13.137.186:5000\/v2.0",
          "interface": "public",
          "region": "RegionOne",
          "region_id": "RegionOne",
          "id": "c402dad761664134ba1e51d346f4c1ce"
        }
      ],
      "type": "identity",
      "id": "7a862c3625a349009af4cfd543f2808f",
      "name": "keystone"
    },
```

and the get_url_endpoint_for_service method does this:

```
def get_endpoint_url_for_service(service_type, service_catalog,
                                 url_type='public'):
    for service in service_catalog:
        if service['type'] == service_type:
            for endpoint in service['endpoints']:
                if endpoint['interface'] == url_type:
                    return endpoint['url']
```

The problme is when we call this method from the ```get_keystone_client``` method, we are asking for the admin interface.  There are 2, and we end up getting the first match back(which is the v2 one).

We just need to add a bit of logic to ensure we always return the v3 admin endpoint for keystone specifically.

stevelle:
needs a forward port still

stevelle:
merged and backport merged. closing.

PR #323, opened 2015-08-07
add 'version' param to get_url_endpoint_for_service method

mancdaz:
Sometimes we have multiple versions of a particular endpoint type for a service.  This change allows us to pass a 'version' which is the last part of the url. e.g. for keystone we have both the v2 and v3 endpoints deployed, and in the catalog they look like this:

http://172.29.236.3:35357/v3
http://172.29.236.3:35357/v2.0

To ensure we return the v3 endpoint, you'd pass version='v3', to get the v2 endpoint, you'd pass version='v2.0'

The default bahaviour is retained, which is to pass no version and return the first match

Closes: #322

### DIFF
--- a/maas/maas_common.py
+++ b/maas/maas_common.py
@@ -208,7 +208,7 @@ else:
         if not endpoint:
             endpoint = get_endpoint_url_for_service('identity',
                                                     auth_ref['catalog'],
-                                                    'admin')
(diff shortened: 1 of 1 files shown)


prometheanfire:
:+1:

stevelle:
:+1:

Issue #321, opened 2015-08-07
after juno -> kilo upgrade, the auth_ref on disk that maas-common uses is in the old keystone apiv2 format


mancdaz:
On first run of maas plugins after an upgrade, part of the functionality of ```maas_common``` is to check if the current auth_ref stored on disk is valid (and if the token has expired). We are checking the 'new format' auth_ref that keystone api v3 spits out, but the auth_ref on disk is one that api v2 produced before the upgrade.  We therefore get an error when running the ```is_token_expired``` function in ```maas_common```:

https://github.com/rcbops/rpc-openstack/blob/master/maas/plugins/maas_common.py#L329-L338

(Apologies, I lost the actual error. You'll have to trust me, it's a stacktrace and a KeyError)

We have 2 options:

- either have the upgrade script basically ```rm /root/.auth_ref.json```
- add logic to the function in maas_common to handle both formats inside this function (yes, I'm changing the name of the token parameter to auth_ref. Because this isn't a token. It's an auth_ref):
```
diff --git a/maas/maas_common.py b/maas/maas_common.py
index 80375fb..9e8d091 100755
--- a/maas/maas_common.py
+++ b/maas/maas_common.py
@@ -326,10 +326,13 @@ class MaaSException(Exception):
     """Base MaaS plugin exception."""


-def is_token_expired(token):
+def is_token_expired(auth_ref):
     for fmt in ('%Y-%m-%dT%H:%M:%SZ', '%Y-%m-%dT%H:%M:%S.%fZ'):
         try:
-            expires = datetime.datetime.strptime(token['expires_at'], fmt)
+            expires = datetime.datetime.strptime(auth_ref['expires_at'], fmt)
+            break
+        except KeyError:
+            expires = datetime.datetime.strptime(auth_ref['token']['expires'], fmt)
             break
         except ValueError as e:
             pass
```

mancdaz:
FYI, I prefer the rm option, since it's cleaner and less risky.

mancdaz:
hmm, except that doesn't take care of all nodes, only the deployment host...

mancdaz:
actually on second thoughts, patching the maas_common file to handle both formats is more seamless in the upgrade process.

mancdaz:
on third thoughts, this bites us in a few more places.  We just need to throw in an ansible one-liner right at the end of the upgrade-script to delete that file from all nodes.

stevelle:
Backport didn't link up, but here it is.
[...]

PR #325, opened 2015-08-07
remove auth_ref on disk after upgrade finishes

mancdaz:
The auth_ref is in a format that the new plugins can't read.  We remove
it at the end of the script.  The next time a plugin is called, a new
auth_ref gets generated in the new format, so this is a one time thing
after upgrade.

Closes: #321


prometheanfire:
:+1: lgtm

nrb:
+1, looks like a reasonable change.

PR #327, opened 2015-08-07
remove auth_ref on disk after upgrade finishes

stevelle:
The auth_ref is in a format that the new plugins can't read.  We remove
it at the end of the script.  The next time a plugin is called, a new
auth_ref gets generated in the new format, so this is a one time thing
after upgrade.

(cherry picked from commit b4b39b0608c683c8f9282635d4037f0b4e08ab22)


nrb:
Port, +1

prometheanfire:
:+1:

Issue #324, opened 2015-08-07
if an upgrade fails part way through, we need to be clear about what is left to do


mancdaz:
the ```run-upgrade.sh``` script from osad does a decent job of saying what's left to run, if it fails part way through.  

However, because ```run-upgrade.sh``` is actually being called from within  ```upgrade.sh```, there's also stuff from ```upgrade.sh``` that needs to be finished off - that is, re-running ```deploy.sh``` without deploying osad again, so that we update all the support/backup/maas stuff too)

d34dh0r53:
@claco @mancdaz Can this be closed?

mancdaz:
@d34dh0r53 I'm not sure. Did we add anything to our upgrade.sh to fix this? Seems like the last 2 attempts were abandoned.

mancdaz:
@d34dh0r53 I don't think this issue is fixed is it?

odyssey4me:
The OSA review that relates to this (I think) was recently abandoned by @nrb - https://review.openstack.org/237689 - if this functionality is desired then it should be easy enough to restore, rebase and update. I would suggest that perhaps now that the new upgrade script is the QE'd method that the old one should be removed.

nrb:
No, it's not - should I reopen that upstream patch, but only target it for the current process?

mancdaz:
That upstream patch relates to failures in the osa playbooks. This issue
was raised because we don't have anything in the rpc-openstack
deploy/upgrade scripts that accounts for failures outside the osa upgrade
script

nrb:
Ah, yeah, good point.

I think we need both.

nrb:
Does this need to be backported to Kilo, since we're looking at Liberty work now?

mancdaz:
@nrb I'd consider it a bug (i.e. not a feature, because kilo is now stable), and given that near enough nobody has yet been through the juno -> kilo upgrade, it would be worth having in kilo

mancdaz:
@d34dh0r53 I'd suggest this is made a medium/high priority targeted at 11.1.3 - fixed in master and backported to kilo, of course

stevelle:
This looks like it should be closed one way or another since the upstream change was abandoned and Juno->Kilo upgrades are running individual operations.

b3rnard0:
@d34dh0r53 Sounds like this should be closed.

git-harry:
[...]

PR #329, opened 2015-08-07
Print remaining steps if upgrade.sh fails

rackertom:
If the upgrade script fails to complete successfully, print a
warning for the user and tell them what remaining steps need to
be run manually.

Closes: #324

### DIFF
--- a/scripts/upgrade.sh
+++ b/scripts/upgrade.sh
@@ -14,7 +14,26 @@
 # limitations under the License.
 #
 # (c) 2015, Nolan Brubaker <nolan.brubaker@rackspace.com>
-set -eux pipefail
+set -ux pipefail
+
+FAILED=0
+
+function run_or_print() {
+        command="$@"
+        if [ $FAILED -ne 0 ]; then
+                echo "  ${command}"
+        else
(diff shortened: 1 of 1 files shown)


scripts/upgrade.sh
```diff
@@ -14,7 +14,26 @@
...
+                echo "  ${command}"
+        else
+                eval "$command"
+                FAILED=$?
+                if [ $FAILED -ne 0 ]; then
+                    echo "******************** FAILURE ********************"
+                    echo "The upgrade script has failed. Please rerun the following task to continue"
+                    echo "Failed on task ${command}"
+                    echo "Do NOT rerun the upgrade script!"
+                    echo "Please execute the remaining tasks:"
```
nrb: Since we're not printing the remaining tasks with this approach, we may want to be clear that the user will have to look at the script to find them.

rackertom: As we discussed in the channel, we actually are printing the remaining tasks so that the user won'y have to look at the script's contents.

rackertom:
Abandoning this PR and creating a new one with a wrapper script created by Evan.

PR #330, opened 2015-08-07
Allow users to resume upgrade execution by using resume.sh wrapper

rackertom:
This update includes a wrapper script written by Evan which receives a list of commands to execute on stdin. If an error occurs while executing a step, the script informs the user and gives instructions for resuming.

Closes: #324

### DIFF
--- a/scripts/resume.sh
+++ b/scripts/resume.sh
@@ -1,0 +1,60 @@
+#!/usr/bin/env bash
+
+# Script originally written by Evan Callicoat
+#
+# Execute each command from stdin. If the command fails
+# output a sensible error message and include instructions
+# to allow the user to resume execution at the next step.
+
+# Set options
+set -euo pipefail
+
+# Initial step count
+STEP=0
+
+# Handle errors
+handle_error() {
+    # Store command that caused error
+    cmd="$BASH_COMMAND"
+
+    # If we didn't run any steps yet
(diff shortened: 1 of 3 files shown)


Apsu:
Looks about right

nrb:
Two small things that are mostly clerical - It'd be preferable to have this squashed, and also remove the issue number from the commit message.

The code looks good and makes sense to me, though.

prometheanfire:
do we care about licensing?

prometheanfire:
other then that and squashing I like it

mancdaz:
sorry to come in late on this, but wouldn't it be possible to just have all this logic in the upgrade script and have a '--resume' flag that we pass if we need to resume? Otherwise we just have wrappers wrapping wrappers all over the place

resume.sh > upgrade.sh > run-upgrade.sh / deploy.sh

nrb:
Yeah, the concept was that we make resume.sh generic enough to use elsewhere...however, I think if this isn't going to make 11.0.0, we should just move the logic to an Ansible playbook, at this point.

rackertom:
The more I think about this approach, the less I like it. This method means we need to make changes in two places many times which we will mess up. In addition, there are tons of corner cases where this won't work (loops, if statements, etc.).

:-1:

d34dh0r53:
Abandoned

Issue #326, opened 2015-08-07
after upgrades, some maas checks have old names


mancdaz:
in kilo, the nova_spice container is renamed to nova_console. After an upgrade, we have 2 issues.  Some checks that are pointing at a non-existent container, and some checks that are newly created at the new nova_console container, but are named after the old container:

remaining checks after an upgrade, for a container that no longer exists:

```
nova_consoleauth_check--heat-node1_nova_spice_console_container-3dd09091
nova_spice_console_check--heat-node1_nova_spice_console_container-3dd09091
```

And I had 2 new checks that were now pointing at the new nova_console container:

```
nova_consoleauth_check--heat-node1_nova_console_container-b14bb7aa
nova_spice_console_check--heat-node1_nova_console_container-b14bb7aa
```

We need to add a doc step somewhere (or something else?) to delete those old checks. But I wonder, do we want to rename the nova-spice-console-check to nova-console-check

prometheanfire:
iirc, the expectation is that maas checks are removed as part of the upgrade so that they can be created from scratch.  If that is the case then this bug can be closed.

claco:
@prometheanfire expectation by whom? Not sure we spell it out in the docs that the deployer needs to delete those. Do we automate the delete as part of the upgrade? Can you sift through the docs w/ a docs person to see if we mention this as part of the 11.0 upgrade docs?

prometheanfire:
the memory comes from talking with support a couple of months ago about monitoring.  could easily be wrong.

claco:
easy to verify. please reach out to them and find out. :-)

mancdaz:
Deleting the checks that are now against a non existent container is easy.
Deleting the others that are created with the wrong name won't help,
because they will be recreated again with the wrong name. The ansible that
creates the checks needs to be updated.

prometheanfire:
ah, also, support doesn't delete checks between upgrades

claco:
@mancdaz @prometheanfire 

So for me, this means:

```
# add docs release notes to delete
nova_consoleauth_check--heat-node1_nova_spice_console_container-3dd09091
nova_spice_console_check--heat-node1_nova_spice_console_container-3dd09091
```

Because the idea is that they're pluggable in the future, we should make the sec one one generic like the first:

```
nova_consoleauth_check--heat-node1_nova_console_container-b14bb7aa
nova_console_check--heat-node1_nova_console_container-b14bb7aa
```

The first half being a higher priority at the moment.

[...]

PR #333, opened 2015-08-10
Rename spice console check

stevelle:
Fix name of maas check created through playbook and related tests to
reflect the general purpose rather than the technology choice used for
nova consoles.

Related to #326


prometheanfire:
lgtm :+1:

nrb:
+1, upgrade related.

PR #337, opened 2015-08-12
Set `ssl_check` to `true` to match previous behavior

### LINKED ISSUE #336: Kilo - MaaS SSL Check Not Created
The SSL Check is no longer getting setup by default. A new flag was added to control the setup, and its default is set to false. This check was getting setup by default in prior versions.

https://github.com/rcbops/rpc-openstack/blob/master/rpcd/playbooks/roles/rpc_maas/tasks/main.yml#L50

https://github.com/rcbops/rpc-openstack/blob/master/rpcd/playbooks/roles/rpc_maas/defaults/main.yml#L91

byronmccollum:
Set `ssl_check` to `true` to ensure check gets created.

Fixes #336

### DIFF
--- a/rpcd/playbooks/roles/rpc_maas/defaults/main.yml
+++ b/rpcd/playbooks/roles/rpc_maas/defaults/main.yml
@@ -88,7 +88,7 @@ maas_target_alias: public0_v4
 #
 # ssl_check: This flag indicates if the ssl-related checks and alarms should be deployed.
 #
-ssl_check: false
+ssl_check: true
 
 #
 # dell_check: This flag indicates if the Dell-related hardware monitoring checks and alarms should


stevelle:
Trying to figure out why this is set to false now but if someone knows and can answer that would be helpful.

andymcc:
I think this was set to false because we were defaulting to not using SSL on endpoints. Happy for it to change though.

byronmccollum:
Except for Horizon, which is SSL enabled.

claco:
But that is usually not SSL terminated in the LB. With the others, the setting has wider ramifications.

byronmccollum:
The new `ssl_check` setting is used in exactly one place, a remote check for Horizon's SSL cert. Equivalent checks for API endpoints don't even exist.

prometheanfire:
Looks like this is related to this bug.  https://github.com/rpcdocs/rpcdocs/pull/174

Does this happen with our current tests? I thought this was fixed.

prometheanfire:
ping?

prometheanfire:
ping @byronmccollum

byronmccollum:
@prometheanfire So...there are a couple things at play.

Horizon is configured for SSL access, regardless of any of these others changes.

#249 works around an issued with the Cloud Monitoring infrastructure. That fix involves changing the check protocol to HTTPS, and specifying the full check URL to eliminate any redirects, which triggers the aforementioned issue with Cloud Monitoring (the issues has more details).

In v10, there is an SSL check for Horizon, which monitors the expiration of Horizon's SSL certificate. This check was getting created my default.

In v11, a new flag was introduced `ssl_check`, and has been defaulted to `false`. This PR just reinstates the previous behavior of v10 by setting the value to `true`.

prometheanfire:
ok then, rebase plese

prometheanfire:
:+1:

prometheanfire:
it seems that it's only used here, so limited exposure.

    rpcd/playbooks/roles/rpc_maas/tasks/main.yml:    inventory_hostname == groups['utility'][0] and ssl_check == true

stevelle:
:+1: lgtm now that we have an explain

Issue #344, opened 2015-08-14
Allow capability to patch OSAD submodule files.


nrb:
There may be times where it is desirable for rpc-openstack to change the SHA of the os-ansible-deployment submodule, but not necessarily match a tag. In order to pull in improvements between tags with minimal code disruption, we should provide some sort of patching mechanism.

nrb:
Closing this, as the change is merged.

PR #345, opened 2015-08-14
Add patcher role

nrb:
There may arise times where it is beneficial for the rpc-openstack
project to use a SHA unrlated to an os-ansible-deployment tag, but still
apply some patches to the os-ansible-deployment files.

This role will patch files in the os-ansible-deployment submodule of
rpc-openstack. It does so prior to an actual deployment of the stack,
and thus must use the local machine, not one from Ansible's inventory.

Also, when run during upgrades, Ansible in Juno doesn't have the
'patch' module, so we upgrade that prior.

This initial version includes a patch incorporating the following fixes
from upstream:

    https://review.openstack.org/#/c/212021/

Addresses #344


nrb:
Whoops - included the wrong patch. Closing this PR.

PR #346, opened 2015-08-14
Add patcher role

nrb:
There may arise times where it is beneficial for the rpc-openstack
project to use a SHA unrlated to an os-ansible-deployment tag, but still
apply some patches to the os-ansible-deployment files.

This role will patch files in the os-ansible-deployment submodule of
rpc-openstack. It does so prior to an actual deployment of the stack,
and thus must use the local machine, not one from Ansible's inventory.

Also, when run during upgrades, Ansible in Juno doesn't have the
'patch' module, so we upgrade that prior.

The first patch included is functionally equivalent to this upstream
change:
    https://review.openstack.org/#/c/212021/

Address #344


sigmavirus24:
Eh, I'm -0.5 on this. `httpd.patch` isn't the greatest name for this patch file. Can we use a better name for it that's more descriptive?

Apsu:
:+1:

miguelgrinberg:
:+1:

sigmavirus24:
:+1:

PR #353, opened 2015-08-19
Clean up MaaS and local scripts for Flake8 (kilo backport)

nrb:
Fix all reported hacking errors

Addresses #240 
Addresses #272
(cherry picked from commit e00db31c320e22627282c384eca6acd2d600040c)

### DIFF
--- a/maas/maas_common.py
+++ b/maas/maas_common.py
@@ -13,6 +13,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from __future__ import print_function
 
 import contextlib
 import datetime
@@ -34,7 +35,7 @@ TOKEN_FILE = '/root/.auth_ref.json'
 
 
 try:
-    from cinderclient.client import Client as c_client
+    from cinderclient import client as c_client
     from cinderclient import exceptions as c_exc
 
 except ImportError:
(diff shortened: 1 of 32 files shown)


maas/maas_common.py
```diff
@@ -71,7 +72,7 @@ def get_cinder_client(previous_tries=0):
         return cinder
 
 try:
-    from glanceclient import Client as g_client
+    import glanceclient as g_client
```
d34dh0r53: Little nit here, later on line 121 we're importing novaclient.client as nova_client, we should probably keep with that convention and call this glance_client.

maas/maas_common.py
```diff
@@ -169,8 +171,8 @@ def get_nova_client(auth_token=None, bypass_url=None, previous_tries=0):
         return nova
 
 try:
-    from keystoneclient.v3 import client as k_client
     from keystoneclient.openstack.common.apiclient import exceptions as k_exc
+    from keystoneclient.v3 import client as k_client
```
d34dh0r53: Same as above, I'd like to see this called keystone_client.

maas/maas_common.py
```diff
@@ -231,8 +233,8 @@ def get_keystone_client(auth_ref=None, endpoint=None, previous_tries=0):
 
 
 try:
-    from neutronclient.neutron import client as n_client
     from neutronclient.common import exceptions as n_exc
+    from neutronclient.neutron import client as n_client
```
d34dh0r53: Change to neutron_client

d34dh0r53:
A couple of nits but otherwise looks good, I'll give this a +1 and ignore my nits if this is time sensitive.

nrb:
@d34dh0r53 While I agree with you, this is also a backport, so I'm hesitant to make the change in this PR. We can file a clean up issue for renaming the clients, though.

Apsu:
:+1:

d34dh0r53:
:+1:

Issue #253, opened 2015-07-15
Need to add a DEPLOY_TEMPEST variable to deploy.sh so we can deploy tempest in rpc-openstack


jacobwagner:
Currently we have no way to tell rpc-openstack that we want tempest deployed with osad. A simple DEPLOY_TEMPEST variable and if block inside deploy.sh would help alleviate this.

PR #359, opened 2015-08-24
Disable running the MaaS playbooks by default.

### LINKED ISSUE #355: Move MaaS playbooks out of upgrade process
The MaaS checks likely need manual intervention, and thus should be moved out of the automated upgrade script process.

nrb:
Per support's request, the MaaS playbooks have been defaulted to not
running in the deploy script.

This affects both new deploys and upgrades, since the upgrade script
calls deploy.sh to lay down the new RPC components in kilo.

Doing the change this way means that both deploys and upgrades behave in
a unified manner.

(cherry picked from commit 37489583d7c710cfeb6d6e4ececb290466f73352)

Fixes #355


d34dh0r53:
:+1:

prometheanfire:
do we have a issue to link the tempest change to?

other then that, :+1:

nrb:
The tempest inclusion is per https://github.com/rcbops/rpc-openstack/issues/253

Issue #360, opened 2015-08-24
Disable input prompt during upgrade


nrb:
Right now, our `upgrade.sh` script does some pre-setup, then runs OSAD's `run-upgrade.sh`. This has a side-effect of appearing to do work, then stopping 'in the middle' to look for input.

We should probably silence the input prompt when running `run-upgrade.sh`.

claco:
Backport: https://github.com/rcbops/rpc-openstack/pull/363

d34dh0r53:
Closing as all PRs have merged.

PR #361, opened 2015-08-24
Disable prompt on run-upgrade.sh

nrb:
This change makes the script uninterrupted (in the happy path case), so
that users don't have to enter input during the process.

Address #360


ionosphere80:
Looks good, thanks!

prometheanfire:
:+1: lgtm

PR #363, opened 2015-08-24
Disable prompt on run-upgrade.sh

claco:
This change makes the script uninterrupted (in the happy path case), so
that users don't have to enter input during the process.

(cherry picked from commit e58b261af9160a0b0cadb6be978ae37f17e03a16)


prometheanfire:
:+1:

ionosphere80:
Works for me.

Issue #354, opened 2015-08-20
Juno to Kilo upgrade: lxc-container can not mount /var/run/netns and losing all namespaces


nrb:
Tracking upstream bug - https://bugs.launchpad.net/openstack-ansible/+bug/1487130

odyssey4me:
Does the neutron container work properly after the setup-openstack playbook is run? If so, then is this really an issue worth holding back for. While it is more ideal to have the upgrade run in a manner that's as online as possible, it's my understanding that the upgrades are done in a maintenance window and therefore down time is expected. As long as it works afterwards, this should be something we improve but it shouldn't hold back RPC11's release.

nrb:
It works properly after a container restart. @Apsu was looking into this on Friday, we think it might have been some of the temporary cleanup playbooks in the upgrade script.

claco:
@odyssey4me Same quesitons were asked the day we experienced it. It has been deemed a release blocker for support.

nrb:
In terms of the actual down time - users are told to expect short periods network interruption in the order of seconds while containers, rabbitmq, mariadb, and neutron itself restart. In our environment, though, the ping test to a neutron IP failed for 29 minutes.

cloudnull:
Did the issue resolve itself after the 29 minutes or was it only resolved after the neutron server containers were restarted? Do all of the neutron servers and agents need to be restarted?

nrb:
It solved itself over the course of the upgrade; those 29 minutes were basically going from the beginning of the script to the actual neutron upgrade steps.

cloudnull:
I would imagine that this issue is being caused by the addition to the openstack-lxc template as found here: https://github.com/stackforge/os-ansible-deployment/blob/kilo/playbooks/roles/lxc_container_create/tasks/container_create.yml#L80 

Would it be possible to review the neutron server lxc configs to see if the app aromor profile with has more than one entry. If so the last one in config is the only one loaded and I'd image that the change in the aa profile for the lxc container build process, which reiterates through all containers, is causing the issues and resolves itself when the neutron container becomes "unconfined" here: https://github.com/stackforge/os-ansible-deployment/blob/kilo/playbooks/os-neutron-install.yml#L27

cloudnull:
related fix that I believe will solve this issue: https://review.openstack.org/216301

nrb:
Thanks @cloudnull

claco:
Closing as fix went into upstream OSA prior to r11.0.0 release.

PR #364, opened 2015-08-24
Add patch for container configurations

nrb:
This is a patch for the upstream bug
https://bugs.launchpad.net/openstack-ansible/+bug/1487130.

It is included here in order to have it available more quickly for
rpc-openstack.

In terms of impact, this change moves the container configuration to
each service's installation. This way, the container is not brought down
early on in the upgrade process and left to sit until late. The
interruption window is far shorter in this way.

Addresses #354

### DIFF
--- a/rpcd/patches/lxc-container-config-lp-bug-1487130.patch
+++ b/rpcd/patches/lxc-container-config-lp-bug-1487130.patch
@@ -1,0 +1,551 @@
+diff --git c/playbooks/galera-install.yml w/playbooks/galera-install.yml
+index 77958f8..c8a74e3 100644
+--- c/playbooks/galera-install.yml
++++ w/playbooks/galera-install.yml
+@@ -67,6 +67,15 @@
+   max_fail_percentage: 20
+   user: root
+   pre_tasks:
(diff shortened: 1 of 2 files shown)


ionosphere80:
Looks good.

stevelle:
:+1:

nrb:
The patch included does *not* apply cleanly. I'm currently trouble shooting.

nrb:
Ok, the most recent version of this PR includes an LXC config file that cleanly applies.

cloudnull:
Patch looks fine to me however i can not :+1:  as this is work I originally created. That said it should solve the issues that we reported with neutron.

nrb:
Rebased this, as testing found that I missed some files in the previous version, namely the lxc_container_create role's changes.

nrb:
I've gotten a full AIO greenfield deploy to finish with this code as it stands. The containers did not get their full configuration until it was time to configure the service they contained, which is what we were expecting.

sigmavirus24:
:+1:

odyssey4me:
:+1:

Apsu:
:+1: Looks good to me now, and can verify it worked for me too.

PR #365, opened 2015-08-25
Change output redirect to pipe

nrb:
The output redirect will overwrite the run-upgrade script, which is not
what we want. This change properly sends the output to the stdin.

Addresses #360


miguelgrinberg:
:+1: 

Due to the severity of this problem, I'm merging this.

PR #366, opened 2015-08-25
Fix for auto-prompt in upgrade script

miguelgrinberg:
Tiny bug in auto-prompt logic.


miguelgrinberg:
This is a duplicate. Closing.

PR #368, opened 2015-08-25
Change output redirect to pipe

### LINKED ISSUE #360: Disable input prompt during upgrade

nrb:
The output redirect will overwrite the run-upgrade script, which is not
what we want. This change properly sends the output to the stdin.

(cherry picked from commit 7a11f099384db434b18c6117fe9dcd61c6143fc4)

Fixes #360

### DIFF
--- a/scripts/upgrade.sh
+++ b/scripts/upgrade.sh
@@ -36,7 +36,7 @@ openstack-ansible -i "localhost," patcher.yml
 
 # Do the upgrade for os-ansible-deployment components
 cd ${OSAD_DIR}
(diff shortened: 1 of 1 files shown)


d34dh0r53:
Assuming this passes gate :+1 from me

sigmavirus24:
Oh I see the problem. :+1:

Issue #369, opened 2015-08-25
Last step in upgrade script should use file module


sigmavirus24:
Otherwise, if `/root/.auth_ref.json` doesn't exist, then the upgrade script will fail and cause confusion.

nrb:
Is there an issue for this?

PR #370, opened 2015-08-25
Use file module to remove /root/.auth_ref.json

sigmavirus24:
Otherwise the upgrade.sh script looks as if it fails and causes
confusion.

Addresses #369


nrb:
:+1:

cloudnull:
:+1: LGTM

PR #374, opened 2015-08-26
Add patch for container configurations

sigmavirus24:
This is a patch for the upstream bug
https://bugs.launchpad.net/openstack-ansible/+bug/1487130.

It is included here in order to have it available more quickly for
rpc-openstack. There were conflicting hunks removed from
os-cinder-install.yml, os-glance-install.yml, and os-nova-install.yml
since they were only style changes, but were problematic to apply via
`patch`.

In terms of impact, this change moves the container configuration to
each service's installation. This way, the container is not brought down
early on in the upgrade process and left to sit until late. The
interruption window is far shorter in this way.

(cherry picked from commit 14aefd38406214b4dde69ecc93605b16de5aa221)


odyssey4me:
:+1: looks like the same thing to me - straightforward backport

cloudnull:
:+1:  LGTM!

PR #377, opened 2015-08-26
[WIP] Default all containers to unconfined profile

sigmavirus24:
Currently, the default profile still causes too much downtime and is
unstatisfactory. By defaulting this to unconfined, we hope to see
minimal downtime.

Addresses #354


prometheanfire:
:+1:

PR #378, opened 2015-08-27
Remove task that removes apparmor profile

sigmavirus24:
This is a backport of 2bec5f01a5a036367c3a25f762255656d4d9371a from
os-ansible-deployment. This prevents the problem where the neutron
containers lose network connectivity causing problems for running
instances.

Addresses #354


Apsu:
:+1:

d34dh0r53:
:+1:

PR #380, opened 2015-08-28
Ceph support

git-harry:
This adds support for Ceph, including:
  - deploying a Ceph cluster
  - integrating with openstack-ansible Ceph support
  - configuring monitoring and logging

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -11,11 +11,14 @@ export DEPLOY_ELK=${DEPLOY_ELK:-"yes"}
 export DEPLOY_MAAS=${DEPLOY_MAAS:-"no"}
 export DEPLOY_TEMPEST=${DEPLOY_TEMPEST:-"no"}
 export DEPLOY_CEILOMETER=${DEPLOY_CEILOMETER:-"no"}
+export DEPLOY_CEPH=${DEPLOY_CEPH:-"no"}
 export FORKS=${FORKS:-$(grep -c ^processor /proc/cpuinfo)}
 
 source /opt/rpc-openstack/openstack-ansible/scripts/scripts-library.sh
 OA_DIR='/opt/rpc-openstack/openstack-ansible'
 RPCD_DIR='/opt/rpc-openstack/rpcd'
+RPCD_VARS='/etc/openstack_deploy/user_extras_variables.yml'
+RPCD_SECRETS='/etc/openstack_deploy/user_extras_secrets.yml'
 
 # begin the bootstrap process
 cd ${OA_DIR}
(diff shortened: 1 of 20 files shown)


git-harry:
@nrb - the travis build appears to have failed because it is testing the third-party roles that are added as submodules by this pull request. What needs to be done to exclude them from the testing?

scripts/deploy.sh
```diff
@@ -49,7 +72,8 @@ fi
 which openstack-ansible || ./scripts/bootstrap-ansible.sh
 
 # ensure all needed passwords and tokens are generated
-./scripts/pw-token-gen.py --file /etc/openstack_deploy/user_extras_secrets.yml
```
claco: The bootstrap does user_secrets I thought, so the user_extras_secrets afterwords was intentional? So changing this back might be patch/merge snafu?

git-harry: @claco line 76 './scripts/pw-token-gen.py --file $RPCD_SECRETS' does the same thing as the line that was removed. pw-token-gen.py is only run by the bootstrap-aio.sh and run-upgrade.sh scripts in openstack-ansible.

mattt416:
Thanks @git-harry .  We will need to bump osad's sha however we'll need to wait for https://review.openstack.org/#/c/209537/ to merge and get backported first?

nrb:
I'm loathe to bump master's SHA until we get 11.0 actually released, but that's likely to be something for further discussion. In terms of excluding the external roles, we'll likely have to add a grep or find excluding those to [the ansible-lint line]ªhttps://github.com/rcbops/rpc-openstack/blob/master/scripts/linting.sh#L54).

Probably the [syntax checking line](https://github.com/rcbops/rpc-openstack/blob/master/scripts/linting.sh#L49), too.

git-harry:
@nrb - in the end we needed to delete the roles before ansible-lint runs - https://github.com/git-harry/rpc-openstack/blob/ceph-squash/scripts/linting.sh#L51 because the alternative was to exclude 5 playbooks.

nrb:
Ok, that works too.

mattt416:
Looks like ansible-ceph-common sha will need to be bumped to include https://github.com/ceph/ansible-ceph-common/commit/92f9f72bf94d79a4c988fea7c2d7a2da19b4edc7.

d34dh0r53:
Hey guys,

I'm getting the following error during the `ceph-client | Install ceph packages` play:
````bash
failed: [d34d-test1_glance_container-6879f982] => (item=({'client': [u'glance'], 'component': 'glance_api', 'service': ['glance-api']}, 'python-ceph')) => {"attempts": 5, "failed": true, "item": [{"client": ["glance"], "component": "glance_api", "service": ["glance-api"]}, "python-ceph"]}
stderr: E: There are problems and -y was used without --force-yes

stdout: Reading package lists...
Building dependency tree...
Reading state information...
The following extra packages will be installed:
  libboost-system1.54.0 libboost-thread1.54.0 libcephfs1 liblttng-ust-ctl2
  liblttng-ust0 libnspr4 libnss3 libnss3-nssdb librados2 librbd1 liburcu1
  python-cephfs python-rados python-rbd
The following NEW packages will be installed:
  libboost-system1.54.0 libboost-thread1.54.0 libcephfs1 liblttng-ust-ctl2
  liblttng-ust0 libnspr4 libnss3 libnss3-nssdb librados2 librbd1 liburcu1
  python-ceph python-cephfs python-rados python-rbd
0 upgraded, 15 newly installed, 0 to remove and 0 not upgraded.
Need to get 11.9 MB of archives.
After this operation, 30.0 MB of additional disk space will be used.
WARNING: The following packages cannot be authenticated!
  libcephfs1 librados2 librbd1 python-rados python-rbd python-cephfs
  python-ceph

msg: Task failed as maximum retries was encountered
````
Probably needs something like this:
````yaml
- apt-key: url=https://ceph.com/git/?p=ceph.git;a=blob_plain;f=keys/release.asc state=present
````
but not sure which play needs that.

git-harry:
@d34dh0r53 - I think this is the upstream issue that will be addressed by https://review.openstack.org/#/c/228894/

d34dh0r53:
@git-harry cherry picking that patch did the trick but now I'm running into another issue where one of my ceph_osd containers is creating a journal file that is filling up the disk:
````
root@d34d-test1_ceph_osd_container-29bb9dc4:/var/lib/ceph/osd/mydir1# ls -al
total 4024896
drwxr-xr-x  3 root root       4096 Sep 30 16:30 .
drwxr-xr-x  3 root root       4096 Sep 30 16:30 ..
-rw-r--r--  1 root root        580 Sep 30 16:30 activate.monmap
-rw-r--r--  1 root root          3 Sep 30 16:30 active
-rw-r--r--  1 root root         37 Sep 30 16:30 ceph_fsid
drwxr-xr-x 74 root root       4096 Sep 30 16:30 current
-rw-r--r--  1 root root          0 Sep 30 16:30 fiemap_test
-rw-r--r--  1 root root         37 Sep 30 16:30 fsid
-rw-r--r--  1 root root 5368709120 Sep 30 16:30 journal
-rw-------  1 root root         56 Sep 30 16:30 keyring
-rw-r--r--  1 root root         21 Sep 30 16:30 magic
-rw-r--r--  1 root root          6 Sep 30 16:30 ready
-rw-r--r--  1 root root          4 Sep 30 16:30 store_version
-rw-r--r--  1 root root         53 Sep 30 16:30 superblock
-rw-r--r--  1 root root          0 Sep 30 16:30 upstart
-rw-r--r--  1 root root          2 Sep 30 16:30 whoami
````
This may be related to being on an AIO, are there configuration parameters that need to be adjusted for an AIO?  The one setting I found in `ceph-common` for journal size was already at 0 (which may mean unlimited, not sure).

Thanks

git-harry:
@d34dh0r53 - when building an AIO some of the variables do get updated/added. The journal size is being set to 5120 when on an AIO [1]. It looks like you are using a flavour with an ephemeral disk of at least 250 GB, when that is the case the OSA scripts put the containers into logical volumes. All prior testing had been done with smaller instances or with dedicated OSD hosts.

[1] - https://github.com/rcbops/rpc-openstack/pull/380/files#diff-93518fda10b0403d3c5c20b4df4740a6R54

mattt416:
@d34dh0r53 , what @git-harry said.  I've always built this on an AIO where LVM isn't used for containers, so never hit this.  I can see it being a problem though since our container_size is limited to 5G.  Should we just decrease the journal size to 1G so that this should work irrespective of AIO instance?  Also @git-harry, can you please update scripts/deploy.sh to overwrite `journal_size` and `raw_multi_journal` rather than append?  Or shall we handle this in a subsequent patch?

prometheanfire:
does this need to be squashed?

git-harry:
@prometheanfire - I think the first three commits should be kept separate. The work logically broke along those lines so I thought it made things easier to view. The rest of the commits are recent changes that I didn't squash down to make it clear that new changes had been introduced. The pull request has been around for so long I didn't want anyone who'd previously reviewed it to think any changes in the SHAs were just due to rebasing. If you guys would prefer it squashed, I don't have an issue with reducing it down to three commits.

mattt416:
I like the idea of squashing it into 3 commits.  @git-harry do you want to include https://github.com/rcbops/rpc-openstack/pull/430 into this PR since it's documentation that probably should accompany the work, or prefer to leave it as is in a separate commit?

git-harry:
@mattt416 I ripped out the Ceph-related bits from #430 and included them here. I've also squashed everything back down to the original three commits.

mattt416:
:+1:

prometheanfire:
cool, lgtm :+1: but needs a rebase

d34dh0r53:
:+1:

mancdaz:
Good work peeps. I'm going to merge, so we can backport to kilo and pop some tags.

Issue #381, opened 2015-08-28
Rename os-ansible-deployment to openstack-ansible


nrb:
Per https://review.openstack.org/#/c/200730/, the `stackforge/os-ansible-deployment` project will be moved to `openstack/openstack-ansible`.  This is scheduled to happen on September 11, 2015. (See Andreas Jager's comment at Aug 27 2:04 PM in that same review).

This is a priority issue, since the installations will be broken until this addressed once the change happens.

This will affect the following files and directories in addition to the `os-ansible-deployment` submodule directory, per an `ack` check.
```
.gitmodules:1:[submodule "os-ansible-deployment"]
.gitmodules:2:	path = os-ansible-deployment
.gitmodules:3:	url = https://git.openstack.org/stackforge/os-ansible-deployment
README.md:4:# os-ansible-deployment integration
README.md:8:[os-ansible-deployment](https://github.com/stackforge/os-ansible-deployment)
README.md:20:[os-ansible-deployment](http://git.openstack.org/cgit/stackforge/os-ansible-deployment/tree/?h=juno)
README.md:27:The Kilo branch of os-ansible-deployment does not include any rpc-maas
README.md:43:run and then rerun the appropriate plays in os-ansible-deployment after this
README.md:67:2. Unless doing an AIO build, prepare the os-ansible-deployment configuration.
README.md:69:     `cp -R os-ansible-deployment/etc/openstack_deploy /etc/openstack_deploy`
README.md:96:To run an upgrade of an existing os-ansible-deployment installation:
rpcd/etc/openstack_deploy/user_extras_variables.yml:37:rpc_repo_path: /opt/rpc-openstack/os-ansible-deployment
rpcd/etc/openstack_deploy/user_extras_variables.yml:88:# This relies on the apt_package_pinning role from os-ansible-deployment.
rpcd/playbooks/ansible.cfg:13:# Set the path to the folder in os-ansible-deployment which holds the dynamic
rpcd/playbooks/ansible.cfg:15:inventory = ../../os-ansible-deployment/playbooks/inventory/
rpcd/playbooks/ansible.cfg:17:# Set the path to the folder in os-ansible-deployment which holds the dynamic
rpcd/playbooks/ansible.cfg:19:#hostfile = ../../os-ansible-deployment/playbooks/inventory/
rpcd/playbooks/ansible.cfg:21:# Set the path to the folder in os-ansible-deployment which holds the
[...]

PR #382, opened 2015-08-28
Update openstack-ansible name references.

nrb:
This change is to account for the name change happening upstream. All
references in tree to `os-ansible-deployment` are changed to
`openstack-ansible`, and `OSAD` is changed to `OA`, mostly for variable
names.

This should **NOT** merge until after the repository rename on Sept 11 2015.

Addresses #381.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,11 @@
 # rpc-openstack
 Optional add-ons for Rackspace Private Cloud
 
-# os-ansible-deployment integration
+# openstack-ansible integration
 
 The rpc-openstack repo includes add-ons for the Rackspace Private Cloud product
 that integrate with the 
-[os-ansible-deployment](https://github.com/stackforge/os-ansible-deployment)
+[openstack-ansible](https://github.com/openstack/openstack-ansible)
 set of Ansible playbooks and roles.
 These add-ons extend the 'vanilla' OpenStack environment with value-added
 features that Rackspace has found useful, but are not core to deploying an
@@ -17,14 +17,14 @@ In Juno, the MaaS plugins code is contained in a separate repo,
 [rpc-maas](https://github.com/rcbops/rpc-maas), and the Ansible code to deploy
 and configure the checks and alarms using these plugins is contained in tree
 with the
-[os-ansible-deployment](http://git.openstack.org/cgit/stackforge/os-ansible-deployment/tree/?h=juno)
+[openstack-ansible](http://git.openstack.org/cgit/openstack/openstack-ansible/tree/?h=juno)
 repo with the Juno branch.
 
 As of Kilo, both the Maas plugins, and the Ansible code to deploy and
 configure the checks and alarms, are contained in the
 [rpc-openstack](https://github.com/rcbops/rpc-openstack) repo.
 
-The Kilo branch of os-ansible-deployment does not include any rpc-maas
+The Kilo branch of openstack-ansible does not include any rpc-maas
 support directly any longer.
 
 # Ansible Playbooks
@@ -40,7 +40,7 @@ Heat templates for commonly deployed applications.
 * `kibana.yml` - Setup Kibana on the Kibana hosts for the logging dashboard.
 * `logstash.yml` - deploys a logstash host. If this play is used, be sure to
 uncomment the related block in user_extra_variables.yml before this play is
-run and then rerun the appropriate plays in os-ansible-deployment after this
+run and then rerun the appropriate plays in openstack-ansible after this
 play to ensure that rsyslog ships logs to logstash. See steps 11 - 13 below
 for more.
 * `repo-build.yml` - scans throug the YAML files in the source tree and builds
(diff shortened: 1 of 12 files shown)


nrb:
Also of note, since this changed all the reference, the git submodule won't work on Travis until the switch happens on openstack-infra.

odyssey4me:
:+1:

openstack-ansible
```diff
@@ -0,0 +1 @@
+Subproject commit 4f779033bc199bd2dadcd554ac7d61c80cd39617
```
claco: @nrb @odyssey4me shouldn't this be the same sha upstream? The goal is RC18 is the same as RC17, where the only addition testing is to ensure the rename is fine.

rpcd/playbooks/local_only_inventory
```diff
@@ -0,0 +1,2 @@
+[all]
```
odyssey4me: is this file an intentional addition?

scripts/deploy.sh
```diff
@@ -6,19 +6,19 @@ set -o pipefail
 export ADMIN_PASSWORD=${ADMIN_PASSWORD:-"secrete"}
 export DEPLOY_AIO=${DEPLOY_AIO:-"no"}
 export DEPLOY_HAPROXY=${DEPLOY_HAPROXY:-"no"}
-export DEPLOY_OSAD=${DEPLOY_OSAD:-"yes"}
+export DEPLOY_OA=${DEPLOY_OA:-"yes"}
```
odyssey4me: I'd suggest using OSA instead of OA, but it's a matter of preference really.

README.md
```diff
@@ -1,11 +1,11 @@
 # rpc-openstack
 Optional add-ons for Rackspace Private Cloud
 
-# os-ansible-deployment integration
+# openstack-ansible integration
 
 The rpc-openstack repo includes add-ons for the Rackspace Private Cloud product
 that integrate with the 
-[os-ansible-deployment](https://github.com/stackforge/os-ansible-deployment)
+[openstack-ansible](https://github.com/stackforge/openstack-ansible)
```
sigmavirus24: stackforge -> openstack

README.md
```diff
@@ -17,14 +17,14 @@ In Juno, the MaaS plugins code is contained in a separate repo,
 [rpc-maas](https://github.com/rcbops/rpc-maas), and the Ansible code to deploy
 and configure the checks and alarms using these plugins is contained in tree
 with the
-[os-ansible-deployment](http://git.openstack.org/cgit/stackforge/os-ansible-deployment/tree/?h=juno)
+[openstack-ansible](http://git.openstack.org/cgit/stackforge/openstack-ansible/tree/?h=juno)
```
sigmavirus24: same

rpcd/playbooks/local_only_inventory
sigmavirus24: Good catch @odyssey4me

sigmavirus24:
Please update the last two references and remove that `local_only_inventory` file (I think?).

nrb:
@sigmavirus24 @odyssey4me Updated. I left the OA variable name for now.

rpcd/playbooks/local_only_inventory
nrb: Nope, a mistake.

odyssey4me:
I'm perhaps not the best reviewer of this change, but it looks good to me. :+1:

mancdaz:
Looks good. @mattt416 pointed out that the change of directory for the submodule will affect all gating and scripts that use that dir.  Is it worth adding a symlink for os-ansible-deployment > openstack-ansible to make that transition easier?

git-harry:
@mancdaz - I think it would be better just to get the change over and done with in one go, so unless there is some need for long-lived backward compatibility I don't think a symlink is a good idea.

mancdaz:
@git-harry it's more the need for short-lived backward compatibility, just to ensure that any QE/gating scripts are not broken at this very critical time for testing.

An alternative is to not make the change now - github redirects are in place and will allow everything to continue without making any changes at all. Just saying.

hughsaunders:
Seems like a few were missed in rpcd:

```
rpcd/playbooks/patcher.yml:20:- name: Patch OSAD files
rpcd/playbooks/group_vars/all.yml:21:# Definitions for the repo server, these should match your OSAD
rpcd/etc/openstack_deploy/user_extras_variables.yml:36:# Where is the os-ansible deployment repo?
```

hughsaunders:
Ship It. recursive clones of rpc-openstack are failing

nrb:
https://github.com/rcbops/jenkins-rpc/pull/453 Was made to update the QE tests, too.

mancdaz:
Pending the few comments made by @hughsaunders, I'm :+1:

nrb:
Updated to address @hughsaunders's comments.

sigmavirus24:
:+1:

Issue #383, opened 2015-08-28
lxc config patch has different formatting causing duplicate mount entry lines


sigmavirus24:
On upgrade, this causes `lxc.mount.entry` to be in the container config twice which breaks containers.

claco:
Closing as we're no longer carrying patches in r11.0.0 release.

PR #384, opened 2015-08-28
Fixes formatting causing duplicate config entry

sigmavirus24:
The update to the formatting around the = sign was causing duplicate
entries in the lxc container config which causes containers to not
restart on upgrade.

Addresses #383


sigmavirus24:
cc @ionosphere80 @Apsu

Apsu:
:+1:

claco:
Closing PR as the issue is no longer relevant.

Issue #386, opened 2015-08-31
Update OSAD/OA SHA and remove patches


nrb:
rc14 and rc15 are carrying some OSAD patches and a pinned SHA that is fairly far behind upstream Kilo.

This issue will un-pin it and move us forward for testing the latest patches.

miguelgrinberg:
I deployed this patch in an AIO without any problems. :+1:

claco:
While we should just keep on merging kilo so it's consistent, we might want to rebump this week when this makes it: https://review.openstack.org/#/c/218572/

nrb:
Yeah, I agree. For what it's worth, the relevant kilo patch is here: https://review.openstack.org/#/c/217014/.

Also, once that merges openstack-ansible will be cutting 11.2, which would be a good time to pin.

mancdaz:
11.2.1 was a fast follow tag to fix a mistake in 11.2 in osad.  I'm going to bump the sha to 11.2.1, and use this issue as a reference.

PR #391, opened 2015-09-04
bump sha forward to match 11.2.1 in osad

mancdaz:
Related: #386


hughsaunders:
:+1:

nrb:
We missed bumping https://github.com/rcbops/rpc-openstack/blob/kilo/rpcd/playbooks/group_vars/all.yml#L19

PR #394, opened 2015-09-04
move rpc-release forward for new rc17

mancdaz:
fixes: #393


nrb:
+1, small change but important.

PR #396, opened 2015-09-14
Update openstack-ansible name references.

mancdaz:
This change is to account for the name change happening upstream. All
references in trtee to `os-ansible-deployment` are changed to
`openstack-ansible`, and `OSAD` is changed to `OA`, mostly for variable
names.

Addresses: #381 

(cherry picked from commit 8e0b097035ac3c37f37d2813a5b7774e65a08b2c)

Conflicts:
	openstack-ansible

PR #397, opened 2015-09-14
Update openstack-ansible name references.

nrb:
This change is to account for the name change happening upstream. All
references in trtee to `os-ansible-deployment` are changed to
`openstack-ansible`, and `OSAD` is changed to `OA`, mostly for variable
names.

Address #381


nrb:
This supersedes #396. That PR did not properly move the os-ansible-deployment directory, where as this one does.

Apsu:
Rename looks good, :+1:

claco:
:+1:  cloned w/ submodule. compared diffs and shas from master.

PR #399, opened 2015-09-15
Update release to match tag.

nrb:
The `rpc-release` variable is dropped into `/etc/rpc-release`, and thus should match the release's tag.


claco:
:+1:

stevelle:
:+1:

Issue #400, opened 2015-09-17
DEPLOY_OSAD: deploy.sh should run repo-install.yml over repo-build.yml


BjoernT:
When DEPLOY_OSAD is set, we should not build locally the pip wheels. They should rather just be mirrored from the rackspace mirror, which the playbook repo-install.yml seems to do. This will significantly reduce the first install time.

odyssey4me:
You're reading this wrong - the repo-build there is for the ELK stack packages to be built, not the rest of the stuff which is copied via rsync in https://github.com/rcbops/rpc-openstack/blob/master/scripts/deploy.sh#L77

byronmccollum:
Just to clarify a bit further...the Kilo branch of OA rsyncs, but Master now builds locally.

https://github.com/openstack/openstack-ansible/blob/kilo/playbooks/repo-install.yml#L17
https://github.com/openstack/openstack-ansible/blob/master/playbooks/repo-install.yml#L17

The change in Master is from this spec: https://github.com/openstack/openstack-ansible-specs/blob/master/specs/liberty/remove-upstream-repo-dependency.rst

BjoernT:
So I realized that the repo-clone-mirror does clone the whole mirror including all releases. This is also not optimal, especially when you only need one version pulled from the mirror

odyssey4me:
That is a very different issue to what's been reported in this one. I would recommend closing both this issue and the PR. If you're looking to optimize the repo sync then I'd suggest registering a new issue with that request.

PR #401, opened 2015-09-17
Use repo-install over repo-build for RPC installs.

BjoernT:
This favors rsyncs from mirror.rackspace.com over actually building all pip wheels locally
This closes bug #400


odyssey4me:
:-1: The line being changed here is the line which builds the ELK stack python packages which are not held in the upstream repo.

BjoernT:
So this change is correct for the kilo branch but not master due to the recent changes.  I will submit it for the kilo branch

Issue #342, opened 2015-08-14
Additional MaaS checks need to be added to test if metadata proxy responds on port 80


jefftelesford:
Please read the below ticket. We need to add another MaaS monitor that will connect to the neutron metadata proxy service on port 80 to see if it responds. The ticket describes an occurrence of this problem which the metadata proxy service was running but wasn't binded to port 80. The customer requested that we add monitoring to catch this event. My manager Charles also feels this should be installed along with our other monitoring checks during deployments or upgrades.

TICKET: <redacted>

Best Regards,

Jeff Telesford
OpenStack Engineer
Rackspace Hosting

toxick:
The neutron-ns-metadata-proxy service was running, but not listening:

**root@infra-node1:/opt/os-ansible-deployment/rpc_deployment# ansible neutron_agent -m shell -a "ip netns exec qdhcp-a5ad7a1d-d3a6-4180-8d61-07a23f6fb449 netstat -ntlp"**
infra-node2_neutron_agents_container-544efdd7 | success | rc=0 >>
Active Internet connections (only servers)
Proto Recv-Q Send-Q Local Address           Foreign Address         State       PID/Program name
tcp        0      0 0.0.0.0:80              0.0.0.0:*               LISTEN      20706/python
tcp        0      0 172.22.89.19:53         0.0.0.0:*               LISTEN      31527/dnsmasq
tcp        0      0 169.254.169.254:53      0.0.0.0:*               LISTEN      31527/dnsmasq
tcp6       0      0 fe80::f816:3eff:fe16:53 :::*                    LISTEN      31527/dnsmasq

**infra-node1_neutron_agents_container-1c74674d | success | rc=0 >>**
**Active Internet connections (only servers)**
**Proto Recv-Q Send-Q Local Address           Foreign Address         State       PID/Program name**
**tcp        0      0 172.22.110.222:53       0.0.0.0:*               LISTEN      2186/dnsmasq**
**tcp        0      0 169.254.169.254:53      0.0.0.0:*               LISTEN      2186/dnsmasq**
**tcp6       0      0 fe80::f816:3eff:fec3:53 :::*                    LISTEN      2186/dnsmasq**

infra-node3_neutron_agents_container-69cf97bb | success | rc=0 >>
Active Internet connections (only servers)
Proto Recv-Q Send-Q Local Address           Foreign Address         State       PID/Program name
[...]

PR #403, opened 2015-09-17
Add maas external check for neutron metadata proxy

stevelle:
The neutron metadata agent proxy provides access from within neutron
network namespaces to the neutron metadata agent and thus to the
various sources of metadata provided to instances on 169.254.169.254
at port 80 when cloud-init is used.

The metadata agent has monitoring in RPC but has been reported to be
running but non-responsive on some occasions (see issue #342). The
new check tests the agent's proxy from within the namespace to
monitor responsiveness. Because the agent will use the client host
IP to look for metadata but the request is not originating from an
instance a 404 Not Found response is expected from the agent.

Because local checks are expected to operate from the host rather
than inside a container, and the host does not have ready access to
neutron network namespaces, the monitor must probe for an appropriate
container and resorts to executing a curl request through the network
namespace within the selected container for each network that is
expected to be proxied (only those networks which have ports). This is
deemed easier to verify and understand through command shell
operations than alternative solutions.

This solution seems to work regardless of the value of OSA's
nova_force_config_drive setting or the use of the config-drive feature
in Nova.

Addresses #342

### DIFF
--- a/maas/plugins/neutron_metadata_local_check.py
+++ b/maas/plugins/neutron_metadata_local_check.py
@@ -1,0 +1,89 @@
+#!/usr/bin/env python
+
+# Copyright 2014, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import argparse
+import shlex
+import subprocess
+
+from maas_common import get_neutron_client
+from maas_common import metric_bool
+from maas_common import print_output
+from maas_common import status_err
+from maas_common import status_ok
+
+# identify the first active neutron agents container on this host
+# network namespaces can only be accessed from within neutron agents container
+FIND_CONTAINER = shlex.split('lxc-ls -1 --running .*neutron_agents')
+SERVICE_CHECK = 'ip netns exec %s curl -fvs 127.0.0.1:80'
+
+
+def check(args):
+    # identify the container we will use for monitoring
+    try:
+        containers_list = subprocess.check_output(FIND_CONTAINER)
+        container = containers_list.splitlines()[0]
+    except (IndexError, subprocess.CalledProcessError):
+        status_err('no running neutron agents containers found')
+
+    network_endpoint = 'http://{host}:9696'.format(host=args.neutron_host)
+    try:
+        neutron = get_neutron_client(endpoint_url=network_endpoint)
+
+    # not gathering api status metric here so catch any exception
+    except Exception as e:
+        status_err(str(e))
+
+    # only check networks which have a port with DHCP enabled
+    ports = neutron.list_ports(device_owner='network:dhcp')['ports']
+    nets = set([p['network_id'] for p in ports])
+
+    # perform checks for each identified network
+    failures = []
+    for net_id in nets:
+        namespace = 'qdhcp-%s' % net_id
+        service_check_cmd = SERVICE_CHECK % namespace
+        command = shlex.split('lxc-attach -n %s -- %s' % (container,
+                                                          service_check_cmd))
+        try:
+            subprocess.check_output(command, stderr=subprocess.STDOUT)
+        except subprocess.CalledProcessError as e:
+            # HTTP 404 response indicates the service is responsive.
+            # this is the expected response because the maas testing host IP
+            # is used to look up metadata and no metadata exists for this IP
+            if '404 Not Found' not in e.output:
+                failures.append(net_id)
+
+    is_ok = len(failures) == 0
+    metric_bool('neutron-metadata-agent-proxy_status', is_ok)
+
+    if is_ok:
+        status_ok()
+    else:
+        status_err('neutron metadata agent proxies fail on host %s '
+                   'net_ids: %s' % (container, ','.join(failures)))
+
+
+def main(args):
+    check(args)
+
+
+if __name__ == "__main__":
+    with print_output():
+        parser = argparse.ArgumentParser(description='Check neutron proxies')
+        parser.add_argument('neutron_host',
+                            type=str,
+                            help='Neutron API hostname or IP address')
+        main(parser.parse_args())
--- a/rpcd/playbooks/roles/rpc_maas/tasks/local.yml
+++ b/rpcd/playbooks/roles/rpc_maas/tasks/local.yml
@@ -195,6 +195,21 @@
 
 - include: local_setup.yml
   vars:
+    check_name: neutron_metadata_agent_proxy_check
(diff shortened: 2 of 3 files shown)


maas/plugins/neutron_metadata_local_check.py
```diff
@@ -0,0 +1,87 @@
...
+    # not gathering api status metric here so catch any exception
+    except Exception as e:
+        status_err(str(e))
+
+    # gather networks to check
+    # only check networks which have a port, otherwise the proxy won't answer
+    ports = neutron.list_ports()['ports']
+    target_nets = [p['network_id'] for p in ports]
+    nets = [n for n in neutron.list_networks()['networks']
+            if n['id'] in target_nets]
```
stevelle: this can be simplified: eliminate list_networks() call since I am only using the network_id.

maas/plugins/neutron_metadata_local_check.py
```diff
@@ -0,0 +1,87 @@
...
+    try:
+        neutron = get_neutron_client(endpoint_url=network_endpoint)
+
+    # not gathering api status metric here so catch any exception
+    except Exception as e:
+        status_err(str(e))
+
+    # gather networks to check
+    # only check networks which have a port, otherwise the proxy won't answer
+    ports = neutron.list_ports()['ports']
```
stevelle: need to ensure we only list ports with dhcp enabled

maas/plugins/neutron_metadata_local_check.py
```diff
@@ -0,0 +1,87 @@
...
+
+from maas_common import get_neutron_client
+from maas_common import metric_bool
+from maas_common import print_output
+from maas_common import status_err
+from maas_common import status_ok
+
+# identify the first active neutron agents container on this host
+# network namespaces can only be accessed from within neutron agents container
+FIND_CONTAINER = 'lxc-ls -1 --active .*neutron_agents|head -n 1'
```
stevelle: eliminate '| head -n1' by only reading first line of output below

stevelle:
I did quite a bit of testing, but it is possible with the various tweaks and optimizations done that I have a failure case or alarm state not working correctly anymore.

@mattt416 your eyes would be very much appreciated on this one when you have a chance.

maas/plugins/neutron_metadata_local_check.py
```diff
@@ -0,0 +1,88 @@
...
+
+from maas_common import get_neutron_client
+from maas_common import metric_bool
+from maas_common import print_output
+from maas_common import status_err
+from maas_common import status_ok
+
+# identify the first active neutron agents container on this host
+# network namespaces can only be accessed from within neutron agents container
+FIND_CONTAINER = shlex.split('lxc-ls -1 --active .*neutron_agents')
```
mattt416: Hi @stevelle, what's the difference between `--active` and `--running`?

maas/plugins/neutron_metadata_local_check.py
```diff
@@ -0,0 +1,88 @@
...
+    # only check networks which have a port with DHCP enabled
+    ports = neutron.list_ports(device_owner='network:dhcp')['ports']
+    nets = set([p['network_id'] for p in ports])
+
+    # perform checks for each identified network
+    failures = []
+    for net_id in nets:
+        namespace = 'qdhcp-%s' % net_id
+        service_check_cmd = SERVICE_CHECK % namespace
+        command = shlex.split('ssh %s "%s"' % (container, service_check_cmd))
```
mattt416: We probably need to disable host key checking here -- on a fresh install, my plugins all failed to run as ssh is waiting for confirmation on accepting the key.

mattt416: Also @stevelle -- is it assumed that every infra node can ssh to instances on itself?  I know the deployment host can obviously ssh to everything, but I'm not sure if that ssh key on the deployment host will always be on each infra node?  Perhaps it's better to do an lxc-attach instead?

rpcd/playbooks/roles/rpc_maas/tasks/local.yml
```diff
@@ -195,6 +195,21 @@
 
 - include: local_setup.yml
   vars:
+    check_name: neutron_metadata_agent_proxy_check
+    check_details: file=neutron_metadata_local_check.py,args=args={{ internal_vip_address }}
```
mattt416: args=args={{ ... }} isn't correct, my check got deployed w/out any args.

mattt416:
A few minor issues @stevelle -- I'll have another crack once those are sorted.

maas/plugins/neutron_metadata_local_check.py
stevelle: I had a thought about the host key checking but didn't rebuild my test instance to check it.  Good call.

In my work on Swift multi-location it seemed like all infra nodes can ssh to containers on itself (esp. if host key checking is disabled) but I inevitably took shortcuts here. I would love to have someone verify that this works is Swift is disabled on a multi-node but I don't have the resources available to do a Heat deploy for that.

rpcd/playbooks/roles/rpc_maas/tasks/local.yml
stevelle: I cleaned up some extra args and appear to have left a bit.  Good catch.

maas/plugins/neutron_metadata_local_check.py
stevelle: good call, active is a superset. changing to running

maas/plugins/neutron_metadata_local_check.py
mattt416: Hi @stevelle, on rpc-heat we only write the private key to controller1 -- but we could easily update this to write it to all controller nodes however it's probably better to check if support deploy the key on all controller nodes and go from there.

rpcd/playbooks/roles/rpc_maas/tasks/local.yml
```diff
@@ -195,6 +195,21 @@
 
 - include: local_setup.yml
   vars:
+    check_name: neutron_metadata_agent_proxy_check
+    check_details: file=neutron_metadata_local_check.py,args={{ internal_vip_address }}
+    check_period: "{{ maas_check_period }}"
+    check_timeout: "{{ maas_check_timeout }}"
+    alarms:
+      - { 'name': 'neutron_metadata_agent_proxy_status', 'criteria': ':set consecutiveCount={{ maas_alarm_local_consecutive_count }} if (metric["neutron-metadata-agent-proxy_status"] != 1) { return new AlarmStatus(CRITICAL, "neutron-metadata-agent-proxy non-responsive"); }' }
```
mattt416: Hi @stevelle , metric["neutron-metadata-agent-proxy_status"] doesn't exist, the plugin is writing 'neutron_metadata_agent_proxies_on_host_%s' % container.  I suppose you'd need to set up an alarm for each container returned by the check?

On a side note, I'm not sure any of the existing checks/alarms support running multiple containers for a service on a single host.  This isn't ideal but just something to think about as you implement this.

stevelle: For now the check only runs on the first container it finds, so I did not go so far as to support duplicate containers on a host yet. This doesn't seem necessary so pulling back a bit.

mattt416:
Did a fresh multi-node deploy and this check went green across my 3 controllers nodes.  LGTM!

prometheanfire:
fixed in #412

Issue #404, opened 2015-09-18
Refactor juno branch os-ansible-deployment to openstack-ansible


jacobwagner:
in the dynamic_inventory script the path to the rpc_repo is wrong due to the recent move from os-ansible-deployment to openstack-ansible. The common vars need to be fixed to mirror this change.

https://github.com/rcbops/rpc-openstack/blob/juno/rpc_deployment/vars/common.yml#L16

PR #405, opened 2015-09-18
Updated common vars to mirror the repo location change.

jacobwagner:
Due to the recent location change for openstack-ansible the common
vars for the juno rpc-openstack needed to be updated to mirror the
upstream change. This was causing the dynamic inventory script to
fail due to a unkown file location

This fixes issue #404


sigmavirus24:
Travis is angry because we don't have it configured for this branch. :+1: from me anyway

claco:
:+1:

Issue #407, opened 2015-09-23
HP monitoring plugin doesn't monitor raid controller status


mancdaz:
Currently the plugin monitors disk/cpu/dimm status, but not the raid controller itself.  This has affected customers who've had a raid controller fail. 

My first assumption was that a failed raid controller means the underlying raid array is innaccessible, and so no alerts would get out anyway, but on reading it seems that some arrays can fail partially, meaning they don't recognise some of the disks in the array, but do recognise enough to keep the array accessible.  Therefore, it's worth throwing a controller monitor into the mix (it's really not much overhead to do so anyway).

PR #408, opened 2015-09-23
add hp monitors for disk controller/cache/battery

mancdaz:
The current HP hardware monitors only cover cpu/dimm/disk.

This commit adds status monitoring via hpssacli for
controller/cache/battery status too

Fixes: #407


prometheanfire:
lgtm

:+1:

d34dh0r53:
:+1: looks good to me as well

PR #413, opened 2015-09-29
Add maas external check for neutron metadata proxy - kilo

### LINKED ISSUE #342: Additional MaaS checks need to be added to test if metadata proxy responds on port 80

prometheanfire:
The neutron metadata agent proxy provides access from within neutron
network namespaces to the neutron metadata agent and thus to the
various sources of metadata provided to instances on 169.254.169.254
at port 80 when cloud-init is used.

The metadata agent has monitoring in RPC but has been reported to be
running but non-responsive on some occasions (see issue #342). The
new check tests the agent's proxy from within the namespace to
monitor responsiveness. Because the agent will use the client host
IP to look for metadata but the request is not originating from an
instance a 404 Not Found response is expected from the agent.

Because local checks are expected to operate from the host rather
than inside a container, and the host does not have ready access to
neutron network namespaces, the monitor must probe for an appropriate
container and resorts to executing a curl request through the network
namespace within the selected container for each network that is
expected to be proxied (only those networks which have ports). This is
deemed easier to verify and understand through command shell
operations than alternative solutions.

This solution seems to work regardless of the value of OSA's
nova_force_config_drive setting or the use of the config-drive feature
in Nova.

Fixes #342

### DIFF
--- a/maas/plugins/README.md
+++ b/maas/plugins/README.md
@@ -193,6 +193,19 @@ Hostname or IP address of service to test
     ...
 
 ***
+#### neutron_metadata_local_check.py
+
+##### Description:
+polls the neutron metadata agent proxies in each network namespace with DHCP enabled to ensure the agent is responsive.
+
+##### mandatory args:
+Hostname or IP address of Neutron API service
+##### example output:
+
+    metric neutron-metadata-agent-proxy_status uint32 1
+    ...
+
+***
 ***
 ### OTHER CHECKS
 ***
--- a/maas/plugins/neutron_metadata_local_check.py
+++ b/maas/plugins/neutron_metadata_local_check.py
@@ -1,0 +1,89 @@
+#!/usr/bin/env python
+
+# Copyright 2014, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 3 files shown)


d34dh0r53:
:+1: LGTM

git-harry:
@prometheanfire I think it makes more sense to reference the commit from which this was cherry picked. Appending kilo to the end of the commit message title doesn't provide any additional information.

prometheanfire:
A direct cherry-pick wasn't possible as the originating files changed location.  see #414 for more info...

not having monitoring merged is making backports hard

claco:
Personally, I vote we always cherry-pick -x for backports. If it's not clean, fix the merge conflicts, and commit with the "Conflicts" comments provided by git uncommented. It;s the best way to ensure track-ability.

prometheanfire:
ok, cherry picked

d34dh0r53:
Looks good to me, but I'd like to get it tested first.  Any volunteers?

prometheanfire:
this was already done in master I thought.  manual tests SUCK

prometheanfire:
rebased and it does look like it was tested

stevelle:
:+1: genuine to the original
this looks like the second, no dissents other than process at this point, so pushing the button.

PR #416, opened 2015-09-29
backport of rcbops/rpc-openstack/pull/333

prometheanfire:
testing wasn't backported so wasn't a straight cherrypick.


d34dh0r53:
:+1:

git-harry:
-1 the commit message 5e20294 needs to be improved, it should include the relevant information from the original commit. Also it looks like there is a merge commit as part of the pull request, which should be removed.

prometheanfire:
If we are going to block on merge commits, then we need to block the 'update branch' button from showing up.  Because that's what I hit...

prometheanfire:
closing in favor of #417

Issue #221, opened 2015-07-06
Standalone Swift - ERROR (EndpointNotFound): internalURL endpoint for compute service not found


byronmccollum:
The `rpc_support` role assumes Nova is installed, and tries registering a support keypair. In a Standalone Swift configuration, this will fail as there are no Nova endpoints to communicate with.

```
TASK: [rpc_support | Add rpc_support key to nova] ***************************** 
skipping: [163f7de7-infra1]
skipping: [163f7de7-object1]
skipping: [163f7de7-object3]
skipping: [163f7de7-infra1_keystone_container-23f8c901]
skipping: [163f7de7-infra1_swift_proxy_container-fa8061b9]
skipping: [163f7de7-object2]
skipping: [163f7de7-infra1_galera_container-22c355d4]
skipping: [163f7de7-infra1_memcached_container-469b23e0]
skipping: [163f7de7-infra1_rabbit_mq_container-bcf9683e]
<172.29.239.239> ESTABLISH CONNECTION FOR USER: root
skipping: [163f7de7-infra1_repo_container-e6a7a307]
<172.29.239.239> REMOTE_MODULE command . /root/openrc nova keypair-add --pub-key /root/.ssh/rpc_support.pub rpc_support #USE_SHELL
<172.29.239.239> EXEC ssh -C -vvv -o ControlMaster=auto -o ControlPersist=60s -o ControlPath="/root/.ansible/cp/ansible-ssh-%h-%p-%r" -o StrictHostKeyChecking=no -o KbdInteractiveAuthentication=no -o PreferredAuthentications=gssapi-with-m
ic,gssapi-keyex,hostbased,publickey -o PasswordAuthentication=no -o ConnectTimeout=120 172.29.239.239 /bin/sh -c 'LANG=en_US.UTF-8 LC_CTYPE=en_US.UTF-8 /usr/bin/python'
failed: [163f7de7-infra1_utility_container-4df0fd1f] => {"changed": true, "cmd": ". /root/openrc\n nova keypair-add --pub-key /root/.ssh/rpc_support.pub rpc_support", "delta": "0:00:00.631306", "end": "2015-07-06 21:06:09.953687", "rc": 1
, "start": "2015-07-06 21:06:09.322381", "warnings": []}
stderr: ERROR (EndpointNotFound): internalURL endpoint for compute service not found
```

nrb:
In order to minimize impact on QE, I'm going to set this as blocked and push it to Kilo 11.1.0

[...]

PR #233, opened 2015-07-08
Make RPC Support role Standalone Swift Friendly

mattt416:
Only create support keypair if nova service is present. Only create support security group if neutron service is present.

Fixes #221

(cherry picked from commit e739459a1cd1fc97de103fd92f390b514be61fdd)

PR #222, opened 2015-07-07
Make RPC Support role Standalone Swift Friendly

byronmccollum:
Only create support keypair if nova service is present. Only create support security group if neutron service is present.

Fixes #221

PR #418, opened 2015-09-30
Standalone Swift Skip Key Distribute Retrieve Tasks

byronmccollum:
This PR adds the distribute and retrieve keypair tasks to the list of skipped tasks when Nova is not installed.

Combined with PR #222, fixes bug #221 for master
Backport combined with PR #233, fixes bug #221 for kilo

### DIFF
--- a/rpcd/playbooks/roles/rpc_support/tasks/main.yml
+++ b/rpcd/playbooks/roles/rpc_support/tasks/main.yml
@@ -49,11 +49,11 @@
 
 - include: distribute_auth_key.yml
   when: >
-    inventory_hostname == groups['utility'][0]
+    groups['nova_api_os_compute']|length > 0 and inventory_hostname == groups['utility'][0]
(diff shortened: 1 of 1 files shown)


byronmccollum:
Currently building a fresh full RPC, and swift only RPC to verify no failures, and no ill side effects.

byronmccollum:
Does what it says on the tin...

rackertom:
@byronmccollum Is the situation here that the group is defined, but empty? Or are we also checking for the case where the group is not defined here?

byronmccollum:
@rackertom I *think* the dynamic inventory creates the host group, regardless if anything is assigned to that group. I also don't have any empty host groups defined in my `openstack_user_config.yml`.

rackertom:
Awesome. :+1:

prometheanfire:
:+1: when rebased

Issue #216, opened 2015-07-06
MaaS testing


git-harry:
Currently there is little MaaS testing. Proper testing needs to be in place to help develop the solution futher.

nrb:
@jpmontez @jacobwagner @klevenstein This has QE and doc impacts; just making you guys aware of it. Particularly the doc generation script which I know was a wish list item.

This will also likely be useful to backport into rpc-maas/juno, but I haven't looked at it in depth yet.

prometheanfire:
just needs backports?

nrb:
I believe so. Looks like @git-harry may have made a commit for it, but not a PR.

d34dh0r53:
@git-harry do you have the commits for the backports of this?

git-harry:
@d34dh0r53 - no, I'll put one together

prometheanfire:
done in #414

prometheanfire:
along with your other shame

git-harry:
I see nothing shameful in what I've done

prometheanfire:
it was a joke :P

rackertom:
@prometheanfire @git-harry Since the PR has been merged, can we mark this as closed?

prometheanfire:
not sure

#352 should probably be done too

PR #420, opened 2015-09-30
MaaS testing

prometheanfire:
Currently there is no regular testing of the MaaS playbooks. This
commit attempts to address that.

generate-definitions.py is used to create files that define the MaaS
configuration for a set of specified entities.

The files in maas/testing/tests/integration are per-MaaS-playbook
reference definitions files and were created using
generate-definitions.py. These need updating when the MaaS configuration
is altered.

compare-definitions.py is used to compare the definitions defined in the
this repo with those generated for the deployment being tested.

generate-docs.py prints tables of the checks, alarms and metrics
defined in one or more definitions files. The format of the output is
reStructuredText. To print the tables for the definitions set:

  ./generate-docs.py tests/integration/*

The test-maas.yml playbook can be used to test a deployments against
the reference definitions files. It prepares the hosts for testing and
then uses generate-definitions.py and compare-definitions.py to test
the deployed MaaS configuration against the one found in
maas/testing/tests/integration. As part of the setup a script is copied
to the hosts to allow HP hardware testing to be configured. Obviously
this does not test that the plugin works with the HP binaries, but that
that checks/alarms are configured as expected.

Closes-issue: https://github.com/rcbops/rpc-openstack/issues/216
(cherry picked from commit 738440d3ea06968e9bd6c5f6cbe1a1e5422a8d95)

### DIFF
--- a/maas/testing/requirements.txt
+++ b/maas/testing/requirements.txt
@@ -1,0 +1,3 @@
+rackspace-monitoring
+requests
+tabulate
--- a/maas/testing/tests/integration/maas_cdm
+++ b/maas/testing/tests/integration/maas_cdm
@@ -1,0 +1,1291 @@
+{
+    "BLOCK_STORAGE": {
+        "active_suppressions": [],
+        "checks": {
+            "cpu--BLOCK_STORAGE": {
+                "active_suppressions": [],
+                "alarms": {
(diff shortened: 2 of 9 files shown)


git-harry:
+1

rackertom:
LGTM :+1:

PR #421, opened 2015-09-30
cinder_service_check.py not using --host option

prometheanfire:
The script cinder_service_check.py defines an optional host parameter
that is designed to filter the results, returned by the cinder API, so
that only services on the specified host are returned. The MaaS setup
role, that configures the checks that use this script, uses '--host'.
The script does not use the host argument if supplied.

This patch fixes the script so that the results returned are filtered
by host, where host includes any backends/pools on that host.

Closes-bug: https://github.com/rcbops/rpc-openstack/issues/200
(cherry picked from commit 6c4d898591bdf2cfdeed5c8d564dda622252daf6)

### DIFF
--- a/maas/plugins/cinder_service_check.py
+++ b/maas/plugins/cinder_service_check.py
@@ -77,8 +77,7 @@ def check(auth_ref, args):
         if service['status'] == 'enabled' and service['state'] != 'up':
             service_is_up = False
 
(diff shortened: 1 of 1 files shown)


d34dh0r53:
:+1:

git-harry:
+1

PR #425, opened 2015-09-30
Only create support keypair is nova service is present. Only create s…

prometheanfire:
…upport security group if neutron service is present.

(cherry picked from commit e739459a1cd1fc97de103fd92f390b514be61fdd)


d34dh0r53:
What's up with the commit message on this one? :-1: (update commit message)

prometheanfire:
it was a cherry pick, the commit message is the same as the one that merged before...

d34dh0r53:
Uggh :+1:

git-harry:
+1

Issue #136, opened 2015-06-02
nova_spice_console group no longer exists


andymcc:
This has been fixed in master, but needs changing in kilo to match up.

nrb:
This has been merged into master and kilo, closing.

PR #90, opened 2015-04-29
Rename nova_spice_console to nova_console

nrb:
In Kilo, we renamed the nova_spice_console group to nova_console. This change keeps our f5 config consistent with that code base.

PR #427, opened 2015-09-30
 Rename nova_console to nova_spice_console in the f5-config 

prometheanfire:
The config name changed in an earlier commit, we are fixing the script
so that the naming is consistent.

Closes-Bug: https://github.com/rcbops/rpc-extras/issues/46
(cherry picked from commit 02659163494670921876943b3205076dfaa1ba44)

### DIFF
--- a/scripts/f5-config.py
+++ b/scripts/f5-config.py
@@ -252,7 +252,7 @@ POOL_PARTS = {
         'make_public': True,
         'hosts': []
     },
-    'nova_console': {
+    'nova_spice_console': {
         'port': 6082,
         'backend_port': 6082,
         'mon_type': '/' + PART + '/' + PREFIX_NAME + '_MON_HTTP_NOVA_SPICE_CONSOLE',


d34dh0r53:
:-1: I don't think this has everything covered in the commit message, unless the files are already in the `kilo` branch and this is just a bug fix in which case the message should be  updated to reflect that fact.

prometheanfire:
It was cherry picked from that commit, this was the only thing not merged, kinda odd, but I can modify the commit message if you want

d34dh0r53:
Yeah, can you modify the commit message just to people don't scratch their heads later.

git-harry:
-1

It looks like a mistake was made and the change was never merged into master. I think this commit is a regression. The change was introduced into kilo by [1]. [2] was meant to introduce the change into master but it looks like that was closed by mistake. [3] shows how [1] was done on the assumption the change was already introduced into master.

So it looks like master requires a patch to bring it in to line with kilo.

[1] - https://github.com/rcbops/rpc-openstack/pull/137
[2] - https://github.com/rcbops/rpc-openstack/pull/90
[3] - https://github.com/rcbops/rpc-openstack/issues/136

prometheanfire:
fair enough, gonna reverse this and submit a patch to master then

PR #443, opened 2015-10-05
Check json_resp before calling max()

rackertom:
The `json_resp` variable needs to be checked for an empty condition before using `max()`. From the `max()` docs:

>If one positional argument is provided, iterable must be a non-empty iterable


d34dh0r53:
:-1: @rackertom can you squash this please :smile:

maas/plugins/rabbitmq_status.py
```diff
@@ -98,6 +98,9 @@ def main():
 
     if r.ok:
         resp_json = r.json()  # Parse the JSON once
+        if not resp_json:
+            # Returned JSON contains no channels
+            status_err('Rabbitmq returned empty channels list')
```
git-harry: I think this should report this information in the metrics instead of exiting the script. As a general rule the scripts should only return status_err when they fail to collect data. In this case it seems that some metric should get set to 0 because there are no channels.

rackertom:
@git-harry The reason I chose to error out here is that if there are no channels, then Rabbit has not been configured to do anything. So checking the status of an unconfigured RabbitMQ shouldn't return success in any form. Any subsequent checks for channel content would also fail, so the utility of continuing is minimal at best IMO. But I'm willing to be convinced otherwise.

rackertom:
Ok @d34dh0r53, hopefully I've corrected the branch and fork so this is in sync with upstream master.

git-harry:
@rackertom - gathering of a metric and the decision about the meaning of a metric are separate, one is the responsibility of the plugin and I believe the other should be for the alarm. I am under the impression this is the principle that has been applied when creating all the other plugins in the repo.

rackertom:
@git-harry Of course. I agree with what you're saying, but that isn't the situation here. The situation in this case is that no channels have been configured so there is nothing to actually do. No further checks make sense to perform because performing them would be impossible. There doesn't exist, IMO, a sane deployment of RabbitMQ that should be monitored *and* does not have any channels configured.

Basically this would be the same as putting a tachometer on an engine that doesn't run. Sure, you can do it and it'll look nice. Even though it will show the correct RPM, it isn't actually functioning because it is impossible for the engine to actually do anything.

stevelle:
This is getting in the way of testing RPC-R, as reported by Manny via email

rackertom:
The latest change I submitted does not return an error if no valid JSON is returned. We just need reviews so this can be merged.

sigmavirus24:
There's 2 commits in this PR. Shouldn't they be squashed?

PR #457, opened 2015-10-06
add script to check if a given commit is contained in a built wheel

prometheanfire:
It's somewhat difficult to determine if a deployed wheel of an
openstack project (say neutron) contains a particular upstream commit,
because of the way we build wheels outside of the containers.

This commit adds a script to be run from the deployment host that does
the following things:

1) connects to all containers of type 'project' and determines the sha
from which the wheel was built
2) clones git repo locally (/tmp by default) to allow interrogation of
git history
3) checks to see if 'supplied commit sha' (the commit that you are
interested in) is contained in the wheel.

optional:
- supply a parent sha and use that instead of getting the
parent sha from the wheel in the containers.
- supply a host group instead of defaulting to <PROJECT>_all for the
  ansible run

Closes: #195
(cherry picked from commit 0c728a71afeadc9bc636023c04301fdb09933547)

### DIFF
--- a/scripts/checksha.py
+++ b/scripts/checksha.py
@@ -1,0 +1,218 @@
+#!/usr/bin/env python
+# Copyright 2015, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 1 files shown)


d34dh0r53:
:+1:

sigmavirus24:
Looks like a straightforward cherry-pick. :+1:

Issue #460, opened 2015-10-07
Neutron L3 Router Monitoring


mrhillsman:
Related to customer request for monitoring of Neutron L3 router in order to assist with resolving network outage issue that could have been resolved faster with an alert.

---
###### For each pull request, update the branch line to reference it.
- [ ] master
- [ ] mitaka-13.0
- [ ] liberty-12.2
- [ ] kilo
##### Upgrade impact. Consider the impact on moving between major, minor and patch releases.
- [ ] This change does not impact upgrades
- [ ] This change addresses a potential upgrade impact
- [ ] The upgrade impact is being tracked by another issue.

The issue tracking the upgrade impact is:
##### Testing impact
- [ ] An issue/pull-request has been create or exists to track the testing gap, and is referenced below
- [ ] This change includes additional testing
- [ ] This change is tested by the existing setup

Additional testing links:
##### Release note
- [ ] This change includes a release note.
- [ ] This change does not require a release note.
##### Does your change require documentation?
- [ ] Yes, and I submitted an issue to [Docs issues](https://github.com/rackerlabs/docs-rpc/issues)
- [ ] Yes, and I contributed a documentation patch to [docs-rpc](https://github.com/rackerlabs/docs-rpc)
- [ ] No documentation required for this change

Additional documentation links (please add links if you have ticked yes):

claco:
It seems like we're adding a new plugin, but we're not also adding checks/alarms?

BjoernT:
We do not need in particular checks. Whenever we exit this script with a >0 return code, it'll generate a alarm anyway

prometheanfire:
@stevelle you said you'd do the check, just making a note here

stevelle:
Without a check, this script wont be run, so it needs something more. Also a squash.
With the pending work to refactoring for maas (https://github.com/rcbops/rpc-openstack/tree/refactor/maas-templates) the method of scheduling a check will be much simpler. 

I will put together a template to be added to this issue that will work with the refactor to schedule a check for this.

mancdaz:
@hughsaunders looks like he is going to pick up the PR

JCallicoat:
The PR was closed. We still need a solution here.

claco:
@JCallicoat @BjoernT Refresh my memory. Was this the one discussed previously that needed some magic determining how to monitor given if things are in, or out of a Router in Neutron?

BjoernT:
[...]

PR #461, opened 2015-10-07
Create neutron_l3_router_local_check.py

codebauss:
Add MaaS monitoring plugin for Neutron L3 router. Addresses issue #460

### DIFF
--- a/maas/plugins/neutron_l3_router_local_check.py
+++ b/maas/plugins/neutron_l3_router_local_check.py
@@ -29,7 +29,7 @@ def check(auth_ref):
     endpoint = get_endpoint_url_for_service('network',
                                             auth_ref['serviceCatalog'],
                                             url_type='internalURL')
-    # TODO set this from args.version to future-proof // version = args.version
+    # TODO: set this from args.version to future-proof // version = args.version
     version = 'v2.0'
     api_endpoint = '{endpoint}/{version}'.format(endpoint=endpoint,
                                                  version=version)
@@ -60,11 +60,11 @@ def check(auth_ref):
     metric_bool('neutron_api_status', is_active)
 
     for router in routers:
-        if(router['external_gateway_info'] == None):
+        if router['external_gateway_info'] is None:
             continue
 
         router_status = router['status']
-        router_name = (router['name']).replace(" ","").lower()
+        router_name = (router['name']).replace(" ", "").lower()
         failed = []
 
         # If router_status is ACTIVE, perform ping check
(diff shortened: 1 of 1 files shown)


git-harry:
@codebauss please address the issues highlighted by the testing [1]. There is also a lot of commented out code, please remove it if it is not required.

[1] https://travis-ci.org/rcbops/rpc-openstack/builds/84024930#L521

codebauss:
@git-harry passing travis-ci and cleaned up unnecessary code/comments

BjoernT:
@codebauss : I would assume we need to execute these pings inside the qrouter namespace since not all routers are accessible directly. Most indeed aren't. Additionally we need to test the default gateway of the router namespace so we now traffic goes into the router and can leave the router. Thoughts ?

prometheanfire:
9 commits still
needs rebase
installs the file but the file isn't used that I can see

prometheanfire:
@stevelle you said you'd do the check, just making a note here

maas/plugins/neutron_l3_router_local_check.py
```diff
@@ -0,0 +1,103 @@
...
+
+        # If router_status is ACTIVE, perform ping check
+        if(router_status == 'ACTIVE'):
+            import os
+
+            # IP address is the WAN interface of the router
+            gateway = router['external_gateway_info']
+            ip_address = gateway['external_fixed_ips'][0]['ip_address']
+
+            rc = os.system('ping -c1 -W3 ' + ip_address + ' > /dev/null')
```
stevelle: I would rather see this use subprocess.call instead of os.system (see also neutron_metadata_local_check.py for an example) and don't bother sending output to dev/null (subprocess will route it to a pipe which you can safely ignore).

Recommend to use shlex.split to split up the ping command and params also.

stevelle:
This is an untested template but should be really close to the target, based on the current state of this PR:
https://gist.github.com/stevelle/1ffd4a7da907f00ffa58

Add this to the PR and the check/alarms will be taken care of post-refactor.

hughsaunders:
Hey @codebauss, are you planning to update this PR? I'll start working on this if I don't see a response by the 15th.

codebauss:
Hey @hughsaunders and @stevelle sorry for the delay in response. I am very new to all of this unfortunately so need a little guidance in getting this resolved.

hughsaunders:
Hey @codebauss, no problem, I've started working on a version incorporating @stevelle's comments: https://github.com/rcbops/rpc-openstack/compare/master...hughsaunders:neutron_l3_router_local_check Am in the process of testing that.

codebauss:
@stevelle ok great, I had actually made the suggested change to subprocess.call but not sure about "squashing". I would definitely like to learn however. I will keep a closer look out for updates, again, apologies for the delay in response.

hughsaunders:
While testing my branch of this PR, I made the following observations:
  * neutron_api_status metric is not needed as it is provided by neutron_api_local_check
  * router ping functionality could also be rolled into neutron_api_local_check
  * Alarm criteria is checking neutron_router_ping for an int, but the metric returns string

PR #480, opened 2015-10-13
Lowering max_retries from 5 to 2 to catch build errors faster

BjoernT:
This change will cause OSA to fail more early in case repeating errors occur. It has been our experience if playbooks fail more than one, there's usually a issue we have to take care off. There is then no need to wait for 3 more failures before we even get a chance to fix and rerun the playbooks again.


d34dh0r53:
@BjoernT can you please file an issue for this and reference it in the PR?

odyssey4me:
:-1: As expressed upstream in https://bugs.launchpad.net/openstack-ansible/+bug/1496870 the retry function is only implemented in run-playbooks.sh, which is built for AIO's and gate checks. This is not something that belongs in any other use-case.

BjoernT:
odyssey4me: sure it is via https://github.com/openstack/openstack-ansible/blob/kilo/scripts/scripts-library.sh#L20, the install_bits loops through. I had to wait 9h for deploy.sh to actually hard fail

git-harry:
I don't think deploy.sh should be using install_bits. That function was created for use by the OSA gating as was the successerator. I believe it would be better to update the deploy.sh script to remove use of install_bits.

rackertom:
@git-harry @BjoernT @odyssey4me Should we abandon this in favor of making the deploy script simply use `openstack-ansible` instead of `install_bits`? I've got a test patch worked up that could be put in for review here pretty quickly.

git-harry:
@rackertom do it

PR #481, opened 2015-10-13
Lowering max_retries from 5 to 2 to catch build errors faster

BjoernT:
This change will cause OSA to fail more early in case repeating errors occur. It has been our experience if playbooks fail more than one, there's usually a issue we have to take care off. There is then no need to wait for 3 more failures before we even get a chance to fix and rerun the playbooks again.


odyssey4me:
:-1: As expressed upstream in https://bugs.launchpad.net/openstack-ansible/+bug/1496870 the retry function is only implemented in run-playbooks.sh, which is built for AIO's and gate checks. This is not something that belongs in any other use-case.

rackertom:
@git-harry @BjoernT @odyssey4me Should we abandon this in favor of making the deploy script simply use openstack-ansible instead of install_bits? I've got a test patch worked up that could be put in for review here pretty quickly.

git-harry:
@rackertom do it

Issue #495, opened 2015-10-15
keystone_api_local_check failes when v3 enabled


BjoernT:
.//keystone_api_local_check.py   1.2.3.4

status error 7/dist-packages/keystoneclient/session.py", line 397, in request\n    raise exceptions.from_response(resp, method, url)\nUnauthorized: The request you have made requires authentication. (HTTP 401) (Request-ID: req-817c5f06-8d05-42ce-ac0e-82d1b05cb3e8)\n
Traceback (most recent call last):
  File "./keystone_api_local_check.py", line 77, in <module>
    main(args)
  File "./keystone_api_local_check.py", line 67, in main
    check(args)
  File "./keystone_api_local_check.py", line 51, in check
    user_count = len(keystone.users.list())
  File "/usr/local/lib/python2.7/dist-packages/keystoneclient/utils.py", line 318, in inner
    return func(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/keystoneclient/v3/users.py", line 108, in list
    **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/keystoneclient/base.py", line 73, in func
    return f(*args, **new_kwargs)
  File "/usr/local/lib/python2.7/dist-packages/keystoneclient/base.py", line 361, in list
    self.collection_key)
  File "/usr/local/lib/python2.7/dist-packages/keystoneclient/base.py", line 113, in _list
    resp, body = self.client.get(url, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/keystoneclient/adapter.py", line 170, in get
    return self.request(url, 'GET', **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/keystoneclient/adapter.py", line 206, in request
    resp = super(LegacyJsonAdapter, self).request(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/keystoneclient/adapter.py", line 95, in request
    return self.session.request(url, method, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/keystoneclient/utils.py", line 318, in inner
    return func(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/keystoneclient/session.py", line 397, in request
    raise exceptions.from_response(resp, method, url)
[...]

PR #496, opened 2015-10-15
Bugfix for #495

BjoernT:
Fixing #495 which add the missing domain for a user list API request This does only apply to v3 identity API which is already hard coded in this script.


BjoernT:
will rebase

PR #497, opened 2015-10-15
Fixing #495 which add the missing domain for a user list API request

BjoernT:
This does only apply to v3 identity API which is already
hard coded in this script.


d34dh0r53:
:+1:

prometheanfire:
:+1:

PR #500, opened 2015-10-16
Add check validation step to maas role

stevelle:
Validation is meant to verify that for each check file applied there
is a check generated. It must rely on each host having the correct
check files applied as the alternative to this would mean duplicating
significant logic.

For filesystem checks, the file on the filesystem will use a '.'
instead of a '/' but otherwise each check file in the filesystem
shall have a name that matches the label of the check as recorded
server-side.

The rpc-maas-tool has added a checks mode and a --tabs boolean option
to enable this validation


d34dh0r53:
:+1:

PR #502, opened 2015-10-16
F5 configuration and monitor script updates

miguelgrinberg:
These are changes proposed by Jonathan Almaleh, who appears to be git impaired.

### DIFF
--- a/scripts/f5-config.py
+++ b/scripts/f5-config.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 #
 # (c) 2014, Kevin Carter <kevin.carter@rackspace.com>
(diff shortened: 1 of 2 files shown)


hughsaunders:
Another PR that needs networking attention (@rackertom @Apsu)

rackertom:
@hughsaunders I'm looking through the script. White space changes make this fun.

rackertom:
That entire python script needs a major refactor, or we need to upgrade our F5 infra so we can use miguel's f5 api work. The shell script is rough in a lot of spots too. But :shipit: 

This gets a :+1: from me with the note that future bugs are coming.

Issue #504, opened 2015-10-19
openstack-ansible has renamed the old run-upgrade script to run-upgrade-old.sh


mancdaz:
In order to continue to consume the QE tested upgrade path, we should alter upgrade.sh to use run-upgrade-old.sh

prometheanfire:
Do you have a reason for this?  why?

mancdaz:
@prometheanfire because the RPC product should consume a tested path, and is an explicit ask from support. The new set of broken out scripts that run-upgrade.sh now points at are an untested path.

prometheanfire:
ok, just wanted to be explicit in the bug report

stevelle:
both merged.

PR #505, opened 2015-10-19
bump osa sha and consume run-upgrade-old.sh script

mancdaz:
openstack-ansible has moved the original run-upgrade.sh script to
run-upgrade-old.sh. In order to continue to consume the QE tested path,
we should consume this instead of run-upgrade.sh, which now points at a
new, un(QE)tested upgrade path.

Partial: #504


prometheanfire:
:+1:

stevelle:
:+1:

Issue #486, opened 2015-10-14
neutron_metadata_local_check.py is not executable


stevelle:
This is a low-priority change but might as well.  Current behavior looks something like this:

```bash
  82642 ok: [rpc-maas-node4] => (item=/usr/lib/rackspace-monitoring-agent/plugins/heat_api_local_check.py)
  82643 ok: [rpc-maas-node4] => (item=/usr/lib/rackspace-monitoring-agent/plugins/horizon_check.py)
  82644 ok: [rpc-maas-node4] => (item=/usr/lib/rackspace-monitoring-agent/plugins/hp_monitoring.py)
  82645 ok: [rpc-maas-node4] => (item=/usr/lib/rackspace-monitoring-agent/plugins/keystone_api_local_check.py)
  82646 ok: [rpc-maas-node4] => (item=/usr/lib/rackspace-monitoring-agent/plugins/maas_common.py)
  82647 ok: [rpc-maas-node4] => (item=/usr/lib/rackspace-monitoring-agent/plugins/memcached_status.py)
  82648 ok: [rpc-maas-node4] => (item=/usr/lib/rackspace-monitoring-agent/plugins/neutron_api_local_check.py)
  82649 changed: [rpc-maas-node4] => (item=/usr/lib/rackspace-monitoring-agent/plugins/neutron_metadata_local_check.py)
  82650 ok: [rpc-maas-node4] => (item=/usr/lib/rackspace-monitoring-agent/plugins/neutron_service_check.py)
  82651 ok: [rpc-maas-node4] => (item=/usr/lib/rackspace-monitoring-agent/plugins/nova_api_local_check.py)
  82652 ok: [rpc-maas-node4] => (item=/usr/lib/rackspace-monitoring-agent/plugins/nova_api_metadata_local_check.py)
  82653 ok: [rpc-maas-node4] => (item=/usr/lib/rackspace-monitoring-agent/plugins/nova_service_check.py)
  82654 ok: [rpc-maas-node4] => (item=/usr/lib/rackspace-monitoring-agent/plugins/openmanage.py)
```

rackertom:
I will backport to Kilo as soon as #487 is merged

PR #507, opened 2015-10-19
Backport 487 to kilo

rackertom:
This plugin was added recently but did not have the mode set correctly. It gets fixed by the maas role but shouldn't be incorrect in the first place.

Resolves: #486
(cherry picked from commit 6abd4fa)


stevelle:
:+1:

prometheanfire:
:+1:

PR #508, opened 2015-10-19
Move verify of checks and alarms to own play

stevelle:
Moving these to their own play allows us more independance in timing
verification runs to ensure that the agent has time to competely sync
the server side checks before we try to verify.

Documents the new playbook in the README and adds a 30 second delay
after setup_maas.yml before the new playbook is run, when deploying
with scripts/deploy.sh with env value DEPLOY_MAAS=yes


stevelle:
this had a few issues, opening a new PR with fixes plus the upgrade script

PR #512, opened 2015-10-20
Maas upgrade play

stevelle:


sigmavirus24:
This needs to rebased as well

rpcd/playbooks/verify-maas.yml
```diff
@@ -0,0 +1,76 @@
...
+      with_items:
+        - "{{ expected_checks.stdout_lines }}"
+      when:
+        - expected_checks is defined
+        - maas_after_alarms is defined
+        - maas_after_alarms.stdout.find("{{ item }}") == -1
+      tags:
+        - maas-verify
+  vars_files:
+    - "{{ rpc_repo_path }}/playbooks/roles/rpc_maas/defaults/main.yml"
```
andymcc: This links to the openstack-ansible repo by default so fails out.

rpcd/playbooks/upgrade/maas_20151013141500.yml
```diff
@@ -0,0 +1,92 @@
...
+      local_action: lineinfile create=yes line="{{ item }}" dest=/tmp/maas_20151013141500_prior_checks.tab
+      with_items: "{{ maas_prior_checks.stdout_lines | sort }}"
+  tasks:
+    #
+    # Remove all checks for all entities
+    #   checks set up through CLI they cannot be managed by the agent through
+    #   template files
+    #
+    - name: Delete existing checks and alarms
+      script: "{{ maas_rpc_dir }}/scripts/rpc-maas-tool.py --force delete"
```
andymcc: I can't get this to work I think its because I'm using dedicated hosts - I'd be interested to see QE/testing on another lab to see if this works/doesn't work:

Exception: {'<REDACTED>', 'message': 'Checks may not be deleted on an entity managed by Cloud Databases.', 'code': 403, 'type': 'forbiddenError', 'details': 'Checks may not be deleted on an entity managed by Cloud Databases.'}

rpcd/playbooks/verify-maas.yml
stevelle: odd, I didn't have an issue testing that.

andymcc: rpcd/playbooks/group_vars/all.yml:rpc_repo_path: /opt/rpc-openstack/openstack-ansible

That's the default so I think that would never work, since the role exists in /opt/rpc-openstack/rpcd

sigmavirus24: Very odd, but @andymcc is correct: https://github.com/rcbops/rpc-openstack/blob/0684b32db78f72453dd676da48df8fe74953e156/rpcd/playbooks/group_vars/all.yml#L16

scripts/deploy.sh
```diff
@@ -133,6 +133,8 @@ install_bits horizon_extensions.yml
 # deploy and configure RAX MaaS
 if [[ "${DEPLOY_MAAS}" == "yes" ]]; then
   install_bits setup-maas.yml
+  sleep 30
```
sigmavirus24:
```sh
SHEP=30
HOW_MANY_SHEPS_DOES_IT_TAKE_TO_VERIFY_MAAS_WORKS=1

sleep $(( $SHEP * $HOW_MANY_SHEPS_DOES_IT_TAKE_TO_VERIFY_MAAS_WORDS ))
```

^ Should refactor to that

README.md
```diff
@@ -90,6 +93,8 @@ above.
   1. edit `/etc/openstack_deploy/user_extras_variables.yml` to add credentials
   2. run the MaaS setup plays:
      `cd /opt/rpc-openstack/rpcd/playbooks && openstack-ansible setup-maas.yml`
+  3. run the MaaS verify play:
```
sigmavirus24: We should probably explain why these should be run separately (timing)

rpcd/playbooks/upgrade/maas_20151013141500.yml
andymcc: Side note - I can delete it using raxmon-checks-delete so im interested what the difference is between that call and the rackspace_monitoring call used in the roc-maas-tool.py

rpcd/playbooks/verify-maas.yml
stevelle: right, and it should have shown up in my testing.  
I must have introduced a change somewhere along the way that broke this.

scripts/deploy.sh
stevelle: hue hue

README.md
stevelle: optionally, we include the verify play with retries and delays at the end of the setup play, similar to how I did in the upgrade play.  That might be more straight forward.  Thoughts?

sigmavirus24: That may be. I just don't want us to trigger our "more and more verbose as retries number gets higher" behaviour for this stuff

stevelle: Fair point. Added a note to the README to address this.

rpcd/playbooks/upgrade/maas_20151013141500.yml
stevelle: rewrote to use raxmon-checks-delete to see if that fixes it

stevelle:
closing this PR and opening a new one with cleaned up commits

Issue #516, opened 2015-10-20
Checking additional RabbitMQ metrics: fd and processes and sockets


BjoernT:
Currently we are not alarming on critical metrics like file descriptors or number of processes.
These are critical metrics which need to be monitored and are currently gathered by the rabbitmq_status.py script

BjoernT:
Please ignore the first 3 commits, only #517 is valid. Looks like didn't notice that I removed those previous commits

d34dh0r53:
:+1:

claco:
Master merged. Is someone working backports on this yet?

BjoernT:
Sure I can do it, the backport is actually my original pull request unless we want to convert the kilo  and juno branch as well to that new agent yaml format

stevelle:
With the backport of template driven maas checks which @sigmavirus24 initiated (above), these new checks will be added to kilo, so hold off on that one @BjoernT.  We will need a backport of this to juno however.

mancdaz:
The juno backport was completed https://review.openstack.org/#/c/236027/

PR #517, opened 2015-10-20
#516 Adding new RabbitMQ alarms (fd,proc,sockets)

BjoernT:
This fix is adding alarms for existing metrics
- fd_used
- proc_used
- socket_used
and will set off alarms when any of these metrics reach 90% of the
current configured total limit. Those metrics are already exposed
as part of the rabbitmq_status.py check.

Closes-Bug: #516


d34dh0r53:
:+1:

nrb:
+1

andymcc:
@robb-romans I know you were working on MaaS docs, this will add new MaaS checks - so if we are documenting a list of checks these should be added.

PR #519, opened 2015-10-21
 Upgrade maas to use template configuration

stevelle:
Adds a playbook to be run as a one-time operation to convert maas to
 use template based configuration. The playbook records a fingerprint
 of when it runs under /etc/openstack_deploy/migrations and will not
 run again if this file is present.

 The play will generate .tab files on the deploy host under /tmp for
 reporting the state of checks before and after upgrade.

### DIFF
--- a/rpcd/playbooks/maas-20151013-1-requirements.yml
+++ b/rpcd/playbooks/maas-20151013-1-requirements.yml
@@ -1,0 +1,43 @@
+---
+# Copyright 2015, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Part of MaaS Migration: maas-20151013.yml
+- name: Ensure MaaS Migration Requirements
+  hosts: hosts
+  user: root
+  tasks:
+    - name: Ensure migrations index exists
+      local_action: file path={{ migrations_dir }} state=directory
+      when: inventory_hostname == groups['hosts'][0]
+    - name: Locate evidence of prior MaaS migration
+      local_action: stat path={{ migrations_dir }}/maas-20151013
+      when: inventory_hostname == groups['hosts'][0]
+      register: maas_upgraded_result
+      changed_when: False
+    - name: Stop if migration has been run
+      fail: msg="MaaS has been previously run migration 20151013"
+      when:
+        - maas_upgraded_result is defined
+        - maas_upgraded_result.stat.exists | bool
+    - name: Locate existing Rackspace monitoring agent
+      stat: path=/etc/rackspace-monitoring-agent.cfg
+      register: raxmon
+      changed_when: False
+    - name: Halt if monitoring agent not found
+      fail: msg="MaaS is not installed and configured already"
+      when:
+       - raxmon is defined
+       - not raxmon.stat.exists
+
--- a/rpcd/playbooks/maas-20151013-2-pre-migrate.yml
+++ b/rpcd/playbooks/maas-20151013-2-pre-migrate.yml
@@ -1,0 +1,29 @@
+---
+# Copyright 2015, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Part of MaaS Migration: maas-20151013.yml
+- name: Record Prior Checks before MaaS Migration
+  hosts: hosts[0]
+  user: root
+  tasks:
+    - name: Inspect existing checks - all entities
+      script: "{{ maas_rpc_scripts_dir }}/rpc-maas-tool.py --tab True checks"
+      register: maas_prior_checks
+      changed_when: False
+    - name: Record report of existing checks - all entities
+      local_action: lineinfile create=yes line="{{ item }}" dest=/tmp/maas_20151013_prior_checks.tab
+      with_items: "{{ maas_prior_checks.stdout_lines | sort }}"
+  vars_files:
+    - "roles/rpc_maas/defaults/main.yml"
--- a/rpcd/playbooks/maas-20151013-3-purge-checks.yml
+++ b/rpcd/playbooks/maas-20151013-3-purge-checks.yml
@@ -1,0 +1,39 @@
+---
+# Copyright 2015, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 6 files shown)


sigmavirus24:
https://travis-ci.org/rcbops/rpc-openstack/builds/86599557#L1233 Looks like the lints are having trouble with verify-maas.yml playbook

stevelle:
Travis was having a significant outage last night so it was just hanging when I tried to get the details so I couldn't tell, but really we just need to merge #518 so that the verify-maas.yml file will exist.

stevelle:
Travis is choking on a ceph subrepo

rpcd/playbooks/maas-20151013-1-requirements.yml
```diff
@@ -0,0 +1,39 @@
...
+      local_action: file path={{ migrations_dir }} state=directory
+      when: inventory_hostname == groups['hosts'][0]
+    - name: Locate evidence of prior MaaS migration
+      local_action: stat path={{ migrations_dir }}/maas-20151013
+      when: inventory_hostname == groups['hosts'][0]
+      register: maas_upgraded_result
+      changed_when: False
+    - name: Stop if migraiton has been run
+      fail: msg="MaaS has been previously run migration 20151013141500"
+      when: maas_upgraded_result.stat.exists | bool
```
andymcc: This will fail on all nodes except hosts[0] because the "maas_upgraded_result" var was only created on host[0]. so we need the "when inventory_hostname == groups['hosts'][0]" clause I'm guessing.

rpcd/playbooks/maas-20151013-2-pre-migrate.yml
```diff
@@ -0,0 +1,29 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Part of MaaS Migration: maas-20151013.yml
+- name: Record Prior Checks before MaaS Migration
+  hosts: hosts
+  user: root
+  tasks:
+    - name: Inspect existing checks - all entities
+      script: "{{ maas_rpc_scripts_dir }}/rpc-maas-tool.py --tab True checks"
```
andymcc: The var "maas_rpc_scripts_dir" is used outside of the rpc_maas role so its not actually present. We'd need to add this to user_extras_variables.yml or group_vars, or some other place where it is accessible as a var outside of the role.

rpcd/playbooks/maas-20151013-2-pre-migrate.yml
```diff
@@ -0,0 +1,29 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Part of MaaS Migration: maas-20151013.yml
+- name: Record Prior Checks before MaaS Migration
+  hosts: hosts
```
andymcc: Perhaps we could change this to hosts[0] since all the tasks in here run against hosts[0] - it would make it simpler because the final task will fail on the other hosts since it tries to evaluate the with_items before the when clause (since with_items var doesn't exist on the other hosts it will fail out on those nodes - which isn't super important but isn't tidy either).

rpcd/playbooks/maas-20151013-3-purge-checks.yml
```diff
@@ -0,0 +1,37 @@
...
+# limitations under the License.
+
+- name: Migrate MaaS checks and alarms to template-driven configuration
+  hosts: hosts
+  user: root
+  pre_tasks:
+    - meta: flush_handlers
+  tasks:
+    - name: Lookup Entity ID
+      shell: raxmon-entities-list | grep {{ inventory_hostname }} | awk '{gsub(/id=/,""); print $2}'
```
andymcc: Could be good to reuse the old MaaS way of doing this - I think ti would be more complete (and tested):
shell: raxmon-entities-list  | grep "label={{ physical_host|quote }}{{ maas_fqdn_extension }} " | sed -e 's/^.* id=\(.*\) label=.*$/\1/g'

More examples from: https://github.com/rcbops/rpc-openstack/blob/master/rpcd/playbooks/roles/rpc_maas/tasks/local_setup.yml

rpcd/playbooks/roles/rpc_maas/defaults/main.yml
```diff
@@ -382,6 +382,7 @@ maas_pip_dependencies:
   - python-memcached
 
 maas_plugin_dir: /opt/rpc-openstack/maas/plugins/
+maas_rpc_scripts_dir: /opt/rpc-openstack/scripts
```
andymcc: See above regarding this var and its use outside of the rpc_maas role.

andymcc:
Some comments/issues above, but they are all pretty easily fixable - and once I made the changes described this worked for me (also got around the fact I couldn't delete checks before!) so LGTM!

One further question, can we make an overall migration playbook (excluding verification and perhaps the step 6 which happens after the verification) - these don't seem like there is a reason to not have these run one after the other? There aren't any required delays between finish of one to start of the next?

stevelle:
cleaned up the concerns @andymcc raised, then rebased. Travis failure still related to the ceph play repo.

rpcd/playbooks/maas-20151013-1-requirements.yml
```diff
@@ -0,0 +1,43 @@
...
+# Part of MaaS Migration: maas-20151013.yml
+- name: Ensure MaaS Migration Requirements
+  hosts: hosts
+  user: root
+  tasks:
+    - name: Ensure migrations index exists
+      local_action: file path={{ migrations_dir }} state=directory
+      when: inventory_hostname == groups['hosts'][0]
+    - name: Locate evidence of prior MaaS migration
+      local_action: stat path={{ migrations_dir }}/maas-20151013
```
sigmavirus24: Do we want to hardcode the dates?

rpcd/playbooks/maas-20151013-1-requirements.yml
```diff
@@ -0,0 +1,43 @@
...
+      when: inventory_hostname == groups['hosts'][0]
+      register: maas_upgraded_result
+      changed_when: False
+    - name: Stop if migraiton has been run
+      fail: msg="MaaS has been previously run migration 20151013141500"
```
sigmavirus24: Same question here

rpcd/playbooks/maas-20151013-1-requirements.yml
```diff
@@ -0,0 +1,43 @@
...
+  tasks:
+    - name: Ensure migrations index exists
+      local_action: file path={{ migrations_dir }} state=directory
+      when: inventory_hostname == groups['hosts'][0]
+    - name: Locate evidence of prior MaaS migration
+      local_action: stat path={{ migrations_dir }}/maas-20151013
+      when: inventory_hostname == groups['hosts'][0]
+      register: maas_upgraded_result
+      changed_when: False
+    - name: Stop if migraiton has been run
```
sigmavirus24: migraiton -> migration?

rpcd/playbooks/maas-20151013-2-pre-migrate.yml
```diff
@@ -0,0 +1,29 @@
...
+- name: Record Prior Checks before MaaS Migration
+  hosts: hosts[0]
+  user: root
+  tasks:
+    - name: Inspect existing checks - all entities
+      script: "{{ maas_rpc_scripts_dir }}/rpc-maas-tool.py --tab True checks"
+      register: maas_prior_checks
+      changed_when: False
+    - name: Record report of existing checks - all entities
+      local_action: lineinfile create=yes line="{{ item }}" dest=/tmp/maas_20151013141500_prior_checks.tab
```
sigmavirus24: Date + time

rpcd/playbooks/maas-20151013-1-requirements.yml
sigmavirus24: Migration name is just the date

rpcd/playbooks/maas-20151013-6-post-migrate.yml
```diff
@@ -0,0 +1,31 @@
...
+  hosts: hosts
+  user: root
+  tasks:
+    - name: Inspect existing checks - all entities
+      script: "{{ maas_rpc_scripts_dir }}/rpc-maas-tool.py --tab True checks"
+      when: inventory_hostname == groups['hosts'][0]
+      register: maas_post_checks
+      changed_when: False
+    - name: Record report of existing checks - all entities
+      local_action: lineinfile create=yes line="{{ item }}" dest=/tmp/maas_20151013141500_post_checks.tab
```
sigmavirus24: Date + time

sigmavirus24:
:+1: Ignoring the ceph problems with Travis

d34dh0r53:
:+1:

PR #524, opened 2015-10-22
Fixed issue where beaver was watching archived .gz files.

absalon-james:
Log messages were being shipped multiple days and multiple times
per day. This fix changes the watch targets for beaver ending in
* to *.log for nova, glance, neutron, heat, and cinder with the
goal of ignoring archived .gz files.

### DIFF
--- a/rpcd/playbooks/beaver.yml
+++ b/rpcd/playbooks/beaver.yml
@@ -22,15 +22,15 @@
   vars:
     beaver_log_monitors:
       - name: "cinder"
-        log_file: "/var/log/cinder/*"
+        log_file: "/var/log/cinder/*.log"
         tags: "cinder,openstack,oslofmt"
         multiline_regex_before: '.*\sTRACE\s.*'
       - name: "nova"
-        log_file: "/var/log/nova/*"
+        log_file: "/var/log/nova/*.log"
         tags: "nova,openstack,oslofmt"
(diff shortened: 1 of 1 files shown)


stevelle:
No bug filed here, so harder to do much with this.
What is the symptom of beaver watching compressed log files>

(CI failure was due to git submodules rewriting git history, unrelated to this change)

absalon-james:
Log messages are repeated multiple times per day and for multiple days. You can see beaver looking at .gz archived log files inside of the beaver log file.

stevelle:
This change needs to be rebased or merged from master, anyway.

If you can take a moment to create an issue with one example line from a beaver log file, I'll get it rebased and nudge this through the system to get it triaged and backported. I presume you would like this in kilo, and we can't keep track of what we fixed or where without an Issue so we're going to need that.

rackertom:
@absalon-james I'd love to see the path actually be * and a regex that excludes .gz and .bz2 files rather than just including *.log. This way, any log file placed in the directory that isn't compressed will be slurped in for processing. Basically include all and blacklist what we know is wrong.

Alternatively, we could make this a regex that includes *, but excludes *.[[:digit:]].* which effectively ignores rotated files but includes everything else.

tomc603:
Closing. duplicates 703

Issue #525, opened 2015-10-22
Add openstack client to hosts where pccommon is running


BjoernT:
The new version of pccommon does rely on the openstack client which we do not install everywhere the pccommon is running. This needs to be fixed.

PR #527, opened 2015-10-22
This fix will add new pip dependencies for the pccommon script

BjoernT:
This fix will add new pip dependencies for the pccommon script
Closes-Bug: #525


git-harry:
@BjoernT I can't see the openstack client being used in the pccommon script. Why is it a dependency?

BjoernT:
The new version will, it's currently in works

stevelle:
the above explains the reason for this.  LGTM +1 when it is rebased

d34dh0r53:
:+1:

Issue #534, opened 2015-10-30
Inconsistent Vars file include 


andymcc:
https://github.com/rcbops/rpc-openstack/blob/master/rpcd/playbooks/setup-maas.yml#L30

We don't do this for any other services, this seems unrequired.

PR #535, opened 2015-10-30
Remove unrequired vars file include

andymcc:
Move the cinder_service_port to maas defaults and remove the vars file
include for cinder default vars.

Fixes: #534

### DIFF
--- a/rpcd/playbooks/roles/rpc_maas/defaults/main.yml
+++ b/rpcd/playbooks/roles/rpc_maas/defaults/main.yml
@@ -127,6 +127,7 @@ maas_horizon_scheme: https
 # maas_heat_cloudwatch_scheme: https
 # maas_swift_proxy_scheme: https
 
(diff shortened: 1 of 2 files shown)


d34dh0r53:
:+1:

prometheanfire:
:+1: needs rebase

git-harry:
@andymcc there seem to be three different approaches to OpenStack service ports in the MaaS templates

- a variable with a specified default (cinder_service_port)
- a variable with an unspecified default (keystone_service_port)
- hardcoded in the template (everything else)

andymcc:
@git-harry sure, which is why only the cinder vars file had been included, because they're not consistent - but I'm not too sure what you're recommending though - we can change it to be in line with the others, but then we are essentially changing functionality for something.

I've rebased and made a new PR https://github.com/rcbops/rpc-openstack/pull/582
Going to close this.

Issue #538, opened 2015-11-02
Invalid syntax in MaaS Migration Playbook


sigmavirus24:
There are a couple when conditions in the upgrade playbooks resulting from the refactor where the condition was written as

```yaml
when:
  - foo
  - bar
```

When this is invalid for Ansible.

PR #539, opened 2015-11-02
Correct syntactically invalid when conditions

sigmavirus24:
It turns out that when you pass a list to when, you get some interesting
errors from Ansible. Using the normal syntax fixes this problem for us.

Addresses #538


rackertom:
:+1: Looks good

d34dh0r53:
:+1:

Issue #485, opened 2015-10-13
Lowering max_retries from 5 to 2 to catch build errors faster


BjoernT:
RPC-openstack utilizes the install_bits function from OSA which is a wrapper around successerator and by default it uses a MAX_RETRIES parameter of 5. 
I'm asking rpc-openstack to fail more early in case repeating errors occur. It has been our experience if playbooks fail more than one, there's usually a issue we have to take care off. There is then no need to wait for 3 more failures before we even get a chance to fix and rerun the playbooks again.

BjoernT:
Pull requests see https://github.com/rcbops/rpc-openstack/pull/480 and  https://github.com/rcbops/rpc-openstack/pull/481

d34dh0r53:
Thanks @BjoernT!

rackertom:
Given the comments in #480 and #481 it seems the intent was never to actually use `scripts-library.sh` in the deploy.sh script, and probably not in any other script in the rpc-openstack repo.

If @git-harry @BjoernT @odyssey4me agree that we should remove `scripts-library.sh` from the `deploy.sh` script, I have a patch I can easily submit for review. This would remove the use of `install_bits` which in turn would also remove the use of `successerator`. This does mean that there would be no automatic retry during a `deploy.sh` run, but it seems like in our case this is actually preferable.

BjoernT:
Yes this is alright by me.

git-harry:
I've closed pull requests #480 and #481. This issue is now pending a pull request from @rackertom

prometheanfire:
just needs backport now

PR #540, opened 2015-11-02
Remove erroneous depencency on scripts-library.sh

rackertom:
rpc-openstack was apparently never intended to depend on functions in the OSA
project's scripts-library.sh utility script. This change removes that
dependence and replaces the `install_bits` call with `run_ansible` instead.

Closes-Bug: #485 
Cherry-pick from 9e7ffb4


prometheanfire:
:+1:

sigmavirus24:
:+1:

Issue #533, opened 2015-10-29
rpc-release file shows incorrect DISTRIB_RELEASE value after Juno->Kilo upgrade


toxick:
I just (ostensibly) completed the Juno->Kilo upgrade in a lab, and found that the DISTRIB_RELEASE value has not been updated in the rpc-release file:

  656  2015-10-28 18:05:42  git clone -b r11.0.3 --recursive  https://github.com/rcbops/rpc-openstack /opt/rpc-openstack

root@<host>:~# cat /etc/rpc-release 
DISTRIB_ID="RPC"
DISTRIB_RELEASE="10.1.15"
DISTRIB_CODENAME="wee-5ive"
DISTRIB_DESCRIPTION="Rackspace Private Cloud"

d34dh0r53:
This should be resolved

byronmccollum:
@toxick Was this a clean upgrade all the way through, or did you encounter any failures along the way...? There is a scenario where I can see this happening, which involves a potentially misleading "what to do next" message in the event of a failure, leaving the cluster in a partially upgraded state.

Upgrading from Juno -> Kilo via RPC's `upgrade.sh` is a 2 step process:

First the [OSA bits get upgraded](https://github.com/rcbops/rpc-openstack/blob/kilo/scripts/upgrade.sh#L39) via OSA's `run-upgrade-old.sh`, then the [RPC bits get upgraded](https://github.com/rcbops/rpc-openstack/blob/kilo/scripts/upgrade.sh#L47) via RPC's `deploy.sh`.

If there is a failure upgrading the OSA bits, `run-upgrade-old.sh` informs you to manually re-run the failed playbook, plus any remaining playbooks. If you did re-run these playbooks, but stopped there, you would have a partially upgraded cluster. The OSA bits would be running Kilo, but the RPC bits would still be running Juno. And that's where it's misleading, because that's not the end of the upgrade. You also need to manually run the [rest of what's in RPC's `upgrade.sh`](https://github.com/rcbops/rpc-openstack/blob/kilo/scripts/upgrade.sh#L42-L51).

If you still have access to this environment, can you check the contents of both `/etc/rpc-release` and `/etc/openstack-release`. I have a feeling OSA will say Kilo, and RPC will say Juno.

byronmccollum:
Unless you've run `deploy.sh` in the mean time...in which case the RPC bits would have been upgraded, and `rpc-release` would reflect the correct version.

toxick:
[...]

PR #541, opened 2015-11-03
Adds a task to remove the previous rpc-release file

d34dh0r53:
Since there is already an existing `/etc/rpc-release` file a new file will never be dropped on upgrade.  This adds a task to remove the existing file if it exists, if no file exists the task will simply pass and a new `rpc-release` will be created.

Closes-Bug: #533

### DIFF
--- a/rpcd/playbooks/roles/rpc_support/tasks/rpc_release.yml
+++ b/rpcd/playbooks/roles/rpc_support/tasks/rpc_release.yml
@@ -13,6 +13,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+- name: Remove previous rpc-release file
+  file:
+    path: "/etc/rpc-release"
+    state: absent
+  tags:
+    - rpc-release
+
 - name: Drop rpc-release file
   template:
     src: "rpc-release.j2"


byronmccollum:
You shouldn't have to remove the file for it to get updated...no?

prometheanfire:
maybe changing to line in file would be better?

byronmccollum:
No, the templated file should get updated regardless.

prometheanfire:
ah right

d34dh0r53:
I'll do more testing with this, the template should probably be updated by Ansible.

git-harry:
-1 I agree with @byronmccollum, the whole point of the template is that it should update if variables change.

sigmavirus24:
I'm also :-1:  on this. Has the root issue been discovered or is #533 still a mystery to us?

byronmccollum:
@sigmavirus24: Speculation -- https://github.com/rcbops/rpc-openstack/issues/533#issuecomment-160813963

hughsaunders:
Ansible will update a template if the variables have been updated, the cause of #533 is not updating the variables. Anyhow, this change won't solve anything, but will cause the template to be marked as 'changed' on every run.

Issue #543, opened 2015-11-04
MaaS refactor upgrade playbooks need to ensure that the maas token is up-to-date


sigmavirus24:
`maas-20151013-2-pre-migrate.yml` runs `rpc-maas-tool.py` on an arbitrary host. That host will have a `/root/.raxrc` with a token that is out of date (in all likelihood). Either `maas-20151013-1-requirements.yml` or `maas-20151013-2-pre-migrate.yml` should run the task that updates the templated `/root/.raxrc` with the auth token from the extras variables file.

This will be necessary to run `2-pre-migrate.yml` and `3-purge-checks.yml` playbooks as part of upgrading from r11.0.0 to r11.x.0 with the MaaS refactor.

cc @xeregin

PR #546, opened 2015-11-04
Re-template raxrc file hosts for MaaS upgrade

sigmavirus24:
When users first deploy MaaS they have to place a token in the correct
variables file. Tokens expire quickly so when they go to upgrade the
MaaS checks, then they'll need to upgrade their MaaS token again. We
need to ensure the token in the variable file is then placed on the
hosts in /root/.raxrc. This allows rpc-maas-tool.py to work as expected
within the playbooks.

Addresses #543


miguelgrinberg:
:+1:

stevelle:
May be superseded by #563 which is simply a rebase

Issue #545, opened 2015-11-04
MaaS post migrate playbook errors fails on hosts that aren't hosts[0]


sigmavirus24:
Running `maas-20151013-6-post-migrate.yml` after a migration fails on the second task because `{{ maas_post_checks.stdout_lines | sort }}` doesn't result in a list or set.

This needs to `| default([])` to prevent spurious errors and accidental errors in running the migration playbooks.

sigmavirus24:
cc @xeregin

PR #548, opened 2015-11-04
Ignore hosts without stdout_lines in maas post-migrate

sigmavirus24:
If stdout_lines wasn't collected calling |sort on it will not produce
anything valuable for with_items (which expects a list or set). So we
need to make sure we default to an empty list in that case.

Addresses #545


prometheanfire:
:+1:

stevelle:
:+1: and merging when it gets rebased

miguelgrinberg:
:+1:

stevelle:
superseded by rebase: https://github.com/rcbops/rpc-openstack/pull/561

Issue #550, opened 2015-11-05
Need to monitor mysql connection count versus configured max


mancdaz:
As a result of some recent occurrences, it was deemed a good idea to be able to monitor mysql connection count versus the configured max, so that we can alert when the node is approaching the configured maximum.

PR #551, opened 2015-11-05
add some connection related monitors to galera_check plugin

mancdaz:
Up to now, we've been grabbing specific matching data from the output
of 'SHOW STATUS'. This commit adds the output from 'SHOW VARIABLES', so
that we can grab the configuration vars too.i Additionally, we no longer
limit the captured data from 'SHOW STATUS'.

Having all of the output from 'SHOW VARIABLES' and 'SHOW STATUS' will
make it easy to add any further metrics/alarms to this plugin in the
future.

Additionally, 'SHOW STATUS' was changed to 'SHOW GLOBAL STATUS' to
ensure we are gathering global stats as opposed to 'this session' stats.

Some new metrics are also added:

current_connections
max_configured_connections
max_seen_connections
percentage_used_connections

Partially-Addresses: #550

### DIFF
--- a/maas/plugins/galera_check.py
+++ b/maas/plugins/galera_check.py
@@ -13,10 +13,10 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
 import optparse
 import shlex
 import subprocess
+
 
 from maas_common import metric
 from maas_common import print_output
@@ -24,7 +24,7 @@ from maas_common import status_err
 from maas_common import status_ok
 
 
-def galera_status_check(arg):
+def galera_check(arg):
     proc = subprocess.Popen(shlex.split(arg),
                             stdout=subprocess.PIPE,
                             stderr=subprocess.PIPE,
@@ -35,7 +35,7 @@ def galera_status_check(arg):
     return ret, out, err
 
 
-def generate_query(host, port):
+def generate_query(host, port, output_type='status'):
     if host:
         host = ' -h %s' % host
     else:
@@ -45,10 +45,12 @@ def generate_query(host, port):
         port = ' -P %s' % port
     else:
         port = ''
-
-    return ('/usr/bin/mysql --defaults-file=/root/.my.cnf'
-            '%s%s -e "SHOW STATUS WHERE Variable_name REGEXP '
-            "'^(wsrep.*|queries)'\"") % (host, port)
+    if output_type == 'status':
+        return ('/usr/bin/mysql --defaults-file=/root/.my.cnf '
+                '%s%s -e "SHOW GLOBAL STATUS"') % (host, port)
+    elif output_type == 'variables':
+        return ('/usr/bin/mysql --defaults-file=/root/.my.cnf '
+                '%s%s -e "SHOW GLOBAL VARIABLES"') % (host, port)
 
 
 def parse_args():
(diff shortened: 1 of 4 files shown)


prometheanfire:
lgtm, gonna wait on this merge to grade the kilo backport

:+1:

mancdaz:
this has merged in rpc-maas

mancdaz:
re-opening since for now we are leaving plugins here

maas/plugins/galera_check.py
```diff
@@ -82,25 +86,35 @@ def print_metrics(replica_status):
            replica_status['wsrep_local_state_uuid'])
     metric('wsrep_local_state_comment', 'string',
            replica_status['wsrep_local_state_comment'])
+    metric('max_configured_connections', 'int64',
+           replica_status['max_connections'], 'connections')
+    metric('current_connections', 'int64',
+           replica_status['Threads_connected'], 'connections')
+    metric('max_seen_connections', 'int64',
+           replica_status['Max_used_connections'], 'connections')
+    metric('percentage_used_connections', 'int64',
```
git-harry: Is this metric required? The other two values are already being recorded and MaaS alarm language can be used to calculate percentages if required.

mancdaz: @git-harry I know, but it's not something that can be visualised or tracked if you don't make it a metric

git-harry: The other metrics track it. An alarm can alert on it. I would expect the visualisation software to allow for the data to be manipulated. My thinking is that the plugins should only be responsible for data collection.

stevelle:
I'm not seeing percent reported correctly (https://github.com/rcbops/rpc-openstack/pull/552#issuecomment-157204257)

mancdaz:
based on feedback, I've removed the percent used. I will add the required template/config changes shortly, so consider this a WIP for now.

mancdaz:
Template changes are now included, so this is good to review.

stevelle:
looks good to me but haven't actually run a test yet.

stevelle:
:+1: validated in lab with help from @xeregin

maas/plugins/galera_check.py
```diff
@@ -13,18 +13,20 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from __future__ import division
```
git-harry: This looks to be unused.

rpcd/playbooks/roles/rpc_maas/templates/galera_check.yaml.j2
```diff
@@ -24,3 +23,14 @@ alarms      :
             if (metric["wsrep_local_state_comment"] != "Synced" ) {
                 return new AlarmStatus(CRITICAL, "Galera cluster node not synced");
             }
+    percentage_used_mysql_connections :
+        label                   : percentage_used_mysql_connections--{{ ansible_hostname }}
+        notification_plan_id    : "{{ maas_notification_plan }}"
+        criteria                : |
+            :set consecutiveCount={{ maas_alarm_local_consecutive_count }}
+            if (percentage(metric["mysql_current_connections"], metric["mysql_max_configured_connections"]) > {{ mysql_connection_warning_threshold }} ) {
+                return new AlarmStatus(WARNING, "Mysql connectons has exceeded {{ mysql_connection_warning_threshold }}% of configured maximum");
```
git-harry: s/connectons/connections/

git-harry: warning and critical conditions should be swapped otherwise the alarm will never go critical.

maas/plugins/galera_check.py
```diff
@@ -45,10 +47,12 @@ def generate_query(host, port):
...
-
-    return ('/usr/bin/mysql --defaults-file=/root/.my.cnf'
-            '%s%s -e "SHOW STATUS WHERE Variable_name REGEXP '
-            "'^(wsrep.*|queries)'\"") % (host, port)
+    if output_type == 'status':
+        return ('/usr/bin/mysql --defaults-file=/root/.my.cnf '
+                '%s%s -e "SHOW GLOBAL STATUS"') % (host, port)
+    elif output_type == 'variables':
+        return ('/usr/bin/mysql --defaults-file=/root/.my.cnf '
+                '%s%s -e "SHOW VARIABLES"') % (host, port)
```
git-harry: Why are you not using 'SHOW GLOBAL VARIABLES' here but introducing GLOBAL for 'SHOW STATUS'?

mancdaz: because the variables we grab do not exist as session variables, so the default action is to pass back the global variables anyway.
Having said that, it's probably not a bad idea to force this to global in case we add a metric in the future and erroneously return the session variable which would not be what we want.

maas/plugins/galera_check.py
mancdaz: another remnant from when I was calculating percentage in the plugin. Will remove

mancdaz:
Thanks for the review @git-harry. All points have been addressed.

git-harry:
+1

mancdaz:
@stevelle @prometheanfire could we get a second +1/merge now I've addressed @git-harry suggestions?

prometheanfire:
:+1:  again

Issue #549, opened 2015-11-04
MaaS and other playbooks need pip_install_options var during upgrades


sigmavirus24:
OpenStack Ansible recently added the `pip_install_options` variable that allows deployers to pass extra options to pip tasks. This is something we need for our own usage of the pip ansible module.

We should mimic, closely, what OpenStack Ansible does ([example](https://github.com/openstack/openstack-ansible/blob/2c00d6b7b51d0ccbcce6b9ec91a61217a1de01dd/playbooks/roles/os_ceilometer/tasks/ceilometer_install.yml#L35)).

cc @toxick

sigmavirus24:
FWIW, @toxick ran into a situation where this was necessary when upgrading from 10.1.15 to r11.0.x. Some nodes had oslo.utils 1.4.1 installed because Juno's requirement looks like `oslo.utils>=1.4.0,<1.5.0` and Kilo's is similar except it looks like `oslo.utils>=1.4.0,!=1.4.1,<1.5.0`. Without being able to specify `pip_install_options` he could not force a downgrade from 1.4.1 to 1.4.0. 1.4.0 had a loose requirement on netaddr (i.e., `netaddr>=0.7.12`) while 1.4.1 had a very strict requirement (`netaddr>=0.7.12,<=0.7.13`) and the r11.0.0 repo directory only had `netaddr==0.7.18` (or `.19`, I honestly forget). So since the oslo.utils dependency was considered satisfied by pip, it couldn't satisfy the netaddr requirement.

byronmccollum:
This is a duplicate of bug #428, and PR #503 can be closed in favor of the more thorough PR #553.

sigmavirus24:
Ah sorry about that. I guessed that we hadn't run into this previously.

Thanks for catching that @byronmccollum

byronmccollum:
No worries...

I'd really like to see this backported to Kilo ASAP.

PR #553, opened 2015-11-05
Convert pip tasks to support pip_install_options

sigmavirus24:
In testing an upgrade from 10.1.15 to r11.0.x, @toxick found that some
packages installed on the host for MaaS need to be forcibly reinstalled
to match requirement constraints on newer packages. The case that
spurred this fix and the original issue was oslo.utils. The exact
information is documented in an issue comment.

Addresses #549


d34dh0r53:
:+1: LGTM

prometheanfire:
:+1:

prometheanfire:
backport?

sigmavirus24:
Backported

PR #557, opened 2015-11-09
use submodule to access maas plugins from rpc-maas repo

mancdaz:
The rpc-maas repo (kilo branch) was recently synchronised with all
plugins that were contained in this repo.  This PR removes the
local plugins and adds rpc-maas as a submodule in the same plugins
directory, meaning usage can continue unaffected.  

The subproject commit sha is currently head of rpc-maas kilo branch


hughsaunders:
PR of maas destruction :trumpet:.  Works for me after submodule init; submodule update, confirmed SHA is indeed head of kilo from rpc-maas.

prometheanfire:
:+1: ok

stevelle:
I'm not in favor of this change yet. It introduces a lot of additional friction for all future maintenance of MaaS in order to save a little friction occasionally.  

Points of friction this re-introduces:
- two separate PRs to separate repos for any added/removed plugin check, and for some modifications to checks that require changes to the configuration as well as the plugin itself.
- potential errors introduced in SHA updates. We have seen how error prone this is in practice with the subrepos we already carry.
- additional branches, issues, and PRs in an additional repo to monitor and manage

An alternative to checkout of the entire rpc-openstack repo in git as `sparse-checkout`

mancdaz:
@stevelle I take the point, but I'm not necessarily in agreement that it's a terrible idea to keep the plugins in their own discreet repo.  We've already been asked by the maas team to do this so that they can point others to our plugins easily, and it makes it much easier for other consumers to clone and contribute.

Using sparse-checkout imho isn't as straightforward as a separate repo. I'd also argue that a sparse checkout is really only beneficial when you have a huge repo which is difficult to traverse - in our case it would probably be just as easy to check out the full thing and just cd into maas/plugins, which is what we were trying to get away from.

Since we've left it pretty late for the team who are requiring this, I'd like us to get to some kind of agreement pretty quickly.

prometheanfire:
Also keep in mind that we originally (or at least I did) want to keep this as a separate submodule as well.

nrb:
I'm guessing time constraints would prevent this right now, but was there discussion of making the MaaS stuff a proper Ansible role, listed on Galaxy? While I appreciate that it's unlikely to happen in the very short term, I feel it would be cleaner to use that tool for distributing the plugins.

That said, I'm not against the submodule approach.

mancdaz:
@nrb surely the ansible role would be where the check/alarm configuration bits live (rather than in rpc-openstack), but we'd still need a place for the plugins to exist independently to foster collaboration from non ansible folk who want to consume them?

Either way, I'm taking your last comment as a +1 for the submodule :)

nrb:
I'm certainly not a -1, at least.

I suppose you're right - there would be 3 layers, in which case the plugins themselves would be a pip package, in my mind.

sigmavirus24:
So RHEL/OSP team needs this but also wants a separate repo for [the templates for the checks and alarms](https://github.com/rcbops/rpc-openstack/tree/master/rpcd/playbooks/roles/rpc_maas/templates). @mancdaz how should we do that?

stevelle:
To be clear, the RPCR team needs both templates and checks. I didn't hear a particular need for two repos, it just seems expedient.

sigmavirus24:
2 seems expedient so we don't have to reorganize rpcos

nrb:
Following on Steve's comments - this does seem like we're adding a lot of places to have to watch for PRs. I think it would totally feasible to see some change manifest that requires patches to OSA, rpc-openstack, and now the plugins. This means it will be harder for our direct users (and us) to track pieces of a change through many different repos.

mattt416:
My main concern here is that we've just pulled these plugins _in_ form a separate repo, and now we're going back.  This has implications on how we manage software releases and issue tracking but also changes the workflow for support on how they create issues and fixes.

I'd hate for this change to happen and then we decide that rpc-maas should actually be in gerrit, and you know we'll just make that change again.  :)

prometheanfire:
I kinda like the idea of doing it like we do the openstack services in osad, define a hash as a variable and check that out.

mancdaz:
Closing for now, as we are not going to do this in the foreseeable future.

PR #561, opened 2015-11-09
Ignore hosts without stdout_lines in maas post-migrate

stevelle:
If stdout_lines wasn't collected calling |sort on it will not produce
anything valuable for with_items (which expects a list or set). So we
need to make sure we default to an empty list in that case.

Addresses #545
Supersedes #548


stevelle:
rebased wrong change.

PR #547, opened 2015-11-04
Ensure raxmon-agent's conf directory exists

sigmavirus24:
When we're setting up the rackspace-monitoring-agent we should ensure
the conf.d directory exists. This should be created as part of the
package install but we found this was an issue on a torn down lab and
prevented a complete upgrade test. This could be an issue on other
environments too during upgrades.

Addresses #544

Issue #544, opened 2015-11-04
MaaS refactor playbooks need to ensure /etc/rackspace-monitoring-agent.conf.d exists


sigmavirus24:
We just ran `setup-maas.yml` in a lab and `/etc/rackspace-monitoring-agent.conf.d` exists prior to trying to place the check yml files in that directory. If they don't exist, setup-maas.yml will fail miserably.

sigmavirus24:
cc @xeregin

PR #562, opened 2015-11-09
Ensure raxmon-agent's conf directory exists

stevelle:
When we're setting up the rackspace-monitoring-agent we should ensure
the conf.d directory exists. This should be created as part of the
package install but we found this was an issue on a torn down lab and
prevented a complete upgrade test. This could be an issue on other
environments too during upgrades.

Addresses #544
Supersedes #547


stevelle:
Clean rebase, no conflicts. Playing on previous +1s I'm merging this.

Issue #564, opened 2015-11-09
Beaver won't start after host restart


BjoernT:
Beaver does not start anymore after a host has been restarted.
The reason for this is the missing /var/run/beaver directory.
Additionally beaver is not deleting the PID file after stopping.

git-harry:
This patch should also clean up [1]; /run is a tmpfs and so it doesn't make much sense for the playbooks/roles to be creating a directory there.

[1] https://github.com/rcbops/rpc-openstack/blob/master/rpcd/playbooks/roles/beaver/tasks/beaver_pre_install.yml#L52

BjoernT:
tmpfs will be cleaned up at reboot automatically

git-harry:
Exactly, that's why I'm suggesting you remove the directory's creation from the role.

d34dh0r53:
@BjoernT @git-harry is the merged fix appropriate for this or are there more steps needed to close this issue?

BjoernT:
Nope we're good

PR #565, opened 2015-11-09
Creating beaver pid directory when missing

BjoernT:
#564 Fixing Beaver PID handling
- Creating Beaver PID directory when missing
- Cleaning Beaver PID after shutting it down

### DIFF
--- a/rpcd/playbooks/roles/beaver/templates/beaver.j2
+++ b/rpcd/playbooks/roles/beaver/templates/beaver.j2
@@ -26,6 +26,13 @@ BEAVER_USER="{{ beaver_daemon_user }}"
 
 do_start() {
     test -f "${BEAVER_BIN}" || exit 0
+
+    local BEAVER_PID_DIRNAME="$(dirname $BEAVER_PID)"
+    if [ ! -d "$BEAVER_PID_DIRNAME" ]; then
+       mkdir -p "${BEAVER_PID_DIRNAME}"
+       chown "${BEAVER_NAME}":"${BEAVER_NAME}" "${BEAVER_PID_DIRNAME}"
+    fi
+
     if is_up
     then
         echo $"${BEAVER_NAME} has already been started."
@@ -60,6 +67,7 @@ do_stop() {
         I="$((I+1))"
     done
     echo '.'
+    test -f ${BEAVER_PID} && rm -f ${BEAVER_PID}
 }
 
 beaver_pid() {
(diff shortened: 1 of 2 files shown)


BjoernT:
Updated beaver_pre_install.yml to remove the pid dir

rpcd/playbooks/roles/beaver/templates/beaver.j2
```diff
@@ -37,6 +37,7 @@ do_start() {
 }
 
 do_stop() {
+    test -d "$( dirname $BEAVER_PID )" || mkdir "$( dirname $BEAVER_PID )" && chown $BEAVER_NAME:$BEAVER_NAME "$( dirname $BEAVER_PID )"
```
prometheanfire: shouldn't this be in start, not stop?

also, doing it all on one line is generally bad practice, can you fix that?

    local BEAVER_PID_DIRNAME="$(dirname $BEAVER_PID)"
    if [[ ! -d $BEAVER_PID_DIRNAME ]]; then
        mkdir -p "${BEAVER_PID_DIRNAME}"
        chown "${BEAVER_NAME}":"${BEAVER_NAME}"
    fi

BjoernT:
Oh yes you're right, cut'n paste victim... let me change it

prometheanfire:
can you make that a proper conditional?

BjoernT:
Yeah "proper" is subjective I personally like on liners... but I'm fine with that

rpcd/playbooks/roles/beaver/templates/beaver.j2
```diff
@@ -60,6 +67,7 @@ do_stop() {
         I="$((I+1))"
     done
     echo '.'
+    test -f ${BEAVER_PID} && rm -f ${BEAVER_PID}
```
prometheanfire: this seems new, why is it needed?

BjoernT: Because beaver is not deleting the file after it stopped which is quite annoying.

prometheanfire: ah, this is part of stop, didn't see that at first

prometheanfire:
:+1:

rpcd/playbooks/roles/beaver/templates/beaver.j2
BjoernT: No worries, I'll update the commit message

git-harry: @BjoernT don't forget to update the commit message.

d34dh0r53:
@BjoernT update the commit message, rebase and :+1:

BjoernT:
Yeah I had it re based this morning and updated the message already. Re based again.

Issue #571, opened 2015-11-10
Raxmon restart handler does not fire consistently on MaaS upgrade


sigmavirus24:
We noticed while running an upgrade of MaaS configuration from r11.0.0 to master we noticed that rackspace-monitoring-agent should have been restarted by a handler but wasn't. It appears we need to use `- meta: flush_handlers` in our playbooks to ensure the handlers are notified and fired appropriately.

See also: http://wherenow.org/ansible-handlers/

sigmavirus24:
cc @stevelle

andymcc:
Since the PR has merged for this, closing the issue out.

PR #572, opened 2015-11-10
Flush handlers to ensure the raxmon agent restarts

sigmavirus24:
During upgrade testing we noticed that the handler which restarts the
rackspace-monitoring-agent service does not always seem to trigger, even
when it should. After reading http://wherenow.org/ansible-handlers/, we
decided to use flush_handlers.

Addresses #571

### DIFF
--- a/rpcd/playbooks/roles/rpc_maas/handlers/main.yml
+++ b/rpcd/playbooks/roles/rpc_maas/handlers/main.yml
@@ -12,9 +12,3 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
-- name: Restart raxmon agent
-  service:
-    name: rackspace-monitoring-agent
-    state: restarted
-  delegate_to: "{{ physical_host }}"
--- a/rpcd/playbooks/roles/rpc_maas/tasks/ensure_filesystem_checks.yml
+++ b/rpcd/playbooks/roles/rpc_maas/tasks/ensure_filesystem_checks.yml
@@ -24,4 +24,3 @@
     - "{{ drives }}"
   when: inventory_hostname in groups["hosts"]
   delegate_to: "{{ physical_host }}"
-  notify: Restart raxmon agent
--- a/rpcd/playbooks/roles/rpc_maas/tasks/ensure_local_checks.yml
+++ b/rpcd/playbooks/roles/rpc_maas/tasks/ensure_local_checks.yml
@@ -26,7 +26,6 @@
     - inventory_hostname in groups["{{ item.group }}"]
     - item.name not in maas_excluded_checks
   delegate_to: "{{ physical_host }}"
-  notify: Restart raxmon agent
 
 - name: Remove checks that are excluded
   file:
@@ -37,4 +36,3 @@
   when:
     - item.name in maas_excluded_checks
   delegate_to: "{{ physical_host }}"
-  notify: Restart raxmon agent
--- a/rpcd/playbooks/roles/rpc_maas/tasks/ensure_remote_backup_checks.yml
+++ b/rpcd/playbooks/roles/rpc_maas/tasks/ensure_remote_backup_checks.yml
@@ -28,7 +28,6 @@
     - inventory_hostname != groups["{{ item.group }}"][0]
     - item.name not in maas_excluded_checks
   delegate_to: "{{ physical_host }}"
-  notify: Restart raxmon agent
 
 - name: Remove checks that are excluded
   file:
@@ -39,4 +38,3 @@
   when:
     - item.name in maas_excluded_checks
   delegate_to: "{{ physical_host }}"
-  notify: Restart raxmon agent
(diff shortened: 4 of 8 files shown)


sigmavirus24:
cc @hughsaunders @stevelle

stevelle:
complete and sufficient :+1:

git-harry:
@sigmavirus24 - While these meta tasks should work fine, it seems a little inefficient and defeats the purpose of the handler. Can you expand on what is happening? My understanding is that the role handlers should be queued up and run at the end of that tasks section of the playbook from which the role is called. If the restart is not happening each run, when a notify is triggered, does that mean there is a bug in Ansible?

The use of flush_handlers allows handlers to be flushed on an ad hoc basis, as I understand it that is generally used when a subsequent section of the role requires the handler to run and so the action cannot be postponed till the end of the playbook tasks. Looking at the patch it seems the meta tasks have been added after each new MaaS configuration file has been created, what requires the new set of checks/alarms to be loaded individually?

sigmavirus24:
@git-harry so symptoms:

- Some hosts have the agent restarted, others don't.
- Running `ansible hosts -m service -a 'name=rackspace-monitoring-agent state=restarted'` fixes this for hosts where it was not restarted

it seems that handlers are supposed to be the best (ansible) practice implementation of what we need to happen here but I'm more in favor of making this a task that happens at the end of `setup-maas.yml`. When I found @hughsaunders's blog post, however, it seemed like this is something that can happen in Ansible (a notification doesn't actually trigger an action) and so we decided to try using the flush_handlers meta action. If you agree that this seems like a better use case for a task (because there's no need to restart the agent every time we template each file) then I'll be happy to update this because that is the approach that I prefer as well.

git-harry:
@sigmavirus24, @hughsaunders blog post relates to failures part way through the run. Is the initial Ansible run failing?

If handlers being missed, because the run fails and the second run doesn't trigger a notify, is a concern for us then I feel like we should never use a handler because there is always the potential for it to fail.

If we want to be sure I think an explicit task, as you suggest, is probably the best way to do it although I'd be inclined to put it at the end of the role instead of the playbook. I guess something could be done with the API or by comparing the timestamps on the files with when the process was started but that's probably more hassle that it's worth.

sigmavirus24:
So it's failing on the initial run. The run reports success across the board which would lead you to believe that the handlers were successfully triggered and run.

git-harry:
@sigmavirus24 ah, I'd say that's something different and possibly worth some additional poking. If the handler is being run but the service is not restarting I would have thought there would be the potential for the same thing to happen with an explicit task given that a handler is itself just a task.

sigmavirus24:
It didn't happen when running `ansible hosts -m service -a 'name=rackspace-monitoring-agent state=restarted'` after the fact so I doubt that restarting the service is the problem itself.

andymcc:
Tested this - it works ok for me. Are we still discussing the approach or not? I'm happy for this to go in if it's working. I didn't personally have any issues with the handlers, and I wasn't able to recreate any, I guess the only downside of this is that if you have to explicitly run the restart task, but our approach to tagging tasks is non-existent anyway, so I don't have a problem with this.

TL;DR: LGTM happy to merge this.

git-harry:
Personally I don't see how this solves anything unless notify is broken, which strikes me as something that should be fixed instead of worked around, or at the very least logged upstream.

sigmavirus24:
(Rebased)

andymcc:
@git-harry the problem would be that we can't recreate the problem at this point, so hard to create an upstream bug about it.
I guess the concern is that if we are seeing issues, then this is a safer and more prudent approach without breaking any existing functionally - so I'm in favour of merging this - especially in the interests of getting an RC so QE can begin testing.

git-harry:
@andymcc except if you can't reproduce it, you don't know what you've fixed or even if you've fixed it.

andymcc:
I think that @sigmavirus24 experienced the issue and didn't after the fix - but I'm not sure what other course of action you would suggest? I think filing an upstream bug with no real evidence/or solution would end in a long winded failure...

git-harry:
@andymcc, the gist of this bug is that sometimes a handler just doesn't run when it's been notified. If that's the case, considering they are used all over the code, why has no one else reported that? The implication is that we should remove all the handlers from the code.

Yes, filing a bug upstream with no real evidence or solution would probably end in long winded failure, yet here it seems to result in a code change.

andymcc:
So again - I'm not sure what the suggestion is? 
If we both agree that filing a bug upstream is a waste of time - and we don't think its necessarily the handler since we would definitely have seen an issue elsewhere if it was (it could be the agent itself, perhaps restarting it later on works better - e.g. a timing issue), and yes I agree we can't be entirely certain this fixes the issue, especially since I've not personally seen the issue. However - if @sigmavirus24 reported the issue and this fixed it in his experience, then I personally would rather this went in than didn't. I'm not sure what other options we have aside from spending a LOT of time testing this further for no real benefit.

git-harry:
Well generally, in my experience, there is an expectation that a bug should be reproducible. If this bug is never seen again, it does not prove this commit fixed anything.

My suggestion is this is a bad methodology for identifying and fixing bugs and that the bug should have been left open to see if it happened on other deployments.

byronmccollum:
I can totally see how the use of successerator and a playbook failure after a handler was triggered, but before the actual handler was executed could leave the environment in a state where some or all agents never get restarted. The second pass from successerator wouldn't trigger a change, and thus wouldn't queue up the handler. Because of this, I've all but given up on handlers, since there is no state retained between playbook runs after a failure.

The removal of successerator has also uncovered several races that previously had gone unnoticed. While I initially wasn't in favor of removing successerator, it has force me to RCA these seemingly erratic failures, and get patches in.

git-harry:
@byronmccollum this bug describes a situation where the Ansible run doesn't fail but the monitoring agent isn't restarted by the handler.

Issue #576, opened 2015-11-11
deploy.sh does not copy all files to /etc/openstack_deploy


d34dh0r53:
The `deploy.sh` script will not copy all files to `/etc/openstack_deploy`.  This is due to the fact that `update_yaml.py` will not merge files that do not exist.

prometheanfire:
if we backport this we also need to cherry pick https://github.com/rcbops/rpc-openstack/commit/9ffc160dfddbbf22c0943f4ad55a769554a89697

PR #577, opened 2015-11-11
Check for files in /etc/openstack_deploy and copy if missing

d34dh0r53:
Before running `update_yaml.py` we check to see if the file exists
in `/etc/openstack_deploy`.  If not we copy the file from the skeleton
directory.

Fixes-Issue: #576

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -36,7 +36,12 @@ if [[ "${DEPLOY_AIO}" == "yes" ]]; then
   if [[ ! -d /etc/openstack_deploy/ ]]; then
     ./scripts/bootstrap-aio.sh
     pushd ${RPCD_DIR}
-        for i in $(find etc/openstack_deploy/ -type f -iname '*.yml'); do ../scripts/update-yaml.py /$i $i; done
+      for filename in $(find etc/openstack_deploy/ -type f -iname '*.yml');
+        if [[ ! -a "/${filename}" ]]; then
+          cp "${filename}" "/${filename}";
+        fi
+        do ../scripts/update-yaml.py "/${filename}" "${filename}";
+      done
     popd
     # ensure that the elasticsearch JVM heap size is limited
     sed -i 's/# elasticsearch_heap_size_mb/elasticsearch_heap_size_mb/' $RPCD_VARS


scripts/deploy.sh
```diff
@@ -36,7 +36,12 @@ if [[ "${DEPLOY_AIO}" == "yes" ]]; then
   if [[ ! -d /etc/openstack_deploy/ ]]; then
     ./scripts/bootstrap-aio.sh
     pushd ${RPCD_DIR}
-        for i in $(find etc/openstack_deploy/ -type f -iname '*.yml'); do ../scripts/update-yaml.py /$i $i; done
+      for i in $(find etc/openstack_deploy/ -type f -iname '*.yml');
```
prometheanfire: use a better variable name then 'i', I'd recommend 'filename'

d34dh0r53:
Fine ;)

scripts/deploy.sh
```diff
@@ -36,7 +36,12 @@ if [[ "${DEPLOY_AIO}" == "yes" ]]; then
   if [[ ! -d /etc/openstack_deploy/ ]]; then
     ./scripts/bootstrap-aio.sh
     pushd ${RPCD_DIR}
-        for i in $(find etc/openstack_deploy/ -type f -iname '*.yml'); do ../scripts/update-yaml.py /$i $i; done
+      for filename in $(find etc/openstack_deploy/ -type f -iname '*.yml');
+        if [[ ! -a /$filename ]]; then
+          cp $filename /$filename;
+        fi
+        do ../scripts/update-yaml.py /$filename $filename;
```
prometheanfire: can you properly quote your variables sir?

not needed in the 'if' but everywhere else...
that's something we likely need to do everywhere in this file.

prometheanfire:
:+1:

d34dh0r53:
Quotes added, please review again :)

scripts/deploy.sh
```diff
@@ -36,7 +36,12 @@ if [[ "${DEPLOY_AIO}" == "yes" ]]; then
   if [[ ! -d /etc/openstack_deploy/ ]]; then
     ./scripts/bootstrap-aio.sh
     pushd ${RPCD_DIR}
-        for i in $(find etc/openstack_deploy/ -type f -iname '*.yml'); do ../scripts/update-yaml.py /$i $i; done
+      for filename in $(find etc/openstack_deploy/ -type f -iname '*.yml');
+        if [[ ! -a /"$filename" ]]; then
+          cp "$filename" /"$filename";
+        fi
+        do ../scripts/update-yaml.py /"$filename" "$filename";
```
prometheanfire: "${foo}"

d34dh0r53:
There is no expansion happening here, why do we need the curly braces?

prometheanfire:
simply best practice

d34dh0r53:
hmm, not sure I agree with that but ok

scripts/deploy.sh
```diff
@@ -36,7 +36,12 @@ if [[ "${DEPLOY_AIO}" == "yes" ]]; then
   if [[ ! -d /etc/openstack_deploy/ ]]; then
     ./scripts/bootstrap-aio.sh
     pushd ${RPCD_DIR}
-        for i in $(find etc/openstack_deploy/ -type f -iname '*.yml'); do ../scripts/update-yaml.py /$i $i; done
+      for filename in $(find etc/openstack_deploy/ -type f -iname '*.yml');
+        if [[ ! -a /"${filename}" ]]; then
```
prometheanfire: escapes are not needed within the brackets iirc, @Apsu right?

d34dh0r53: Those aren't escapes, they are for pathing purposes.  cp a file from `etc/openstack_deploy` to `/etc/openstack_deploy`

scripts/deploy.sh
```diff
@@ -36,7 +36,12 @@ if [[ "${DEPLOY_AIO}" == "yes" ]]; then
...
+          cp "${filename}" /"${filename}";
```
mrda: First off, thanks for using "${...}" notation.  It's just a bit more robust that way.

Not that this is very important in this instance, but the "/" should be inside the quotes, not outside.  In this case it won't make any difference (unless ${filename} starts with a space) but going forward, I think the preferred syntax would be:

cp "${filename}" "/${filename}"

nrb:
+1

prometheanfire:
:+1:

PR #578, opened 2015-11-11
change to galaxy roles

prometheanfire:
youre welcome


stevelle:
confirming, this is the bomb

claco:
:-1: 

If this is a WIP for a bakeoff, maybe we should adjust the PR title accordingly to prevent accidental merge.

Additionally, please supply a real commit message, and convert "it puts the stuff in the places" type things to something more befitting of how we wish to present our work as an organization.

And finally, maybe we should file an issue for "submodules bad" and reference that in the bakeoffs.

prometheanfire:
that's fine, this is for a 'bakeoff' as it were

prometheanfire:
#580

prometheanfire:
I'll update the var names when I get in tomorrow

rpcd/playbooks/repo-fetcher.yml
```diff
@@ -0,0 +1,67 @@
...
+  tasks:
+    - name: lay the repos down
+      git:
+        dest: "{{ item.value.dest }}"
+        repo: "{{ item.value.repo }}"
+        version: "{{ item.value.version }}"
+      with_dict: "{{repos}}"
+    - name: generate rpcd ansible.cfg
+      copy:
+        dest: /opt/rpc-openstack/rpcd/playbooks/ansible.cfg
```
mattt416: We supply rpc-openstack/rpcd/playbooks/ansible.cfg -- should that get removed from the repo?

prometheanfire: it isn't removed, but being changed to a dynamically generated file because we are changing to galaxy roles.

prometheanfire:
recheck

prometheanfire:
23:21:44 jrpcaio-e5e_rabbit_mq_container-7a7ce8a6 : ok=31   changed=20   unreachable=0    failed=1
23:21:44 jrpcaio-e5e_rabbit_mq_container-8c191403 : ok=31   changed=20   unreachable=0    failed=1

not sure why this is happening...

prometheanfire:
recheck

prometheanfire:
recheck

prometheanfire:
recheck

mancdaz:
gate failure...

```
16:08:55 TASK: [rabbitmq_server | Enable queue mirroring] ****************************** 
16:08:56 failed: [jrpcaio-2f7_rabbit_mq_container-09e70d4b] => {"cmd": "/usr/sbin/rabbitmqctl -q -n rabbit list_policies -p /", "failed": true, "rc": 2}
16:08:56 stderr: Error: unable to connect to node 'rabbit@jrpcaio-2f7_rabbit_mq_container-09e70d4b': nodedown
```

prometheanfire:
notmyfault?

mancdaz:
@prometheanfire no, just calling out the failure. @hughsaunders is working on this gate

prometheanfire:
ok, this is good then

mancdaz:
recheck

mancdaz:
recheck

mancdaz:
recheck rabbit failure

prometheanfire:
recheck

prometheanfire:
I've moved test requirements to this repo as we shouldn't really depend on external test requirements and it lets tox work easily.

prometheanfire:
recheck

prometheanfire:
recheck

prometheanfire:
now using the same sha as master

prometheanfire:
21:35:52 jrpcaioiad-9fd_heat_apis_container-a457c08e : ok=7    changed=2    unreachable=0    failed=1   

notmyfault

prometheanfire:
recheck

rpcd/playbooks/repo-fetcher.yml
mancdaz: I think the point is that we supply this in the repo when it is cloned:

https://github.com/rcbops/rpc-openstack/blob/master/rpcd/playbooks/ansible.cfg

If you want to generate it dynamically, remove the supplied one

rpcd/playbooks/repo-fetcher.yml
```diff
@@ -0,0 +1,67 @@
...
+          # plugins required
+          filter_plugins = {{ repos.openstack_ansible.dest }}/playbooks/plugins/filters/
+
+          # Set the path to the folder in openstack-ansible which holds the action
+          # plugins required
+          action_plugins = {{ repos.openstack_ansible.dest }}/playbooks/plugins/actions/
+
+          [ssh_connection]
+          pipelining = True
+
```
mancdaz: Shouldn't this be a template rather than inlining an entire file?

mancdaz:
@prometheanfire I'm all good with moving the ceph roles to being pulled in by ansible-galaxy. I'm not convinced that changing osa to being pulled in by a custom playbook using the git module (which itself requires additional plumbing in the various scripts because ansible isn't installed yet) actually adds any value.

IIRC we'd agreed that roles would be pulled in by galaxy, and anything else would be left as submodules.

rpcd/playbooks/repo-fetcher.yml
prometheanfire: ah, sgtm

rpcd/playbooks/repo-fetcher.yml
prometheanfire: for this, meh

rpcd/playbooks/repo-fetcher.yml
```diff
@@ -0,0 +1,67 @@
...
+
+  vars:
+    repos:
+      openstack_ansible:
+        dest: /opt/rpc-openstack/openstack-ansible
```
cloudnull: I'd like to see this cloned directly to "/opt/openstack-ansible" and then to have our additional plumbing changed to point at that location. In this way we'd be able to keep the deployment inline with what the upstream docs say which will benefit our documentation team as well as our support teams.

prometheanfire: Iteration, I think this is a good idea though.  Also, if cloned there then it can't be a submodule.

cloudnull:
I'd like to see all of the submodules removed and the clone process clone the upstream OSA project into the documented location of /opt/openstack-ansible. this will ensure we can directly benefit from the upstream documentation that already has the deployment directory at /opt/openstack-ansible. To accomplish this we'd need to re-point our additional scripts to /opt/openstack-ansible OR create a symlink to link rpc-openstack/openstack-ansible > /opt/openstack-ansible either works and is simple to accomplish.

prometheanfire:
Ya, that's a good idea, it would be nice to have this happen for the liberty cycle.

claco:
Personally, I would vote Mitaka. Give the OSA roles work time to solidify and let Liberty work just focus on upgrade streamline and automation w/o another big change.

cloudnull:
My 2Cents -- I don't think there's any advantage of having this submodule. As we go into Mitaka we should not be working towards cleaning up the technical debt of Kilo and we should be thinking about how we begin leveraging the newly available roles that have been separated out of the main repo.

BjoernT:
From a support view, we like to get rid of the submodules as fast a possible since they cause enough confusion already. People forget to add --recursive or are not doing any "submodule update" once we change RPC Openstack tags.   Since we seem to use the same tools as OSA now to pull in roles I'm happy to support this change in ANY version, as fast as possible and in my eyes this seems to be the more "fool" prove approach than retaining the submodules. Pushing this ahead to mitaka is way to late in my eyes, knowing the operational risks.

hughsaunders:
recheck

mancdaz:
This has been superseded by https://github.com/rcbops/rpc-openstack/pull/804

cloudnull:
I think #804 and this change set can be done as dependencies of one another which will rid the repository of all submodules.

prometheanfire:
this has not been superseded by #804

mancdaz:
I personally don't think 'because we forget to add ```--recursive```` (which would be obvious immediately because the openstack-ansible directory would be empty) is a good enough reason to make change. If this is the only reason that can be given for this desire to not use a submodule for osa then I can't see how it's justifiable.

Having said that if the majority feel it's worth making the change, then this PR needs overhaul for it to be in a reviewable state.

d34dh0r53:
My vote at this point is :-1:, this has been discussed *ad naseum* and a significant amount of work has gone into #804.  Granted #804 does not totally supersede this but it covers a large portion of this PR and is nearly gating.  IMHO we should close this, merge #804, get a functional liberty to testing and then revisit this.  That being said I think this is a team decision and we should vote on it.

prometheanfire:
yes, I'm waiting to see what else that PR will take from this one before updating it

d34dh0r53:
recheck

prometheanfire:
recheck

prometheanfire:
recheck

rpc-aio passed this time at least

prometheanfire:
recheck

now neither aio pass

timeout on verify maas, are we hitting their API limits?

prometheanfire:
:trollface:

hughsaunders:
:-1:  Too much duplication for no benefit. As @d34dh0r53 said, this has been discussed in meetings, on github, in irc and rcbcore, as a team we decided to keep the osa submodule.

prometheanfire:
neat, reopening makes it get tested again

nrb:
:-1: While I'm not against this idea at some point, I don't think we need to do this *right now*.

prometheanfire:
ok, no code for you

Issue #583, opened 2015-11-12
Bug introduced by 37cc4fbda1551b71b572037f4b6519fc2b512d77


git-harry:
https://github.com/rcbops/rpc-openstack/pull/577 has introduced a bug.

./scripts/deploy.sh: line 40: syntax error near unexpected token `if'

prometheanfire:
fixed, problem isn't in kilo, closing

PR #584, opened 2015-11-12
Fix for loop syntax for updating Ansible config

git-harry:
37cc4fbda1551b71b572037f4b6519fc2b512d77 has introduced a bug because
the for-loop syntax is incorrect. This patch moves the 'do' so that the
loop will run.

Closes-issue: https://github.com/rcbops/rpc-openstack/issues/583


d34dh0r53:
:+1:

nrb:
+1

Issue #586, opened 2015-11-13
kibana role elasticsearch URL is not easily configurable


odyssey4me:
Currently to adjust the elasticsearch URL used by the kibana role only the ``external_lb_vip_address`` can be overridden, and that has wide-ranging consequences.

We should be able to set the whole URL more surgically.

stevelle:
no backport label from this week's triage.
merged, so closing.

PR #587, opened 2015-11-13
Allow Kibana's Elasticsearch public URL to be configurable

odyssey4me:
This patch adds a new configuration variable in order to allow the
public elasticsearch URL that Kibana uses for its javascript bits to
be configurable directly without affecting any other services.

The variable, elasticsearch_public_url, may be set in any of the
/etc/openstack_deploy/user_*.yml files.

Closes-Issue: https://github.com/rcbops/rpc-openstack/issues/586


git-harry:
+1

stevelle:
:+1:

PR #465, opened 2015-10-07
[WIP] Adds bashate linting for rpc-openstack

d34dh0r53:
WIP Do Not Merge
Fixes-Bug: #454

PR #592, opened 2015-11-17
Add bash syntax check to lint script

hughsaunders:
This addition loops over all (ba)sh scripts and runs a bash syntax check
on them. This should help catch some scripting errors.

Also removes the localhost only inventory as it wasn't working as
expected. Ansible was not recieving the input properly and was
generating the skeleton inventory itself. I have made this expclicit by
passing /dev/null as the inventory.

### DIFF
--- a/scripts/linting.sh
+++ b/scripts/linting.sh
@@ -21,9 +21,33 @@ trap "exit -1" ERR
 # Track whether linting failed; we don't want to bail on lint failures
 failed=0
 
-## Main ----------------------------------------------------------------------
-echo "Running Basic Ansible Lint Check"
+## Bash Syntax Check ---------------------------------------------------------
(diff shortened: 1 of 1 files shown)


hughsaunders:
That didn't go well then :(

hughsaunders:
Better, turns out the travis ci version of grep reads stdin if no path is specified, even when -r is given.

nrb:
Could this be done with bashate, like it is upstream? Looks like there's a WIP here: https://github.com/rcbops/rpc-openstack/pull/465

sigmavirus24:
+1 in my opinion for bashate.

hughsaunders:
Opinion seems to go in favour of bashate and @d34dh0r53's #465, so closing this pr.

Issue #595, opened 2015-11-17
linting.sh buries flake8 failures


mancdaz:
For example, there was a failing flake8 line in the PR, but the output from the travis run just says 'failed linting'. eg:

https://travis-ci.org/rcbops/rpc-openstack/builds/91598278#L1980

When running the relevant flake8 line manually locally, I saw the actual failure:

```
# flake8 $(grep -rln -e '^#!/usr/bin/env python' -e '^#!/bin/python' -e '^#!/usr/bin/python' * ) || failed=1
maas/plugins/nova_cloud_stats.py:103:80: E501 line too long (83 > 79 characters)
```

We should probably fix the linting script to output those errors properly

nrb:
Originally, my intention was to let all of the linting run before finishing the script. Should we instead simply remove the `|| failed=1` logic and let the linting checks die on first failure? I think that would be cleaner.

@rackertom @d34dh0r53 you have both been bitten by this too - which would be preferred?

mancdaz:
@nrb My personal preferred option would be to see all places that it is going to fail, in one go, so that I can fix them all in one go. 
If that's too painful, failing at the first failure (but ensuring it actually gets spat out) would work too.

rackertom:
If it's going to fail anyway, I'd prefer it to fail early. But there is obvious utility from getting all of the linting output at once so there isn't a trial-and-error period.

Then again, it would be just as easy to do further lint checks on my local machine like an adult I guess.

nrb:
@mancdaz That was my thinking, too. I'm doing a test PR on my own fork to see how the fail-fast method might work. In practice it's just removing a bunch of code. I think collecting then reporting failures may be more intensive and more code added to handle the collection.

d34dh0r53:
@nrb @mancdaz I think the fail first is the right way to do it, it's a relatively quick test so running locally and iterating through your failures is not too painful.  Plus I think adding a ton of code to catch all of the errors has the possibility to mask certain errors like we're seeing now.

stevelle:
I like the idea of a gate reporting all failures patiently, but getting it right adds complexity and opportunity for it to mask errors or give false positives if things change and the logic for this accumulation does not update in lockstep. I'd take a fail-fast if it's offered.

stevelle:
related 
#612

sigmavirus24:
I know we like having things "all-in-one" but separate travis checks would work well for fast failure without losing the information about the individual checks failing.

[...]

PR #596, opened 2015-11-17
Allow linting.sh to exit on first failure

nrb:
Using the trap to keep the linting script running beyond the first
failure was intended to provide complete coverage on each run of the
linting job, but in hindsight it buried information. The output of the
Ansible commands is long, resulting in a lot of output to parse.

Also, flake8 failures were being eaten by the `|| failed=1` logic, but not
being repeated at the end, so developers ended up having to re-run the
checks locally anyway to find out what exactly failed.

### DIFF
--- a/scripts/update-yaml.py
+++ b/scripts/update-yaml.py
@@ -13,9 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+import yaml
 import argparse
 import errno
-import yaml
 
 
 def parse_args():
--- a/scripts/linting.sh
+++ b/scripts/linting.sh
@@ -16,11 +16,6 @@
 ## Shell Opts ----------------------------------------------------------------
 set -euo pipefail
 
-trap "exit -1" ERR
-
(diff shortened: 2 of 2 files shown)


scripts/update-yaml.py
```diff
@@ -13,9 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+import yaml
 import argparse
 import errno
-import yaml
```
sigmavirus24: Did you intend to commit this so the tests would fail?

nrb: Yes, not sure why this is against this repo/branch - I was intending to demonstrate it against my own master branch until we had settled on the approach.

sigmavirus24: I seem to have missed the discussion of the approach. Is that still on-going or can we fix this up and resolve this issue?

nrb: https://github.com/rcbops/rpc-openstack/issues/595 is where most of it happened.

sigmavirus24:
Closing due to the re-ification that has merged.

PR #601, opened 2015-11-18
Adding information how to configure EMC XtremeIO inside rpc-openstack

BjoernT:
Adding example configuration for EMC XtremeIO storage


nrb:
I really like this idea, though I'm wondering if it would be more helpful upstream. Is there anything specific to Rackspace's deployment that I'm not aware of in here?

BjoernT:
Yeah I figured it's easier to contribute here, in context with Rackspace.
Although it's not specific to rpc-openstack. I'm happy to bring it into the OSA project

Issue #442, opened 2015-10-05
Check for empty rabbitmq channels list


rackertom:
When fetching the channels list from rabbitmq in the `rabbitmq_status.py` maas check, we can potentially fail by not checking the parsed returned json.

rackertom:
Potential patch in pr #443

rackertom:
Master and Kilo merged. Closing.

stevelle:
I know that juno doesn't have plugins stored here, but clicking reopen here until we have a juno backport over in rpc-maas

rackertom:
@stevelle Submitted https://github.com/rcbops/rpc-maas/pull/225 for rpc-maas.

stevelle:
merged to juno

PR #605, opened 2015-11-18
Check json_resp before using max()

rackertom:
Check the JSON response body for valid data when requesting channels.

After discussions with @git-harry and thinking the error over again, I've
decided that it is not technically an error to not have channels configured.
It is an error to receive an empty response when checking the overview and
nodes API endpoints. I will fix those in a later bug report.

Closes-Bug: #442 
Cherry-Pick: d379440


stevelle:
:+1:

sigmavirus24:
:+1:

Issue #610, opened 2015-11-19
MaaS: rabbitmq_status check gives KeyError: 'mem_used'


stevelle:
Error looks like this:
```
status error toring-agent/plugins/rabbitmq_status_new.py", line 172, in
<module>\n    main()\n  File
"/usr/lib/rackspace-monitoring-agent/plugins/rabbitmq_status_new.py",
line 139, in main\n    metrics[k] = {'value': resp_json[0][k], 'unit':
v}\nKeyError: 'mem_used'\n
```

Not sure if this is because OPC is deploying RabbitMQ 3.3.5 or something else, but they are getting a response from the nodes endpoint which does not consistently position the current host first in the nodes array.

e.g.
```
[root@overcloud-controller-1 ~]# cat nodes | python -m json.tool
[
    {
        "name": "rabbit@overcloud-controller-0",
        "running": true,
        "type": "disc"
    },
    {
        "applications": [
            {
                "description": "RabbitMQ AMQP Client",
                "name": "amqp_client",
                "version": "3.3.5"
            },
            {
                "description": "INETS  CXC 138 49",
                "name": "inets",
                "version": "5.9.8"
            },
```
this check running on overcloud-controller-1

stevelle:
http://paste.openstack.org/show/L7B5BB9jOjwTIuunwYYg/ as an example of the response format we get

http://paste.openstack.org/show/XQhuqdKHoeeQ96tDaFrd/ as an example of the response format from OPC

sigmavirus24:
Has anyone backported #611 yet?

stevelle:
@sigmavirus24 no backport yet the change was to support RPC-R and this issue did not originate in RPC-O so I don't know why it has a backport label, personally.

PR #611, opened 2015-11-19

stevelle:

Addresses: #610

### DIFF
--- a/maas/plugins/rabbitmq_status.py
+++ b/maas/plugins/rabbitmq_status.py
@@ -59,7 +59,16 @@ CONNECTIONS_METRICS = {"max_channels_per_conn": "channels"}
 
 def hostname():
     """Return the name of the current host/node."""
-    return subprocess.check_output(['hostname']).strip()
+    return subprocess.check_output(['hostname', '-s']).strip()
+
+
+def rabbit_version(node):
+    if ('applications' in node and 'rabbit' in node['applications']
+            and 'version' in node['applications']['rabbit']):
+        version_string = node['applications']['rabbit']['version']
+        return tuple(int(part) for part in version_string.split('.'))
+    else:
+        return tuple()
 
 
 def parse_args():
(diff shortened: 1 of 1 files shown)


maas/plugins/rabbitmq_status.py
```diff
@@ -135,8 +135,9 @@ def main():
     is_cluster_member = False
     if r.ok:
         resp_json = r.json()
+        current_node = [n for n in resp_json if n['name'].endswith(name)][0]
```
sigmavirus24: Is it plausible for the comprehension to be `[]`?

stevelle: poking at the code says that if the node falls out of the cluster, yes

maas/plugins/rabbitmq_status.py
```diff
@@ -135,29 +141,32 @@ def main():
...
+        if CLUSTERED:
+            if len(r.json()) < CLUSTER_SIZE:
+                status_err('cluster too small')
+            if not is_cluster_member:
+                status_err('{0} not a member of the cluster'.format(name))
+
+        for k, v in NODES_METRICS.items():
+            metrics[k] = {'value': nodes_matching_name[0][k], 'unit': v}
+
+        if not rabbit_version(nodes_matching_name[0]).startswith('3.3'):
```
sigmavirus24: So this will backfire if rabbit 3.2 or 3.1 or pretty much anything that isn't what we want. Might I suggest doing something like

```py
version_string = node['applications']['rabbit']['version']
return tuple(int(component) for component in version_string.split('.'))
```

In `rabbit_version` and then doing:

```py
if rabbit_version(nodes_matching_name[0]) > (3, 4):  # Or whatever version actually introduced this feature
```

Since tuple ordering will work nicely here.

sigmavirus24:
Looks okay to me. :+1:

maas/plugins/rabbitmq_status.py
```diff
@@ -59,7 +59,16 @@
 
 def hostname():
     """Return the name of the current host/node."""
-    return subprocess.check_output(['hostname']).strip()
+    return subprocess.check_output(['hostname', '-s']).strip()
+
+
+def rabbit_version(node):
+    if ('applications' in node and 'rabbit' in node['applications']
```
hughsaunders: maybe clearer as try/catch

hughsaunders:
Tested with AIO and works for me. Discussed with @git-harry that run_queue may not be the right thing here, but that predates this PR.

git-harry:
I logged a bug [1] to track the issue referenced by @hughsaunders 

[1] https://github.com/rcbops/rpc-openstack/issues/626

Issue #598, opened 2015-11-18
Maas: openstack service checks should support v2 or v3


stevelle:
This affects glance-api checks, cinder-api checks, etc.

stevelle:
Scope of change shouldn't be that bad thanks to this
https://github.com/rcbops/rpc-openstack/commit/9729b8b3727d108806d1a6e6b4eeb0a6d66d5d33

stevelle:
outline of likely solution is going to require specifying an extra arg to each check plugin: 
 `--auth-version "v2.0"`

PR #612, opened 2015-11-19
Discover Keystone Auth version to use in plugins

stevelle:
Discover Keystone Auth version to use in plugins
    
    To allow these checks to work in environments with different keystone
    auth versions enabled, we attempt to discover the version of keystone
    auth to use based on the auth_url with 'v3' being the only version
    we look for explicitly, and v2.0 being used as a fallback. This is
    then used to select the correct keystone client version (v3 or v2_0)
    to use when retrieving a current token or retrieving an updated
    service catalog to discover the services being checked.
    
    We could have inspected additional ENV vars to make the determination
    but that introduces and additional requirement which would complicate
    reuse by a minor increment. This solution adds no such barrier.
    
    Addresses: #598

This should to be tested in an OPC sandbox before acceptance.

### DIFF
--- a/maas/plugins/maas_common.py
+++ b/maas/plugins/maas_common.py
@@ -172,7 +172,8 @@ else:
 
 try:
     from keystoneclient.openstack.common.apiclient import exceptions as k_exc
-    from keystoneclient.v3 import client as k_client
+    from keystoneclient.v3 import client as k3_client
+    from keystoneclient.v2_0 import client as k2_client
 except ImportError:
     def keystone_auth(*args, **kwargs):
         status_err('Cannot import keystoneclient')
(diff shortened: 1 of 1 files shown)


sigmavirus24:
Does that maas_common function even take "auth_version"?

stevelle:
fixed long lines

@sigmavirus24 not sure which function you mean
get_endpoint_url_for_service takes 'version' already, which is how v3 support was added
get_keystone_client takes 'auth_version' now

stevelle:
Looks like travis is choking on something...

stevelle:
oh hey, submodules causing a branch conflict. imagine that.

stevelle:
flake8 passes locally.
There are zero playbook impacts to this change, but brew doesn't have a formula for ansible-lint so I haven't run that locally.

Not sure why this is failing and it's late on a Friday night before I vacate for holiday so someone else can work it out.

hughsaunders:
Looks good and tested ok (with v3 and v2 auth url) :+1:

git-harry:
Replaced with [1] due to linting error 'maas/plugins/maas_common.py:176:5: H306  imports not in alphabetical order (keystoneclient.v3.client, keystoneclient.v2_0.client)'

[1] https://github.com/rcbops/rpc-openstack/pull/627

Issue #603, opened 2015-11-18
Maas: memcached check enforces exact patch level 


stevelle:
should add either a less-strict-mode argument or just loosen checking to minor version number

sigmavirus24:
`--service` as an option name doesn't intuitively make sense to me. I know it's longer bug `--disable-version-check` is a better option name if we want to turn off the version check altogether as an option.

I'm okay having a list of VERSIONS that we know the software works with and check explicitly for each of them. We should re-evaluate the versions, however, each time we cut a new release (r11, r12, etc.)

stevelle:
I don't know if I made this clear by my proposed --service arg was meant to flag a specific version string as allowed: e.g. python memcachhed_status.py --service "1.4.15" 172.29.238.249

--service-version would work as well.

we could also --disable-version-check and disable the check, but that is like driving with headlights off, and I didn't want to make that the only option.

stevelle:
paging input from @git-harry as well, as someone who pays attention to maas checks and might have a different perspective on the fix.

This is something that is important for reuse of our check plugins by the RPC-R team for their up-coming soft launch.  The RHEL memcached package is used for OSP is 1.4.15 which causes this check to immediately fail.  When their engineers patch the check file to allow the 1.4.15 memcached service version instead, the checks work fine.

git-harry:
@stevelle, the use of an argument to specify the expected version moves the responsibility for validating a version works with the plugin from the plugin maintainer to the plugin consumer. For rpc-openstack I'm not sure that provides any benefit but for other users, assuming the new version works, I can see how that might be useful because verifying versions not used by rpc-openstack is always likely to be a low priority. I think it would need to include the appropriate warning so that users don't just bump the version without checking the output is consistent.

robb-romans:
The --version or -v flags almost always imply the incantation will simply return the version of the command itself. Using those as you intend would be confusing to me, at least.

sigmavirus24:
I'm increasingly in favor of keeping a list (or set) of versions that we know work and are good and making sure we update them each time we release a version that is on a new supported platform. It's the simplest solution. It ensures it only works on things we've tested rather than whatever the user feels like telling us is there. It's a far less disruptive change as well. (Incantation of the plugin, etc. do not change as a result or diverge based on product these plugins are deployed with.)

stevelle:
[...]

PR #613, opened 2015-11-19
Allow any memcached supported versions

stevelle:
To support OPC reuse of this plugin we now allow any memcached versions
which have been tested to pass the version valiation check.

Addresses: #603

### DIFF
--- a/maas/plugins/memcached_status.py
+++ b/maas/plugins/memcached_status.py
@@ -27,7 +27,7 @@ import memcache
 
 
 VERSION_RE = re.compile('STAT version (\d+\.\d+\.\d+)(?![-+0-9\\.])')
-VERSION = '1.4.14 (Ubuntu)'
+VERSIONS = ['1.4.14 (Ubuntu)', '1.4.15']
 MEMCACHE_METRICS = {'total_items': 'items',
                     'get_hits': 'cache_hits',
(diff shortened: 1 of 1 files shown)


d34dh0r53:
:+1:

mancdaz:
@stevelle maybe to avoid having to do this for each patch release, we could do as @sigmavirus24 suggested and just check for maj/min series, rather than the patch version. This assumes that the packages maintain the api output within a minor release series.

sigmavirus24:
@mancdaz 

- How often do we expect this list of versions to change?
- How many other platforms are we expected to support with these plugins? (Or put differently, do we think there are other platforms we can forsee these plugins being used for?)

mancdaz:
Hi @sigmavirus24 I'm not sure - it comes down to how often ubuntu and RedHat are likely to release new patch versions of it. Probably not that often, but probably worth futureproofing against anyway

As far as other platforms, at this time I only know of these two, but in theory anyone using openstack *could* use them to monitor using maas.

hughsaunders:
Tested and WFM, but I agree that this list looks fragile and not matching patch version will probably save some work.

git-harry:
Discussed with @hughsaunders and @mancdaz, we will revisit this if it becomes a regular occurrence.

Issue #597, opened 2015-11-17
pip conf not correctly building


BjoernT:
In case of the Juno (10.1.16) to Kilo upgrade (r11.1.0rc3), the upgrade script stopped because of a trusty backport repo issue so we continued running the setup-infrastructure and openstack-setup.yml playbooks. After that I wanted to run setup-maas but realized that the pip.conf was missing
the link:

http://172.29.236.207:8181/rpc-releases/r11.1.0rc3/ 

so I rerun repo-pip-setup.yml which now created a more correct pip.conf but the rpc-release version is now wrong:

find-links =
	http://172.29.236.207:8181/rpc-releases/r11.1.0rc2/
	http://172.29.236.207:8181/os-releases/11.2.4/

The reason for this is group_vars/all.yml contains the wrong version at rpc_release

This is more so an issue when running OSA playbooks which can rewrite the pip.conf removing the rpc-releases link again

byronmccollum:
`rpc_release` is incorrect for tags `r11.1.0rc3` and `r11.0.3`...

BjoernT:
There is actually another issue, the rpc_release.link seems to grow instead of being replaced with a new release information and you end up with multiple releases inside the pip.conf:
I can image this being a problem when we downgraded packages in new releases..

[global]
no-index = True
pre = True
timeout = 120
trusted-host =
	172.29.236.207

[install]
upgrade = True
allow_all_external = True
find-links =
	http://172.29.236.207:8181/rpc-releases/r11.1.0rc3/
	http://172.29.236.207:8181/rpc-releases/r11.1.0rc2/
	http://172.29.236.207:8181/os-releases/11.2.4/

BjoernT:
Hi all ! This is still an issue. How are we going to proceed here ?

mancdaz:
@BjoernT it seems there are 2 issues here. Firstly, when rc3 was released we had forgotten to update the rpc_release file.  This is now actually discovered automatically so should not present an issue again:

https://github.com/rcbops/rpc-openstack/blob/kilo/rpcd/playbooks/group_vars/all.yml#L19

When you say 'it is still an issue' what are you referring to?

mancdaz:
closing in favour of the newer issue https://github.com/rcbops/rpc-openstack/issues/1124

PR #619, opened 2015-11-20
Rebuild rpc_release pip links file everytime rather then apending to it

BjoernT:
Rebuild rpc_release pip links file everytime rather then apending to it
Fixes-Issue: #597

### DIFF
--- a/rpcd/playbooks/repo-pip-setup.yml
+++ b/rpcd/playbooks/repo-pip-setup.yml
@@ -24,7 +24,14 @@
   roles:
     - pip_lock_down
   post_tasks:
-    - name: Add new pip link
+    - name: Remove old pip link
+      file:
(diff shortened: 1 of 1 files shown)


d34dh0r53:
:+1:

sigmavirus24:
:+1:

hughsaunders:
This will cause the template to be marked as changed on every run. Would be better to add regexp: "." to the current lineinfile job. :-1:

git-harry:
-1 I think it would be better to use the copy module, it has a content parameter that can be used to specify the contents of the file. That removes the need for deleting the old file and doesn't require a regexp type search for the line. Using copy is essentially doing the job of a simple template file which would be the other alternative but possibly a bit heavyweight for something so simple.

PR #624, opened 2015-11-23
Add MAAS Metric for rabbit queue length

### LINKED ISSUE #633: Rabbitmq Queue Length Metric required
"For the RabbitMQ queue length, the scenario we want to detect is a queue growing due to a failed consumer or consumers that can’t keep up with load."

"Support confirmed (via @bjoern) that monitoring and alerting on the following single metric would be sufficient: ((Total message count across all queues) – (messages in notifications queue))"

Further details: 
https://trello.com/c/widYAV5F/35-add-new-monitor-rabbitmq-queue-length

hughsaunders:
Add MAAS Metric for rabbit queue length

* Adds a single metric: queued_messages_excluding_notifications
* Adds an alarm with configurable threshold (default 100)
* Refactors the rabbitmq plugin to reduce duplicated code

Closes #633
Trello: https://trello.com/c/widYAV5F/35-add-new-monitor-rabbitmq-queue-length

### DIFF
--- a/maas/plugins/rabbitmq_status.py
+++ b/maas/plugins/rabbitmq_status.py
@@ -28,6 +28,7 @@ import requests
 OVERVIEW_URL = "http://%s:%s/api/overview"
 NODES_URL = "http://%s:%s/api/nodes"
 CONNECTIONS_URL = "http://%s:%s/api/connections?columns=channels"
+QUEUES_URL = "http://%s:%s/api/queues"
 
 CLUSTERED = True
 CLUSTER_SIZE = 3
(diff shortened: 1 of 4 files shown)


hughsaunders:
Removed erroneous SHA bump from commit

git-harry:
@hughsaunders rebase required

maas/plugins/rabbitmq_status.py
```diff
@@ -85,79 +86,91 @@ def parse_args():
...
             status_err('cluster too small')
         if not is_cluster_member:
             status_err('{0} not a member of the cluster'.format(name))
 
+
+def _get_queue_metrics(session, metrics, host, port):
+    response = _get_rabbit_json(session, QUEUES_URL % (host, port))
+    all_queued_messages = sum([q['messages'] for q in response])
+    notification_messages = sum([q['messages'] for q in response
+                                if q['name'].startswith('notification.')])
```
mancdaz: this will fail to pick up the correct queue as the notifications queue is ```notifications.info```

mancdaz:
:-1: this doesn't currently work

git-harry:
-1 I'm opposed to the metric queued_messages_excluding_notifications. The commit provides no details but my understanding is that this is being done because the notifications queue is not consumed. Instead of working round the issue we should solve it [1]. This prevents rabbitmq from consuming unnecessary resources when the queue is not required.

[1] https://github.com/openstack/oslo.messaging/blob/3a5db723aa0613a99df6d95eecf6d8cecf2ca79d/doc/source/FAQ.rst

mancdaz:
@hughsaunders did you mean to include a sha bump for rpc-openstack in this commit?

hughsaunders:
Removed sha bump. This is now tested and ready for review.

hughsaunders:
updated metric names in readme

mancdaz:
@git-harry I agree we should turn off notifications if we don't want them.  Turns out we already do, except for neutron. So I've started the review process to get that fixed:

https://review.openstack.org/#/c/252416/

In the interim, and for current deployments, I think we need to go with the current approach of generating a metric that has the notifications messages subtracted.

I've tested on a rpc-openstack master deployment and this current patchset works for me

Therefore, for me right now,  :+1:

andymcc:
Works for me - tested both alarms + new metrics.

PR #625, opened 2015-11-23
Add check/alarm for monitoring network throughput

git-harry:
The new MaaS template is used to create a set of checks and alarms, one
per specified NIC, for monitoring network throughput.

As part of this commit, the conntrack monitor is moved to 'kernel'
monitors, so that the network monitors can be for actual network things.

### DIFF
--- a/rpcd/playbooks/roles/rpc_maas/tasks/kernel.yml
+++ b/rpcd/playbooks/roles/rpc_maas/tasks/kernel.yml
@@ -1,0 +1,18 @@
+---
+# Copyright 2014, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+- include: ensure_local_checks.yml
+  vars:
+    checks: "{{ kernel_checks_list }}"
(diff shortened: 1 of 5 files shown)


rpcd/playbooks/roles/rpc_maas/tasks/kernel.yml
```diff
@@ -0,0 +1,18 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+- include: ensure_local_checks.yml
+  vars:
+    checks: "{{ kernel_checks_list }}"
```
mancdaz: I would just move this up into main, and remove the need for this intermediate file

mancdaz:
Tested latest version and works ok. Aside from potential extra file that's not needed (and I agree there's no clear pattern that defines how to do this), I'm :+1:

git-harry:
@mancdaz I've tweaked this so that it should automatically get the NIC speed. This doesn't work on VMs so the variable net_max_speed will need to be set to test this on a cloud server.

mancdaz:
Tested on cloud by overriding net_max_speed. Seems to work and drop the correct config files/create alarms etc.

Though I have to say @git-harry I pretty much hate the fact that we now just get tons of red ignored failures because /sys/class/net/<name_of_interface>/speed doesn't exist.

hughsaunders:
I deployed this overriding network_checks_lis and net_max_speed. Only got one ignored failure. Difficult to avoid that failure as the file exists and is marked as readable, you don't get the error until you actually try and read. Tested alarm in ui by setting net_max_speed to one, alarmed and cleared successfully. :+1:

rpcd/playbooks/roles/rpc_maas/tasks/kernel.yml
sigmavirus24: I'm not concerned about this being in main. In fact I think this follows the existing pattern well enough to leave it.

sigmavirus24:
:+1: from me too.

Just waiting to hear back from @d34dh0r53 about when we should start merging these.

PR #627, opened 2015-11-24
Discover Keystone Auth version to use in plugins

git-harry:
To allow these checks to work in environments with different
keystone auth versions enabled, we attempt to discover the version
of keystone auth to use based on the auth_url with 'v3' being the
only version we look for explicitly, and v2.0 being used as a
fallback. This is then used to select the correct keystone client
version (v3 or v2_0) to use when retrieving a current token or
retrieving an updated service catalog to discover the services being
checked.

We could have inspected additional ENV vars to make the determination
but that introduces and additional requirement which would complicate
reuse by a minor increment. This solution adds no such barrier.

Addresses: #598

### DIFF
--- a/maas/plugins/maas_common.py
+++ b/maas/plugins/maas_common.py
@@ -172,7 +172,8 @@ else:
 
 try:
     from keystoneclient.openstack.common.apiclient import exceptions as k_exc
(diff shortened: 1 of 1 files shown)


git-harry:
+1

Issue #630, opened 2015-11-24
Elasticsearch Curator cron missing and misnamed fields


d34dh0r53:
In the Elasticsearch Curator cron there are several missing options and a misspelled option:

* `--older_than` should be renamed to `--older-than`
* `delete` should be `delete indices`
* the flag `--time-unit days` needs to be added
* the flag `--timestring %Y.%m.%d` needs to be added
* the flag `--prefix logstash` needs to be added

- [x] master
- [ ] kilo

jacobwagner:
@d34dh0r53 I would also sit with OPS and find out what a good deletion default would be, or maybe we could better document how to override it. In our lab we were seeing about 100G a day, so @ 2TB that would mean a full disk in ~18 days. This brought the elasticsearch service down with caused the upgrade ELK plays to fail when trying to connect to elasticsearch. Trickle effect = bad times for all.

d34dh0r53:
@jacobwagner yeah, I agree we should probably default to a lower value, but this seems like a very deployment dependent setting.  The proper value is going to be based on the size of the environment as well as compliance factors for the customer.

d34dh0r53:
Note to self, create `kilo` backport

PR #631, opened 2015-11-24
Add required parameters to ES curator and fix typo

d34dh0r53:
This adds the following required parameters to ElasticSearch curator
* `--time-unit days`
* `--timestring %Y.%m.%d`
* `--prefix logstash`
* `indices`

And changes the flag `--older_than` to `--older-than`.

Closes-Bug: #630

### DIFF
--- a/rpcd/playbooks/roles/elasticsearch/tasks/elasticsearch_post_install.yml
+++ b/rpcd/playbooks/roles/elasticsearch/tasks/elasticsearch_post_install.yml
@@ -69,7 +69,7 @@
     minute: 0
     hour: 0
     user: "root"
-    job: "/usr/local/bin/curator --host {{ hostvars[inventory_hostname]['container_address'] }} delete --older_than {{ elasticsearch_prune_days }}"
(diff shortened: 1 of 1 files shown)


sigmavirus24:
Looks good to me (:+1:) but I haven't tested this and it's no clear how to do so.

tomc603:
@d34dh0r53 Some questions...

The default for `--time-unit` is `days`, but explicitly specifying (probably) won't hurt.
The default for `--prefix` is `logstash-`. Did we intend to use `logstash` instead?

d34dh0r53:
@tomc603 so I could not get it to work manually without specifying the `--prefix logstash` parameter.

tomc603:
:+1:

d34dh0r53:
I added impacts doc because a special note should be added to the documentation indicating that `{{ elasticsearch_prune_days }}` should be set based on the size of the environment and the storage available on the logging server.  A new cluster operator could determine the storage used by elasticsearch for a few days and then set the value accordingly.

hughsaunders:
recheck - failed on quota

tomc603:
recheck - quota again

Issue #634, opened 2015-11-26
rpc-support playbook must configure pip.conf


BjoernT:
In case the rpc-support playbook is started after setup-openstack, the pip configuration needs to be configured. Otherwise it will pull down pip packages directly from upstream

d34dh0r53:
This needs to be backported

BjoernT:
Backport added

andymcc:
Closing this - both PRs have merged.

PR #635, opened 2015-11-26
Configure pip.conf prior to running the rpc-support playbook

BjoernT:
This fix makes sure that a proper pip.conf is in place so it can
pull down pip packages from the local repo clones rather than
upstream. Usually rpc-support is executed after the repo-pip-setup.yml
playbook but this fix will include repo-pip-setup.yml as a safety
precaution

Closes-Bug: #634


d34dh0r53:
:+1:

sigmavirus24:
:+1:

Issue #637, opened 2015-11-27
Add details for Ceph vars


andymcc:
As a result of this Docs issue, we want to add more info on the ceph vars
https://github.com/rpcdocs/rpcdocs/issues/421

PR #638, opened 2015-11-27
Add details for Ceph vars in user_extras_variables

andymcc:
Adds a comment above each ceph var to indicate the variables use.

Fixes Issues: #637

### DIFF
--- a/rpcd/etc/openstack_deploy/user_extras_variables.yml
+++ b/rpcd/etc/openstack_deploy/user_extras_variables.yml
@@ -95,18 +95,32 @@ maas_filesystem_critical_threshold: 90.0
 #   - { package: "rabbitmq-server", origin: "www.rabbitmq.com" }
 #   - { package: "*", release: "MariaDB" }
 
-# Ceph
+# Ceph options
+# fsid is the unique identifier for your object store.
 fsid: '{{ fsid_uuid }}'
+# directory for backing up ceph keys.
 fetch_directory: /etc/openstack_deploy/ceph_fetch
+# Use stable version of ceph
 ceph_stable: true
+# Specify ceph release name
 ceph_stable_release: hammer
+# Enable OpenStack support inside the ceph-ansible playbooks
 openstack_config: true
+# Use raw journal devices
 raw_multi_journal: true
+# Set the journal size to: "Size of journal device / number of devices for which it is a journal"
+# E.g. Given a 400G journal disk with 5 disks using it as their journal device, the journal size should be 80G each or 80000
 journal_size: 80000
+# Default number of replicas for a pool
 pool_default_size: 3
+# Default min number of replicas for ceph to consider the state to be not degraded.
 pool_default_min_size: 2
+# The % of disk used before an osd is considered full - Ceph will be marked critical and stop functioning if an OSD reaches this %
 mon_osd_full_ratio: .90
+# The % of disk used before an osd is considered nearfull - Ceph will still work but will return a HEALTH_WARN.
 mon_osd_nearfull_ratio: .80
+# Determines whether we use secure cluster flags.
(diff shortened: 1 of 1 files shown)


asettle:
Thank you @andymcc :)

robb-romans:
@andymcc @asettle  Documented here: https://github.com/rpcdocs/rpcdocs/pull/445

sigmavirus24:
:+1:

BjoernT:
I had to manually add cinder_ceph_client_uuid inside the user_extras_secrets otherwise the MAAS playbooks did not work that should be mentioned too

rpcd/etc/openstack_deploy/user_extras_variables.yml
```diff
@@ -95,18 +95,32 @@ maas_filesystem_critical_threshold: 90.0
...
+# Ceph options
+# fsid is the unique identifier for your object store.
 fsid: '{{ fsid_uuid }}'
+# directory for backing up ceph keys.
 fetch_directory: /etc/openstack_deploy/ceph_fetch
+# Use stable version of ceph
 ceph_stable: true
+# Specify ceph release name
 ceph_stable_release: hammer
+# Enablie openstrack support inside the ceph-ansible playbooks
```
mattt416: s/Enablie/Enable/g

rpcd/etc/openstack_deploy/user_extras_variables.yml
```diff
@@ -95,18 +95,32 @@ maas_filesystem_critical_threshold: 90.0
...
 openstack_config: true
+# Use raw journal devices
 raw_multi_journal: true
+# Set the journal size to: "Size of journal device / number of devices for which it is a journal"
+# E.g. Given a 400G journal disk with 4 disks using it as their journal device, the journal size should be 100G each or 100000
```
mattt416: "Given a 400G journal disk with 5 disks using it as their journal device, the journal size should be 80G each or 80000"

May as well use the ratio that we'll have in prod.

andymcc:
@mattt416 fixed those typos, thanks for the spot!

@BjoernT the "cinder_ceph_client_uuid" var is in the user_secrets.yml in the openstack-ansible repo, so if you have deployed properly you should have this (regardless of whether you are deploying ceph or not) - since its required for openstack-ansible deployments that are using ceph regardless of whether we deployed the ceph or not:

$ (git::master) grep -ir "cinder_ceph_client_uuid" *
etc/openstack_deploy/openstack_user_config.yml.example:#           rbd_secret_uuid: "{{ cinder_ceph_client_uuid }}"
etc/openstack_deploy/user_secrets.yml:cinder_ceph_client_uuid:

This would mean you don't need to add it to user_extras_secrets.yml on new deploys as long as you follow the normal steps (which are to copy over user_secrets.yml) from the openstack-ansible repository.

BjoernT:
Please bear in mind that the user_extras_secrets file is not copied when doing an upgrade from RPC 10 to 11 and adding CEPH after the upgrade.
This is still a manual step, necessary for this particular situation.

andymcc:
@BjoernT that var isn't in the user_extras_secrets.yml even for new deploys - its in user_variables.yml inside openstack-ansible, but there is a note in the docs (once https://github.com/rpcdocs/rpcdocs/pull/431 merges) that mentions ensuring that var exists. Does that work?

robb-romans:
It seems like there are two separate issues here. One is to more fully document the variables used by the RPC deploy, and is the subject of this PR. The other issue covered by https://github.com/rpcdocs/rpcdocs/pull/431 is what variables the deployer should set when doing an automated installation, a manual installation, and an upgrade.

mancdaz:
LGTM

Issue #569, opened 2015-11-10
neutron metadata check uses incorrect address


toxick:
neutron_metadata_local_check.py uses 127.0.0.1:80 for testing metadata proxy when it should be using 169.254.169.254:80:

[ distributed version ]
**# grep ^SERVICE_CHECK ./neutron_metadata_local_check.py**
SERVICE_CHECK = 'ip netns exec %s curl -fvs 127.0.0.1:80'
**# ./neutron_metadata_local_check.py 172.24.240.49**
status error neutron metadata agent proxies fail on host [clip]_neutron_agents_container-7576f37a net_ids: 993b54c7-10aa-43b1-80e7-67a579a02eea

[...fix fix fix...]
**# grep ^SERVICE_CHECK ./neutron_metadata_local_check.py**
SERVICE_CHECK = 'ip netns exec %s curl -fvs 169.254.169.254:80'
**# ./neutron_metadata_local_check.py 172.24.240.49**
status okay
metric neutron-metadata-agent-proxy_status uint32 1

git-harry:
I'm unable to reproduce this bug. On my test cluster the proxy is listening on both:

# tail -n4 /var/log/neutron/neutron-ns-metadata-proxy-9097ef9d-d0e9-4980-8926-d7e1a115cc80.log
2015-11-17 13:25:07.177 4613 INFO neutron.wsgi [-] (4613) accepted ('169.254.169.254', 41353)
2015-11-17 13:25:07.208 4613 INFO neutron.wsgi [-] 169.254.169.254 - - [17/Nov/2015 13:25:07] "GET / HTTP/1.1" 404 278 0.029188
2015-11-17 13:26:11.590 4613 INFO neutron.wsgi [-] (4613) accepted ('127.0.0.1', 59397)
2015-11-17 13:26:11.628 4613 INFO neutron.wsgi [-] 127.0.0.1 - - [17/Nov/2015 13:26:11] "GET / HTTP/1.1" 404 278 0.036527

That said I believe 169.254.169.254 is the address used by the nova instances and so it does strike me as more appropriate to be using that IP address. @stevelle, I see you wrote the plugin, can you give any additional details.

stevelle:
I tested both 169.254.169.254 and 127.0.0.1 and they worked in my limited testing. That said, 169.254.169.254 is an appropriate target address and I'm +1 on a change to the host here.

BjoernT:
[...]

PR #640, opened 2015-11-30
Disable neutron metadata proxy check/alarm

stevelle:
This maas check does not work in our common configuration and the
strategy being used is ill-fit for the varied networks it needs to
test.

Functional Backport of: #639
Addresses: #569


rackertom:
Also :+1:

d34dh0r53:
:+1:

PR #644, opened 2015-12-02

stevelle:

We ensure we check that the RabbitMQ version is at least 3.5.x before
checking the lengths of queues on all nodes since we know that this
check is not valid in version 3.3.5 and further noodling into the
version where that feature becomes available can be deferred based on
need.

Addresses: #610
(cherry picked from commit 78bdd1d39cd845359ac62b5a837f57439cf62cc3)

### DIFF
--- a/maas/plugins/rabbitmq_status.py
+++ b/maas/plugins/rabbitmq_status.py
@@ -59,7 +59,16 @@ CONNECTIONS_METRICS = {"max_channels_per_conn": "channels"}
 
 def hostname():
     """Return the name of the current host/node."""
-    return subprocess.check_output(['hostname']).strip()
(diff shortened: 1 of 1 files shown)


stevelle:
I don't really think this needs a backport, see also the trello board.

sigmavirus24:
If we don't back port this, backporting https://github.com/hughsaunders/rpc-openstack/commit/005648d6fb5005c7d788e08aa071440a0a489051 will be nightmarish (but doable).

stevelle:
ok, that's a find reason to backport this one.

sigmavirus24:
I'm not sure we need to backport all of those anymore. /cc @mancdaz

PR #647, opened 2015-12-02
Reify linting for multi-portion gating

sigmavirus24:
Fixes-bug: #595 

--

This splits the gating out into two linting scripts:

- `scripts/linting-pep8.sh`
- `scripts/linting-ansible.sh`

And it changes `scripts/linting.sh` to use `tox` to isolate those dependencies and run them in virtual environments for us.

This also updates our Travis-CI configuration to run tox for each of the two checks. One check may fail while another may pass and the output from both will be easily readable from both. Travis will report "one check" to GitHub, but it has two "sub-checks" that it runs - one for hacking/flake8, one for ansible-lint.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,18 @@
 language: python
 
-script: ./scripts/linting.sh
+before_script:
+- pip install tox
+
+script: tox
 
 notifications:
     email: false
+
+matrix:
+  include:
+    - python: 2.7
+      env: TOXENV=flake8
+    - python: 2.7
+      env: TOXENV=ansible-lint
+
+sudo: false
--- a/scripts/linting-ansible.sh
+++ b/scripts/linting-ansible.sh
@@ -1,0 +1,48 @@
+#!/usr/bin/env bash
+# Copyright 2015, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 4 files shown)


sigmavirus24:
I'm not convinced this is working at the moment but I can't figure out why tox is being difficult.

scripts/linting.sh
```diff
@@ -16,49 +16,4 @@
...
-  echo "Running Lint Check"
-  ansible-lint --version || failed=1
-  ansible-lint *.yml || failed=1
-popd
-
-if [[ $failed -eq 1 ]]; then
-  echo "Failed linting"
-  exit -1
-fi
+python -m tox
```
nrb: Could we perhaps remove this file entirely, and created 2 separate jobs? I think that highlights the failures more nicely.

d34dh0r53: :+1: I like that idea

sigmavirus24: I left this here as a shim for people. I don't understand the "and created 2 separate jobs"

nrb: My question was answered in IRC, but for those who missed it - Travis is running 2 separate build jobs, but GitHub only sees a single service per 'check'.

So, what I wanted was a clear :+1: or :-1: per lint job, ansible or python. This change gives us that.

nrb:
:+1: for this, I think this makes the goal of better identifying failures even more obvious than simply failing fast, like I had originally tried.

I think @rackertom and @mancdaz should take a look, too.

rackertom:
Taking a glance at it, I like it. :+1:

d34dh0r53:
:+1:

stevelle:
is this not merged yet? :shipit:

sigmavirus24:
Shall I do the backporterating?

nrb:
@sigmavirus24 I approve of backporting this.

PR #652, opened 2015-12-03

sigmavirus24:

We ensure we check that the RabbitMQ version is at least 3.5.x before
checking the lengths of queues on all nodes since we know that this
check is not valid in version 3.3.5 and further noodling into the
version where that feature becomes available can be deferred based on
need.

Addresses: #610
(cherry picked from commit 78bdd1d39cd845359ac62b5a837f57439cf62cc3)


sigmavirus24:
This blocks a backport of https://github.com/hughsaunders/rpc-openstack/commit/005648d6fb5005c7d788e08aa071440a0a489051

sigmavirus24:
Dupe of #644

Issue #655, opened 2015-12-08
Add Swift replication time alarms


andymcc:
We should add replication time arlarms for swift - the metrics already exist

- [x] master
- [x] kilo

d34dh0r53:
Please also create a `kilo` backport for this

andymcc:
Will do once the initial patch merges.

d34dh0r53:
This has merged into `master` please create the `kilo` backport.

PR #656, opened 2015-12-08
Add MaaS Alarms for swift replication time

andymcc:
Add a rate increase alarm for object/container/account replication
Add an average time threshold alarm for object/container/account
replication.

Closes-Issue: #655


prometheanfire:
where were these check values sourced from, are they sane defaults?

andymcc:
@prometheanfire not particularly, we don't have well calculated defaults since we haven't got any production data to go by - but as a start this will work and we can adjust the defaults at a later point if they turn out to be flawed.

tomc603:
Given we have no historical data to base these on, and that we need to start somewhere, I'm OK with this. :+1:

d34dh0r53:
:+1:

PR #660, opened 2015-12-09
Add /etc/ansible/roles to roles_path

### LINKED ISSUE #659: role_path should include /etc/ansible/roles
OSA is starting to import external roles and put them in /etc/ansible/roles, so that should be in the search path.

hughsaunders:
Openstack Ansible now imports some externala roles to /etc/ansible/roles,
These must be in the roles path in rpc-openstack to be available to
rpc-openstack plays.

closes #659


odyssey4me:
:+1: from me - this is a good pre-emptive step

prometheanfire:
:+1: as well

prometheanfire:
dunno why I didn't merge this earlier

PR #662, opened 2015-12-09
Upgrade the upgrade script

cloudnull:
This change removes the old upgrade script in favor of the new upgrade
script. This should prove to be more stable and supportable for all
future upgrades.

### DIFF
--- a/scripts/upgrade.sh
+++ b/scripts/upgrade.sh
@@ -36,7 +36,7 @@ openstack-ansible -i "localhost," patcher.yml
 
 # Do the upgrade for openstack-ansible components
 cd ${OA_DIR}
-echo 'YES' | ${OA_DIR}/scripts/run-upgrade-old.sh
+echo 'YES' | ${OA_DIR}/scripts/run-upgrade.sh
 
 # Prevent the deployment script from re-running the OA playbooks
 export DEPLOY_OA="no"


d34dh0r53:
:+1:

sigmavirus24:
Is this for 11.1?

I don't think this has backport potential. I don't think kilo has a SHA new enough to have the new upgrade script.

d34dh0r53:
@sigmavirus24 Yes, this is for 11.1 and beyond

cloudnull:
@sigmavirus24 this can/could go in at anytime the script has been there for a while. 

The script based on the submodule sha from RPCO kilo:
  https://github.com/openstack/openstack-ansible/blob/132953b/scripts/run-upgrade.sh

sigmavirus24:
Has QE tested the new upgrade process at all? Or support?

cloudnull:
Support yes, QE no. The patch that made the new process was here: https://review.openstack.org/#/c/224137

sigmavirus24:
@BjoernT @seancarlisle if either of (or preferably both) you are in favor of this, I'll merge it happily.

BjoernT:
:+1:  in favor of the upgrade playbooks

sigmavirus24:
I saw that Sean +1'd the original review. :+1: from me as well.

Issue #38, opened 2015-03-16
Remove virtual disks from maas disk utilisation check


BjoernT:
With the recent implementation to iterate over all SCSI devices to install maas utilisation alarms, we introduced a new error. This maas  disk utilisation does now include cinder attached volumes if you run those playbooks after a system has been handed over. We basically need to amend this logic to exclude all disks with the model "VIRTUAL-DISK"

BjoernT:
The playbook maas_local.yml is affected from this issue

andymcc:
backport is here https://review.openstack.org/#/c/255833/

d34dh0r53:
Can we retarget the backport for the `kilo` branch?

nrb:
@d34dh0r53 The backport was already done via https://github.com/rcbops/rpc-openstack/pull/718

prometheanfire:
ok, closing

PR #663, opened 2015-12-10
Don't monitor Cinder Volume LVs for disk util.

andymcc:
The disk utilisation monitoring will monitor cinder volumes if they are
attached to an instance and you run the MaaS play again.

We need to avoid this by ensuring that the model of moniitored disks is
not 'VIRTUAL-DISK'.

Closes-Issues: #38


mancdaz:
:+1:

sigmavirus24:
:+1:

Issue #155, opened 2015-06-04
MaaS plays expect pre-generated entities


stevelle:
> We made the conscious decision when putting these plays together not to create missing entities as 
> these should be created at server provisioning and missing entities could mean other provisioning-
> related errors. We felt it was best to bail and have the operator investigate. Also, if you misconfigure
> some variables in ansible you could end up with a slew of newly created entities which are unneeded.

This expectation needs to be captured in the role's readme file and maybe an additional note in the tasks/mainl.yml about why at the point where they are read or another solution to allow them to be more safely created should be discussed to remove this expectation.

I'm not seeing anything related to this topic in the RPC install guide either but that could be a pebsak.

See also: #147 and #152

Alternative to documenting this is to improve the design to allow entities to be auto-generated more safely.

nrb:
This is largely a documentation task, but it's worth clarifying history for future reference.

mancdaz:
@nrb @stevelle is there anything to be done here?

nrb:
@mancdaz Documentation in the README/docs really, nothing that I know of in code.

stevelle:
what @nrb said

andymcc:
I'll create a docs issue and we can work to add this in.

andymcc:
Happy to just go with the PR - I think it explains failures well enough to not have to create a separate docs issue.

mancdaz:
@andymcc agree - with a second +1 on that review we can merge and close this off.

andymcc:
Closing this issue out.

PR #664, opened 2015-12-10
Update fail messages for MaaS entity discovery

git-harry:
Entities should alway exist for the customers hardware and are created
when it is assigned to the account. This commit updates the error
messages generated to be clearer about the action support should take.

The use of entites is not RPC specific and so no additional
documentation should be required.

Closes-issue: https://github.com/rcbops/rpc-openstack/issues/155

### DIFF
--- a/rpcd/playbooks/roles/rpc_maas/tasks/raxmon_agent_install.yml
+++ b/rpcd/playbooks/roles/rpc_maas/tasks/raxmon_agent_install.yml
@@ -42,9 +42,14 @@
   shell: "raxmon-entities-list | grep ' label={{ entity_name }} provider=Rackspace Monitoring ...>$' | wc -l"
   register: entity_count
 
+- name: Zero entities with label {{ entity_name }} exist
+  fail:
+    msg: "Zero entities with the label {{ entity_name }} exist. Entities should be created as part of the hardware provisioning process, if missing, please consult the internal documentation for associating one with the device."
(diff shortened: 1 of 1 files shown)


mancdaz:
:+1:

rpcd/playbooks/roles/rpc_maas/tasks/raxmon_agent_install.yml
```diff
@@ -42,9 +42,14 @@
...
+- name: Zero entities with label {{ entity_name }} exist
+  fail:
+    msg: "Zero entities with the label {{ entity_name }} exist. Entities should be created as part of the hardware provisioning process, if missing, please consult the internal documentation for associating one with the device."
+  when: entity_count.stdout|int == 0
+
 - name: At most one {{ entity_name }} entity exists
   fail:
-    msg: "Entity count of {{ entity_count.stdout }} != 1 for entity with the label {{ entity_name }}"
+    msg: "Entity count of {{ entity_count.stdout }} != 1 for entity with the label {{ entity_name }}. Reduce the entity count to one by deleting or re-labelling the duplicate entities."
   when: entity_count.stdout|int != 1
```
andymcc: maybe change this to be > 1 ? Since we are already checking the case of if it = 0.
Although I guess it can't fail twice and I'm not sure if it can ever = -1.

andymcc:
Minor maybe not worth changing. Otherwise happy to merge.

Issue #668, opened 2015-12-12
KeyError when using v2 Auth in MaaS checks


stevelle:
`status error line 360, in get_auth_ref\n    if
is_token_expired(auth_ref):\n  File
"/usr/lib/rackspace-monitoring-agent/plugins/maas_common.py", line 345,
in is_token_expired\n    expires =
datetime.datetime.strptime(token['expires_at'], fmt)\nKeyError:
'expires_at'
`

d34dh0r53:
Needs `kilo` backport

d34dh0r53:
Prematurely closed this

PR #669, opened 2015-12-12
Fix expires_at vs expires mismatch in v2 auth token

stevelle:
Just checking for either label on the key works as a solution
for both versions. Just tested by pulling down a patched
maas_commons.py and in the interpreter running get_auth_ref()
against devstack (Auth v2) and an OSA master instance (Auth v3).

Addresses: #668


stevelle:
fix is insufficient. 
:-1: 
```status error 361, in get_auth_ref
    if is_token_expired(auth_ref): File "/usr/lib/rackspace-monitoring-agent/plugins/maas_common.py", line 346, in is_token_expired
    expires = datetime.datetime.strptime(expire_expr, fmt)
TypeError: must be string, not None```

mancdaz:
superseded by https://github.com/rcbops/rpc-openstack/pull/683

Issue #599, opened 2015-11-18
ceph playbooks fail quite often during initial install


BjoernT:
The ceph playbooks fail quite often during initial install with either downloading from upstream and comparing MD5 hashes or this error:

TASK: [ceph-mon | copy keys to the ansible server] ****************************
changed: [233d28af-infra1_ceph_mon_container-295016f6] => (item=/etc/ceph/ceph.client.admin.keyring)
changed: [233d28af-infra1_ceph_mon_container-295016f6] => (item=/etc/ceph/ceph.client.cinder-backup.keyring)
changed: [233d28af-infra1_ceph_mon_container-295016f6] => (item=/etc/ceph/ceph.client.cinder.keyring)
changed: [233d28af-infra1_ceph_mon_container-295016f6] => (item=/etc/ceph/ceph.client.glance.keyring)
fatal: [233d28af-infra1_ceph_mon_container-295016f6] => Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/ansible/runner/__init__.py", line 586, in _executor
    exec_rc = self._executor_internal(host, new_stdin)
  File "/usr/local/lib/python2.7/dist-packages/ansible/runner/__init__.py", line 815, in _executor_internal
    complex_args=complex_args
  File "/usr/local/lib/python2.7/dist-packages/ansible/runner/__init__.py", line 1036, in _executor_internal_inner
    result = handler.run(conn, tmp, module_name, module_args, inject, complex_args)
  File "/usr/local/lib/python2.7/dist-packages/ansible/runner/action_plugins/fetch.py", line 141, in run
    os.makedirs(os.path.dirname(dest))
  File "/usr/lib/python2.7/os.py", line 157, in makedirs
    mkdir(name, mode)
OSError: [Errno 17] File exists: '/etc/openstack_deploy/ceph_fetch/98cc124b-d0cc-4e2e-8a9e-55367e12a0ad/var/lib/ceph/bootstrap-osd'

changed: [233d28af-infra1_ceph_mon_container-8d55b576] => (item=/etc/ceph/ceph.client.admin.keyring)
changed: [233d28af-infra1_ceph_mon_container-8d55b576] => (item=/etc/ceph/ceph.client.cinder-backup.keyring)
changed: [233d28af-infra1_ceph_mon_container-8d55b576] => (item=/etc/ceph/ceph.client.cinder.keyring)
[...]

PR #672, opened 2015-12-14
Bump ceph-common and ceph-mon SHAs

mattt416:
This commit bumps the ceph-common and ceph-mon role SHAs to their
latest commits. This fixes an issue where a fetch of a file from a
remote host to local host was running across all MON nodes
simultaneously, resulting in a potential race condition.

Note that the fix above is only in ceph-mon, however additional
changes to ceph-mon require that ceph-common gets bumped also.

Closes issue #599


hughsaunders:
I came across this issue when testing rpc-aio gate, example: http://jenkins.propter.net/job/RPC-AIO/26/console 

recheck - to see if this patch gets past that issue - should do, looks sensible.

hughsaunders:
This PR gets the AIO passed the point it was failing at previously so :+1: from me.

PR #673, opened 2015-12-14
Discover rpc-openstack version

git-harry:
Currently Ansible determines the version of rpc-openstack in use by
reading the variable rpc_release. This variable needs to be updated
whenever a new version is tagged. The have been a number of occasions
where this has been missed and so rpc_release does not match the version
of code being used.

Tags are being used to identify different releases within the project.
rpc_release is essentially duplicating that information. This commit
adds a new module called git_repo_info that is used to discover the
current tag and so remove the requirement for manual updates. If
required the dynamic value can be overridden.

Closes-issue: https://github.com/rcbops/rpc-openstack/issues/533

### DIFF
--- a/rpcd/playbooks/group_vars/all.yml
+++ b/rpcd/playbooks/group_vars/all.yml
@@ -15,8 +15,8 @@
 # This is the location where the rpc repository was checked out
 rpc_repo_path: /opt/rpc-openstack/openstack-ansible
 
-# rpc-openstack version
-rpc_release: r11.1.0rc5
+# rpc-openstack version, this should be discovered by git_repo_info
+rpc_release: "{{ rpc_openstack_repo is defined | ternary((rpc_openstack_repo | default({}))['version'], 'UNKNOWN_RELEASE') }}"
 
 # Definitions for the repo server, these should match your openstack-ansible
 # deployment
--- a/rpcd/playbooks/library/git_repo_info
+++ b/rpcd/playbooks/library/git_repo_info
@@ -1,0 +1,120 @@
+#!/usr/bin/env python
+# Copyright 2015, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+DOCUMENTATION = """
+---
+module: git_repo_info
+short_description:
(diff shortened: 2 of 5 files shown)


andymcc:
This works for me code wise all good.
Would it be wise to add a comment around how the module works? Its not immediately obvious that it will create an ansible fact/var that is named "{{repo-name}}_repo" which makes it a bit confusing as a module - i.e. im not too clear what the output of using that module is by just looking at it.

Other than that, LGTM and we can get this merged when somebody else agrees.

hughsaunders:
recheck - failed on glance/ceph conflict.

odyssey4me:
This looks good to me. I've not tested it, but it looks conceptually good and I'd like, in fact, to see this in OSA!

rpcd/playbooks/group_vars/all.yml
```diff
@@ -15,8 +15,8 @@
 # This is the location where the rpc repository was checked out
 rpc_repo_path: /opt/rpc-openstack/openstack-ansible
 
-# rpc-openstack version
-rpc_release: r11.1.0rc5
+# rpc-openstack version, this should be discovered by git_repo_info
+rpc_release: "{{ rpc_openstack_repo is defined | ternary((rpc_openstack_repo | default({}))['version'], 'UNKNOWN_RELEASE') }}"
```
nrb: I'm a little confused how the value here actually gets set. There's no reference to the rpc_openstack_repo value outside of this file, so it's not registered in the Ansible discover roles, nor in the Python module.

Am I missing a link here?

nrb: Ah, actually, I see it.

rpcd/playbooks/library/git_repo_info
```diff
@@ -0,0 +1,91 @@
...
+        self.module = module
+        self.params = self.module.params
+
+    def gather_facts(self):
+        """Get information about RPC release."""
+        repo = {}
+        repo['version'], repo['sha'] = self.get_repo_version()
+        repo['name'] = (self.params['name'] or
+                        os.path.split(self.params['path'])[-1])
+        name = '%s_repo' % (repo['name'].replace('-', '_'))
```
nrb: Could we explicitly set a variable value through Ansible's `register`? This seems a little bit like black magic.

d34dh0r53:
:+1:

nrb:
I'm -1 on the current implementation; I think the variable that is produced from the module should be explicitly registered in the Ansible code, rather than appearing in the namespace as a side effect.

nrb:
@git-harry Would you mind looking into @andymcc and I's comments?

git-harry:
@nrb @andymcc, I've updated the documentation in the module to make it clearer how the module works.

@nrb this is a facts module, I don't think the variable appearing in the namespace is a side effect, it is the purpose of the module.

nrb:
@git-harry Thanks for the update. I still believe it's not as obvious as it could be from the code, but the documentation helps, so I'll at least rescind my -1.

prometheanfire:
:+1: ok, merging

PR #674, opened 2015-12-15
Add MaaS compatibility with Keystone v2.0

git-harry:
Currently the OpenStack MaaS plugins only support v3. The OSP project
requires that the plugins support Keystone v2.0.

This commit updates maas_common.py and the plugins so that both Keystone
v2.0 and v3 are supported.

Closes-Issue: https://github.com/rcbops/rpc-openstack/issues/668

### DIFF
--- a/maas/plugins/cinder_api_local_check.py
+++ b/maas/plugins/cinder_api_local_check.py
@@ -20,7 +20,6 @@ import collections
 import ipaddr
 # Technically maas_common isn't third-party but our own thing but hacking
 # consideres it third-party
-from maas_common import get_auth_ref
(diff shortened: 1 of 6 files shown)


sigmavirus24:
I can't comment on those lines directly because GitHub doesn't allow it, but this does not address the problem in #668 of `expires_at` not being the correct attribute on the token. This also does not address the disparities between Keystone v2.0's `serviceCatalog` and v3's `catalog` structures.

See #683 for a far smaller fix that addresses those problems directly while avoiding adding more global state and refactoring things in a fix that *might* need to be backported.

Issue #676, opened 2015-12-15
Disk utilisation override var overlaps with ceph var.


andymcc:
This causes the disk utilisation checks to fail when running ceph, as the wrong disks are monitored and the path is incorrect.

d34dh0r53:
Merged into `master`, please create a `kilo` backport for `11.1.1`

PR #678, opened 2015-12-15
Adjust disk utilisation devices variable

andymcc:
The current variable "devices" overlaps with ceph, in that we need to
specify a list of "devices" for ceph to use, and this var is then taken
to be used by the disk utilisation monitoring.

This causes incorrect (and failing) disk utilisation MaaS
checks/alarms to be setup.

Fixes-Issue: #676


git-harry:
+1

d34dh0r53:
:+1: needs a refactor and can we create a `kilo` backport please?

andymcc:
Will add a kilo backport when this merges.

d34dh0r53:
Merged into `master`, please create a `kilo` backport for `11.1.1`

Issue #675, opened 2015-12-15
Ceph OSD stats checks fail to get metrics


andymcc:
The ceph osd check fails to get metrics because the string passed in is incorrect.

d34dh0r53:
Needs a `kilo` backport

andymcc:
Ahh sorry missed you'd already done it @byronmccollum - I'll close off my branch and I've +2'd your kilo backport.

byronmccollum:
Cheers

mancdaz:
The backport got pulled in by master-kilo sync PR https://github.com/rcbops/rpc-openstack/pull/718. Closing backport PR and this issue

PR #679, opened 2015-12-15
Fix ceph_osd stats template file

andymcc:
The ceph_osd checks/alarms were not returning metrics because of
additional quoting around the OSD number list.

Additionally, the metric requested should be "osd.x_up" rather than just
"osd.x".

This patch fixes both those issues.

Fixes-Issue: #675


git-harry:
+1

d34dh0r53:
:+1: LGTM

Issue #677, opened 2015-12-15
Network util checks fail when non-default network devices are in use


andymcc:
We may need to just document this, but the network utilisation checks fail and a var may need to be set to correct the device names for network interfaces.

andymcc:
Looks like we don't add/remove the correct maas checks either, and we also aren't running the tasks against physical hosts, but rather against all containers (and then delegating to physical host).

andymcc:
@asettle @robb-romans this is probably a docs issue too.

The issue is as follows:
There is no sensible way to discover the physical interfaces so we default the network_checks_list variable to consist of the interfaces eth0, eth1, eth2 and eth3: (https://github.com/rcbops/rpc-openstack/blob/master/rpcd/playbooks/roles/rpc_maas/defaults/main.yml#L318-L322) if the physical host itself does not have those specific interfaces (they are either named differently or simply don't exist, OR there are additional interfaces to monitor), we need to set that variable in /etc/openstack_deploy/user_extras_variables.yml

For example, if we had the interfaces em0 em1 em2 em3, we would set the following var in /etc/openstack_deploy/user_extras_variables.yml:
network_checks_list:
  - { name: "em0", group: "hosts", max_speed: "{{ net_max_speed }}", rx_pct_warn: "{{ net_rx_pct_warn }}", rx_pct_crit: "{{ net_rx_pct_crit }}", tx_pct_warn: "{{ net_tx_pct_warn }}", tx_pct_crit: "{{ net_tx_pct_crit }}"}
  - { name: "em1", group: "hosts", max_speed: "{{ net_max_speed }}", rx_pct_warn: "{{ net_rx_pct_warn }}", rx_pct_crit: "{{ net_rx_pct_crit }}", tx_pct_warn: "{{ net_tx_pct_warn }}", tx_pct_crit: "{{ net_tx_pct_crit }}"}
  - { name: "em2", group: "hosts", max_speed: "{{ net_max_speed }}", rx_pct_warn: "{{ net_rx_pct_warn }}", rx_pct_crit: "{{ net_rx_pct_crit }}", tx_pct_warn: "{{ net_tx_pct_warn }}", tx_pct_crit: "{{ net_tx_pct_crit }}"}
[...]

PR #680, opened 2015-12-15
Fix issues with Network Utilisation MaaS Checks

andymcc:
This PR fixes the following issues:

* The network tasks were run against ALL containers, this should only be
run against physical hosts.
* Adds a test to see if an interface exists on a host, and errors with a
message to indicate what to do if the interface does not exist.
* Fixes the logic around adding/removing MaaS checks based on check
name.
* Fixes agent restart task to only run against physical hosts.

TODO: Fix maas check exclude list logic, currently it will perform this
per check type, and that doesn't work particularly well. Will be fixed
in a separate PR/issue.

Fixes-Issue: #677

### DIFF
--- a/rpcd/playbooks/roles/rpc_maas/tasks/main.yml
+++ b/rpcd/playbooks/roles/rpc_maas/tasks/main.yml
@@ -34,6 +34,8 @@
     inventory_hostname in groups['hosts']
 
 - include: network.yml
+  when: >
+    inventory_hostname in groups['hosts']
 
 - include: kernel.yml
 
@@ -65,3 +67,5 @@
     groups['ceph_all'] is defined
 
 - include: restart_raxmon.yml
+  when: >
+    inventory_hostname in groups['hosts']
--- a/rpcd/playbooks/roles/rpc_maas/tasks/network.yml
+++ b/rpcd/playbooks/roles/rpc_maas/tasks/network.yml
@@ -13,6 +13,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+- name: Test that network interfaces exist
+  fail:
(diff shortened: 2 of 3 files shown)


rpcd/playbooks/roles/rpc_maas/tasks/network.yml
```diff
@@ -31,8 +39,7 @@
     - discover_nic_speed.results
   when:
     - inventory_hostname in groups["{{ item.0.group }}"]
-    - item.name not in maas_excluded_checks
-  delegate_to: "{{ physical_host }}"
```
git-harry: What is the issue caused by using delegate_to?

rpcd/playbooks/roles/rpc_maas/tasks/main.yml
```diff
@@ -65,3 +67,5 @@
     groups['ceph_all'] is defined
 
 - include: restart_raxmon.yml
```
git-harry: The file contains a task using delegate_to

andymcc:
The problem with the restart job is that for EVERY container (and host) we are running a task for "restart agent service" delegated to the physical host.
Since we don't run the agents on the containers, there is no need to restart the service for each container.

Similarly for networking - we don't monitor network interfaces on the containers themselves, so there is no point gathering information about the container's networks, or delegating anything to the physical host.

git-harry:
+1

hughsaunders:
recheck - failed on pub cloud issue.

tomc603:
recheck - build timed out

nrb:
Looks like this branch needs to be rebased.

tomc603:
I'll rebase it tonight assuming it wasn't rolled into the Bit Branch Merge (tm).

tomc603:
This PR has been replaced by #736

Issue #681, opened 2015-12-15
maas_excluded_checks list doesn't work as intended


andymcc:
We have some logic issues with how we implemented "maas_excluded_checks" - this should ideally be a "include" task that runs at the end and removes maas_excluded_checks files, at the moment its run at the end of some of the "ensure_x_checks.yml" plays and is mostly inconsistent.

- [x] master
- [x] kilo

d34dh0r53:
This needs a `kilo` backport, it was missed in #718

PR #682, opened 2015-12-15
Put maas_excluded_checks removal into its own task

andymcc:
The logic around removing maas checks based on the maas_excluded_checks
list is flawed in that it runs after every "check/alarm" setup task, and
removes checks based on the checks it's setting up.

We need to remove checks in the list if they exist at all, regardless of
whether they are being set up or not.

This PR moves the removal task into its own included task file and will
remove any checks that exist with the listed checkname (for remote
checks) or checkname-inventory_hostname (for local checks).

Fixes-Issue: #681

### DIFF
--- a/rpcd/playbooks/roles/rpc_maas/tasks/ensure_remote_backup_checks.yml
+++ b/rpcd/playbooks/roles/rpc_maas/tasks/ensure_remote_backup_checks.yml
@@ -28,13 +28,3 @@
     - inventory_hostname != groups["{{ item.group }}"][0]
     - item.name not in maas_excluded_checks
   delegate_to: "{{ physical_host }}"
-
-- name: Remove checks that are excluded
-  file:
-    path: "/etc/rackspace-monitoring-agent.conf.d/{{ item.name }}.yaml"
-    state: absent
-  with_items:
-    - "{{ checks }}"
-  when:
-    - item.name in maas_excluded_checks
-  delegate_to: "{{ physical_host }}"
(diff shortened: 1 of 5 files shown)


git-harry:
+1

mancdaz:
@andymcc this won't work for the network checks as they have a slightly different conf file naming convention - that's why I'd added that separate stanza to the networks file.

andymcc:
@mancdaz the stanza in the network checks didnt work at all... I fixed it up, and I tested this specifically with network checks - can you show an example of where it didnt work?

andymcc:
Here's my maas_excluded_checks var that I tested to remove checked for eth2/eth3 on my install, which worked as intended with the above patch.

maas_excluded_checks:
  - network_throughput-eth2
  - network_throughput-eth3

d34dh0r53:
:+1: looks good, can we create a backport for `kilo` as well?

andymcc:
I can put a backport in once we get this merged - do we just need to re-run the gate?

d34dh0r53:
`master` has been merged, we need a `kilo` backport for `11.1.1`.

andymcc:
we'll need to backport @mancdaz PR first https://github.com/rcbops/rpc-openstack/pull/713 
But it looks like there is another plan to backport these? 
@mancdaz can you confirm how this is working/what needs to be done?

mancdaz:
If this was in master at the time https://github.com/rcbops/rpc-openstack/pull/718 merged, then it won't need a kilo backport because it would have been brought in to kilo as a result of that large merge. 

If it wasn't in master at that time, but has merged in since, then it will need a kilo backport

d34dh0r53:
@mancdaz this was merged after #718, so I think it needs a `kilo` backport.

mancdaz:
@d34dh0r53 backport was done here https://github.com/rcbops/rpc-openstack/pull/734

PR #683, opened 2015-12-15
Make maas_common compatible with Keystone v2 and v3

sigmavirus24:
This is an alternative to #674 that avoids the large-scale refactoring and attempts to reconcile the minor differences in auth refs from Keystone v2.0 and v3.

This also supersedes #669

Addresses #668


stevelle:
This looks good, seems to pass all my smoketests. 
:+1:

d34dh0r53:
:+1: LGTM

Issue #689, opened 2015-12-19
Instance --> metadata agents connection doesn't work for RPC AIOs


hughsaunders:
Instance --> neutron metadata agents don't work for an AIO deployed via the RPC deploy.sh script. 

I think this is because the RPC deploy script doesn't call run-playbooks, so doesn't use the section that drops the iptables rule necessary to make instance<-->metadata agent connections work: 

https://github.com/openstack/openstack-ansible/blob/master/scripts/run-playbooks.sh#L118-L129

PRs:
---
- [x] master: #690 
- [x] kilo: #694

nrb:
Correct, not calling `run-playbooks.sh` was a conscious decision when I first wrote it, but this is indeed a problem currently.

tomc603:
Per my comment in #694 the desired change was already merged into Kilo. It actually looks like the same commit was applied to both Master and Kilo but I can't quite figure out if it happened all at once or if the commit was applied to a branch that was then branched for other work and subsequently applied.

PR #690, opened 2015-12-19

hughsaunders:

Related: #689

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -114,6 +114,19 @@ if [[ "${DEPLOY_OA}" == "yes" ]]; then
   # setup the infrastructure
   run_ansible setup-infrastructure.yml
 
+  # This section is duplicated from OSA/run-playbooks as RPC doesn't currently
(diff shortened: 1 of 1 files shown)


hughsaunders:
recheck

hughsaunders:
AIO passed with tempest, please review & merge this, other AIOs will fail without it.

scripts/deploy.sh
```diff
@@ -114,6 +114,17 @@ if [[ "${DEPLOY_OA}" == "yes" ]]; then
   # setup the infrastructure
   run_ansible setup-infrastructure.yml
 
+  # When running in an AIO, we need to drop the following iptables rule in any neutron_agent containers
+  # to that ensure instances can communicate with the neutron metadata service.
```
andymcc: really minor comment - but "to ensure that" 
Since it hasnt merged yet may as well just fix that.

Otherwise LGTM

andymcc:
LGTM :+1:

mattt416:
LGTM!

PR #694, opened 2015-12-21

hughsaunders:

Related: #689

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -114,6 +114,19 @@ if [[ "${DEPLOY_OA}" == "yes" ]]; then
   # setup the infrastructure
   run_ansible setup-infrastructure.yml
 
+  # This section is duplicated from OSA/run-playbooks as RPC doesn't currently
+  # make use of run-playbooks. (TODO: hughsaunders)
+  # When running in an AIO, we need to drop the following iptables rule in any neutron_agent containers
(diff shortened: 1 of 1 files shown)


hughsaunders:
This is a backport of #690

nrb:
Have not run this, but backport and looks good, so :+1:

d34dh0r53:
@hughsaunders can you target this to the `kilo-r11.1.0` branch?

mancdaz:
:+1: on the code from me (it even passed an aio with tempest!)

hughsaunders:
@d34dh0r53 should I target both kilo and kilo-r11.1.0 or just kilo-r11.1.0 ?

d34dh0r53:
This should be just targeted at `kilo` now.  Thanks!

prometheanfire:
recheck

prometheanfire:
recheck

tomc603:
I'm going to rebase this and PR it again.

tomc603:
As I dug into this and wondered why I couldn't cherry pick and why I didn't see the expected change when doing a rebase from @hughsaunders fork, I realized this change was already merged into Kilo, probably as a result of branching from this branch, then having the resulting branch merged?

In either case, upstream seems to be carrying the desired change already. I'm closing this and marking it done in the parent Issue.

PR #697, opened 2015-12-23

hughsaunders:

Related: #689


tomc603:
:+1:

prometheanfire:
not sure the target branch should still exist (been out)

Issue #658, opened 2015-12-08
Logstash overrun with messages around heat


BjoernT:
RPC 11 all versions:

      "_source":{"tags":["heat","openstack","oslofmt","logstash-input-tcp"],"@version":1,"@timestamp":"2015-12-08T01:26:15.063Z","host":"xx","file":"/var/log/heat/heat-api.log-20151123.gz","message":"2015-11-22 06:35:47.269 39419 INFO heat.openstack.common.policy [req-c48c0b9e-dac2-4484-b357-cb3d649b51a5 - admin] Can not find policy directory: policy.d","type":"file","logdate":"2015-11-22 06:35:47.269","pid":"39419","loglevel":"INFO","module":"heat.openstack.common.policy","logmessage":"[req-c48c0b9e-dac2-4484-b357-cb3d649b51a5 - admin] Can not find policy directory: policy.d","received_at":"2015-12-07 20:26:15 -0500"}

Currently we seem to create 200G messages per day on a small cluster

BjoernT:
I will add more messages if I can dig though this haystack

d34dh0r53:
We can either fix this by creating the directory or adding a filter to ignore this message.  The former is the quickest fix but the latter is probably more correct.

tomc603:
I've been playing with this for a while now, and I've got some questions. As noted in the OSA version of this bug, this will impact anything that has a deprecated policy.d situation. The messages don't constantly spew out, only when the item in question (glance, heat, etc.) attempt to take an action that would trigger the policy nonsense.

I've had an AIO standing up for about a day and I've watched the logs roll through. If we're looking to reduce the amount of log data then we should switch the default logging from INFO to WARN or higher. Most of the INFO level logs are absolutely useless unless you're using them for auditing purposes. On an installation with an active glance or heat, I could see there being lots of these policy.d messages, but can I get a sense from @BjoernT how many he's seeing compared to other messages? They seem to be a small minority compared to the other noise my stack is generating.

[...]

PR #698, opened 2015-12-23
Add filter for policy.d missing messages

tomc603:
Add a filter in logstash 02-general.conf file to filter policy.d missing message from any source. Right now this is a problem at least for Heat and Glance, but there could be other services that will or do produce this error.

### DIFF
--- a/rpcd/playbooks/roles/logstash/templates/02-general.conf
+++ b/rpcd/playbooks/roles/logstash/templates/02-general.conf
@@ -1,5 +1,8 @@
 filter {
   if "oslofmt" in [tags] {
+    if "Can not find policy directory: policy.d" in [message] {
+      drop{}
+    }
     grok {
       match => { "message" => "^%{TIMESTAMP_ISO8601:logdate}%{SPACE}%{NUMBER:pid}?%{SPACE}?(?<loglevel>AUDIT|CRITICAL|DEBUG|INFO|TRACE|WARNING|ERROR) \[?\b%{NOTSPACE:module}\b\]?%{SPACE}?%{GREEDYDATA:logmessage}?" }
       add_field => { "received_at" => "%{@timestamp}" }


hughsaunders:
recheck

sigmavirus24:
I wonder why travis stopped working here. Let me investigate

sigmavirus24:
There we go

git-harry:
Why not just create the missing directory? Then there's nothing to ignore.

tomc603:
There was a lot of conversation about exactly this. @sigmavirus24 convinced
me that it was not the best way forward because in the future, more
services will be impacted, and it wasn't the intent of the policy framework
in the first place. This is from memory.

sigmavirus24:
@git-harry Creating the directory implies (to operators) that it may actually be used by oslo.policy, which is not the long-term case. In fact, I'm rather certain oslo.policy stopped loading policies from that directory in Liberty (if not then it will be Mitaka). Creating the directory implies it should be used and I'd rather not have our customers start wondering why the policy files they placed there stopped working if they ever upgrade.

git-harry:
Which versions of rpc-openstack/openstack-ansible does this affect? Why not create the directories where appropriate and remove them when no longer supported? That strikes me as a normal upgrade task the code should account for. I don't know what load ignoring the log line equates to but 200G of messages on a small cluster sounds like a lot of unnecessary processing by logstash.

sigmavirus24:
> Why not create the directories where appropriate and remove them when no longer supported? 

Will removing files created by customers without telling them will go over well?

tomc603:
@git-harry I do not believe these account for 200G of messages. See https://github.com/rcbops/rpc-openstack/issues/658#issuecomment-167704195 and https://github.com/rcbops/rpc-openstack/issues/658#issuecomment-167798823

Having said that, I made this check as simple and exact as possible. Is there a detriment you see with this patch? I agree it's not a perfect solution, but I also recognize that there isn't a perfect solution for this situation. @BjoernT and @sigmavirus24 have gone back and forth on this, as have I. And this simple patch after stewing on this problem for a week myself, seems to be the simplest solution. And removing this clause in the future poses no risk to customers and no unexpected human behavior. Adding a misleading directory which we then delete at a later date is just begging for misunderstanding and misuse.

BjoernT:
I was surprised to see outdated messages in a recent logstash index and that caught my interest :
The file /var/log/nova/nova-cert.log-20151109.gz was parsed to we seem to parse rotated archives into ES again, causing all the duplicates ?

"_source":{"tags":["nova","openstack","oslofmt","logstash-input-tcp"],"@version":1,"@timestamp":"2016-01-04T22:53:37.957Z","host":"<ip>","file":"/var/log/nova/nova-cert.log-20151109.gz","message":"2015-11-01 23:46:12.275 559 ERROR oslo_messaging._drivers.impl_rabbit [req-85ed2552-187f-4dcb-9031-35568d70d21f - - - - -] AMQP server on  <ip>:5672 is unreachable: [Errno 111] ECONNREFUSED. Trying again in 1 seconds.","type":"file","logdate":"2015-11-01 23:46:12.275","pid":"559","loglevel":"ERROR","module":"oslo_messaging._drivers.impl_rabbit","logmessage":"[req-85ed2552-187f-4dcb-9031-35568d70d21f - - - - -] AMQP server on <ip>:5672 is unreachable: [Errno 111] ECONNREFUSED. Trying again in 1 seconds.","received_at":"2016-01-04 17:53:37 -0500"}
    }, {
      "_index" : "logstash-2016.01.04",

nrb:
I'm +1 on this - having a directory just deleted on upgrade doesn't seem like a good idea to me.

sigmavirus24:
I'm also +1 on this.

PR #703, opened 2016-01-04
Only parse *.log files for cinder,nova,glance,heat and neutron

BjoernT:
The current beaver config does parse all files under /var/log/nova for example
including .gz files which potentially adds redundant log information
into Elasticsearch.
This fix will only seach for /var/log/{cinder,nova,glance,heat,neutron}/*.log
log files.

Related-Bug: #658

### DIFF
--- a/rpcd/playbooks/beaver.yml
+++ b/rpcd/playbooks/beaver.yml
@@ -22,15 +22,15 @@
   vars:
     beaver_log_monitors:
       - name: "cinder"
-        log_file: "/var/log/cinder/*"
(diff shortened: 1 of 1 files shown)


tomc603:
This bug is a duplicate of https://github.com/rcbops/rpc-openstack/pull/524, and implemented exactly the same. We should close one and merge the other and backport where necessary.

@d34dh0r53 opinions?

tomc603:
I'm :+1: on this

andymcc:
LGTM :+1:

PR #710, opened 2016-01-08
Do not create RabbitMQ monitoring user if RabbitMQ is not installed.

byronmccollum:
This is needed when deploying stand alone Swift, and the RabbitMQ
affinity is set to 0 to prevent its installation.

### DIFF
--- a/rpcd/playbooks/roles/rpc_maas/tasks/main.yml
+++ b/rpcd/playbooks/roles/rpc_maas/tasks/main.yml
@@ -23,7 +23,7 @@
 
 - include: rabbitmq_user.yml
   when: >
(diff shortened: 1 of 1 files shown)


byronmccollum:
recheck

byronmccollum:
Actually, this might only be needed for `kilo`...

byronmccollum:
So, this was fixed in `kilo r11.1` series, but not `kilo r11.0` series, which is now dead.

```
    _____
  /~/~   ~\
 | |       \
 \ \        \
  \ \        \
 --\ \       .\''
--==\ \     ,,i!!i,
    ''"'',,}{,,
```

PR #714, opened 2016-01-08
add some connection related monitors to galera_check plugin

mancdaz:
Up to now, we've been grabbing specific matching data from the output
of 'SHOW STATUS'. This commit adds the output from 'SHOW VARIABLES', so
that we can grab the configuration vars too. Additionally, we no longer
limit the captured data from 'SHOW STATUS'.

Having all of the output from 'SHOW VARIABLES' and 'SHOW STATUS' will
make it easy to add any further metrics/alarms to this plugin in the
future.

Additionally, 'SHOW STATUS' was changed to 'SHOW GLOBAL STATUS' to
ensure we are gathering global stats as opposed to 'this session' stats.

Some new metrics are also added:

current_connections
max_configured_connections
max_seen_connections

2 new alarms were added to the config template to alert on used vs max
mysql connections

Partial #550 

(cherry picked from commit 5b24196daa6112f711912bea928542f9ae0a23c9)

### DIFF
--- a/maas/plugins/README.md
+++ b/maas/plugins/README.md
@@ -307,6 +307,9 @@ connects to an individual member of a galera cluster and checks various statuses
     metric wsrep_cluster_status string primary
     metric wsrep_local_state_uuid string 67e41d08-165d-11e4-9d87-7e94ef43b302
     metric wsrep_local_state_comment string synced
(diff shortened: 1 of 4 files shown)


mancdaz:
closing as doing a larger merge commit to synchronise kilo and master branches

PR #715, opened 2016-01-08
update rabbitmq monitoring (queue length metrics, checks and alarms)

mancdaz:
This backport PR necessarily has 2 commits as they depend on each other.  See individual commits for details, but essentially they:

- update the way we search for a particular rabbit node in the output from the api
- add a check to monitor for growing queue lengths, as requested by support

Partial: #644 
Partial: #633


mancdaz:
closing as doing a larger merge commit to synchronise kilo and master branches

PR #717, opened 2016-01-09
Fix ceph_osd stats template file

byronmccollum:
The ceph_osd checks/alarms were not returning metrics because of
additional quoting around the OSD number list.

Additionally, the metric requested should be "osd.x_up" rather than just
"osd.x".

This patch fixes both those issues.

Fixes-Issue: #675
(cherry picked from commit 9fbeb5e7e3fdaf5bbced437bb1d635f6cdb6f25f)


andymcc:
Thanks @byronmccollum LGTM!

mancdaz:
already backported in https://github.com/rcbops/rpc-openstack/pull/718

PR #718, opened 2016-01-11
Synchronise master and kilo branches

mancdaz:
There are a number of commits that went into master that we did not backport to kilo.  Rather than have a mass of individual pull requests with a single cherry-pick in each one, this is a single pull request with all commits.


git-harry:
+1 the two branches appear to be the same except for the rpc_release variable because that is still r11.1.0rc5 in master.

mancdaz:
NB: rpc-aio is currently a WIP and so it's comments shouldn't be used as a basis for decision making.

prometheanfire:
ah, that's good to know, was holding back the other PRs too

prometheanfire:
also, given that 11.1 is the last kilo I think this should merge, not sure if all at once though.

prometheanfire:
:+1:

d34dh0r53:
:+1:

Issue #704, opened 2016-01-06
Verify MAAS fails in AIO


hughsaunders:
Since enabling maas, verify-maas has caused all jobs to fail, despite maas itself deploying ok (according to ansible). 

Verify maas must be reliable, or it/maas will have to be disabled.

byronmccollum:
See #621 and #687

Issue #695, opened 2015-12-22
deploy.sh disables MAAS when building an AIO


hughsaunders:
Deploy.sh has an env var for DEPLOY_MAAS, however this is ignored when DEPLOY_AIO is true. For the gate use case it would be useful to be able to deploy MAAS on an AIO. Jenkins scripts that run before deploy.sh will handle sedding creds into user_extras_variables. 

- [x] master #696
- [x] kilo #738

hughsaunders:
fixed in master and kilo, closing.

PR #719, opened 2016-01-11
Fix verify-maas for RPC AIO

hughsaunders:
Currently all RPC AIO gate jobs are failing (they have been since MAAS)
was enabled in #695), this patch resolves this outstanding issues with
maas.

1) Move verify-maas logic from playbook into rpc-maas tool using the new
"verify" command which is a single command healthcheck for a maas
deployment. Verify includes the following:
  * compare config checks & alarms against api registered items, report any
    items that are found in config files but not succesfully registered.
  * List checks that have no alarms (apart from the default ping check)
  * List alarms currently in error status. Also re-executes checks
    associated with erroring alarms in order to show error messages.
  * List excluded alarms that are present.

Rpc-maas-tool verify is used in the verify_maas playbook, but can also
be used as a standalone tool for checking the state of a deployment.

2) Some fixes to the network_throughput config for AIO compatibility
 * Only check eth0 by default. This create work for physical deployments
   as the NIC list will need to be overridden anyway.
   Fix max_speed for vitual nics. Currently when speed detection fails,
   max speed is set to 0, so if a check succeeds to get a
   transmit/recieve value, it instantly alarms as that is likely to be
   >0. This patch sets the in template default to 1000 if a value
   isn't set in user_extras_variables and speed detection fails.

Related: #704

### DIFF
--- a/rpcd/playbooks/verify-maas.yml
+++ b/rpcd/playbooks/verify-maas.yml
@@ -16,61 +16,24 @@
 - name: Verify MaaS setup
   hosts: hosts
   user: root
-  pre_tasks:
-    # Raxmon client doesn't make it very easy to just get the full check or
-    # alarm names due to limited choices in formatting of output
-    - name: Enumerate all checks found
-      script: "{{ maas_rpc_scripts_dir }}/rpc-maas-tool.py --tab True --prefix {{ inventory_hostname }} checks"
-      register: maas_after_checks
-      changed_when: False
-      failed_when: maas_after_checks.rc != 0
-      tags:
-        - maas-verify
-    - name: Enumerate expected checks
-      shell: ls -1 /etc/rackspace-monitoring-agent.conf.d/ | sed -e 's/\.yaml$//'
-      when:
-        - maas_after_checks is defined
-      register: expected_checks
-      changed_when: False
-      tags:
-        - maas-verify
-    - name: Enumerate all alarms found
-      script: "{{ maas_rpc_scripts_dir }}/rpc-maas-tool.py --tab True --prefix {{ inventory_hostname }} alarms"
-      when:
-        - expected_checks is defined
-      register: maas_after_alarms
-      changed_when: False
-      failed_when: maas_after_alarms.rc != 0
-      tags:
-        - maas-verify
   tasks:
-    - name: Verify presence of expected checks
-      fail: msg="Expected check ({{ item }}) not found"
-      with_items:
-        - "{{ expected_checks.stdout_lines }}"
-      when:
-        - maas_after_checks is defined
-        - maas_after_checks.stdout.find("{{ item }}") == -1
-      tags:
-        - maas-verify
-    - name: Verify absence of excluded checks
-      fail: msg="Excluded check ({{ item }}) found"
-      with_items:
-        - "{{ maas_excluded_checks }}"
-      when:
-        - maas_after_checks is defined
-        - maas_after_checks.stdout.find("{{ item }}") != -1
-      tags:
-        - maas-verify
-    - name: Verify presence of expected alarms
-      fail: msg="Expected alarm for check ({{ item }}) not found"
-      with_items:
-        - "{{ expected_checks.stdout_lines }}"
-      when:
-        - expected_checks is defined
-        - maas_after_alarms is defined
-        - maas_after_alarms.stdout.find("{{ item }}") == -1
-      tags:
-        - maas-verify
+    - name: "Verify Checks & Alarms are registered"
+      shell: >
+        {{ maas_rpc_scripts_dir }}/rpc-maas-tool.py verify-created
+        --entity {{ inventory_hostname }}
+        {% for ec in maas_excluded_checks %} --excludedcheck {{ec}}
+        {% endfor %}
+      register: verify_maas
+      failed_when: verify_maas.rc != 0
+      changed_when: False
+      until: verify_maas.rc == 0
+      retries: 10
+      delay: 60
+
+    - name: "Verify Check & Alarm Status"
+      shell: >
+        {{ maas_rpc_scripts_dir }}/rpc-maas-tool.py verify-status
+        --entity {{ inventory_hostname }}
+
   vars_files:
     - "roles/rpc_maas/defaults/main.yml"
--- a/scripts/rpc-maas-tool.py
+++ b/scripts/rpc-maas-tool.py
@@ -13,199 +13,607 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+
+from __future__ import print_function
+
+import argparse
+import ConfigParser
+import datetime
+import logging
+import os
+from Queue import Queue
+import re
+import sys
+import yaml
+
+import concurrent.futures
+
 from rackspace_monitoring.drivers import rackspace
 from rackspace_monitoring.providers import get_driver
 from rackspace_monitoring.types import Provider
 
-import argparse
-import ConfigParser
-import re
-import sys
+DEFAULT_CONFIG_FILE = '/root/.raxrc'
+logging.basicConfig(level=logging.DEBUG,
+                    datefmt="",
+                    format="%(message)s",
+                    stream=sys.stdout)
+LOGGER = logging.getLogger(__name__)
 
 
-def main(args):
-    config = ConfigParser.RawConfigParser()
-    config.read('/root/.raxrc')
+class RpcMaas(object):
+    """Class representing a connection to the MAAS Service"""
 
-    driver = get_driver(Provider.RACKSPACE)
-    conn = _get_conn(config, driver)
+    def __init__(self, entity_match='', entities=None,
+                 config_file=DEFAULT_CONFIG_FILE):
+        self.entity_label_whitelist = entities
+        self.entity_match = entity_match
+        self.config_file = config_file
+        self.driver = get_driver(Provider.RACKSPACE)
+        self._get_conn()
+        self._get_overview()
+        self._add_links()
+        self._filter_entities()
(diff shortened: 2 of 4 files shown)


rpcd/playbooks/roles/rpc_maas/templates/network_throughput.yaml.j2
```diff
@@ -1,4 +1,4 @@
-{% set max_speed = (item.0.max_speed | default(0, boolean=true) or item.1.stdout) | int * 10 ** 6 / 8 %}
+{% set max_speed = (item.0.max_speed | default(item.1.stdout | default(1000, boolean = true), boolean = true)) | int * 10 ** 6 / 8 %}
```
git-harry: Support should never need to set this value, if a value isn't automatically discovered something is probably broken and so it is better that they are notified than a default value gets popped in. It is simple enough to set the variable in the overrides file if required.

rpcd/playbooks/verify-maas.yml
```diff
@@ -16,61 +16,22 @@
...
+        --entityprefix {{ inventory_hostname }}
+        {% for ec in maas_excluded_checks %} --excludedcheck {{ec}}
+        {% endfor %}
+      register: verify_maas
+      failed_when: verify_maas.rc != 0
+      changed_when: False
+      until: verify_maas.rc == 0
+      retries: 10
+      delay: 30
+    - name: debug verify_maas output
```
git-harry: Has this been left here intentionally?

hughsaunders: Nope, removed, good spot.

rpcd/playbooks/roles/rpc_maas/templates/network_throughput.yaml.j2
hughsaunders: They don't have to set it, autodetect is still the default, all I've done is add sensible default for when autodetect fails. IMO 0 is not a sensible default as it will usually cause alarms.

git-harry: Yes, auto-detect is the default for a production deployment. What your value of 1000 does is hide potential failure, it's essentially simplistic error handling. The only reason auto-detect should fail is if the deployment is on virtualised hardware, for those situations the variable net_max_speed exists so that a value can be specified in the overrides file.

scripts/rpc-maas-tool.py
```diff
@@ -13,199 +13,575 @@
...
+
+    def __init__(self, entity_prefix='', config_file=DEFAULT_CONFIG_FILE):
+        self.entity_prefix = entity_prefix
+        self.config_file = config_file
+        self.driver = get_driver(Provider.RACKSPACE)
+        self._get_conn()
+        self._get_overview()
+        self._add_links()
+        self.entities = [e['entity'] for e in self.overview
+                         if entity_prefix in e['entity'].label]
```
git-harry: If you want to use a prefix then in should probably be replaced with the startswith string method. My preference would be to modify the tool to take a list of entities. Ansible already know the names and it is easy enough to supply them to the script so that the command is exact. The removes the potential for an erroneous match to the prefix.

scripts/rpc-maas-tool.py
```diff
@@ -13,199 +13,575 @@
...
+        """
+        return self.entities
+
+    def get_checks(self, check_prefix=''):
+        """List checks for entities matching a prefix"""
+
+        checks = []
+        for entity in self.entities:
+            checks.extend([c for c in entity.checks
+                           if check_prefix in c.label])
```
git-harry: check_prefix isn't being used by a prefix.

scripts/rpc-maas-tool.py
```diff
@@ -13,199 +13,575 @@
...
+        return checks
+
+    def get_alarms(self, entry, check):
+        """Get list of alarms
+
+        Params:
+            entry: overview dictionary for one entity.
+
+        This function adds a state field to each alarm object
+        using information form the 'latest_alarm_states' entry key.
```
git-harry: from

scripts/rpc-maas-tool.py
```diff
@@ -13,199 +13,575 @@
...
+            alarm_states = sorted(
+                [als for als in entry['latest_alarm_states']
+                    if als.alarm_id == alarm.id],
+                key=lambda x: x.timestamp
+            )
+            if alarm_states:
+                alarm.state = alarm_states[-1].state
+            else:
+                alarm.state = "UNKNOWN"
+            alarms.append(alarm)
```
git-harry: break

scripts/rpc-maas-tool.py
```diff
@@ -13,199 +13,575 @@
...
+        This function adds a state field to each alarm object
+        using information form the 'latest_alarm_states' entry key.
+        """
+        alarms = []
+        for alarm in entry['alarms']:
+            if not alarm.check_id == check.id:
+                continue
+            # add state to the alarm object from latest_alarm_states
+            alarm_states = sorted(
+                [als for als in entry['latest_alarm_states']
```
git-harry: sorted takes an iterable so this could be a generator

scripts/rpc-maas-tool.py
```diff
@@ -13,199 +13,600 @@
...
+    Parse them as yaml and store that.
+    """
+    def __init__(self, agentconfdpath):
+        self.agentconfdpath = agentconfdpath
+        self.checks = self._parse_maas_configs()
+
+    def _parse_maas_configs(self):
+        """Read all config files in agentconfdpath"""
+        self.checks = {}
+        for path in list(os.walk(self.agentconfdpath))[0][2]:
```
git-harry: 'list(os.walk(self.agentconfdpath))[0][2]' is hard to read, at least for me. Why not use os.listdir instead? The directory shouldn't contain any sub-directories but it you're worried you could always catch the exceptions.

scripts/rpc-maas-tool.py
```diff
@@ -13,199 +13,600 @@
...
+            return 1
+
+    def excluded_checks(self):
+        """List checks that are present but shouldn't be"""
+        present_but_excluded_checks = []
+        checks = self.rpcm.get_checks(self.args.entityprefix)
+        for check in checks:
+            for exclude in self.args.excludedcheck:
+                # ping is a default check and may not have an alarm
+                if exclude in check.label:
```
git-harry: startswith

scripts/rpc-maas-tool.py
```diff
@@ -13,199 +13,600 @@
...
+
+        Check that all checks found in config files are registered with The
+        maas api.
+        """
+
+        api_checks = self.rpcm.get_checks(self.args.checkprefix)
+        missing_checks = [configcheck for configcheck in
+                          self.rpcmac.checks.keys()
+                          if configcheck not in
+                          [apicheck.label for apicheck in api_checks]]
```
git-harry: sets are nice for this type of operation, e.g.

set(self.rpcmac.checks) - set(apicheck.label for apicheck in api_checks)

scripts/rpc-maas-tool.py
```diff
@@ -13,199 +13,600 @@
...
-            print("\t".join([entity.id, entity.label, o.id, o.label,
-                             keys["filename"]]))
-    else:
-        print('Entity %s (%s):' % (entity.id, entity.label))
-        for o in objects:
-            print(' - %s' % o.label)
-
+            # run checks that have at least one failed alarm
+            # in most situations this much quicker than executing all checks
+            checks_with_failed_alarms = set([a.check for a in failed_alarms])
```
git-harry: No need for the list when defining the set.

scripts/rpc-maas-tool.py
hughsaunders: done

scripts/rpc-maas-tool.py
hughsaunders: Thanks, os.listdir is much better, done.

scripts/rpc-maas-tool.py
hughsaunders: done

scripts/rpc-maas-tool.py
hughsaunders: done

scripts/rpc-maas-tool.py
```diff
@@ -13,199 +13,605 @@
...
+            LOGGER.info("The following alarms are present in config files but "
+                        "are not registered with the maas api: "
+                        "%(missing_alarms)s "
+                        % {'missing_alarms':
+                            [a['label'] for a in missing_alarms]})
+            return 1
+
+    def verify_created(self):
+        """Verify that all checks and alarms have been created"""
+        LOGGER.info("--- %(datestamp)s ---"
```
git-harry: If everything if as expected this line gets added to the logs in isolation, e.g.

--- 2016-01-13 15:03:16.200597 ---
--- 2016-01-13 15:03:20.225372 ---
Checks and Alarms with failures:
 - Entity:en6D9A7qgR:aio4
   ! Check:agent.network:chAKzXzVfb:network_throughput_eth2--aio4: Completed:False Status:No such interface: eth2
     ! Alarm:alPhX1WQBP:Network transmit rate on eth2:CRITICAL
     ! Alarm:alScgsNWKx:Network receive rate on eth2:CRITICAL

hughsaunders: Added a success line for verify-created

mancdaz:
:+1:  this works for me, though I question whether as part of this we should be checking that every alarm is 'OK' and failing if not. Seems like the wrong place to check that, kind of - we mostly want to check that the checks and alarms got created properly in maas, not whether they are alerting or not.

Issue #688, opened 2015-12-18
Elastic Search not configured to auto start on containers


hybridpollo:
On recent kilo deployments I have observed that the elastic search daemon is not automatically started when stopping/starting the containers. This is a different behavior observed from previous releases of OSAD. Is this behavior intentional on Kilo ? 

Here are some recent examples from two different Kilo releases. Notice that where the startup scripts is missing the symlink to the elastic search start up script.  There are also no upstart jobs for elastic search on those containers with the same issue.

````
root@xxxxxx-logging01:~# cat /etc/openstack-release

DISTRIB_ID="OASD"
DISTRIB_RELEASE="11.2.5"
DISTRIB_CODENAME="1AndOne=11"
DISTRIB_DESCRIPTION="OpenStack Cloud"

root@xxxxxx-logging01:~# lxc-attach -n xxxxxx-logging01_elasticsearch_container-abfea3b1
root@xxxxxx-logging01_elasticsearch_container-abfea3b1:~# ls -l /etc/rc2.d/
total 4
-rw-r--r-- 1 root root 677 Jun 14 2015 README
lrwxrwxrwx 1 root root 15 Jul 29 2014 S20rsync -> ../init.d/rsync
lrwxrwxrwx 1 root root 13 Dec 11 17:30 S23ntp -> ../init.d/ntp
lrwxrwxrwx 1 root root 29 Dec 11 17:30 S37iptables-persistent -> ../init.d/iptables-persistent
lrwxrwxrwx 1 root root 21 Dec 11 17:30 S99grub-common -> ../init.d/grub-common
lrwxrwxrwx 1 root root 18 Jul 29 2014 S99ondemand -> ../init.d/ondemand
lrwxrwxrwx 1 root root 18 Jul 29 2014 S99rc.local -> ../init.d/rc.local

root@xxxxxx-logging01:~# cat /etc/rpc-release

DISTRIB_ID="RPC"
DISTRIB_RELEASE="r11.0.4"
DISTRIB_CODENAME="wee-5ive"
DISTRIB_DESCRIPTION="Rackspace Private Cloud powered by OpenStack"

root@xxxxxx-logging01:~# lxc-attach -n xxxxxx-logging01_elasticsearch_container-c24d95f3
root@xxxxxx--logging01_elasticsearch_container-c24d95f3:~# ls -l /etc/rc2.d/
total 4
-rw-r--r-- 1 root root 677 Jun 14 2015 README
[...]

PR #723, opened 2016-01-11
Enable the Elasticsearch service during install, but do not start it

tomc603:
Enable the Elasticsearch service in the install playbook, but defer starting until the service handler is triggered.

Relates to: #688

(cherry-pick from af141bc)


d34dh0r53:
:+1:

andymcc:
:+1:

PR #727, opened 2016-01-13
Only parse *.log files for cinder,nova,glance,heat and neutron 

BjoernT:
The current beaver config does parse all files under /var/log/nova for example
including .gz files which potentially adds redundant log information
into Elasticsearch.
This fix will only seach for /var/log/{cinder,nova,glance,heat,neutron}/*.log
log files.

Related-Bug: #658


tomc603:
recheck

BjoernT:
recheck

d34dh0r53:
:+1: LGTM

PR #728, opened 2016-01-13
Adjust disk utilisation devices variable

git-harry:
The current variable "devices" overlaps with ceph, in that we need to
specify a list of "devices" for ceph to use, and this var is then taken
to be used by the disk utilisation monitoring.

This causes incorrect (and failing) disk utilisation MaaS
checks/alarms to be setup.

Fixes-Issue: #676
(cherry picked from commit d499449d84cf73cbaf7975aeaa921a50af1dbdfe)


prometheanfire:
lgtm :+1:

PR #730, opened 2016-01-13
Put maas_excluded_checks removal into its own task

git-harry:
The logic around removing maas checks based on the maas_excluded_checks
list is flawed in that it runs after every "check/alarm" setup task, and
removes checks based on the checks it's setting up.

We need to remove checks in the list if they exist at all, regardless of
whether they are being set up or not.

This PR moves the removal task into its own included task file and will
remove any checks that exist with the listed checkname (for remote
checks) or checkname-inventory_hostname (for local checks).

Fixes-Issue: #681
(cherry picked from commit 477012bc8f309edcb26a1aeb239c0eb118c546ea)


mancdaz:
this backport got duplicated and merged in #734

Issue #732, opened 2016-01-13
Ceph needs testing


hughsaunders:
Currently the RPC AIO gate doesn't deploy/test ceph. Ceph is going into production and requires testing.

hughsaunders:
Required PR: https://github.com/rcbops/rpc-openstack/pull/760

hughsaunders:
Some backports may also be required. Will get a master job going first.

prometheanfire:
this has been merged, closing

prometheanfire:
does this need to go into 12.0.x as well?  If so, please update the top comment with the correct format.

git-harry:
@prometheanfire it already is

PR #733, opened 2016-01-13
Add Swift toggle to deploy.sh

hughsaunders:
Curently setup-openstack is called which includes setup-swift.
In order to make swift optional its necessary to avoid setup-openstack
and call the next layer of playbooks directly.

This is introduced so that swift can be disabled for a ceph gate scenario

Related: #732


mancdaz:
recheck

hughsaunders:
recheck - failed on keyserver

stevelle:
Looks good, didn't test.

prometheanfire:
:+1: I dare

PR #736, opened 2016-01-14
Fix issues with Network Utilisation MaaS Checks

tomc603:
This PR fixes the following issues:

* The network tasks were run against ALL containers, this should only be
run against physical hosts.
* Adds a test to see if an interface exists on a host, and errors with a
message to indicate what to do if the interface does not exist.
* Fixes the logic around adding/removing MaaS checks based on check
name.
* Fixes agent restart task to only run against physical hosts.

TODO: Fix maas check exclude list logic, currently it will perform this
per check type, and that doesn't work particularly well. Will be fixed
in a separate PR/issue.

Replaces PR: #680
Fixes-Issue: #677

### DIFF
--- a/rpcd/playbooks/roles/rpc_maas/tasks/main.yml
+++ b/rpcd/playbooks/roles/rpc_maas/tasks/main.yml
@@ -34,6 +34,8 @@
     inventory_hostname in groups['hosts']
 
 - include: network.yml
+  when: >
+    inventory_hostname in groups['hosts']
 
 - include: kernel.yml
 
@@ -67,3 +69,5 @@
 - include: maas_exclude.yml
 
 - include: restart_raxmon.yml
+  when: >
+    inventory_hostname in groups['hosts']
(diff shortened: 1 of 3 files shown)


git-harry:
recheck

git-harry:
+1

mancdaz:
Tested by adding a known non-existing interface, and got the expected failure:

```
failed: [aiotest] => (item={'tx_pct_crit': u'60', 'rx_pct_warn': u'40', 'max_speed': u'', 'group': 'hosts', 'name': 'eth4', 'tx_pct_warn': u'40', 'rx_pct_crit': u'60'}) => {"failed": true, "item": {"group": "hosts", "max_speed": "", "name": "eth4", "rx_pct_crit": "60", "rx_pct_warn": "40", "tx_pct_crit": "60", "tx_pct_warn": "40"}}
msg: The specified network interfaces eth4 doesn't exist, consider setting the 'network_checks_list' variable to override these interfaces
```

:+1:

PR #738, opened 2016-01-14
Fix verify-maas for RPC AIO

hughsaunders:
Currently all RPC AIO gate jobs are failing (they have been since MAAS)
was enabled in #695), this patch resolves this outstanding issues with
maas.

1) Move verify-maas logic from playbook into rpc-maas tool using the new
"verify" commands which are a healthcheck for a maas deployment.

  verify-created:
      * compare config checks & alarms against api registered items,
        report any items that are found in config files but not succesfully
        registered.
      * List checks that have no alarms (apart from the default ping
        check)
      * List excluded alarms that are present.

    verify-Status:
      * List alarms currently in error status. Also re-executes checks
        associated with erroring alarms in order to show error messages.

Rpc-maas-tool verify-check and verify-status are used in the verify_maas
playbook, but can also be used as standalone tools for checking the
state of a deployment.

2) Some fixes to the network_throughput config for AIO compatibility:
 * Only check eth0 by default. This doesn't create work for physical
   deployments as the NIC list will need to be overridden anyway.
 * Set net_max_speed in deploy.sh for AIOs as the speed of virtual NICs
   cannot be easily determined.

Related: #704


hughsaunders:
Straight cherry-pick from master.

git-harry:
+1

mattt416:
LGTM

PR #740, opened 2016-01-14
Discover rpc-openstack version

prometheanfire:
Currently Ansible determines the version of rpc-openstack in use by
reading the variable rpc_release. This variable needs to be updated
whenever a new version is tagged. The have been a number of occasions
where this has been missed and so rpc_release does not match the version
of code being used.

Tags are being used to identify different releases within the project.
rpc_release is essentially duplicating that information. This commit
adds a new module called git_repo_info that is used to discover the
current tag and so remove the requirement for manual updates. If
required the dynamic value can be overridden.

Closes-issue: https://github.com/rcbops/rpc-openstack/issues/533
(cherry picked from commit 2c4e3c4821d591ead5e2826e7d27f156225d81ff)

### DIFF
--- a/rpcd/playbooks/group_vars/all.yml
+++ b/rpcd/playbooks/group_vars/all.yml
@@ -15,8 +15,8 @@
 # This is the location where the rpc repository was checked out
 rpc_repo_path: /opt/rpc-openstack/openstack-ansible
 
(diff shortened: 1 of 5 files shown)


d34dh0r53:
:+1: straightforward backport

Issue #626, opened 2015-11-23
run_queues is not a rabbitmq metric


git-harry:
run_queues is used by the rabbitmq plugin to check that the cluster is working [1]. This metric is not giving information about the rabbitmq message queues [2]. This section of the plugin should be revisited so that is testing as intended.

[1] https://github.com/rcbops/rpc-openstack/blob/master/maas/plugins/rabbitmq_status.py#L144
[2] http://erlang.org/pipermail/erlang-questions/2012-February/064141.html

- [x] Liberty #745
- [x] Kilo

d34dh0r53:
This is not in fact fixed

d34dh0r53:
Needs a backport

PR #745, opened 2016-01-15
Remove incorrect run_queues test

hughsaunders:
The run_queues metric is load metric from the erlang environment, it
isn't directly related to rabbitmq message queues, though it is related
to node load, so a busy broker may have a higher run_queues.

However as this is a measure of load, it is not a good value for
checking replication across rabbit cluster members. This test has been
removed.

This patch also adds a couple of checks to ensure that all nodes are
clustered succesfully.

Related: #626


prometheanfire:
:+1: this is much cleaner as well

hughsaunders:
recheck failed on apt

d34dh0r53:
:+1:

Issue #747, opened 2016-01-17
Ceph Cinder Volume Creation Fails


byronmccollum:
Creating a Ceph Cinder Volume fails with error: `Error: error calling conf_read_file: errno EACCES`

Ceph OSD Host -- `cinder-volume.log`

```
2016-01-17 01:39:39.655 25796 INFO cinder.volume.flows.manager.create_volume [req-0efef6a0-99bf-40f0-9991-088fd2acd953 e558bb61b8974f2f98f0cd7344bed2a8 c289b84b69434801a363fa8703f28a4e - - -] Volume 04d762cd-17c6-488a-b98e-7cd9417069ca: being created as raw with specification: {'status': u'creating', 'volume_size': 1, 'volume_name': u'volume-04d762cd-17c6-488a-b98e-7cd9417069ca'}
2016-01-17 01:39:39.688 25796 WARNING cinder.volume.manager [req-0efef6a0-99bf-40f0-9991-088fd2acd953 e558bb61b8974f2f98f0cd7344bed2a8 c289b84b69434801a363fa8703f28a4e - - -] Task 'cinder.volume.flows.manager.create_volume.CreateVolumeFromSpecTask;volume:create' (8e671707-b9e6-4f58-adc6-793aa8c49769) transitioned into state 'FAILURE' from state 'RUNNING'
2016-01-17 01:39:39.688 25796 TRACE cinder.volume.manager Traceback (most recent call last):
2016-01-17 01:39:39.688 25796 TRACE cinder.volume.manager   File "/usr/local/lib/python2.7/dist-packages/taskflow/engines/action_engine/executor.py", line 67, in _execute_task
2016-01-17 01:39:39.688 25796 TRACE cinder.volume.manager     result = task.execute(**arguments)
2016-01-17 01:39:39.688 25796 TRACE cinder.volume.manager   File "/usr/local/lib/python2.7/dist-packages/cinder/volume/flows/manager/create_volume.py", line 639, in execute
2016-01-17 01:39:39.688 25796 TRACE cinder.volume.manager     **volume_spec)
2016-01-17 01:39:39.688 25796 TRACE cinder.volume.manager   File "/usr/local/lib/python2.7/dist-packages/cinder/volume/flows/manager/create_volume.py", line 613, in _create_raw_volume
2016-01-17 01:39:39.688 25796 TRACE cinder.volume.manager     return self.driver.create_volume(volume_ref)
[...]

PR #748, opened 2016-01-17
Adjust permission of /etc/ceph to allow Cinder to create Ceph Volumes

byronmccollum:
Cinder Volume needs access to /etc/ceph to create Ceph Volumes. This
change sets `conf_directory_mode` to adjust permissions accordingly.

Note: This depends on an upstream pull request for the ceph-common role

https://github.com/ceph/ceph-ansible/pull/485

Fixes: #747

### DIFF
--- a/rpcd/playbooks/group_vars/all.yml
+++ b/rpcd/playbooks/group_vars/all.yml
@@ -25,3 +25,6 @@ openstack_repo_url: "http://{{ internal_lb_vip_address }}:{{ repo_server_port }}
 
 # Location to store a catalog of migrations completed
 migrations_dir: /etc/openstack_deploy/migrations
+
+# Ceph Conf Directory Mode
+conf_directory_mode: 755


byronmccollum:
The role `ceph-osd` depends on `ceps-common`, but I don't know how to pass the role variables to `ceps-common` via `ceph-osd` to make this work.

byronmccollum:
Yeah so this won't work. Ansible does validation on the group, and errors if it doesn't exist, which it won't until after Cinder installs.

byronmccollum:
Setting permission to match what OpenStack Ansible cep_client role sets (root:root:755). This ensure that whichever playbook was last right, the same permissions will be applied.

byronmccollum:
Added `conf_directory_mode=755` to `ceph-common` role in `setup-mass.yml`...

byronmccollum:
If I add `conf_directory_mode` to `user_extra_varables.yml`, will the role pick it up? Currently the role is being applied twice, once as a dependency, once explicitly.

hughsaunders:
works for me :+1:

prometheanfire:
we will need to wait til the shas are updated to merge this then

hughsaunders:
@prometheanfire this could be merged in anticipation - won't break anything.

prometheanfire:
ya, this just sets a variable

prometheanfire:
:+1:

PR #749, opened 2016-01-18
[WIP] Point kilo branch of rpc-openstack at head of kilo branch of OSA

mancdaz:
openstack-ansible hasn't been bumped forward since November. This
commit brings it forward to the head of osa kilo branch:

11.2.6-55-g3208b2a (11.2.6 tag plus 55 commits). 

Since we only need a sha, and can't
guarantee when osa will tag 11.2.7, it's worth doing this now so we
have a full 2 weeks to work with this new sha and fix any issues that
arise because of it, so that it can be included in the r11.1.2 release.

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -14,6 +14,7 @@ export DEPLOY_CEILOMETER=${DEPLOY_CEILOMETER:-"no"}
 export DEPLOY_CEPH=${DEPLOY_CEPH:-"no"}
 export FORKS=${FORKS:-$(grep -c ^processor /proc/cpuinfo)}
 export ANSIBLE_PARAMETERS=${ANSIBLE_PARAMETERS:-""}
+export BOOTSTRAP_OPTS=${BOOTSTRAP_OPTS:-""}
 
 OA_DIR='/opt/rpc-openstack/openstack-ansible'
 RPCD_DIR='/opt/rpc-openstack/rpcd'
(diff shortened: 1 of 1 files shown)


mancdaz:
recheck - jenkins concurrency issues

hughsaunders:
recheck

mancdaz:
w00t! gating ftw

d34dh0r53:
So, `11.1.2` is going to be released on or shortly after `1/20/2016` as it will contain some security fixes that will be pulled in from upstream OSA.  OSA will require SHA bumps as well, I would like to hold off on the SHA bumps for RPCO until OSA has incorporated the security fixes that we need.

prometheanfire:
this will need to be resubmitted when upstream has the fixes

mancdaz:
Just testing another sha bump which includes the pip patch to fix issues with pip 8.0.0. 

@d34dh0r53 If it works, it's maybe worth considering including this sha bump just now to unblock gating, with the expectation that there will be no tag made until we bump again to include the security fixes. Otherwise nothing will gate, so other work on kilo can't be done/included.

hughsaunders:
Recheck , jenkins java stacktrace fun.

hughsaunders:
recheck (regex doesn't cope with capitals)

hughsaunders:
recheck

hughsaunders:
Such tick, so green. :+1:

odyssey4me:
:+1: This SHA is the last commit before the tagged OSA 11.2.7 release (which happened earlier today).

d34dh0r53:
:+1:

PR #751, opened 2016-01-19
remove submodules and use native ansible-y bits

prometheanfire:
original pr was here https://github.com/rcbops/rpc-openstack/pull/578


prometheanfire:
the 'default' test needs to be retargeted from master to liberty I don't have perms to do that.

PR #753, opened 2016-01-19
Document deploy.sh Environment Variables

byronmccollum:


prometheanfire:
not sure if this had a master PR before the change, but please resubmit to master :|

Issue #754, opened 2016-01-19
Add Ansible 2.x linting job (Liberty and above)


nrb:
In order to get ahead of potential issues, we should have a Travis CI job that checks our Ansible work against Ansible 2.x's syntax, so we're not stuck with a massive clean up task later.

The job wouldn't necessarily be a blocker, but rather an early warning to contributors that they may need to make the syntax cross-version compatible.

In my view, this should only be for Liberty work and later, not backported to Kilo.

PR #755, opened 2016-01-19
Add Ansible linting under Ansible 2.x

nrb:
This commit adds a new tox environment for linting playbooks and roles
against Ansible 2.x. It also modifies the linting-ansible.sh script so
that it is compatible with both the 1.x and 2.x series.

This commit is not intended to provide full Ansible 2.x support. Rather,
it is to serve as an early warning system so that the playbooks and
roles aren't written in an incompatible manner prior to actually
switching.

Fixes: #754

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,5 +14,7 @@ matrix:
       env: TOXENV=flake8
     - python: 2.7
       env: TOXENV=ansible-lint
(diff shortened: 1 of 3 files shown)


nrb:
Liberty was renamed back to master, closing this PR and opening a new one.

Issue #724, opened 2016-01-11
RPC/DB thread settings missing


BjoernT:
Apparently we renamed certain settings from juno like keystone_db_max_pool_size to keystone_database_max_pool_size in kilo and the current settings on the customer side do get lost during a upgrade. So we need to add proper default values for kilo and higher like: 

rpc_conn_pool_size: 180
rpc_response_timeout: 180
rpc_thread_pool_size: 180

keystone_database_max_pool_size: 120
keystone_database_pool_timeout: 60

﻿neutron_api_workers: 64
neutron_rpc_conn_pool_size: "{{ rpc_conn_pool_size }}" 
neutron_rpc_response_timeout: "{{ rpc_response_timeout }}"
neutron_rpc_thread_pool_size: "{{ rpc_thread_pool_size }}"
neutron_db_max_overflow: 60
neutron_db_max_pool_size: 120
neutron_db_pool_timeout: 60

nova_rpc_conn_pool_size: "{{rpc_conn_pool_size}}"
nova_rpc_response_timeout: "{{rpc_response_timeout}}"
nova_rpc_thread_pool_size: "{{rpc_thread_pool_size}}"

# committed
- [x] Master
- [x] Kilo

# released
- [x] Master
- [ ] Kilo

BjoernT:
We also need to add a documentation fix to rename existing 
 keystone_db_max_pool_size and keystone_db_pool_timeout, possible others as part of the Juno to Kilo upgrade

BjoernT:
@klevenstein 

Addition to the docs would be:

```
sed -i -e 's/keystone_db_max_pool_size:/keystone_database_max_pool_size:/g' \
        -e 's/keystone_db_pool_timeout:/keystone_database_pool_timeout:/g' \
        -e 's/neutron_db_pool_size:/neutron_db_max_pool_size:/g' \
        -e 's/neutron_db_pool_timeout:/neutron_db_pool_timeout:/g' /etc/rpc_deploy/user_variables.yml
```

odyssey4me:
Alternatively we could just rename the vars to the same names as those found in juno?

BjoernT:
I'm open to any suggestion but we also have 11.2.6 release so there might be installs using the new names. I can't tell which poison to pick ...

prometheanfire:
@BjoernT please backport to kilo

BjoernT:
OK, just cherry-picked

prometheanfire:
merged everywhere

PR #725, opened 2016-01-11
Adding new default values for neutron/nova/keystone

BjoernT:
Adding new default values for neutron/nova/keystone SQLAlchemy/Olso Thread Pool Settings for bug #724

PR #756, opened 2016-01-19
Adding new default values for neutron/nova/keystone DB/RPC pools

BjoernT:
Adding new default values for neutron/nova/keystone SQLAlchemy/Olso Thread Pool Settings for bug #724


prometheanfire:
@BjoernT click reopen https://github.com/rcbops/rpc-openstack/pull/725

PR #757, opened 2016-01-19
Add Ansible linting under Ansible 2.x

nrb:
This commit adds a new tox environment for linting playbooks and roles
against Ansible 2.x. It also modifies the linting-ansible.sh script so
that it is compatible with both the 1.x and 2.x series.

This commit is not intended to provide full Ansible 2.x support. Rather,
it is to serve as an early warning system so that the playbooks and
roles aren't written in an incompatible manner prior to actually
switching.

Fixes: #754

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,5 +16,7 @@ matrix:
       env: TOXENV=flake8
     - python: 2.7
       env: TOXENV=ansible-lint
+    - python: 2.7
+      env: TOXENV=ansible-lint ANSIBLE_VERSION='>=2.0'
 
 sudo: false
--- a/tox.ini
+++ b/tox.ini
@@ -4,10 +4,11 @@ skipsdist = True
 envlist = flake8,ansible-lint
 
 [testenv]
+passenv = ANSIBLE_VERSION
 basepython = python2.7
 deps =
-    ansible>=1.9.1,<2.0.0
     -rtest-requirements.txt
+    ansible{env:ANSIBLE_VERSION:>=1.9.1,<2.0.0}
 
 [testenv:flake8]
 commands =
--- a/README.md
+++ b/README.md
@@ -134,3 +134,13 @@ To run an upgrade of an existing openstack-ansible installation:
 Please note the following behaviors that are **destructive**:
     * `/etc/rpc_deploy` will be deprecated and the file structure moved to 
       `/etc/openstack_deploy`.
+
+# Linting
+
+If you would like to lint against a version of ansible that is not the
+default, set the `ANSIBLE_VERSION` environment variable to the proper pip
+version specification:
+
+```
+ANSIBLE_VERSION='>=2.0' tox -e ansible-lint
+```
--- a/scripts/linting-ansible.sh
+++ b/scripts/linting-ansible.sh
@@ -21,13 +21,10 @@ if [[ -z "$VIRTUAL_ENV" ]] ; then
     echo "WARNING: Not running hacking inside a virtual environment."
 fi
 
-# Put local inventory in a var so we're not polluting the file system too much
-LOCAL_INVENTORY='[all]\nlocalhost ansible_connection=local'
-
 pushd rpcd/playbooks/
   echo "Running ansible-playbook syntax check"
   # Do a basic syntax check on all playbooks and roles.
-  ansible-playbook -i <(echo $LOCAL_INVENTORY) --syntax-check *.yml --list-tasks
+  ansible-playbook -i 'localhost,' --syntax-check *.yml --list-tasks
   # Perform a lint check on all playbooks and roles.
   ansible-lint --version
   echo "Running ansible-lint"


sigmavirus24:
:+1:

d34dh0r53:
Rebase please

tox.ini
```diff
@@ -16,6 +16,16 @@ commands =
 [testenv:ansible-lint]
 commands =
     {toxinidir}/scripts/linting-ansible.sh
+deps=
+    -rtest-requirements.txt
+    ansible>=1.9.1,<2.0.0
+
+[testenv:ansible2-lint]
+commands =
+    {toxinidir}/scripts/linting-ansible.sh
```
sigmavirus24: This can be (more simply)

```
commands = {[testenv:ansible2-lint]commands}
```

That way if one of them changes, they both change automatically.

.travis.yml
```diff
@@ -16,5 +16,7 @@ matrix:
       env: TOXENV=flake8
     - python: 2.7
       env: TOXENV=ansible-lint
+    - python: 2.7
+      env: TOXENV=ansible2-lint
```
sigmavirus24: An alternative approach to using a separate testenv in tox would be to set up another environment variable, e.g., `ANSIBLE_VERSION=">=2.0"`

And then have `ansible-lint` simply do something like this for requirements:

```
passenv=ANSIBLE_VERSION
deps=
    -rtest-requirements.txt
    ansible{env:ANSIBLE_VERSION:>=1.9.1,<2.0}
```

nrb: Thanks, I'll update this.

tox.ini
nrb: Yep, that's worth doing.

.travis.yml
nrb: Based on my understanding, If I take this approach, I can't trigger ansible 2.x testing without an alternative invocation (e.g. `ANSIBLE_VERSION=">=2.0" tox -e ansible-lint`, which is generally breaking people expectations of interacting with tox locally (or mine, at least).

I think I would favor separate environments solely to keep calling environments consistent. I may have missed something here that gets around that, though.

sigmavirus24: So in that case what you could do is this:

```ini
[testenv:ansible2-lint]
setenv=
    ANSIBLE_VERSION=">=2.0"
deps=
    {[testenv:ansible-lint]deps}
commands=
    {[testenv:ansible-lint]commands}
```

But yes, otherwise that would need to be the invocation to do 2.0 testing

nrb: see my updated version, which I think is a compromise between the two you suggested.

I'm not sure I see much value derived from adding the environment variables, since we're only dealing with 2 versions.

It could simplify the two ansible environments to just using `{[testenv]deps}`, I suppose...

nrb:
```
[testenv]
setenv=
    ANSIBLE_VERSION=""
basepython = python2.7
deps =
    -rtest-requirements.txt
    ansible{env:ANSIBLE_VERSION}

[testenv:ansible-lint]
setenv=
    ANSIBLE_VERSION=>=1.9.1,<2.0.0
commands =
    {toxinidir}/scripts/linting-ansible.sh
deps=
    {[testenv]deps}

[testenv:ansible2-lint]
setenv=
    ANSIBLE_VERSION=>=2.0
commands = {[testenv:ansible-lint]commands}
deps=
    {[testenv]deps}
```
Here's what it would look like if the ansible-lint jobs were completely 'templated'. My personal preference would be just to explicitly set the ansible version on it's own line because it's more readable; the current 1.9.1 to 2.0 range looks somewhat messy as an environment var to my eyes.

sigmavirus24: So I'd caution against calling it templated. This isn't templates so much as it is substitution.

Also if the deps do not change between testenvs you don't need to specify them, e.g.,

```ini
[testenv]
setenv=
    ANSIBLE_VERSION=""
basepython = python2.7
deps =
    -rtest-requirements.txt
    ansible{env:ANSIBLE_VERSION}

[testenv:ansible-lint]
setenv=
    ANSIBLE_VERSION=>=1.9.1,<2.0.0
commands =
    {toxinidir}/scripts/linting-ansible.sh

[testenv:ansible2-lint]
setenv=
    ANSIBLE_VERSION=>=2.0
commands = {[testenv:ansible-lint]commands}
```

sigmavirus24: That said, it still misses the point of allowing the user to specify an override for the ansible-lint testenv (which I think is preferable to proliferation of testenvs based on versioning of dependencies).

nrb: I thought that the vars didn't need to be specified if unchanged, but for some reason I was under the impression that had caused failures in the past. I don't have proof of them, though, so I'll try again.

I take your point that there shouldn't be an environment for every single minor version of every dependency. This one is meant to be temporary, too.

With that in mind, I'll rewrite some of this to maybe only have one, such that updating it is as simple as removing a travis env and updating the version number, which is a combination of what you've already proposed.

sigmavirus24:
I'm :+1: once the gate passes

nrb:
Added one last change, and I think I'm satisfied now. I marked the job as an ignored failure so travis won't make the build red just because of an ansible 2.0 linting error.

sigmavirus24:
I'm +0 on allowing ansible 2.0 to fail. If we already have compatibility we should just enforce it. I don't see a downside to not enforcing it.

claco:
I agree with @sigmavirus24. If we're already achieved it, and that's where we're going, let's enforce now. WE can always shut it off in the rare circumstance we must do something in 1.9 in a 2.x incompatible way.

nrb:
Fair enough. I've pushed new changes that keep the ansible 2 job voting based on feedback.

sigmavirus24:
:+1: I somehow doubt that the ceph failure is related to this.

prometheanfire:
recheck

tests must pass, no exceptions

hughsaunders:
recheck - failed on tempest/neutronportnotfound

prometheanfire:
yay, tests passed

:+1:

PR #758, opened 2016-01-19
Added patch to redefine the OS nova service sha

cloudnull:
This change patches our nova service sha such that we're able to ensure
the BDM issue [0] is resolved for our customers. The patch file pulls
in a refspec [1] for an upstream change in the stable/kilo branch for nova
which resolves the issue. Sadly because this change is so late in the kilo 
lifecycle its unlikely upstream will accept the changeset. To ensure all customers
 upgrading to kilo or deploying green-field Kilo have a deployment that works as 
expected the patch has been created to redefine the nova sha git source and sha. 
This does not impact the deployment process for normal stable support workflow 
however it does ensure anyone building the wheels, locally or remotely have correct 
package sets. The change simply instructs our wheel building process to create a
wheel using the stable sha as the base and the refspec out of gerrit cherr-picked
on top.

[0] - https://bugs.launchpad.net/nova/+bug/1457527
[1] - https://review.openstack.org/#/c/269421


prometheanfire:
:+1:

d34dh0r53:
:+1:

PR #759, opened 2016-01-19
 Adding new default values for neutron/nova/keystone DB/RPC pools

BjoernT:
Adding new default values for neutron/nova/keystone SQLAlchemy/Olso Thread Pool Settings for bug #724

### DIFF
--- a/rpcd/etc/openstack_deploy/user_variables.yml
+++ b/rpcd/etc/openstack_deploy/user_variables.yml
@@ -13,18 +13,30 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Nova overrides
+# SQLAlchemy/Olso Thread Pool Settings
+rpc_conn_pool_size: 180
+rpc_response_timeout: 180
+rpc_thread_pool_size: 180
+
+keystone_database_max_pool_size: 120
+keystone_database_pool_timeout: 60
+
+neutron_api_workers: 64
+neutron_rpc_conn_pool_size: "{{ rpc_conn_pool_size }}"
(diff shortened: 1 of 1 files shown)


prometheanfire:
nova_cross_az_attach: False was removed?

BjoernT:
nova_cross_az_attach should be set already

claco:
Isn't the default in Kilo True, which we don't want?

http://docs.openstack.org/kilo/config-reference/content/list-of-compute-config-options.html

BjoernT:
Yes but we set it on purpose to false via https://github.com/rcbops/rpc-openstack/blob/master/rpcd/etc/openstack_deploy/user_variables.yml#L17, which was not back ported to kilo.
The reason for this is, if the customer places cinder in different DCs, he can't allocate nova instance and cinder volumes across DCs. That's why the cinder AZs should line up with nova. In case when it's not we have to enable it

claco:
That's my confusion. Once this patchset goes in, we will no longer be setting it to false as that line is being removed? https://github.com/rcbops/rpc-openstack/pull/759/files#diff-7507cc9de5006d33bbff4104099c0955L17

BjoernT:
I see, I just fixed that. Thanks for checking

prometheanfire:
lgtm :+1:

claco:
Good catch @prometheanfire!

prometheanfire:
they say I'm 'detail oriented', no clue which direction that is though.

tomc603:
Looks good. Nice catch @prometheanfire. :+1:

prometheanfire:
two :+1: , merging

Issue #763, opened 2016-01-20
v3 env missing for rpcd/playbooks/roles/rpc_maas/templates/openrc


BjoernT:
It seems we should add v3 domain and auth version settings inside the openrc-maas similar to openrc ? 
Also adding endpoint type selection and a new setting for insecure API calls, the insecure openrc setting is not used for MAAS yet until OSA fixes the availability of openrc_insecure inside the host vars
```
export OS_USER_DOMAIN_NAME=Default
export OS_PROJECT_DOMAIN_NAME=Default

# For openstackclient
export OS_IDENTITY_API_VERSION=3
export OS_AUTH_VERSION=3
```

PR #764, opened 2016-01-20
Adding Keystone V3 settings into openrc-maas along with endpoint type selection

BjoernT:
Adding V3 settings into openrc-maas, similar to standard openrc.
Also adding endpoint type selection and a new setting for insecure API calls (OSA fix pending LP #1537117)
Also fixing nova_exc.Unauthorized error for nova to reauthenticate instead
of only expecting AttributeError

Closes-Bug: #763 
Closes-Bug: #765

### DIFF
--- a/maas/plugins/maas_common.py
+++ b/maas/plugins/maas_common.py
@@ -28,7 +28,16 @@ import traceback
 AUTH_DETAILS = {'OS_USERNAME': None,
                 'OS_PASSWORD': None,
                 'OS_TENANT_NAME': None,
-                'OS_AUTH_URL': None}
+                'OS_AUTH_URL': None,
+                'OS_USER_DOMAIN_NAME': None,
+                'OS_PROJECT_DOMAIN_NAME': None,
+                'OS_IDENTITY_API_VERSION': None,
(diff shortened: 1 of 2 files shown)


maas/plugins/maas_common.py
```diff
@@ -351,13 +378,22 @@ def get_service_catalog(auth_ref):
                         auth_ref.get('serviceCatalog'))
 
 
+def get_endpoint_type(auth_details):
```
sigmavirus24: Can we just do

```py
def get_endpoint_type(auth_details):
    endpoint_type = auth_details['OS_ENDPOINT_TYPE']
    if endpoint_type == 'publicURL':
        return 'public'
    if endpoint_type == 'adminURL':
        return 'admin'
    return 'internal'
```

BjoernT: sure

sigmavirus24:
Other than my one comment I have no other review comments. Thanks @BjoernT

prometheanfire:
Generally seems alright, would like to get someone who's done more maas work to review first though.

d34dh0r53:
:+1:

PR #766, opened 2016-01-22
Do not create RabbitMQ monitoring user if RabbitMQ is not installed.

byronmccollum:
This is needed when deploying stand alone Swift, and the RabbitMQ
affinity is set to 0 to prevent its installation.

### DIFF
--- a/rpcd/playbooks/roles/rpc_maas/tasks/main.yml
+++ b/rpcd/playbooks/roles/rpc_maas/tasks/main.yml
@@ -23,7 +23,7 @@
 
 - include: rabbitmq_user.yml
   when: >
-    inventory_hostname == groups['rabbitmq_all'][0]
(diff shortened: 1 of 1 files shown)


byronmccollum:
Creating this in the event an `r11.0.5` release happens for the recent pip 8 issues.

hughsaunders:
recheck

prometheanfire:
recheck

prometheanfire:
actually, this needs to go into kilo/master, not zzz-archive-kilo

byronmccollum:
@prometheanfire This is already fixed in the r11.1 series and on. I opened this targeting the r11.0 series (zzz-archive-kilo) in the off change that an r11.0.5 release would be needed to fix all the pip issues. (Un)fortunately, the pip issues resolved themselves, and I don't think y'all will be releasing any more versions on the r11.0 series....so close maybe?

prometheanfire:
sgtm, closing

Issue #767, opened 2016-01-22
Difficult to inject variables when building an AIO


hughsaunders:
Currently the rpc deploy script checks if /etc/openstac_deploy exists and skips large sections of setup if it does. This makes it difficult to pre-populate that directory with config. 

When running the deploy script there are no opportunities after /etc/openstack_deploy is created to insert extra variables. Currently the only option is to checkout rpc-openstack, modify or insert variables into the tree, then run the deploy script which will create /etc/openstack_deploy and copy in the changes. 

I propose that we test for an individual file within /etc/openstack_deploy so that the directory itself can be created before running the script and pre-populated with variables. This will help with gate jobs as we can have a variables file for each scenario and copy the relevant one into /etc/openstack_deploy before running deploy.sh. 

---
 - [ ] Master #768 
 - [ ] Kilo

d34dh0r53:
Love it :+1:

prometheanfire:
can merge earlier if it's ready, this is upgrade related

nrb:
I like the idea, though I had some feedback on the implementation in the PR.

I'd like to be sure that the RPC defaults carried in tree always win over OSA defaults, but user-defined values beat both.

PR #768, opened 2016-01-22
Use existing vars file in deploy.sh

hughsaunders:
The rpc deploy script currently has a checks whether some variable files
exist in /etc/openstack_deploy. If they don't the repo version is copied
in. The file is then merged with the repo version via update-yaml, the
repo values taking precedence.

IMHO this is backwards, user provided values should override values
provided in the repo. There is a larger problem though, which is that
this code for checking/updating vars files is only run if
/etc/openstack_deploy doesn't exist, so there will never be any user
specified files to read/update. This is probably why the reverse
precedence hasn't been detected.

What this patch does:
  * Check for the existence of
    /etc/openstack_deploy/openstack_user_config.yaml rather than
    /etc/openstack_deploy when deciding whether to bootstrap an aio.
    This allows a user to precreate that directory and supply their own
    vars files before running deploy.sh

  * Flip the precedence so that user provided values override repo
    provided values

Related: #767

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -50,14 +50,23 @@ if [[ "${DEPLOY_AIO}" == "yes" ]]; then
   fi
   # force the deployment of haproxy for an AIO
   export DEPLOY_HAPROXY="yes"
-  if [[ ! -d /etc/openstack_deploy/ ]]; then
+  if [[ ! -e /openstack ]]; then
     ./scripts/bootstrap-aio.sh
     pushd ${RPCD_DIR}
-      for filename in $(find etc/openstack_deploy/ -type f -iname '*.yml'); do
-        if [[ ! -a "/${filename}" ]]; then
-          cp "${filename}" "/${filename}";
+      for relative_path in $(find etc/openstack_deploy/ -type f -iname '*.yml'); do
+        # If relative_path exists in /etc/openstack_deploy then it is merged
+        # into then repo version overriding the repo value when there is a
+        # conflict.
+        if [[ -e "/${relative_path}" ]]; then
+          # Output is written to the base file
+          #                         Base                 Override
+          ../scripts/update-yaml.py "${relative_path}"   "/${relative_path}";
+          #                         repo/rpcd/etc        /etc/openstack_deploy
         fi
-        ../scripts/update-yaml.py "/${filename}" "${filename}";
+          # Copy (possibly updated) config files from repo to
+          # /etc/openstack_deploy
+          cp "${relative_path}" "/${relative_path}";
+          git checkout ${relative_path}
       done
     popd
     # ensure that the elasticsearch JVM heap size is limited


scripts/deploy.sh
```diff
@@ -31,14 +31,22 @@ cd ${OA_DIR}
...
+        if [[ -e "/${relative_path}" ]]; then
+          # Output is written to the base file
+          #                         Base                 Override
+          ../scripts/update-yaml.py "${relative_path}"   "/${relative_path}";
+          #                         repo/rpcd/etc        /etc/openstack_deploy
         fi
-        ../scripts/update-yaml.py "/${filename}" "${filename}";
+          # Copy (possibly updated) config files from repo to
+          # /etc/openstack_deploy
+          cp "${relative_path}" "/${relative_path}";
```
prometheanfire: This dirties up the repo meaning a it will need to be cleaned up before pulling down again.

prometheanfire: can we add a `git checkout "${relative_path}"` here to reset the file?

hughsaunders: done

prometheanfire:
rebased to pass travis, all other tests were passing

hughsaunders:
recheck

hughsaunders:
recheck apt fail

scripts/deploy.sh
```diff
@@ -50,14 +50,23 @@ if [[ "${DEPLOY_AIO}" == "yes" ]]; then
   fi
   # force the deployment of haproxy for an AIO
   export DEPLOY_HAPROXY="yes"
-  if [[ ! -d /etc/openstack_deploy/ ]]; then
+  if [[ ! -e /openstack ]]; then
```
prometheanfire: The commit message mentioned that this would change to `/etc/openstack_deploy/openstack_user_config.yaml` not `/openstack`.  Unless I'm missing something, what gives?

nrb:
This has a documentation impact for users. At the very least the readme should let the users know that's happening.

scripts/deploy.sh
```diff
@@ -50,14 +50,23 @@ if [[ "${DEPLOY_AIO}" == "yes" ]]; then
...
-        if [[ ! -a "/${filename}" ]]; then
-          cp "${filename}" "/${filename}";
+      for relative_path in $(find etc/openstack_deploy/ -type f -iname '*.yml'); do
+        # If relative_path exists in /etc/openstack_deploy then it is merged
+        # into then repo version overriding the repo value when there is a
+        # conflict.
+        if [[ -e "/${relative_path}" ]]; then
+          # Output is written to the base file
+          #                         Base                 Override
+          ../scripts/update-yaml.py "${relative_path}"   "/${relative_path}";
```
nrb: How does this account for overrides that RPC specifically put in place? The idea was that we could override the OSA defaults with out own, should we diverge. As I'm reading this, OSA wins, provided the user hasn't customized this.

I'm struggling with the same problem during upgrades for when users upgrade env.d files; I'm not sure a simple copy is going to work.

nrb:
I agree with your statement that user-provided values should win, but in the case they aren't, RPC's defaults need to override OSAs.

As this patch stands, I don't think that's accomplished. Originally when we created rpc-openstack as a wrapper project, I was thinking we use something besides user_variables.yml to specify RPC variables, but that user_variables.yml would always 'win' in alphabetic precedence when given to openstack-ansible.

Provided Ansible's precedence works as documented, I think this would give us a the layering we want: user-provided values accepted first, then RPC, then OSA.

nrb:
I know others we against having separate variable files for the "same" values, but to me, it makes sense to have *users* edit user_variables and code that we control edit its own file - that way, we don't have two sources of info feeding into the same file, thus making it hard to determine which edits are really the winners.

tomc603:
My initial reading of this made more sense, but as I reread it I think there is some ambiguity because of typos perhaps.

What I thought the intent was:
- If vars file are found in /etc/openstack_deploy, we merge the _new_ vars files from the checked out repo with the _existing_ vars files in /etc/openstack_deploy
  - variable precedence is OSA (lowest), RPC-O repo defaults, existing values, user preferences (highest)
- If no vars files are found, simply deploy files from the checked out repo, and override with user preferences

Given this possibly incorrect understanding of the situation, we need to swap the "base" and "override" arguments so that RPC-O wins over OSA. This still completely ignores situations where values in OSA deploys have been overridden by users/support, though, so that's still a problem. Additionally, I agree with with @prometheanfire comment [here](https://github.com/rcbops/rpc-openstack/pull/768#discussion-diff-52795930)- the check for /openstack seems incorrect.

My gut feeling is that given the way things are, the fix for situations where support has overridden OSA values, support is going to have to port those into the vars files after the `update-yaml` process runs. So they would need to back up those vars files _before_ running the `upgrade` or `deploy` scripts, which is a change in process and needs to be spelled out clearly in the docs and in every conversation we have with support until all customers are upgraded (yikes). This certainly isn't optimal, but it would be difficult to try to glean the intent of a deployment with a simple script that has no knowledge of what values were default versus overrides at deploy time, and how those values have changed over time.

I also agree that working in the checked out repo dirties it with unadded/uncommitted changes, but that doesn't really matter unless someone attempts to deploy a clean RPC-O deployment from a dirty git tree. Probably not a very big issue.

Ultimately, I think the best solution to this problem (not the bug/PR) is to clearly document the precedence we use for OSA/RPC default files, then modify the `update-yaml` script to reflect that better. Overrides by customers/support should go in a `user_${filename}.yml` file and override the RPC-O values at runtime. This way the defaults are kept clean, the intent of user configuration is clear in the future, and munging files becomes less of a problem for us.

scripts/deploy.sh
hughsaunders: I changed to checking /openstack because I wanted to be able to create /etc/openstack_deploy/openstack_user_config.yaml before running deploy.sh, and not have it overridden. Will update the commit message.

scripts/deploy.sh
hughsaunders: This isn't about OSA vs RPC. At this point OSA hasn't been checked out yet. What this means is that if you create /etc/openstack_deploy and put some configs in there before running deploy.sh, those values will take precedence over values that are in the rpc-openstack repo. This allows overriding values in the gate. 

What actually happens is that the repo version is read, its updated with any values from /etc, then the repo version is written out to disk. That means the merged version is now in tree, so its copied out to etc, then reset via git so we don't leave a dirty repo.

hughsaunders:
This script (rpc/scripts/deploy.sh) runs very early. At this point OSA has not dropped any values in to /etc, so there is no question of OSA/RPC precedence. It only allows values that exist in /etc to override values provided in the rpc repo. Scripts/deploy shouldn't be running on existing deploys where support have changed stuff or OSA has already dropped its defaults in /etc.

prometheanfire:
upgrade.sh runs deploy.sh

nrb:
As was said above, upgrade.sh calls deploy.sh. I'd have to check the environment vars it sets though, not sure if it actually tries to copy the etc directory over or not.

That said, support is not using upgrade.sh nor deploy.sh in the upgrade process; they are copying the lines out manually.

I do understand what you're saying now in terms of a fresh install, though; if someone or something created the directory prior to deploy.sh, then that should win.

hughsaunders:
Ah, I wasn't aware that upgrade runs deploy. So when would we want to override values in /etc with values from the rpc repo?

nrb:
I think that would happen during version upgrades, so when the environment/config changes, or variables are added/removed.

Though that might be better off being in a separate process...

prometheanfire:
I think part of the problem that we could have is once a config is laid down, how do we know which values in it are user set?  What may be easiest is to have a user_extra_zzz_variables or something that is totally unmanaged by us and would override anything in OSA or RPC-O.  This would reduce the merging we have to just one level (OSA and RPC-O).

nrb:
I agree that an extra file should be useful, but I think that file should be `user_variables.yml`, since that was its original purpose as I understand it.

Ultimately that's a debate over semantics, but the problem @prometheanfire brings up is real in upgrades for both OSA (which I'm struggling with for env.d changes) and RPC - having a single flat file that RPC, OSA, and deployers without tagging who did what becomes a sticky issue.

I'd almost wonder if some of this is more in scope for OSA to define, with us using RPC as a test bed for downstream projects.

prometheanfire:
ya, my comment was more for OSA to define it

hughsaunders:
@git-harry pointed out that this change is within the AIO section so doesn't affect production upgrades.

git-harry:
I've only skimmed this conversation but it looks like there is a misunderstanding here - the patch is only relevant to AIOs. The idea of the update-yaml.py script is to override the vars dropped by the OSA bootstrap process when rpc-o is carrying a different default for that var.

nrb:
> the patch is only relevant to AIOs.

Good point, I missed that in the context of the diff.

prometheanfire:
I do think we should probably take time to fully rethink how we are handling the config (lifecycle wise).

d34dh0r53:
recheck

tomc603:
There seems to be a rabbitmq failure on the last build.

recheck

tomc603:
recheck

prometheanfire:
is this still being worked or should it be closed?

Issue #783, opened 2016-01-25
switch config_drive_format default to vfat


prometheanfire:
this will allow us to do migrations

https://github.com/openstack/nova/blob/master/nova/virt/libvirt/driver.py#L6350-L6364

git-harry:
I've removed the milestone because, to my knowledge, this not been planned to be part of the release.

b3rnard0:
This was added a while back for consideration into 13.0. It's likely a candidate to be discussed for consideration in 13.2.

prometheanfire:
I wonder what our efficiency is on working on this (a one line change)

git-harry:
@prometheanfire the reason I've changed this is because I'm attempting to tidy up the issues in the repo. It may feel like things are going backwards but what we're hopefully achieving here is drawing a line in the sand so that going forward we are more efficient and predictable with better tracking of what is being added to the product.

This change seems to be for enabling live migrations, that is relatively significant in function and it strikes me as appropriate to have testing in place for it. What I've done is reclassify this as a medium enhancement and move it to u-suk-dev so that we can make sure it is properly tracked.

PR #784, opened 2016-01-25
change the default config_drive_format to vfat

prometheanfire:
this will allow migrations to work
https://github.com/openstack/nova/blob/c9c0b1b4b5acc1ea3c2bbb694aa79d7e4a7087e3/nova/virt/libvirt/driver.py#L6237-L6251

Connects #783

### DIFF
--- a/rpcd/etc/openstack_deploy/user_variables.yml
+++ b/rpcd/etc/openstack_deploy/user_variables.yml
@@ -33,6 +33,10 @@ nova_rpc_conn_pool_size: "{{ rpc_conn_pool_size }}"
 nova_rpc_response_timeout: "{{ rpc_response_timeout }}"
 nova_rpc_thread_pool_size: "{{ rpc_thread_pool_size }}"
 
+nova_nova_conf_overrides:
+  DEFAULT:
+    config_drive_format: "vfat"
+
 # Nova config overrides
 nova_cross_az_attach: False
 


hughsaunders:
recheck

BjoernT:
:+1:  We do this change quite often at customer environments who want to retain the live migration and config drives. But there is one known issue. Images with coreos are not able to mount vfat filesystems (at least stock) so we probably add a release note about those restrictions.

d34dh0r53:
I think we will need release notes around Windows images as well.

d34dh0r53:
Note to self, added `impacts-gating` as we may want to add a live migration test to the AIO.

BjoernT:
Yeah, Windows seem to rely more on EC2 metadata than config drive but that depends on the used cloud-init agent. But it's true we need to include this information too

prometheanfire:
@d34dh0r53 it looks like we can do this if we build nova in containers, but then we wouldn't be able to test cinder.  Both testing this (and cinder) might be better for when we do our multi-node test-release process.

prometheanfire:
recheck

prometheanfire:
Talked more with support and it seems like this is something that's already being done sometimes.

prometheanfire:
so, what are the obstacles to getting this merged?

BjoernT:
Support has no issues as long as we document this enough

d34dh0r53:
:+1:

prometheanfire:
recheck_upgrade

git-harry:
I've closed this to keep the repo in good order. The pull request is referenced in https://github.com/rcbops/u-suk-dev/issues/245, which is the new issue tracking this work so when it becomes approved this code is available for use/reference.

Issue #777, opened 2016-01-25
setuptools > 19.4 is breaking holland (and it's plugins) installs


mancdaz:
This is the issue when attempting to do something like this with setuptools >= 19.4:
```
pip install plugins/holland.backup.random/
```

```
Processing ./plugins/holland.backup.random
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 3, in <module>
      File "/usr/local/lib/python2.7/dist-packages/setuptools/__init__.py", line 11, in <module>
        from setuptools.extern.six.moves import filterfalse, map
      File "/usr/local/lib/python2.7/dist-packages/setuptools/extern/__init__.py", line 1, in <module>
        from pkg_resources.extern import VendorImporter
      File "/usr/local/lib/python2.7/dist-packages/pkg_resources/__init__.py", line 3130, in <module>
        @_call_aside
      File "/usr/local/lib/python2.7/dist-packages/pkg_resources/__init__.py", line 3116, in _call_aside
        f(*args, **kwargs)
      File "/usr/local/lib/python2.7/dist-packages/pkg_resources/__init__.py", line 3155, in _initialize_master_working_set
        add_activation_listener(lambda dist: dist.activate())
      File "/usr/local/lib/python2.7/dist-packages/pkg_resources/__init__.py", line 956, in subscribe
        callback(dist)
      File "/usr/local/lib/python2.7/dist-packages/pkg_resources/__init__.py", line 3155, in <lambda>
        add_activation_listener(lambda dist: dist.activate())
      File "/usr/local/lib/python2.7/dist-packages/pkg_resources/__init__.py", line 2655, in activate
        declare_namespace(pkg)
      File "/usr/local/lib/python2.7/dist-packages/pkg_resources/__init__.py", line 2225, in declare_namespace
        _handle_ns(packageName, path_item)
      File "/usr/local/lib/python2.7/dist-packages/pkg_resources/__init__.py", line 2194, in _handle_ns
        path.sort(key=sort_key)
      File "/usr/local/lib/python2.7/dist-packages/pkg_resources/__init__.py", line 2192, in sort_key
        return sys_path.index(_normalize_cached(os.sep.join(parts)))
    ValueError: '/tmp/pip-l8H5uM-build' is not in list
```

@hughsaunders and I traced it back to this commit in setuptools which was released with 19.4:

[...]

PR #788, opened 2016-01-26
pin setuptools to 19.3

mancdaz:
this commit pins setuptools to 19.3 in the galera containers.
setuptools>=19.4 does not allow holland installs to work the way we are
currently doing them (local pip installs).

This will unblock the kilo gate and allow us to work on the better
solution of having the holland pip wheels get built by the repo builder
and install that way, rather than a local pip install.

Partial: #777

### DIFF
--- a/rpcd/etc/openstack_deploy/user_variables.yml
+++ b/rpcd/etc/openstack_deploy/user_variables.yml
@@ -62,3 +62,9 @@ ceph_mons: >
   {% endfor -%}
   {% endif -%}
   {{ _var }}
+
+galera_pip_packages:
+  - MySQL-python
+  - python-memcached
+  - pycrypto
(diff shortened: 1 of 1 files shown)


d34dh0r53:
:+1:

d34dh0r53:
recheck

rpcd/etc/openstack_deploy/user_variables.yml
```diff
@@ -62,3 +62,9 @@ ceph_mons: >
   {% endfor -%}
   {% endif -%}
   {{ _var }}
+
+galera_pip_packages:
```
git-harry: It would be helpful to have a comment here referencing the issue tracking the eventual fix.

d34dh0r53:
This was superseded by #790

PR #792, opened 2016-01-26
[WIP] Add colorized output to jenkins console

d34dh0r53:
According to Apsu the AnsiColor Plugin has been installed on our
Jenkins server.  And according to [`this`](https://major.io/2014/06/25/get-colorful-ansible-output-in-jenkins/) page from a highly suspect
developer loosely affiliated with our team all we should need to
do to get it working is to export `ANSIBLE_FORCE_COLOR` with a
true value.


d34dh0r53:
stupid subprojects

Issue #789, opened 2016-01-26
Additional maas refactoring required


BjoernT:
When using a v2 keystone client, the token inside  is_token_expired can be accessed without the token.get function. 
Additionally I want to remove redundant auth code inside the individual plugins so that we can use auth code over the  maas_common function only. That also enables proper endpoint selection of the individual client like nova etc.

- [x] Liberty #794

BjoernT:
PR following

d34dh0r53:
Adding to `r11.1.4` per @BjoernT

prometheanfire:
backport?

BjoernT:
RPC-R could consume the changes from master so I would not really need a backport but since we already cherry-picked it I don't mind back porting it.

PR #794, opened 2016-01-27
RPC-R support plus ip argument and glance_api_local_check rewriting

BjoernT:
Changes : 
- glance_api_local_check: utilize glanceclient now instead of requests
- auth_ref['auth_token'] reference has been updated to keystone.auth_token to support v2 and v3 token format
- auth_ref['project']['id'] reference has been updated to keystone.tenant_id to support v2 and v3 extraction
- IP argument now optional on many checks to fall to query the service endpoint rather than individual IP (except cinder)
- This adds support for RPC-R which still uses keystone v2 client/token format

Closes-Bug: #789

### DIFF
--- a/maas/plugins/glance_api_local_check.py
+++ b/maas/plugins/glance_api_local_check.py
@@ -16,55 +16,48 @@
 
 import argparse
 import collections
-
 import ipaddr
+import time
+
+from glanceclient import exc as exc
 from maas_common import get_auth_ref
-from maas_common import get_keystone_client
+from maas_common import get_glance_client
 from maas_common import metric
 from maas_common import metric_bool
 from maas_common import print_output
 from maas_common import status_err
 from maas_common import status_ok
-import requests
-from requests import exceptions as exc
+
 
 IMAGE_STATUSES = ['active', 'queued', 'killed']
 
 
 def check(auth_ref, args):
-    # We call get_keystone_client here as there is some logic within to get a
-    # new token if previous one is bad.
-    keystone = get_keystone_client(auth_ref)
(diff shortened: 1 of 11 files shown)


BjoernT:
recheck
another cirrus image issue....

BjoernT:
another cirrus image issue....

BjoernT:
recheck

hughsaunders:
recheck

hughsaunders:
lint fail is being addressed in #813

d34dh0r53:
:+1:

maas/plugins/glance_api_local_check.py
```diff
@@ -16,55 +16,48 @@
...
-    auth_token = keystone.auth_token
-    api_endpoint = 'http://{ip}:9292/v1'.format(ip=args.ip)
-
-    s = requests.Session()
-
-    s.headers.update(
-        {'Content-type': 'application/json',
-         'x-auth-token': auth_token})
+    GLANCE_ENDPOINT = (
+        'http://{ip}:9292/v1'.format(ip=args.ip)
```
prometheanfire: do we want to hardcode protocol and port?

prometheanfire:
looks good, I think doing another pass eventually to add support for https might make sense eventually.

:+1:

maas/plugins/glance_api_local_check.py
BjoernT: The endpoints are hard coded everywhere so I did not change it, I fixed the more important stuff to use the glance client rather than then requests. We can roll this up in a later cycle and fix all the checks, by adding a port/protocol option

PR #795, opened 2016-01-27
Feature colorize Ansible output

d34dh0r53:
According to `Apsu` our jenkins server has the plugin installed to colorize ansible output, and according to this [`shady`](https://major.io/2014/06/25/get-colorful-ansible-output-in-jenkins/) site I found on the internet all we should have to do is to export `ANSIBLE_FORCE_COLOR` with a `true` value.


d34dh0r53:
recheck

d34dh0r53:
It works!
![Example of colored jenkins console.html][example]

[example]: https://820612b8facd2e0fd750-33886355402e2d315a125d039414e445.ssl.cf5.rackcdn.com/JenkinsCodeHighlighting.jpg

nrb:
+1, verified the colors show up in my browser.

prometheanfire:
can you verify that the plugin doesn't have xss issues like @cloudnull mentioned?

:-1: til then

prometheanfire:
ok, was confirmed fixed

:+1:

nrb:
The confirmed fix is at https://github.com/dblock/jenkins-ansicolor-plugin/issues/14.

PR #750, opened 2016-01-18
Point master branch of rpc-openstack at head of liberty branch of OSA

mancdaz:
Point master branch of rpc-openstack at head of liberty branch of OSA

In order to begin development towards a liberty release, we need to
switch our master branch to point at osa liberty. This commit bumps
forward to the head of osa liberty

12.0.5-4-ga91fa43 (12.0.5 plus 4 commits)

This commit also contains necessary changes to have the gate run and
pass linting, including:

- use ephemeral disk on cloud server when AIO
- bootstrap ansible before bootstrap aio (which in liberty is ansible)
- run lint using test-requirements.txt instead of dev-requirements.txt
- pass a list of locations that should be scanned for files containing
  information about what packages to build (allows us to define
  /opt/rpc-openstack/rpcd so that all of the rpc packages can be built at
  the same time by the osa repo-builder and we don't need separate rpc
  plays to build our own packages)
- correct the location of the horizon-manage.py file due to venv use
- pin setuptools==19.3 to allow holland to install
- pin elasticsearch<2.1.0 as this is a requirement of latest
  elasticsearch-curator

NOTE: I have not included the patch to make the holland wheels get built, rather than installing locally.  Instead, I have gone with the less intrusive option to pin setuptools and allow the old method to work.  Since this PR is an effort to do the minimum to get a working gate for liberty, I felt that work could be done in a subsequent patch

PR #796, opened 2016-01-27
 Point master branch of rpc-openstack at head of liberty branch of OSA

prometheanfire:
In order to begin development towards a liberty release, we need to
switch our master branch to point at osa liberty. This commit bumps
forward to the head of osa liberty

12.0.5-13-ge002596 (12.0.5 plus 13 commits)

This commit also contains necessary changes to have the gate run and
pass linting, including:

- use ephemeral disk on cloud server when AIO
- bootstrap ansible before bootstrap aio (which in liberty is ansible)
- run lint using test-requirements.txt instead of dev-requirements.txt
- pass a list of locations that should be scanned for files containing
  information about what packages to build (allows us to define
  /opt/rpc-openstack/rpcd so that all of the rpc packages can be built at
  the same time by the osa repo-builder and we don't need separate rpc
  plays to build our own packages)
- remove unneeded rpc plays to build rpc repo
- correct the location of the horizon-manage.py file due to venv use
- pin setuptools==19.3 to allow holland to install
- pin elasticsearch<2.1.0 as this is a requirement of latest
  elasticsearch-curator
- slight var modification to drop the correct maas openrc file
- add swift as a pip package for maas - we need it on host for swift
  monitoring via swift-recon
- remove with_items usage from rpc_maas pip package installation to
  prevent it from expanding list items as variables (specifically swift)

Please note, this patch includes a change to the way rpc wheels are built.
Previously, we had plays in rpc-o to build wheels for rpc and place
them in a separate rpc repo. Because in liberty osa, yaprt has gone
away, we would have needed to rework those plays. There were 2 choices:

- rewrite the plays to use the new wheel building machinery
- utilise upstream patch https://review.openstack.org/#/c/271233/ to
  allow the osa wheel building process to build packages for us, and so
remove unneccessary duplication of plays in rpc-o.

I chose the latter, as it seemed the best choice to keep the work in
this patch minimal


prometheanfire:
replaces #750

hughsaunders:
recheck - failed to get a node to test on.

hughsaunders:
recheck

mancdaz:
recheck

mancdaz:
recheck

hughsaunders:
recheck

hughsaunders:
recheck

prometheanfire:
To be clear, this replaced #750 because it the source for this is a branch on this repo. :D

mancdaz:
recheck

mancdaz:
```
rackspace_cloud_api_key: cc52ff2e158a4f4f84e57c65c8b7e7e0
rackspace_cloud_auth_url: https://lon.identity.api.rackspacecloud.com/v2.0/
rackspace_cloud_password: Manch3st3rmabar0k
rackspace_cloud_tenant_id: 10001878
rackspace_cloud_username: darrenbirkett
```

mancdaz:
recheck

d34dh0r53:
recheck

mancdaz:
jenkins is a liar. This is working on my AIOs (I just ran 4 in a row).

It appears to be failing at the point of installing swift on the host - needed for the swift monitoring plugins (which is indeed part of this patch):

```
20:03:51 changed: [jrpcaioiad-f68] => (item=rackspace-monitoring-cli)
20:03:51 changed: [jrpcaioiad-f68] => (item=ipaddr)
20:03:51 failed: [jrpcaioiad-f68] => (item={u'mount_point': u'/srv', u'storage_network': u'br-storage', u'storage_policies': [{u'policy': {u'default': True, u'index': 0, u'name': u'default'}}], u'replication_network': u'br-storage', u'drives': [{u'name': u'swift1.img'}, {u'name': u'swift2.img'}, {u'name': u'swift3.img'}], u'part_power': 8}) => {"failed": true, "item": {"drives": [{"name": "swift1.img"}, {"name": "swift2.img"}, {"name": "swift3.img"}], "mount_point": "/srv", "part_power": 8, "replication_network": "br-storage", "storage_network": "br-storage", "storage_policies": [{"policy": {"default": true, "index": 0, "name": "default"}}]}, "parsed": false}
20:03:51 OpenSSH_6.6.1, OpenSSL 1.0.1f 6 Jan 2014
20:03:51 debug1: Reading configuration data /etc/ssh/ssh_config
20:03:51 debug1: /etc/ssh/ssh_config line 19: Applying options for *
20:03:51 debug1: auto-mux: Trying existing master
20:03:51 debug1: mux_client_request_session: master session id: 2
20:03:51 Traceback (most recent call last):
20:03:51   File "<stdin>", line 1973, in <module>
20:03:51   File "<stdin>", line 300, in main
20:03:51 AttributeError: 'dict' object has no attribute 'startswith'
20:03:51 
```

which happens here:

https://github.com/rcbops/rpc-openstack/commit/521f5197701ab1ef393ed3fda87c7e4308a286f8#diff-1d0825392161b9f338fa4db5ddd23713R414

but I'm not sure why it's failing

hughsaunders:
This seems to be happening because with_flattened expands strings as variables: https://gist.github.com/hughsaunders/91974e4f4053a81f1285

hughsaunders:
removed with_flattened usage in rpc_maas pip package installation.

hughsaunders:
recheck timed out on ceilometer

rpcd/playbooks/roles/rpc_maas/tasks/package_install.yml
```diff
@@ -26,6 +26,4 @@
     name: "{{ item }}"
     state: present
     extra_args: "{{ pip_install_options | default('') }}"
-  with_flattened:
-    - maas_pip_packages
-    - maas_pip_dependencies
+  with_items: maas_pip_packages + maas_pip_dependencies
```
sigmavirus24: Shouldn't this be

```yaml
with_flattened:
  - "{{ maas_pip_packages }}"
  - "{{ maas_pip_dependencies }}"
```

Based on the example [here](https://docs.ansible.com/ansible/playbooks_loops.html#flattening-a-list)

rpcd/playbooks/roles/rpc_maas/defaults/main.yml
```diff
@@ -411,6 +411,7 @@ maas_apt_packages:
 maas_pip_packages:
   - rackspace-monitoring-cli
   - ipaddr
+  - swift
```
sigmavirus24: Why is swift a pip package?

rpcd/playbooks/roles/rpc_maas/tasks/package_install.yml
d34dh0r53: I think this may be the cause of what we're [`seeing`](http://jenkins.propter.net/job/RPC-AIO/870/consoleFull):

```
15:20:00 changed: [jrpcaioiad-2c] => (item=ipaddr)
15:20:00 failed: [jrpcaioiad-2c] => (item={u'mount_point': u'/srv', u'storage_network': u'br-storage', u'storage_policies': [{u'policy': {u'default': True, u'index': 0, u'name': u'default'}}], u'replication_network': u'br-storage', u'drives': [{u'name': u'swift1.img'}, {u'name': u'swift2.img'}, {u'name': u'swift3.img'}], u'part_power': 8}) => {"failed": true, "item": {"drives": [{"name": "swift1.img"}, {"name": "swift2.img"}, {"name": "swift3.img"}], "mount_point": "/srv", "part_power": 8, "replication_network": "br-storage", "storage_network": "br-storage", "storage_policies": [{"policy": {"default": true, "index": 0, "name": "default"}}]}, "parsed": false}
15:20:00 OpenSSH_6.6.1, OpenSSL 1.0.1f 6 Jan 2014
15:20:00 debug1: Reading configuration data /etc/ssh/ssh_config
15:20:00 debug1: /etc/ssh/ssh_config line 19: Applying options for *
15:20:00 debug1: auto-mux: Trying existing master
15:20:00 debug1: mux_client_request_session: master session id: 2
15:20:00 Traceback (most recent call last):
15:20:00   File "<stdin>", line 1973, in <module>
15:20:00   File "<stdin>", line 300, in main
15:20:00 AttributeError: 'dict' object has no attribute 'startswith'
15:20:00 
15:20:00 ok: [jrpcaioiad-2c] => (item=requests)
```

sigmavirus24:
So the AIO is failing with

```
15:20:00 changed: [jrpcaioiad-2c] => (item=rackspace-monitoring-cli)
15:20:00 changed: [jrpcaioiad-2c] => (item=ipaddr)
15:20:00 failed: [jrpcaioiad-2c] => (item={u'mount_point': u'/srv', u'storage_network': u'br-storage', u'storage_policies': [{u'policy': {u'default': True, u'index': 0, u'name': u'default'}}], u'replication_network': u'br-storage', u'drives': [{u'name': u'swift1.img'}, {u'name': u'swift2.img'}, {u'name': u'swift3.img'}], u'part_power': 8}) => {"failed": true, "item": {"drives": [{"name": "swift1.img"}, {"name": "swift2.img"}, {"name": "swift3.img"}], "mount_point": "/srv", "part_power": 8, "replication_network": "br-storage", "storage_network": "br-storage", "storage_policies": [{"policy": {"default": true, "index": 0, "name": "default"}}]}, "parsed": false}
15:20:00 OpenSSH_6.6.1, OpenSSL 1.0.1f 6 Jan 2014
15:20:00 debug1: Reading configuration data /etc/ssh/ssh_config
15:20:00 debug1: /etc/ssh/ssh_config line 19: Applying options for *
15:20:00 debug1: auto-mux: Trying existing master
15:20:00 debug1: mux_client_request_session: master session id: 2
15:20:00 Traceback (most recent call last):
15:20:00   File "<stdin>", line 1973, in <module>
15:20:00   File "<stdin>", line 300, in main
15:20:00 AttributeError: 'dict' object has no attribute 'startswith'
15:20:00 
15:20:00 ok: [jrpcaioiad-2c] => (item=requests)
15:20:00 changed: [jrpcaioiad-2c] => (item=lxml)
15:20:00 changed: [jrpcaioiad-2c] => (item=python-cinderclient)
15:20:00 changed: [jrpcaioiad-2c] => (item=python-glanceclient)
15:20:00 changed: [jrpcaioiad-2c] => (item=python-heatclient)
15:20:00 ok: [jrpcaioiad-2c] => (item=python-keystoneclient)
15:20:00 changed: [jrpcaioiad-2c] => (item=python-neutronclient)
15:20:00 changed: [jrpcaioiad-2c] => (item=python-novaclient)
15:20:00 ok: [jrpcaioiad-2c] => (item=python-memcached)
15:20:00 
````

In `package_install.yml` for MaaS. That, however, does not look like something we'd see from installing `swift` (which is the missing item in that list). This does, however, look like some kind of fact gathering we would have seen when trying to template out disk checks with MaaS. @andymcc do you have any ideas where this might be coming from?

**Comment by sigmavirus24** [comment]
> So the "swift" item in mass_pip_packages gets replaced with the swift dict from conf.d/swift global_overrides. I think this didn't happen previously because swift wasn't in the list of pip packages required by maas. 

Why is it there now?

File: rpcd/playbooks/roles/rpc_maas/defaults/main.yml (hunk 2 above)
Comment (prometheanfire) [inline rpcd/playbooks/roles/rpc_maas/defaults/main.yml]: this is my only concern with this

**Comment by mancdaz** [comment]
Because we need it on the host as it supplies swift-recon and
swift-dispersion-report for the monitoring plugins

File: rpcd/playbooks/roles/rpc_maas/templates/openrc [hunk 3]
```diff
@@ -4,7 +4,7 @@ export OS_ENDPOINT_TYPE={{ openrc_os_endpoint_type }}
 export OS_USERNAME={{ maas_keystone_user }}
 export OS_PASSWORD={{ maas_keystone_password }}
 export OS_TENANT_NAME={{ keystone_admin_tenant_name }}
-export OS_AUTH_URL={{ keystone_service_internalurl_v3 }}
+export OS_AUTH_URL={{ openrc_os_auth_url }}
```
prometheanfire: did we ever figure out why it couldn't use the keystone var?

rpcd/playbooks/roles/rpc_maas/tasks/package_install.yml
hughsaunders: I removed with_flattened because it expands strings as variables, so "swift" was getting replaced with the swift config dict - see above comment

hughsaunders:
:+1 it's passing!

rpcd/playbooks/roles/rpc_maas/templates/openrc
hughsaunders: The variable was renamed in OSA Liberty

hughsaunders:
check

hughsaunders:
recheck

mancdaz:
it's passing twice!

rpcd/playbooks/roles/rpc_maas/defaults/main.yml
mancdaz: in kilo, swift got installed on the host (as a pip package) by the os_swift role from osa.  I think this was probably a side effect more than anything on purpose. Now that things have moved into venvs, this no longer gets done, so we need to explicitly install it for maas.

See https://gist.github.com/mancdaz/6052ed1810ca1c521b41

rpcd/etc/openstack_deploy/user_variables.yml
```diff
@@ -62,3 +62,20 @@ ceph_mons: >
...
+pkg_locations:
+  - ../
+  - /etc/ansible/roles
+  - /etc/openstack_deploy
+  - /opt/rpc-openstack/rpcd
+
+#NOTE(mancdaz) there may be a better place to override this
+horizon_venv_tag: "{{ openstack_release }}"
+
+galera_pip_packages:
```
git-harry: It would be nice to have a comment here explaining why this has been overridden

rpcd/etc/openstack_deploy/user_variables.yml
```diff
@@ -62,3 +62,20 @@ ceph_mons: >
   {% endfor -%}
   {% endif -%}
   {{ _var }}
+
+# locations for the package builder to look for files that contain pip packages and git repos to build from
```
git-harry: Why did all this repo build stuff need to be done for liberty? The commit message doesn't give an explanation.

mancdaz:
recheck - image upload timeout

hughsaunders:
recheck - had to take jenkins down :(

rpcd/playbooks/roles/rpc_maas/tasks/package_install.yml
sigmavirus24: Please read the docs. The documentation shows that the proper way to do this is with what I posted. Granted it's more necessary for nested lists but my example should be the correct way of doing this and we were still seeing the swift config dict being interpolated into the dependencies even with your change (based on what Dave and I saw in the logs).

mancdaz:
recheck - jenkins is a liar

prometheanfire:
:+1:

rpcd/etc/openstack_deploy/user_variables.yml
prometheanfire: For package locations, osa is now importing external roles.

rpcd/playbooks/roles/rpc_maas/tasks/package_install.yml
prometheanfire: I'd be willing to leave this as is and have a follow on patch that changes it to what @sigmavirus24 suggested

    with_flattened:
      - "{{ maas_pip_packages }}"
      - "{{ maas_pip_dependencies }}"

sigmavirus24: So while the documentation says I'm right, this is actually a bug in ansible: https://github.com/ansible/ansible/issues/12249 It silently looks up strings in `with_flattened` but not `with_items` and will be fixed in a release "Soon" (post-2.0).

rpcd/playbooks/roles/rpc_maas/defaults/main.yml
sigmavirus24: Can we add a comment explaining this at least?

sigmavirus24:
> jenkins is a liar

Guess we can't trust the gate then, eh? :trollface:

sigmavirus24:
I'll be +1 if we add a comment explaining swift's purpose in the `maas_pip_packages` list.

rpcd/playbooks/roles/rpc_maas/defaults/main.yml
git-harry: So swift hosts are going to have swift installed in two places? That strikes me as rather confusing and creating the potential for bugs. I think it would be better to get the maas plugin(s) to use the venv's Python interpreter. Maybe a line in file task would suffice for now to achieve that with a new card to look at the more general case.

git-harry: So I see it's actually calling a number of swift Python scripts e.g. swift-recon. In that case the plugin can be adjusted to accept an additional optional argument specifying the venv's bin directory.

d34dh0r53:
@git-harry @nrb can you come up with alternatives to installing swift for MaaS, I would really like to get this merged ASAP so that we can continue on with the Libertification. If we can't come up with a timely solution I will merge this and we can fix it in the existing code.

git-harry:
@d34dh0r53 I commented in line an alternative method for fixing this - by passing the path to the venv's bin directory as an argument to the plugins.

cloudnull:
I will add here that the MaaS issues are not just swift related. Every single openstack service on a greenfield deployment of Liberty will be installed within a venv. This means that every single mass check that relies on a given client lib will need to have the ability to reach into the venv to run. Currently, when upgrading, if a MaaS checks are passing without the additional venv capabilities we're getting a false positive because old code had not been cleaned up on hosts or within a container. 

I don't think the MaaS issues should hold up the transfer of the "master" branch to use OSA liberty. By moving this forward we can enable other folks to begin more of the Liberty work and we can iterate on the MaaS issues at that time.

claco:
Likely also worth noting we need to retain the non-venv ability as well for RPC-R.

mancdaz:
I will also add that in my observation, actually only swift was missing on
the host for the swift plugins to work in liberty. Everything else was fine.

That said, this is supposed to be a 'minimum change needed to get things
going' so that other liberty work can continue. This works as it is right
now. I'd also agree that is should be merged as is, and other work/changes
can be built on top of it, knowing that they have to continue to pass the
gate.

rpcd/playbooks/roles/rpc_maas/templates/openrc
nrb: I seemed to have missed that when looking through variable name changes for the upgrade process.

Does this need to be renamed in the process of an upgrade?

prometheanfire: yes, it was renamed upstream

Issue #798, opened 2016-02-01
Verify MAAS should wait before checking alarm status


hughsaunders:
Currently the verify maas playbook loops until all checks & alarms are registered, then immediately checks that there are no active alarms. 

This doesn't make much sense as the alarm status will be checked before most of the checks have had time to run. 

---
- [x] master: #799 
- [x] kilo:

prometheanfire:
maybe even impacts gating, given some of the timeouts related to checking mass

prometheanfire:
please backport

PR #799, opened 2016-02-01
Pause before checking alarm status

hughsaunders:
The Verify-Maas playbook has two tasks:
1) Wait for all checks & alarms to be registered
2) Check that no alarms are triggered.

The problem is the step 2 happens before most of the checks have had
time to run. This patch introduces a two minute wait, to ensure all
checks have had time to run. (default check interval is 60s).

Related: #798

### DIFF
--- a/rpcd/playbooks/verify-maas.yml
+++ b/rpcd/playbooks/verify-maas.yml
@@ -30,6 +30,10 @@
       retries: 10
       delay: 60
 
+    - name: "Allow MAAS time to run checks before verifying their status"
+      pause:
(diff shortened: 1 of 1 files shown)


hughsaunders:
recheck

hughsaunders:
recheck

prometheanfire:
:+1: lgtm

tomc603:
I hate this, and I feel like we should be polling for results instead of blindly setting timeouts and waits all over the place. But, since this is the best we have for now. :+1:

Issue #814, opened 2016-02-09
RPC-AIO Doesn't need to test with duplicate containers.


hughsaunders:
The OSA gate uses container affinity > 1 to test replication code paths. We don't need to reproduce this in RPC. When we eventually have multi node gating, then we will again have multiple containers of each type and will be adding something else on top of OSA as OSA doesn't have multinode testing.

Currently most kilo RPC-AIO gate jobs are failing due to lsync not completing on time, I intend to fix that, but reducing container affinity will also make it go away. 

There is also a cirros image upload failure that may be related to multiple keystones, which may be prevented by reducing affinity.

prometheanfire:
when we have a multi-node gate then I'd be fine reducing coverate, until then I don't think reducing coverage is a good idea.  We do set some different values for DB stuff than upstream OSA for instance.

:-1: til we have working multinode

nrb:
While I can somewhat understand the argument that we don't have to retest large swaths of functionality, I feel like we should keep testing multiple containers as long as we can.

Is the problem that we lack resources on the AIOs we're building, or is it just getting around the lsync issue?

hughsaunders:
consensus is that we do need to carry on testing multiple containers where applicable, at least untill we have multinode gating.

PR #815, opened 2016-02-09
Set affinities to 1

hughsaunders:
This should prevent some failure cases (repo sync, keystone catalog
sync). It doesn't reduce test coverage as multiple containers are tested
in OSA.

Related: #814

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -16,7 +16,11 @@ export DEPLOY_SWIFT=${DEPLOY_SWIFT:-"yes"}
 export FORKS=${FORKS:-$(grep -c ^processor /proc/cpuinfo)}
 export ANSIBLE_PARAMETERS=${ANSIBLE_PARAMETERS:-""}
 export ANSIBLE_FORCE_COLOR=${ANSIBLE_FORCE_COLOR:-"true"}
-export BOOTSTRAP_OPTS=${BOOTSTRAP_OPTS:-""}
+
+# Use location that doesn't rely on or interfere with future bootstrapping
+BOOTSTRAP_VARS_FILE='/tmp/rpc_bootstrap_vars.yml'
+# add any bootstrap vars passed in via env to the vars file.
+echo ${BOOTSTRAP_OPTS:-} |sed 's/=/: /' >> $BOOTSTRAP_VARS_FILE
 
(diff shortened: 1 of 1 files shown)


prometheanfire:
Changes we make downstream cause me to believe it is not valid logic to trust upstream tests and reduce our coverage downstream.  It would be better to find the cause of the failures downstream and fix it here.  If upstream is not having these issues and we are worries me.

hughsaunders:
So setting all the affinities to 1 didn't work because maas checks that we have 3 containers for some things :( will try reducing keystone to see if it solves the cirros upload issue.

hughsaunders:
recheck - it passed, but I want to see if it passes again..

hughsaunders:
recheck - one more time

hughsaunders:
recheck failed on apt

hughsaunders:
recheck failed sub module checkout, have tweaked Jenkins job config

hughsaunders:
recheck

hughsaunders:
recheck apt fail

hughsaunders:
28 AIOs have run on this patch and none have failed on cirros image upload, so reducing the keystone container affinity seems to solve the problem.

odyssey4me:
If this solves the problem, then all this has done is isolate the root cause into a particular domain. The next step is going to have to be to figure out why this is an issue here, and not in OSA.

prometheanfire:
I agree, particularly as RPC-O and OSA diverge we should keep test coverage.  This was useful for narrowing down the problem though.

prometheanfire:
:-1: would like to test at least somewhat close to what we deploy, also, these don't seem to be causing failures (mostly cirros timeouts now)

d34dh0r53:
recheck - testing cirrros fail

nrb:
-1; I would rather we actually find the issue with multiple instances than simply reducing the affinity if at all possible.

Were this for resource constraint reasons, that might be more convincing for me.

hughsaunders:
I agree. Possible cause: keystone caching, specifically multiple memcaches?

prometheanfire:
given the closure of #814 I'm going to close this as well.  If we wish to continue discussing this we should probably do so in a bug.

prometheanfire:
I left the branch around to reopen, please close if you don't reopen.

PR #817, opened 2016-02-09
Remove incorrect run_queues test

prometheanfire:
The run_queues metric is load metric from the erlang environment, it
isn't directly related to rabbitmq message queues, though it is related
to node load, so a busy broker may have a higher run_queues.

However as this is a measure of load, it is not a good value for
checking replication across rabbit cluster members. This test has been
removed.

This patch also adds a couple of checks to ensure that all nodes are
clustered succesfully.

Descriptions of the variables that are used for the new tests:

Cluster_links:
A list of the other nodes in the cluster. For each node, there are
details of the TCP connection used to connect to it and statistics on
data that has been transferred.

Partitions:
List of network partitions this node is seeing.

From: https://github.com/rabbitmq/rabbitmq-management/blob/626f857dcab1532d6a7006cd7cd16aa824445969/priv/www/doc/stats.html#L653

Related: #626
(cherry picked from commit 99052a15535637415ab9098e18bac2735ee771ef)

### DIFF
--- a/maas/plugins/rabbitmq_status.py
+++ b/maas/plugins/rabbitmq_status.py
@@ -146,23 +146,14 @@ def _get_node_metrics(session, metrics, host, port, name):
             status_err('cluster too small')
         if not is_cluster_member:
             status_err('{0} not a member of the cluster'.format(name))
+        if sum([len(n['partitions']) for n in response]):
+            status_err('At least one partition found in the rabbit cluster')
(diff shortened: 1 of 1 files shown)


hughsaunders:
recheck - kilo now has a better chance of passing

tomc603:
This looks okay to me :+1:

d34dh0r53:
:+1:

PR #824, opened 2016-02-11
Add retry & force to support apt installs

hughsaunders:
So far 29 RPC-AIO gate jobs have failed on the rpc support package
install task. Usually most containers succeed but one or two fail. This
patch increases the chances of success by retrying and adding force so
that unauthenticated packages can be installed.

### DIFF
--- a/rpcd/playbooks/roles/rpc_support/tasks/support_preinstall.yml
+++ b/rpcd/playbooks/roles/rpc_support/tasks/support_preinstall.yml
@@ -19,6 +19,11 @@
     state: present
     update_cache: yes
     cache_valid_time: 600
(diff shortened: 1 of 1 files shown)


hughsaunders:
recheck failed on ssh

hughsaunders:
recheck start rsyslog fail

hughsaunders:
recheck

hughsaunders:
This has now run through 12 AIOs, none of which have failed on apt. So this seems to solve that problem.

d34dh0r53:
:+1:

prometheanfire:
ya, :+1: here, nice :D

Issue #186, opened 2015-06-21
megacli


albertcard:
Looking to see if we can get the megacli utility added to our builds as we are having more and more customers being built into CDC without HP or Dell gear. Having this utility would help us with investigation into Drives/Controllers/Raid Arrays/PSU/etc issues and diagnosing issues a little bit quicker. 

http://hwraid.le-vert.net/wiki/DebianPackages

## Branches

- [x] Master #840
- [x] Liberty - >=12.0.1
- [x] Kilo

nrb:
Can I ask how many customers this affects? Not names, just a count.

albertcard:
Looks like we have 28 customers with CDC environments -- This would help with examination of hardware in their chassis/blades. However, we've also found a bit of useful use cases on drive errors -- more verbose than 'omreport' -- for all of our customers in using Dell Chassis which covers all of our customers in house (72 customers)

rackertom:
@albertcard Are all of these environments Debian-based (Debian, Ubuntu, Mint, etc.) or are there also RPM-based distributions (RHEL, SuSE, and the like)?

BjoernT:
+1 we should add this to the rpc_support role and also add the lsi utils. Those are common controllers

prometheanfire:
to be clear, liberty here is 12.0.1

PR #840, opened 2016-02-17
Install HWRaid packages as part of rpc_support

tomc603:
Support has requested that RPC installations receive the most common set of hardware RAID card utilities for diagnostic purposes beyond what is currently installed by default with the OS. This PR introduces two task files which add the RAID utilities package repo and install the `megaraid` and `lsiutil` packages by default.

The utilities installed by default are the two most common for RPC installations, but can be overridden with the variable `hwraid_apt_packages`.

**NOTE** As of right now, these plays are not running as part of the rpc_support role. I would like support to test this in several scenarios before we made the decision to change this. I'm worried that there may be situations where detecting the presence of a tool triggers checks against a RAID card that may not be present. If that is the case, then we will simply require that support override the packages variable at run time, otherwise no packages will be installed at all.

Partial-Fix: #186

### DIFF
--- a/rpcd/playbooks/roles/rpc_support/defaults/main.yml
+++ b/rpcd/playbooks/roles/rpc_support/defaults/main.yml
@@ -72,3 +72,15 @@ holland_packages:
 
 holland_pip_dependencies:
   - MySQL-python
+
+hwraid_apt_repo_url: "http://hwraid.le-vert.net/ubuntu"
+
+hwraid_apt_repos:
+  - { repo: "deb {{ hwraid_apt_repo_url }} {{ ansible_lsb.codename }} main", state: "present" }
+
+hwraid_apt_keys:
+  - { url: "http://hwraid.le-vert.net/debian/hwraid.le-vert.net.gpg.key", state: "present" }
+
+hwraid_apt_packages:
+  - megacli
+  - lsiutil
--- a/rpcd/playbooks/roles/rpc_support/tasks/main.yml
+++ b/rpcd/playbooks/roles/rpc_support/tasks/main.yml
@@ -27,6 +27,14 @@
 
 - include: support_preinstall.yml
 
+- include: raid_preinstall.yml
+  when: >
+    inventory_hostname in groups['hosts']
+
+- include: raid_install.yml
+  when: >
+    inventory_hostname in groups['hosts']
+
 - include: bashrc.yml
 
 - include: histformat.yml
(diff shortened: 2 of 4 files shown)


prometheanfire:
glad to see it has keys, seems fine at first glance

prometheanfire:
:+1:

git-harry:
-1 I'm not clear how this is expected to be used. Assuming the implementation is good, I would say the commit needs a more detailed message and documentation.

tomc603:
@git-harry Thanks for the feedback. I've updated the message. Mind giving it a look?

git-harry:
@tomc603 these are task files not plays/playbooks. How are support expected to run this code when they don't link to main? As it currently is I believe support would need to write their own playbook to make use of this code. What are they testing it on? It sounds like you're trying to introduce an experimental feature, I'm not sure we've taken that approach before. The pull request message has been updated but not the commit message.

tomc603:
@git-harry Sorry, I'm constantly using the wrong words when it comes to tasks inside roles vs plays vs everything. Fixed.

This isn't experimental, support currently use these utilities. Where my concern lies is in scripts that individuals may have which are "too smart"- They check for the presence of a utility and presume that it being there means the hardware it supports is also present.

I suppose this is just me being too gun shy. I'll edit this commit so that the tasks run and the packages are installed. The issue this addresses did request the two most common utilities, so this really is just a case of me overthinking the problem and trying to solve a situation which may not even exist.

git-harry:
@tomc603 ah, okay I understand now. Yeah, I wouldn't over-engineer this. Support have asked for these packages. If it introduces a bug we can deal with it then.

tomc603:
Ok, I think I un-broke the reflog on my side, so everything looks good. @git-harry I've added changes to the `tasks/main.yml` file to execute the tasks in the `raid_preinstall.yml` and `raid_install.yml` files as we discussed. Also cleaned up some of the task names, limited execution to hosts only, and squashed the commits so the PR is clean.

d34dh0r53:
:+1:

tomc603:
@BjoernT Does this look good for you? Just want support eyes on it to be sure.

prometheanfire:
@BjoernT ping, just need an ack before I push the button

BjoernT:
Generally ok but how about hosting the tools on the Rackspace mirrors ?

prometheanfire:
I think that's a change that'll be needed down the line, merging

PR #843, opened 2016-02-18
force removal of aodh and ceilometer conf files

### LINKED ISSUE #842: r12.0.0rc1 - deploy.sh - rm: cannot remove ‘/etc/openstack_deploy/conf.d/aodh.yml’: No such file or directory
These conf.d files don't exist in OSA or RPC....

https://github.com/rcbops/rpc-openstack/blob/master/scripts/deploy.sh#L115-L116

```
rm: cannot remove ‘/etc/openstack_deploy/conf.d/aodh.yml’: No such file or directory 
```

- [x] Master (https://github.com/rcbops/rpc-openstack/pull/843)
- [x] liberty-12.0 (https://github.com/rcbops/rpc-openstack/pull/844)

prometheanfire:
exits 0 when the file is not found, which is just fine here

fixes #842

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -10,7 +10,7 @@ export DEPLOY_OA=${DEPLOY_OA:-"yes"}
 export DEPLOY_ELK=${DEPLOY_ELK:-"yes"}
 export DEPLOY_MAAS=${DEPLOY_MAAS:-"no"}
 export DEPLOY_TEMPEST=${DEPLOY_TEMPEST:-"no"}
-export DEPLOY_CEILOMETER=${DEPLOY_CEILOMETER:-"no"}
+export DEPLOY_CEILOMETER="no"
 export DEPLOY_CEPH=${DEPLOY_CEPH:-"no"}
 export DEPLOY_SWIFT=${DEPLOY_SWIFT:-"yes"}
 export FORKS=${FORKS:-$(grep -c ^processor /proc/cpuinfo)}
@@ -99,6 +99,8 @@ if [[ "${DEPLOY_AIO}" == "yes" ]]; then
   then
     rm /etc/openstack_deploy/conf.d/swift.yml
   fi
+  rm -f /etc/openstack_deploy/conf.d/aodh.yml
+  rm -f /etc/openstack_deploy/conf.d/ceilometer.yml
 fi
 
 
@@ -109,13 +111,6 @@ fi
 # Apply any patched files.
 cd ${RPCD_DIR}/playbooks
 openstack-ansible -i "localhost," patcher.yml
-
-# disable aodh and ceilometer
-export DEPLOY_CEILOMETER=no
-rm /etc/openstack_deploy/conf.d/aodh.yml
-rm /etc/openstack_deploy/conf.d/ceilometer.yml
-rm /etc/openstack_deploy/env.d/aodh.yml
-rm /etc/openstack_deploy/env.d/ceilometer.yml
 
 # begin the openstack installation
 if [[ "${DEPLOY_OA}" == "yes" ]]; then
--- a/scripts/upgrade.sh
+++ b/scripts/upgrade.sh
@@ -26,13 +26,6 @@ cp ${RPCD_DIR}/etc/openstack_deploy/user_variables.yml /tmp/upgrade_user_variabl
 ${BASE_DIR}/scripts/update-yaml.py /tmp/upgrade_user_variables.yml /etc/rpc_deploy/user_variables.yml
 mv /tmp/upgrade_user_variables.yml /etc/rpc_deploy/user_variables.yml
 
-# disable aodh and ceilometer
-export DEPLOY_CEILOMETER=no
-rm /etc/openstack_deploy/conf.d/aodh.yml
-rm /etc/openstack_deploy/conf.d/ceilometer.yml
-rm /etc/openstack_deploy/env.d/aodh.yml
-rm /etc/openstack_deploy/env.d/ceilometer.yml
-
 # Upgrade Ansible in-place so we have access to the patch module.
 cd ${OA_DIR}
 


nrb:
-1;

After reviewing the environment, I don't think this is actually necessary at all - if the groups are not specified in conf.d, they won't be created. So there's no need to remove anything.

This was my mistake in proposing it in the first place.

prometheanfire:
Are you stating that we should check for the file instead of forcing the removal?

nrb:
No - I'm saying we shouldn't need to remove anything (for multiple nodes), since the conf.d files are not copied in place by default - see https://github.com/openstack/openstack-ansible/tree/liberty/etc/openstack_deploy/conf.d.

On AIOs they might be placed there, depending on how overrides are set (https://github.com/openstack/openstack-ansible/blob/liberty/tests/roles/bootstrap-host/tasks/prepare_aio_config.yml#L35-L38)

prometheanfire:
So, that section should be moved to be within the aio check?

nrb:
I don't believe so. We disable Ceilometer twice in the script, regardless of whether it's an AIO or not. https://github.com/rcbops/rpc-openstack/blob/master/scripts/deploy.sh#L13 defaults it to no.

nrb:
Essentially these can be treated in the exact same way as swift; if we enable it, then it'll be there, if not, there shouldn't be anything to touch.

git-harry:
Echoing some of the comments already made, I think this section does make more sense in the aio and, similar to swift, only the conf files should need deleting. I don't think conf overrides with the bootstrap-host role can be used because I don't believe there is a way to delete a key.

I think 'right' approach is a bit confusing because counterintuitively, the deploy script has a section for deleting the env files for elasticsearch, logstash and kibana.

nrb:
Some examples of OSA liberty from a multinode install:

Liberty branch deployed
```
root@605086-infra01:/opt/openstack-ansible# git branch
* liberty
  master
```

Ceilometer and aodh defined in env.d, but not conf.d 

```
root@605086-infra01:/opt/openstack-ansible# ls /etc/openstack_deploy/env.d
ceilometer.yml  extra_container.yml.example  glance.yml  horizon.yml  keystone.yml  neutron.yml  os-infra.yml  rabbitmq.yml  shared-infra.yml  swift.yml
cinder.yml      galera.yml                   heat.yml    infra.yml    memcache.yml  nova.yml     pkg_repo.yml  rsyslog.yml   swift-remote.yml  utility.yml
root@605086-infra01:/opt/openstack-ansible# ls /etc/openstack_deploy/conf.d
ceilometer.yml.aio  ceilometer.yml.example  swift-remote.yml.sample  swift.yml.aio  swift.yml.example
```

No ceilometer listed in the inventory via the management script
```
root@605086-infra01:/opt/openstack-ansible/scripts# ./inventory-manage.py --file /etc/openstack_deploy/openstack_inventory.json --list-groups | grep ceilometer
root@605086-infra01:/opt/openstack-ansible/scripts#
```

No ceilometer containers found across the hosts:
```
root@605086-infra01:/opt/openstack-ansible/playbooks# ansible hosts:is_metal -m shell -a "lxc-ls | grep ceilometer"
605086-infra01 | FAILED | rc=1 >>


605087-infra02 | FAILED | rc=1 >>


605093-storage01 | FAILED | rc=1 >>


605095-storage03 | FAILED | rc=1 >>


605094-storage02 | FAILED | rc=1 >>


605088-infra03 | FAILED | rc=1 >>


605089-logging01 | FAILED | rc=1 >>


605091-compute02 | FAILED | rc=1 >>


605092-compute03 | FAILED | rc=1 >>


605090-compute01 | FAILED | rc=1 >>
```

Verifying nova containers just to show the above command wasn't broken.
```
root@605086-infra01:/opt/openstack-ansible/playbooks# ansible hosts:is_metal -m shell -a "lxc-ls | grep nova"
605093-storage01 | FAILED | rc=1 >>


605086-infra01 | success | rc=0 >>
605086-infra01_nova_api_metadata_container-7d430e35
605086-infra01_nova_api_os_compute_container-d0e823d9
605086-infra01_nova_cert_container-71641fdd
605086-infra01_nova_conductor_container-fddfa95b
605086-infra01_nova_console_container-d94d22e9
605086-infra01_nova_scheduler_container-6a954e9b

605095-storage03 | FAILED | rc=1 >>


605087-infra02 | success | rc=0 >>
605087-infra02_nova_api_metadata_container-a9c2d782
605087-infra02_nova_api_os_compute_container-6095fc84
605087-infra02_nova_cert_container-cdd1c5e8
605087-infra02_nova_conductor_container-6e86b1d7
605087-infra02_nova_console_container-f010beb3
605087-infra02_nova_scheduler_container-533c4cd8

605094-storage02 | FAILED | rc=1 >>


605089-logging01 | FAILED | rc=1 >>


605088-infra03 | success | rc=0 >>
605088-infra03_nova_api_metadata_container-47494efd
605088-infra03_nova_api_os_compute_container-8279ff46
605088-infra03_nova_cert_container-67c4d5b3
605088-infra03_nova_conductor_container-5f546103
605088-infra03_nova_console_container-d9e0f82a
605088-infra03_nova_scheduler_container-485f0641

605091-compute02 | FAILED | rc=1 >>


605092-compute03 | FAILED | rc=1 >>


605090-compute01 | FAILED | rc=1 >>
```

prometheanfire:
Pretty sure I tried deleting just the conf files, but could have been just the env.d files.  I'll try again.

scripts/deploy.sh
```diff
@@ -99,6 +99,8 @@ if [[ "${DEPLOY_AIO}" == "yes" ]]; then
   then
     rm /etc/openstack_deploy/conf.d/swift.yml
   fi
+  rm -f /etc/openstack_deploy/conf.d/aodh.yml
+  rm -f /etc/openstack_deploy/conf.d/ceilometer.yml
```
nrb: Thanks for the update, these two lines look more along the lines of what I'd expect (that is, an optional service being removed from an AIO in the same way Swift might be). The difference between these and Swift is that we don't explicitly support them yet. When that time comes, they can be wrapped in a conditional.

scripts/deploy.sh
```diff
@@ -10,7 +10,7 @@ export DEPLOY_OA=${DEPLOY_OA:-"yes"}
 export DEPLOY_ELK=${DEPLOY_ELK:-"yes"}
 export DEPLOY_MAAS=${DEPLOY_MAAS:-"no"}
 export DEPLOY_TEMPEST=${DEPLOY_TEMPEST:-"no"}
-export DEPLOY_CEILOMETER=${DEPLOY_CEILOMETER:-"no"}
+export DEPLOY_CEILOMETER="no"
```
nrb: I think I get why this was done - so that any callers of this script like a gate check, don't accidentally try to test aodh/ceilometer when the configs are deleted.

This solution works for me, but will mean another edit if/when ceilometer is supported.

prometheanfire: such is the nature of our work lol

nrb:
This new commit looks much better, and gate logs show that ceilometer/aodh aren't being made. Thanks.

I'm also comfortable with this not necessarily receiving multinode tests before merge because it does nothing to change the standard OSA behavior (which is documented in https://github.com/rcbops/rpc-openstack/pull/843#issuecomment-187339894) for the env.d/conf.d files in that scenario. That said, covering the bases before a QE hand off won't hurt, either.

So, +1.

scripts/deploy.sh
stevelle: Until it is supported this seems like the right behavior.

nrb: Agreed on both counts, just pointing it out for posterity.

stevelle:
offering only a +0.5 vote on this, simply because I'm not working in RPC much right now

scripts/deploy.sh
```diff
@@ -99,6 +99,8 @@ if [[ "${DEPLOY_AIO}" == "yes" ]]; then
   then
     rm /etc/openstack_deploy/conf.d/swift.yml
```
byronmccollum: `rm -f` for consistency...?...even though the `-f` is technically not needed anymore, since these files are being added as part of the AIO host-bootstrap role...

prometheanfire: Ya, that could go here, it's not a problem right now though, so I'd like to keep it as is (since it passed, etc).

git-harry:
-1 The code looks fine, the commit message could do with some additional work. Anyone looking at it would struggle to understand why this change is being made.

prometheanfire:
@git-harry updated commit message

git-harry:
@prometheanfire thanks for taking another stab at the commit message. I think it is still a bit light on detail, I believe the important thing about a commit message is to include the what and the why. Something like the following is more helpful to someone in the future:
```
The ability to override the value of DEPLOY_CEILOMETER is removed. This variable is used by openstack-ansible scripts, currently Ceilometer is not supported by rpc-o and so its value should always be 'no'.

The removal of the ceilometer.yml and aodh.yml conf.d files is moved to the AIO section, these files are added by the openstack-ansible bootstrap-aio process which is not run for a customer deployment. Removing these files means there are no hosts that are members of the ceilometer aodh Ansible groups.

The env.d files are not removed, these files define the Ansible group structure but have no effect without the corresponding conf.d files.

'rm' is changed to 'rm -f' so that the command does not error if the file is missing.

There is no upgrade-specific requirement, beyond what deploy.sh does, in terms of Ceilometer and Aodh and so the section has been removed from the upgrade script.
```

prometheanfire:
@git-harry your commit message is there verbatim now.

prometheanfire:
recheck - 17:20:23 glanceclient.exc.HTTPException: 504 Gateway Time-out: The server didn't respond in time. (HTTP N/A)

nrb:
Still +1, thanks for the detailed commit message.

stevelle:
Upgrading my +0.5 to a :+1: 
Not going to push merge though. This needs to wait for gating.

tomc603:
This is all good stuff, and I've confirmed it worked fine on my last multi-node gate. Just noting it needs backport to liberty-12.0 branch.

nrb:
recheck

In RPC-AIO-Ceph
```
0:52:40 TASK: [Verify Checks & Alarms are registered] ********************************* 
20:52:40 failed: [jrpcaioiad-38e] => {"attempts": 10, "changed": false, "failed": true, "failed_when_result": true, "rc": 1, "stdout_lines": ["--- 2016-02-24 20:52:40.392617 ---", "The following checks have config files but are not registered with the maas api: ['lb_api_check_heat_cloudwatch--unspecified_lb'] ", "The following alarms are present in config files but are not registered with the maas api: ['lb_api_alarm_heat_cloudwatch'] "]}
```

Issue #845, opened 2016-02-19
rpc-support: Unnecessary host packages should be removed for old deployments 


BjoernT:
Starting with commit c6798eb778621819e13f0334778ff2d55d7ff2b1 we do not install the host packages on containers anymore but a remove task is missing to remove them from the old containers

- [ ] kilo

prometheanfire:
this was kilo only, just docing that ehre

prometheanfire:
closed based on PR and asking @BjoernT in irc

PR #846, opened 2016-02-19
rpc_support: Remove old host packages from containers

BjoernT:
Closes-Bug: #845


prometheanfire:
How long will this task be around?  How long should cleanup tasks like this stick around?

BjoernT:
I would guess it will be around for quite some time, probably for the lifetime of kilo and we can remove of for liberty

prometheanfire:
If that's the case, please close this bug and reopen one against kilo, master is liberty. :D

PR #853, opened 2016-02-22
rpc_support: Remove old host packages from containers

BjoernT:
Closes-Bug: #845

### DIFF
--- a/rpcd/playbooks/roles/rpc_support/tasks/support_preinstall.yml
+++ b/rpcd/playbooks/roles/rpc_support/tasks/support_preinstall.yml
@@ -33,3 +33,19 @@
   when: inventory_hostname in groups['hosts']
   tags:
     - support_packages
+
+- name: Remove host packages from containers
+  apt:
+    pkg: "{{ item }}"
+    state: absent
+    update_cache: yes
+    cache_valid_time: 600
+  with_items:
+    - linux-crashdump
+    - ntp
+    - ntpdate
+    - linux-headers*
+  when: inventory_hostname in groups['all_containers']
+  tags:
+    - support_packages
+    - remove_host_packages_from_containers


prometheanfire:
:+1:

rpcd/playbooks/roles/rpc_support/tasks/support_preinstall.yml
```diff
@@ -33,3 +33,20 @@
...
     - support_packages
+
+- name: Remove host packages from containers
+  apt:
+    pkg: "{{ item }}"
+    state: absent
+    update_cache: yes
+    cache_valid_time: 600
+  with_items:
+    - debbootstrap
```
git-harry: The package name should be 'debootstrap'. This package seems to be a requirement for the repo containers.

git-harry:
-1 this type of cleanup action has the potential to cause issues. This task is one of the last to run and so has the potential to impact on services by removing packages that are required. While I agree with the intent of this commit the action is too far removed from the playbooks that are written to configure those containers. I think it would be safer to replace the containers than have a separate task blindly modifying them. This task creates an ongoing risk because if any of those containers have one of the listed packages introduced, the rpc-support role will remove them.

rpcd/playbooks/roles/rpc_support/tasks/support_preinstall.yml
BjoernT: @git-harry Thanks fixed that. But I can't find this requirement, where did you find it ?

git-harry: ansible all_containers -a 'dpkg -l debootstrap'

BjoernT: Well we even install tgtd inside the cinder api container. Our package management is not good.
Also debootstrap is not listed anywhere inside OSA.
It seems we to install lxc inside an lxc container which pulls lxc-template which pulls debootstrap so fix OSA first

prometheanfire:
have you tested this (just by running though an ansible shell manually)?

BjoernT:
yes it removes the packages just fine. Now I have to look why we install lxc, lxc-dev inside containers with https://github.com/openstack/openstack-ansible-repo_server/blob/master/defaults/main.yml#L74

BjoernT:
After talking to Kevin, removing debootstrap for now since repo_server was intended to build lxc images but never did

prometheanfire:
this does need a better commit message, preferably mentioning the manual testing that took place along with proof (of some sort) that each of the removed packages won't be installed again like de-bootstrap would have been.

BjoernT:
@prometheanfire This is overkill since we put test protocols in the commit message. It's not worth the effort here, closing it.

Issue #861, opened 2016-02-24
New nova db pool seetings needed


BjoernT:
I just realized that 
https://github.com/rcbops/rpc-openstack/blob/master/rpcd/etc/openstack_deploy/user_variables.yml

does not include the nova DB settings, which needs to be added ASAP

- [x] [Master](https://github.com/rcbops/rpc-openstack/pull/862)
- [x] [liberty-12.0.1](https://github.com/rcbops/rpc-openstack/pull/886)
- [x] [Kilo](https://github.com/rcbops/rpc-openstack/pull/887)

prometheanfire:
@BjoernT what branches should this target and any specific releases?

BjoernT:
all active ones, back to kilo

prometheanfire:
what is the urgency of this?

BjoernT:
We missed it and no larger deployment does function without increased settings

prometheanfire:
thanks, the issue helps in our prioritization.

BjoernT:
yeah if RPC12 would be closer I would say not back port necessary but we will still build cluster with RPC11

PR #862, opened 2016-02-24
Adding missing nova db settings

BjoernT:
This fix will increase nova DB settings we have determined for RPC-O.
The setting is similar to the neutron component, at least to start with.

Closes-Bug: #861


prometheanfire:
Looks good, mind updating the commit message as to why these keys are set to those values?

git-harry:
@BjoernT I agree with @prometheanfire adding new defaults is fine as long as the commit message clearly explains the reasoning. At the moment the message is misleading because the options are described as missing.

BjoernT:
I prefer to keep them separate even though they start with similar values

tomc603:
This looks reasonable. :+1:

prometheanfire:
:+1: @BjoernT please backport to kilo and liberty-12.0

Issue #836, opened 2016-02-16
cinder_volume_check Does Not Respect Multiple Backends


seancarlisle:
This was encountered on r11.0.1. When an environment has multiple Cinder back ends (such as ceph and xtremio) the cinder_volume_check may not fire if only one of the back ends is down.

Steps to reproduce:

* Setup an environment with multiple Cinder back ends such that the result is something like the following:

```
root@infra01:~# cinder service-list
+------------------+----------------------------------------------------------+------+---------+-------+----------------------------+-----------------+
|      Binary      |                           Host                           | Zone |  Status | State |         Updated_at         | Disabled Reason |
+------------------+----------------------------------------------------------+------+---------+-------+----------------------------+-----------------+
| cinder-scheduler |           infra01_cinder_scheduler_container             | nova | enabled |   up  | 2016-02-15T23:39:57.000000 |       None      |
| cinder-scheduler |           infra02_cinder_scheduler_container             | nova | enabled |   up  | 2016-02-15T23:39:53.000000 |       None      |
| cinder-scheduler |           infra03_cinder_scheduler_container             | nova | enabled |   up  | 2016-02-15T23:39:54.000000 |       None      |
|  cinder-volume   |         infra01_cinder_volumes_container@ceph            | nova | enabled |   up  | 2016-02-15T23:39:52.000000 |       None      |
|  cinder-volume   |        infra01_cinder_volumes_container@xtremio          | nova | enabled |   up  | 2016-02-15T23:39:52.000000 |       None      |
|  cinder-volume   |         infra02_cinder_volumes_container@ceph            | nova | enabled |   up  | 2016-02-15T23:39:53.000000 |       None      |
|  cinder-volume   |        infra02_cinder_volumes_container@xtremio          | nova | enabled |   up  | 2016-02-15T23:39:56.000000 |       None      |
|  cinder-volume   |         infra03_cinder_volumes_container@ceph            | nova | enabled |   up  | 2016-02-15T23:39:55.000000 |       None      |
|  cinder-volume   |        infra03_cinder_volumes_container@xtremio          | nova | enabled |   up  | 2016-02-15T23:39:55.000000 |       None      |
+------------------+----------------------------------------------------------+------+---------+-------+----------------------------+-----------------+
```

* Force a situation where one of the back ends is unable to check in with cinder api, but the other can.
* It looks like the check will only compare the first "cinder-volume_check" from the metrics gathered. If this happens to be the service that is still checking in, an alert will never fire.

[...]

PR #863, opened 2016-02-25
Add multiple backend support to cinder_service_check.py

claco:
When the host argument is supplied and the host has multiple cinder
backends configured, duplicate status metrics are returned. MaaS
currently only checks the last value during alarm evaluation. If the
last backend metric is 1 (up) but previous backend metrics are 0 (down),
no MaaS alert will be fired.

This change addresses this by evaluating all services for a host first,
and then aggregating the results into one status metric.

Additionally, each backend on the host now has its own status metric to
aid in troubleshooting and allow for alarms to be set on individual
backends.

Issue #836

### DIFF
--- a/maas/plugins/cinder_service_check.py
+++ b/maas/plugins/cinder_service_check.py
@@ -72,17 +72,32 @@ def check(auth_ref, args):
         status_err('No host(s) found in the service list')
 
     status_ok()
-    for service in services:
-        service_is_up = True
-        if service['status'] == 'enabled' and service['state'] != 'up':
-            service_is_up = False
 
-        if args.host:
-            name = '%s_status' % service['binary']
-        else:
+    if args.host:
+        all_services_are_up = True
+
+        for service in services:
+            service_is_up = True
+
+            if service['status'] == 'enabled' and service['state'] != 'up':
+                service_is_up = False
+                all_services_are_up = False
+
+            if '@' in service['host']:
+                [host, backend] = service['host'].split('@')
+                name = '%s-%s_status' % (service['binary'], backend)
+                metric_bool(name, service_is_up)
+
+        name = '%s_status' % service['binary']
+        metric_bool(name, all_services_are_up)
+    else:
+        for service in services:
+            service_is_up = True
+            if service['status'] == 'enabled' and service['state'] != 'up':
+                service_is_up = False
+
             name = '%s_on_host_%s' % (service['binary'], service['host'])
-
-        metric_bool(name, service_is_up)
+            metric_bool(name, service_is_up)
 
 
 def main(args):


claco:
Builds failed that maas check when the host returned in the services list has no @ at all (354d7f5e-block1), which wasn't a case I was able to test for in the heat stacks I'd run. Adjusting that `split('@')` and re pushing.

claco:
Gate failures unrelated. Something is afoot with SQL-Alchemy in pip land:  http://lists.openstack.org/pipermail/openstack-dev/2016-February/087486.html

```
error in SQLAlchemy-Utils setup command: 'extras_require' must be a dictionary whose values are strings or lists of strings containing valid project/version requirement specifiers.
```

prometheanfire:
recheck - let's see if they fixed gate

sigmavirus24:
I think setuptools released a fixed version. I saw a commit related to this fly by in IRC this morning. The bug was not in sqlalchemy-utils or openstack but in setuptools.

sigmavirus24:
What actually changed is that (as Jason mentions here: https://bitbucket.org/pypa/setuptools/issues/502/packaging-164-does-not-allow-whitepace) he pulled the 20.2.1 release from PyPI. This means that we no longer pull it down.

prometheanfire:
python looks fine, and AIO built/ran meaning maas was tested :+1:

sigmavirus24:
> AIO built/ran meaning maas was tested

There's no test that these metrics were generated/reported to maas though. We'll need a real lab to do that that has cinder configured with *multiple* backends. Since this is meant to fix such a bug, I think relying on the AIO alone in this case is suboptimal.

claco:
What's the recommendation for these types of things where's it's a customer edge case? @seancarlisle  might be able to test it for confirmation. Gating aside, I'm also more than happy to spin a setup w/ MaaS that has multiple backends for some eyeball verifications if that helps.

prometheanfire:
it'd help for the feel goods I guess

sigmavirus24:
We could also add a wishlist item to add actual unit testing to these scripts. I think spinning up labs and asking others to do verification for us is rather inefficient and a poor quality test. That said, I doubt we have time to write some automated testing of these plugins.

maas/plugins/cinder_service_check.py
```diff
@@ -72,17 +72,32 @@ def check(auth_ref, args):
...
+            if service['status'] == 'enabled' and service['state'] != 'up':
+                service_is_up = False
+                all_services_are_up = False
+
+            if '@' in service['host']:
+                [host, backend] = service['host'].split('@')
+                name = '%s-%s_status' % (service['binary'], backend)
+                metric_bool(name, service_is_up)
+
+        name = '%s_status' % service['binary']
```
git-harry: This is making an assumption that service['binary'] is the same across all services. This is probably a safe; the scheduler and volume services used to exist on the same host but that should no longer be the case. I just wanted to point it out incase anyone can think of potential issues.

claco:
I'm going to stand up a stack today for @seancarlisle to take a gander at this change.

sigmavirus24:
I was able to test this on @claco's stack and it works as expected. Just want to make sure that @seancarlisle is okay with the names that @claco is generating for the checks then I will merge-y merge-y

prometheanfire:
@sigmavirus24 is that your :+1: on this?

sigmavirus24:
@prometheanfire you replied to this after normal 9-5. ;)

That's a +0.5. I think @seancarlisle should contribute the other 0.5

prometheanfire:
this is to master, therefore 12.1.0 right now

seancarlisle:
The code looks alright. My only comment would be that the else section will never generate an alert because it will pass a metric for "cinder-volume_on_host_blah." Was this meant as more of a local test?

claco:
@seancarlisle I guess that's the nebulous part. I don't think we never had any alarms on that no-host-passed scenario.

```
https://github.com/rcbops/rpc-openstack/blob/master/rpcd/playbooks/roles/rpc_maas/templates/cinder_volume_check.yaml.j2

https://github.com/rcbops/rpc-openstack/blob/master/rpcd/playbooks/roles/rpc_maas/templates/cinder_scheduler_check.yaml.j2
```

I'm pretty sure both of those pass host, and I don't see and alarms configured that don't pass host.

```
https://github.com/rcbops/rpc-openstack/search?utf8=%E2%9C%93&q=cinder_service_check.py
```

Good catch though.

claco:
Just to clarify for those following, I forgot that the scheduler check uses the scheduler container ip as the host are, so the output is as expected. The alarm is still green.

image

prometheanfire:
once it's confirmed this is intended behavior then merge

@BjoernT ?

BjoernT:
:+1:

sigmavirus24:
@BjoernT confirmed this independently.

Issue #864, opened 2016-02-25
r12.0.0rc1 - Standalone Swift Installation Fails


byronmccollum:
Standalone Swift installation is not possible with v12.0.

Standalone Swift installations do not need RabbitMQ, and deployers are expected to set its affinity to 0 to prevent its installation. However, upstream OSA has a bug were RabbitMQ's presence is assumed, and fails when trying to configure the Keystone service's messaging user and vhost in RabbitMQ.

A review is in flight to correct this, and would need to be incorporated as part of a SHA bump.

https://review.openstack.org/#/c/284255/

An obvious work around would be to set the affinity to something > 0, which allows the installation to continue. The problem with this approach is that MaaS verification then fails, as the checks were never designed to monitor an unused RabbitMQ environment.

```
failed: [e13ce1f6-infra1] => {"changed": true, "rc": 1}
stderr: OpenSSH_6.6.1, OpenSSL 1.0.1f 6 Jan 2014
debug1: Reading configuration data /etc/ssh/ssh_config
debug1: /etc/ssh/ssh_config line 19: Applying options for *
debug1: auto-mux: Trying existing master
debug1: mux_client_request_session: master session id: 2
Shared connection to 172.29.236.1 closed.

stdout: --- 2016-02-25 00:11:08.440042 ---
Checks and Alarms with failures:
 - Entity:eniwgtj4rR:e13ce1f6-infra1
   ! Check:agent.plugin:chpIZrVjgG:rabbitmq_status--e13ce1f6-infra1_rabbit_mq_container-70a33109: Completed:False Status:error ptions.host, options.port)\n  File "/usr/lib/rackspace-monitoring-agent/plugins/rabbitmq_status.py", line 115, in _get_connection_metrics\n    max_chans = max(connection['channels'] for connection in response\nValueError: max() arg is an empty sequen
     ! Alarm:alBdqIaZgB:rabbitmq_proc_used_alarm_status--e13ce1f6-infra1_rabbit_mq_container-70a33109:CRITICAL
     ! Alarm:alCvCEt7lm:rabbitmq_mem_alarm_status--e13ce1f6-infra1_rabbit_mq_container-70a33109:CRITICAL
     ! Alarm:alLgFD4XNy:rabbitmq_qgrowth_excl_notifications--e13ce1f6-infra1_rabbit_mq_container-70a33109:CRITICAL
[...]

Issue #852, opened 2016-02-22
Multinode 12.0 deployments fail on os_cinder | Add in cinder devices types


git-harry:
This is fixed upstream, the OSA SHA needs to be updated to include 9d03db957aca27f28af994131dd25a1a70bd05c8

- [x] Master #865 
- [x] liberty-12.0 (for rc) #870

git-harry:
The review for the commit referenced above - https://review.openstack.org/#/c/276784

prometheanfire:
why is this not going into rc2?  If it's not going into 12.0.0 is it targeted for 12.0.1 or 12.1.0?

jacobwagner:
QE Verified that we are seeing this

TASK: [os_cinder | Add in cinder devices types] *******************************
skipping: [node89_cinder_scheduler_container-1d96acda] => (item=cinder_backends|dictsort)
skipping: [node89_cinder_api_container-15698120] => (item=cinder_backends|dictsort)
skipping: [node84_cinder_api_container-e33ed9d6] => (item=cinder_backends|dictsort)
skipping: [node98_cinder_api_container-61a9ffd8] => (item=cinder_backends|dictsort)
skipping: [node84_cinder_scheduler_container-825f0c67] => (item=cinder_backends|dictsort)
skipping: [node98_cinder_scheduler_container-f7592c57] => (item=cinder_backends|dictsort)
failed: [node118] => (item=(u'lvm', {u'volume_driver': u'cinder.volume.drivers.lvm.LVMISCSIDriver', u'volume_backend_name': u'LVM_iSCSI', u'volume_group': u'cinder-volume-hdd'})) => {"changed": true, "cmd": ". /root/openrc\n /openstack/venvs/cinder-12.0.5/bin/cinder type-create \"lvm\"\n /openstack/venvs/cinder-12.0.5/bin/cinder type-key \"lvm\" set volume_backend_name=\"LVM_iSCSI\"", "delta": "0:00:06.982501", "end": "2016-02-23 21:56:41.884749", "item": ["lvm", {"volume_backend_name": "LVM_iSCSI", "volume_driver": "cinder.volume.drivers.lvm.LVMISCSIDriver", "volume_group": "cinder-volume-hdd"}], "rc": 1, "start": "2016-02-23 21:56:34.902248", "warnings": []}
stderr: ERROR: Service Unavailable (HTTP 503)
ERROR: Service Unavailable (HTTP 503)
[...]

PR #865, opened 2016-02-25
Update OSA Submodule to head of Liberty

hughsaunders:
My motivation for moving forwards is that
362828cea9b831ed7d0cd6f43725fa80f073bbc6 is required for  defcore tests.
However we may as well move to head of Liberty rather than just that commit.


hughsaunders:
Also related: #852

hughsaunders:
recheck - no master swift/full job started.

prometheanfire:
:+1:

prometheanfire:
iirc, this was being held back by https://github.com/rcbops/rpc-openstack/issues/864 is that in contained within this update?

Issue #879, opened 2016-03-02
r11.1.4 ceph nodes with 4GB of RAM will OOM, "Unable to fork"


timolow:
##### What version (commit) do you encounter this issue on?
r11.1.4
##### What type of issue is this (bug, feature, documentation, regression)?
regression/bug

##### What is the expected behavior?
ceph to deploy on OSD nodes with 4GB.

##### What is actually happening?
Ansible injects sysctl values that cause the host machine to OOM and "unable to fork"

r11.1.4 (4194303=4GB)
Mar  1 14:49:24 swift02 ansible-<stdin>: Invoked with name=vm.min_free_kbytes ignoreerrors=True value=4194303 reload=True state=present sysctl_set=False sysctl_file=/etc/sysctl.conf
Mar  1 14:50:01 swift02 cron[3405]: (CRON) error (can't fork)

r11.1.3 (4294967296=4TB)
Mar  1 16:08:07 swift02 ansible-<stdin>: Invoked with name=vm.min_free_kbytes ignoreerrors=True value=4294967296 reload=True state=present sysctl_set=False sysctl_file=/etc/sysctl.conf

##### How can we reproduce this issue?
Pull down r11.1.4 and attempt ceph install on OSDs with 4GB of RAM.

##### How can we verify your issue is addressed?
Adjust sysctl value to X% of Total Ram.

##### If this is a large issue, what can be done immediately to alleviate the issue while working on a longer-term solution?

Remove or adjust value  vm.min_free_kbytes, value: 4194303 

##### What is your evaluation on the importance of this issue, and why?
Minor, production systems should have more then 4GB of RAM.

##### What branches / milestones do you believe this issue should target?
The list below is what is currently supported for backports, remove branches you do not wish to target.
- [x] Master
- [x] liberty-12.0
- [x] kilo

d34dh0r53:
It looks like this is actually a ceph regression: https://github.com/ceph/ceph-ansible/search?utf8=%E2%9C%93&q=vm_min_free_kbytes&type=Code  @timolow can you file this bug with the [`ceph-ansible`](https://github.com/ceph/ceph-ansible) project and link that issue here?

d34dh0r53:
It looks like it may have been introduced here: https://github.com/ceph/ceph-ansible/pull/543 @mattt416 can you work with @timolow on this?

mattt416:
[...]

PR #880, opened 2016-03-03
Bump ceph SHAs

mattt416:
This bump is to include commits [1] and [2] which fix issue #879 which
was introduced by the last SHA bumps.  That issue is limited to OSD
hosts running 4 GB of RAM or below and was therefore not discovered in
our gating.

[1] https://github.com/ceph/ansible-ceph-common/commit/0bfa01a6cb5a934e4c2d203eeceb51af3c3de186
[2] https://github.com/ceph/ansible-ceph-common/commit/739230648581659a337902e7159c974547e94898

Closes issue #879

### DIFF
--- a/ansible-role-requirements.yml
+++ b/ansible-role-requirements.yml
@@ -1,9 +1,9 @@
 - name: ceph-common
   src: https://github.com/ceph/ansible-ceph-common.git
-  version: 033a5e54b97903750586efce67f3fe8ba968bf45
+  version: 1c623611a1d43b4c9116d1a0c21f0bdbdd87a0e7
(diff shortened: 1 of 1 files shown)


prometheanfire:
well, since they squashed to one commit we have to update this everywhere :+1: :(

mattt416:
Shouldn't @prometheanfire -- he was able to sort it all out.

prometheanfire:
oh, good news then

prometheanfire:
this will need backports, so should reopen the issue when this is merged

d34dh0r53:
:+1:

Issue #906, opened 2016-03-18
Update OSA SHA to include latest fixes


git-harry:
##### What version (commit) do you encounter this issue on?
127950076bb7a883e3e25a645adc7eb33e502677
##### What type of issue is this (bug, feature, documentation, regression)?
Bug
##### What is your evaluation on the importance of this issue, and why?
This is required for 12.1.0, it pulls in the latest OSA Liberty fixes including addressing the Neutron port security upgrade issue.
##### What branches / milestones do you believe this issue should target?
The list below is what is currently supported for backports, remove branches you do not wish to target.
- [x] Master

prometheanfire:
closed as this is an upgrade only thing

PR #907, opened 2016-03-18
Update OSA SHA to OSA tag 12.0.8

git-harry:
Update openstack-ansible submodule SHA to pull in the latest fixes from
OSA. This includes the fix to address compute instance build failures
due to the Neutron ml2 port security extension being enabled on a
deployment with existing Neutron networks.

Closes-issue: https://github.com/rcbops/rpc-openstack/issues/906


prometheanfire:
I don't think we should consider #906 closed until we remove the workaround

git-harry:
@prometheanfire issue 906 isn't specifically for tracking the port security issue. We should probably have a separate one if we want to do that.

hughsaunders:
:+1:

prometheanfire:
:+1:

Issue #910, opened 2016-03-21
pywbem fails to build


git-harry:
##### What version (commit) do you encounter this issue on?
r12.0.0rc2 and 27c2ebf977105a1292f789cc1744647baa560b87
##### What type of issue is this (bug, feature, documentation, regression)?
Bug
##### What is the expected behavior?
pywbem should build on the repo server
##### What is actually happening?
The build is failing because a new version (0.8.0) has been published.
##### How can we reproduce this issue?
Attempt a deployment or rebuild the repo server containers.
##### How can we verify your issue is addressed?
The repo build playbooks complete.
##### What is your evaluation on the importance of this issue, and why?
This is high importance because it causes all new builds to fail.
##### What branches / milestones do you believe this issue should target?
The list below is what is currently supported for backports, remove branches you do not wish to target.
- [x] Master
- [x] liberty-12.0

PR #912, opened 2016-03-21
Add constraints file for all Pip packages

git-harry:
Add a constraints file generated by the repo server of an AIO build.

The repo server builds a single set of Python wheels based on the
required packages. These requirements vary in how specific a version
they want of any individual package. The result is that two deployments
build using the same configuration from the same SHA may end up with
different versions of wheels. Generally this does not cause an issue
however sometimes a new version of a Pip package is released that is
incompatible with the others required or has other dependencies beyond
what Pip can satisfy.

These incompatible package versions can cause testing/deployment
failures of releases and release candidates. This slows down the release
of new versions of this project and impacts customers. The constraints
file added by this commit is designed to more tightly control the wheels
that are build and so reduce the number of deployments that are impacted
by new versions of Pip packages.

The addition of this file creates additional overhead for the
development team, it should be regenerated for each new release. There
is also the possibility that it will impact a deployer where they
additional Python packages to be installed that are not compatible with
the constraints imposed by global-requirement-pins.txt. In this
situation it may be necessary for the deployer to modify or remove the
file.

Related-issue: https://github.com/rcbops/rpc-openstack/issues/910


prometheanfire:
If we can get a way for this to be generated by OSA for our use that would be much better then doing it ourselves.

git-harry:
There appear to be bugs in OSA trying to use this process. It's not critical to release so closing.

Issue #897, opened 2016-03-14
[r12.0.0rc2] — Neutron failing to bind ports when creating instances


jpmontez:
##### What version (commit) do you encounter this issue on?
r12.0.0rc2

##### What type of issue is this (bug, feature, documentation, regression)?
Bug

##### What is the expected behavior?
To create a successful VM with the `openstack server create` command

##### What is actually happening?
The instance fails to build. An exception is raised in the Nova and Neutron logs.

The logs show that the internal endpoints aren't being used by the Neutron service, but rather the public ones. Because the firewall doesn't allow public API access, connections fail and exceptions are raised.

##### How can we reproduce this issue?
Within a customer lab behind a firewall w/ no public access to the APIs, try and create a server.

##### How can we verify your issue is addressed?
By logging into the QE team's Sydney lab and using the utility container to try and create a basic server.

`openstack server create --flavor 2 --image cirros --nic net-id=XXX test`

##### If this is a large issue, what can be done immediately to alleviate the issue while working on a longer-term solution?
Possibly by force overriding the endpoint used by the Keystone adaptor

##### What is your evaluation on the importance of this issue, and why?
If the internal URL isn't respected, many services involving Neutron and Keystone will break.

##### What branches / milestones do you believe this issue should target?
The list below is what is currently supported for backports, remove branches you do not wish to target.
- [ ] Master
- [x] liberty-12.0

```
root@627551-infra01_utility_container-2484aa62:~# openstack server create --flavor 2 --image cirros --nic net-id=5fadf69c-136f-499d-9f3d-9e79f9bea358 --wait test
Error creating server: test
```

Relevant Neutron log:
```
2016-03-15 09:09:11.149 3114 ERROR neutron.notifiers.nova [-] Failed to notify nova on events: [{'tag': u'12c3241e-e54e-457c-976c-ef0e4be6b3b9', 'name': 'network-vif-deleted', 'server_uuid': u'f5176db2-4604-4917-9987-303616088744'}]
2016-03-15 09:09:11.149 3114 ERROR neutron.notifiers.nova Traceback (most recent call last):
2016-03-15 09:09:11.149 3114 ERROR neutron.notifiers.nova   File "/openstack/venvs/neutron-12.0.7/lib/python2.7/site-packages/neutron/notifiers/nova.py", line 247, in send_events
2016-03-15 09:09:11.149 3114 ERROR neutron.notifiers.nova     batched_events)
[...]

PR #923, opened 2016-03-24
Make Neutron use internal Nova endpoint

git-harry:
Update Neutron SHA and add configuration override to enable Neutron to
use Nova's internal endpoint.

This removes the need to change the OpenStack-Ansible SHA to address the
issue but will require an upgrade task to remove these changes in a
future release.

Closes-issue: https://github.com/rcbops/rpc-openstack/issues/897

### DIFF
--- a/rpcd/etc/openstack_deploy/user_variables.yml
+++ b/rpcd/etc/openstack_deploy/user_variables.yml
@@ -82,3 +82,8 @@ nova_ceilometer_enabled: False
 swift_ceilometer_enabled: False
 keystone_ceilometer_enabled: False
 tempest_service_available_ceilometer: False
+
+neutron_git_repo: "https://git.openstack.org/openstack/neutron"
+neutron_git_install_branch: "b80fa23ac0b292db84df84855141eb16a4e0fec0"
+neutron_git_dest: "/opt/neutron_{{ neutron_git_install_branch | replace('/', '_') }}"
+neutron_neutron_conf_overrides: {"nova": {"endpoint_type": "internal"}}


d34dh0r53:
:-1: for the time being, I would like to see if OSA is open to bumping their SHAs early.  If not I will change my vote.

git-harry:
@d34dh0r53 I don't think liberty-12.0 should be updating the OSA SHA, I believe it is too much change to be pulling in when the facilities exist to allow for a more targeted approach.

d34dh0r53:
@git-harry I would argue that trying to override a SHA in this manner is more untested than bumping the OSA SHA.  SHA bumps are part of the process and as long as the QE team is aware of the changes they should be fine.

git-harry:
@d34dh0r53 the beauty of this is it's such a small change, it is very easy to tell whether or not it has worked.

@nrb @prometheanfire what are your thoughts on how the Neutron fix should be included.

rpcd/etc/openstack_deploy/user_variables.yml
```diff
@@ -82,3 +82,8 @@ nova_ceilometer_enabled: False
 swift_ceilometer_enabled: False
 keystone_ceilometer_enabled: False
 tempest_service_available_ceilometer: False
+
+neutron_git_repo: "https://git.openstack.org/openstack/neutron"
+neutron_git_install_branch: "b80fa23ac0b292db84df84855141eb16a4e0fec0"
+neutron_git_dest: "/opt/neutron_{{ neutron_git_install_branch | replace('/', '_') }}"
```
nrb: So we need the `neutron_git_repo` and `neutron_git_dest` vars redefined in order to update the neutron branch? That seems unwieldy, but fixing that behavior isn't in scope, more looking to verify that it's necessary.

prometheanfire: `neutron_git_install_branch` is all that's needed, don't think we need to redefine the repo if the change is merged upstream already.

prometheanfire:
:-1: because this overrides too much

rpcd/etc/openstack_deploy/user_variables.yml
nrb: @prometheanfire That's what the initial patchset was, but it's since changed, so I'm wondering if that assumption is false (fwiw, I think that's what should be needed).

Actually, looking at the commit message:

> neutron_git_repo is required so that the lookup filter py_pkgs knows to
look in user_variables.yml for the new SHA.

So it's to account for behavior in OSA.

prometheanfire: oh, odd, but ok

prometheanfire:
removing the -1 for this and adding a -1 to wait for upstream

https://review.openstack.org/#/c/297318

d34dh0r53:
[`297318`](https://review.openstack.org/#/c/297318) has merged which incorporates the upstream neutron SHA which this PR uses.  Closing this out as it's redundant.

PR #935, opened 2016-03-26
Py3 and code cleanup

prometheanfire:
This is split of into multiple commits because it seemed logical, the py3 commit is the main one that should be on it's own because of behaviour changes.

This could be a first pass at things, I mainly did this because I wanted to try out pycharm's py2/3 compat stuff.  It doesn't really hit that maas files because they tend to use modules that are not py3 compat at all (the maas-tool uses configparser and queue).

### DIFF
--- a/scripts/f5-config.py
+++ b/scripts/f5-config.py
@@ -30,7 +30,7 @@ SNAT_POOL = (
     ' %(snat_pool_addresses)s } }'
 )
 
-#Persistance Profile:
+# Persistance Profile:
 PERSISTANCE = [
     r'create ltm persistence source-addr /' + PART + '/' + PREFIX_NAME + '_PROF_PERSIST_IP {'
     r' app-service none defaults-from /Common/source_addr'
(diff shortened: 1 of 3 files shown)


git-harry:
:-1: 

There is no issue tracking this work.

None of your commit messages conform with the guidelines.

There is no variable shadowing happening because the global f is defined after get_config is called. Assuming there was shadowing happening and it was agreed that it's even something the project should worry about, it would be better to create a separate function, say write_config, to remove f from the global namespace.

I am not aware of any requirement for Python 3 compatibility. Assuming it is required I would expect the work to add this functionality be done as part of adding a new gate check. Without any testing these changes compatibility is unlikely to be maintained.

scripts/f5-config.py
```diff
@@ -635,9 +635,12 @@ def main():
                         '/' + PART + '/' + PREFIX_NAME + '_PROF_SSL_%(ssl_domain_name)s { context clientside }'
                     ) % user_args
                 else:
-                    virtual_dict['ssl_profiles'] = '/' + PART + '/' + PREFIX_NAME + '_PROF_SSL_SERVER { context serverside } /' + PART + '/' + PREFIX_NAME + '_PROF_SSL_%(ssl_domain_name)s { context clientside }'% user_args
+                    virtual_dict['ssl_profiles'] = '/' + PART + '/' + PREFIX_NAME + \
```
git-harry: PEP8 prefers brackets over '\' for line continuation.

prometheanfire:
meh, it was just for fun anyway

Issue #924, opened 2016-03-24
cinder_service_check unable to find host in service list


xeregin:
This issue was discovered in r12.0.0rc2
From what I can tell, this is due to the fact that MaaS is configured to look for (in this case) '**573963-storage01@**'
The related entries in the service list are as follows:
  573963-storage01.qe-iad3-2.rpc.rackspace.com@lvm
  573963-storage01.qe-iad3-2.rpc.rackspace.com@lvm_ssd
  573963-storage01.qe-iad3-2.rpc.rackspace.com@emc_vnx

The command that is executed by MaaS is as follows:
cinder_service_check.py --host 573963-storage01 172.20.236.10

The output is as follows:
status error No host(s) found in the service list

With a standard installation of MaaS, it's already in alert status. This would probably be an important fix for the Support side.

xeregin:
Either the scripts setting up the MaaS checks need to include the rest of the FQDN, or the python script that performs the actual check should not include the '@' symbol. Either case should fix this bug...

b3rnard0:
@xeregin Can you please use the new template for this bug (https://github.com/rcbops/rpc-openstack/issues/new)? Thanks in advance.

d34dh0r53:
@git-harry can you look at this during your day?  Based on the issue hopefully it's an easy fix.  After talking with @tbyrom138 we're not treating this as a blocker for RC3.

git-harry:
@d34dh0r53 only if you're happy to wait till Tuesday.

claco:
Personally, I'd think we would want to correct the hostname going into the alarm check setup, and not lter the search using @, since that was on purpose.

@BjoernT @busterswt @cemason @codebauss @timolow Can any of you shed some light on what we see most commonly in our installations? Do most installs not have a sub domain/domain name, or are we using some other work around, etc?  Thanks!

BjoernT:
I would not alter the hostname either and just delimit by the @ sign. Also most installations seem to use the ansible hostnames rather than FQDN, this is more a QE thing

d34dh0r53:
So, I'm going to table this for RC3 as it's not a blocker.  Relevant folks please continue to work this either as a docs issue or a bug.

prometheanfire:
does this get backports and to what?

git-harry:
@prometheanfire this has been targeted at r12.0.0rc4 however seeing as that tag has already been created it will have to go in the next milestone for the liberty-12.0 branch.

[...]

PR #937, opened 2016-03-29
Use ansible_nodename for Cinder service checks

git-harry:
Replace ansible_hostname with ansible_nodename as the host argument
supplied to the cinder_service_check.py MaaS plugin by the configured
checks.

The default value for the hostname used in Cinder is supplied by
socket.gethostname. The Ansible fact that should be equivalent to this
is ansible_nodename.

ansible_hostname == ansible_nodename.split('.')[0]

Closes-issue: https://github.com/rcbops/rpc-openstack/issues/924


git-harry:
check_maas

prometheanfire:
:+1: simple change

PR #940, opened 2016-03-30
Override neutron.conf to add endpoint type

git-harry:
In issue #897 we pulled in an upstream fix which adds the ability
to specify the type of nova endpoint to use. The endpoint specified
should be one of public, internal or admin and will be looked up
in the keystone catalog.  OSA did not add the key in `neutron.conf`
to specify the endpoint type as the default `public` works for most
deployments.  RPC-O deployments on the other hand typically have
the public endpoints inaccessible from the control plane thus
requiring us to override that value.

Partially-fixes: #897

This was cherry picked from the liberty-12.0 branch because the fix was
mistakenly not first committed to master.
(cherry picked from commit aadb2d42a0c07603f95f995dd840919371333080)


git-harry:
recheck_ceph
recheck_swift
recheck_upgrade

d34dh0r53:
:+1:

Issue #916, opened 2016-03-22
Elasticsearch not reachable through tunnel


BjoernT:
##### What version (commit) do you encounter this issue on?

ALL RPC versions

##### What type of issue is this (bug, feature, documentation, regression)?

feature

##### What is the expected behavior?

No hard coded server URLs

##### What is actually happening?

After login into kibana, the Kibana fronted will create HTTP urls to absolute server URLs which can not be reached through SSH tunnels

##### How can we reproduce this issue?

ssh rack@server -L8443:172.29.236.10:8443
and login into https://localhost:8443

##### How can we verify your issue is addressed?

It works

##### If this is a large issue, what can be done immediately to alleviate the issue while working on a longer-term solution?

Small issue

##### What is your evaluation on the importance of this issue, and why?

Normal prio, I fix it myself

##### What branches / milestones do you believe this issue should target?
The list below is what is currently supported for backports, remove branches you do not wish to target.
- [x] Master
- [x] liberty-12.0
- [x] liberty-12.1
- [x] kilo

BjoernT:
PR in progress

prometheanfire:
needs backports

BjoernT:
See PR for backports

prometheanfire:
needs kilo backport as well :D

BjoernT:
@prometheanfire  https://github.com/rcbops/rpc-openstack/pull/943 is for kilo

prometheanfire:
right, just 12.0.1 then

BjoernT:
yeah the branch you secretly created over the weekend ;-)

prometheanfire:
nah, no secret there :P

BjoernT:
Ha, already is part of the 12.1 branch

PR #943, opened 2016-03-30
Remove fixed elasticsearch URL from Kibana

BjoernT:
Remove hard coded elasticsearch URL from the kibana configuration
and the apache proxy server.
The window hostname in which kibana is started, is used to compose the
elasticsearch URL.
This change enables access of the Kibana application through port
tunneling protocols like SSH.

Closes-Bug: #916
(cherry picked from commit 2ad48166c183a6cc6988aa07cc00b7cfbcf38ed3)


prometheanfire:
:+1: straightforward backport

PR #944, opened 2016-03-31
Add Swift toggle to deploy.sh

git-harry:
This is introduced so that swift can be disabled for a ceph gate scenario

/etc/openstack_deploy/conf.d/swift.yml is removed if swift is
disabled so that it's containers are not created. The swift playbooks will
still run, but no hosts will match so the tasks will be skipped.

Related: #732
(cherry picked from commit 2060c4c69586e0de3925a450f65570b2a11c4de8)


hughsaunders:
:+1:

prometheanfire:
:+1:

Issue #945, opened 2016-03-31
Kilo AIO deploys Ceilometer


git-harry:
##### What version (commit) do you encounter this issue on?
kilo
##### What type of issue is this (bug, feature, documentation, regression)?
bug
##### What is the expected behavior?
Ceilometer is not installed when creating an AIO and DEPLOY_CEILOMETER == "no"
##### What is actually happening?
Ceilometer is being configured.
##### How can we reproduce this issue?
Build a Kilo AIO.
##### How can we verify your issue is addressed?
The Ceilometer playbooks will be skipped and no Ceilometer containers will be created.
##### If this is a large issue, what can be done immediately to alleviate the issue while working on a longer-term solution?
N/A
##### What is your evaluation on the importance of this issue, and why?
Low. This should not be causing an issue with production deployments because they are manually creating the config in /etc/openstack_deploy
##### What branches / milestones do you believe this issue should target?
The list below is what is currently supported for backports, remove branches you do not wish to target.
- [x] kilo

prometheanfire:
Is this only kilo?

PR #946, opened 2016-03-31
Prevent deployment of Ceilometer in AIO

git-harry:
Remove the Ceilometer conf.d file so that the Ceilometer groups are
empty thereby preventing the creation of any Ceilometer containers or
the Ceilometer playbooks finding any valid hosts.

This commit should have not impact on production deployments but is
intended to improve the the Kilo gating process.

Closes-issue: https://github.com/rcbops/rpc-openstack/issues/945

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -91,6 +91,10 @@ if [[ "${DEPLOY_AIO}" == "yes" ]]; then
   then
     rm /etc/openstack_deploy/conf.d/swift.yml
   fi
+  if [ "${DEPLOY_CEILOMETER}" != "yes" ]; then
+    rm -f /etc/openstack_deploy/conf.d/ceilometer.yml
+    sed -i 's/swift_ceilometer_enabled:.*/swift_ceilometer_enabled: False/' /etc/openstack_deploy/user_variables.yml
+  fi
 fi
 
 # bootstrap ansible if need be


hughsaunders:
:+1:

d34dh0r53:
:+1:

scripts/deploy.sh
```diff
@@ -85,6 +85,10 @@ if [[ "${DEPLOY_AIO}" == "yes" ]]; then
     sed -i "s/aio1/$(hostname)/" /etc/openstack_deploy/openstack_user_config.yml
     sed -i "s/aio1/$(hostname)/" /etc/openstack_deploy/conf.d/*.yml
   fi
+  if [ "$DEPLOY_CEILOMETER" != "yes" ]; then
```
prometheanfire: should probably be `"${DEPLOY_CEILOMETER}" instead of unbracketed.

scripts/deploy.sh
```diff
@@ -85,6 +85,10 @@ if [[ "${DEPLOY_AIO}" == "yes" ]]; then
...
+    rm -f /etc/openstack_deploy/conf.d/ceilometer.yml
+    sed -i 's/swift_ceilometer_enabled:.*/swift_ceilometer_enabled: false/' /etc/openstack_deploy/user_variables.yml
```
prometheanfire: The current default seems to be set to false, better to be explicit, but if that line is instead commented out we don't actually disable it (since we won't remove the comment).  If we only want to change the variable if it's defined then I suggest starting the search with a `^`.  Also, not a global replace, but that's likely intentional.  Also, also to match the rest of the file it should be `False` not `false`.

prometheanfire:
:-1: til cleanup is done

scripts/deploy.sh
git-harry: This isn't done consistently in the file and isn't required.

prometheanfire: Not technically required but something I was taught to do as good practice.

scripts/deploy.sh
git-harry: The bootstrap-host role overrides the OSA default so it is always set, it this wasn't happening we wouldn't need to override the value.

prometheanfire:
looks good now, need a second :+1: before we can merge this commit

prometheanfire:
recheck_swift

d34dh0r53:
:+1:

Issue #918, opened 2016-03-22
logstash heap to small on systems with > 96G ram


BjoernT:
##### What version (commit) do you encounter this issue on?

ALL RPC version

##### What type of issue is this (bug, feature, documentation, regression)?

bug

##### What is the expected behavior?

Use 1/4th of the installed RAM just for logstash, I might go far as even using 1/2 when using > 64G

##### What is actually happening?

Uses "hard coded" 16G

##### How can we reproduce this issue?

root@infra01:/opt/rpc-openstack/rpcd/playbooks# ssh logging01_logstash_container-a2fb86f1 grep LS_HEAP_SIZE /etc/default/logstash; free -g
LS_HEAP_SIZE=16384m
             total       used       free     shared    buffers     cached
Mem:           125         53         72          0          2         24
-/+ buffers/cache:         26         99
Swap:            7          0          7

##### How can we verify your issue is addressed?

LS_HEAP_SIZE is configured appropriately 

##### If this is a large issue, what can be done immediately to alleviate the issue while working on a longer-term solution?

Small issue

##### What is your evaluation on the importance of this issue, and why?

P3 prio

##### What branches / milestones do you believe this issue should target?
The list below is what is currently supported for backports, remove branches you do not wish to target.
- [x] Master
- [x[ liberty-12.0
- [x] kilo

BjoernT:
PR pending

odyssey4me:
FYI the jre cannot allocate more than 16GB of RAM - that is why the 16GB limit was there. I would recommend that the calculation maintains the 16GB limit otherwise the JRE will try to book the RAM, but can never use it.

BjoernT:
@odyssey4me  We are running already logstash with > 32G RAM. There's no 16GB limit. Not sure where you get this limit from

PR #956, opened 2016-04-04
Revising logstash LS_HEAP_SIZE calculation

BjoernT:
This fix revises the logstash LS_HEAP_SIZE calculation to 1/3rd
of the RAM when having less than 32G RAM available.
If a system has more than 32G RAM available, 1/2 of the RAM
is allocated to logstash.

Closes-Bug: #918
(cherry picked from commit 4c1dee14d88a685ea40f086043d77ce67d192659)


prometheanfire:
:+1: woo

d34dh0r53:
:+1:

Issue #963, opened 2016-04-07
ansible-list is causing the travis-ci checks to fail


git-harry:
##### What version (commit) do you encounter this issue on?
master
##### What type of issue is this (bug, feature, documentation, regression)?
bug
##### What is the expected behavior?
ansible-lint tests pass
##### What is actually happening?
ansible-lint test fail
##### How can we reproduce this issue?
./scripts/linting-ansible.sh with ansible-lint==2.5.0
##### How can we verify your issue is addressed?
travis-ci jobs start working
##### What is your evaluation on the importance of this issue, and why?
High, it break the travis check jobs
###### Remove branches (lines) you do not wish to target.
- [x] Master
- [x] liberty-12.1
- [x] liberty-12.0

git-harry:
# ansible-lint *.yml
Traceback (most recent call last):
  File "/tmp/al/bin/ansible-lint", line 125, in <module>
    sys.exit(main(sys.argv[1:]))
  File "/tmp/al/bin/ansible-lint", line 112, in main
    matches = runner.run()
  File "/tmp/al/local/lib/python2.7/site-packages/ansiblelint/__init__.py", line 222, in run
    skip_list=set(self.skip_list)))
  File "/tmp/al/local/lib/python2.7/site-packages/ansiblelint/__init__.py", line 134, in run
    matches.extend(rule.matchtasks(playbookfile, text))
  File "/tmp/al/local/lib/python2.7/site-packages/ansiblelint/__init__.py", line 62, in matchtasks
    for task in ansiblelint.utils.get_action_tasks(yaml, file):
  File "/tmp/al/local/lib/python2.7/site-packages/ansiblelint/utils.py", line 368, in get_action_tasks
    if 'include' not in task.keys()]
  File "/tmp/al/local/lib/python2.7/site-packages/ansiblelint/utils.py", line 326, in normalize_task
    return normalize_task_v2(task)
  File "/tmp/al/local/lib/python2.7/site-packages/ansiblelint/utils.py", line 258, in normalize_task_v2
    action, arguments, result['delegate_to'] = mod_arg_parser.parse()
  File "/tmp/al/local/lib/python2.7/site-packages/ansible/parsing/mod_args.py", line 304, in parse
    raise AnsibleParserError("no action detected in task. This often indicates a misspelled module name, or incorrect module path.", obj=self._task_ds)
ansible.errors.AnsibleParserError: no action detected in task. This often indicates a misspelled module name, or incorrect module path.

[...]

Issue #961, opened 2016-04-06
Tracking of upstream 1566915


d34dh0r53:
##### What version (commit) do you encounter this issue on?
`r12.1.0rc1`

##### What type of issue is this (bug, feature, documentation, regression)?
bug

##### What is the expected behavior?
N/A

##### What is actually happening?
Tracking of upstream [`1566915`](https://bugs.launchpad.net/openstack-ansible/+bug/1566915)

##### How can we reproduce this issue?
N/A

##### How can we verify your issue is addressed?
A fix for [`1566915`](https://bugs.launchpad.net/openstack-ansible/+bug/1566915) is merged and SHAs bumped.

##### If this is a large issue, what can be done immediately to alleviate the issue while working on a longer-term solution?
N/A

##### What is your evaluation on the importance of this issue, and why?
High

###### Remove branches (lines) you do not wish to target.
- [x] Master
- [x] liberty-12.1

major:
![Bug](https://i.imgflip.com/1222z7.jpg)

sigmavirus24:
Cry me an upgrade

prometheanfire:
backport?

git-harry:
@prometheanfire the above commit reference links back to it. It's pr #969

prometheanfire:
ah, that just popped up

git-harry:
@prometheanfire the pull request description didn't reference this issue only the commit did. I've updated so the larger bold link appears in addition to the less useful commit one.

mancdaz:
master commit and liberty-12.1 backport have both merged, closing

PR #969, opened 2016-04-08
Update OSA SHA and constrain ansible-lint version

git-harry:
Both commits are required for checks to pass.

https://github.com/rcbops/rpc-openstack/issues/961
https://github.com/rcbops/rpc-openstack/issues/963


prometheanfire:
:+1:

Issue #967, opened 2016-04-07
memcache ttl too short on kilo -> liberty for running rpc-support play


b3rnard0:
##### What version (commit) do you encounter this issue on?
V12.1.0-rc1

##### What type of issue is this (bug, feature, documentation, regression)?
bug,regression

##### What is the expected behavior?
play succeeds

##### What is actually happening?
play fails

##### How can we reproduce this issue?
make rpc-support play take a long time

##### How can we verify your issue is addressed?
never hit missing rpc_support ssh key fetching from memcache

##### If this is a large issue, what can be done immediately to alleviate the issue while working on a longer-term solution?
rerun the play

##### What is your evaluation on the importance of this issue, and why?
medium/low?

###### Remove branches (lines) you do not wish to target.
- [x] Master #977 
- [x] liberty-12.1 #978
- [x] liberty-12.0 #1010 
- [x] kilo #1014 

@sigmavirus24 can you confirm this?

BjoernT:
The default TTL should be 24h so I'm not sure if that is correct, the object could have been evicted too.
Personally I would prefer to get rid of memcache a distribution and spend the extra time to implement this via ansible facts

prometheanfire:
nope :D

https://github.com/openstack/openstack-ansible-plugins/blob/master/library/memcached#L87

BjoernT:
Lol, of course our driver alters the default ttl :-)

txgunslinger77:
We seem to be running into this problem again/still in r12.0.1, with the rpc-support play failing with memcache errors, so we still need someone to look at this please. We get these quite often with very little idea as to the solution or work aorund.

prometheanfire:
ok, reopening as a request to backport to the 12.0 series

b3rnard0:
@git-harry If you can make the eval that this is a backport candidate for r12.0.2, it would be much appreciated.

txgunslinger77:
For a little more context, I'm running deploy.sh and get the following error at the following task (for brevity, I only included one sample error, there are many more just like it):

TASK: [rpc_support | Retrieve authorized keys] ********************************
[...]

PR #977, opened 2016-04-11

prometheanfire:

related bug #967


mancdaz:
:+1:

git-harry:
👍

PR #978, opened 2016-04-11

prometheanfire:

related bug #967

(cherry picked from commit 68621a5b8375a386f33f82a82a5760f0da5d1ba4)


prometheanfire:
recheck_swift

mancdaz:
:+1:  pending master patch being merged, unchanged.

git-harry:
👍

Issue #982, opened 2016-04-12
Beaver 33.3.0 requires mosquitto which has been renamed paho-mqtt


hughsaunders:
##### What version (commit) do you encounter this issue on?
master, kilo

##### What type of issue is this (bug, feature, documentation, regression)?
bug

##### What is the expected behavior?
Wheels build ok

##### What is actually happening?
`Create OpenStack-Ansible requirement wheels` fails because beaver requires mosquitto which cannot be found. 

```
06:33:42   Could not find a version that satisfies the requirement mosquitto>=1.1 (from beaver==33.3.0->-c /var/www/repo/os-releases/12.0.10/requirements_constraints.txt (line 40)) (from versions: )
06:33:42 No matching distribution found for mosquitto>=1.1 (from beaver==33.3.0->-c /var/www/repo/os-releases/12.0.10/requirements_constraints.txt (line 40))
```

Updated Beaver packages have been released that depend on paho-mqtt instead of mosquitto, so we probably need to increase our beaver cap. 

##### How can we reproduce this issue?
Attempt to build an aio from kilo, master or liberty branches.

##### How can we verify your issue is addressed?
Beaver wheel builds ok.

##### What is your evaluation on the importance of this issue, and why?
Critical - breaking the gate and presumably production builds.

###### Remove branches (lines) you do not wish to target.
- [x] Master
- [x] liberty-12.1
- [x] liberty-12.0
- [x] kilo

git-harry:
This has been addressed by adding the mosquitto package [1] to rpc-repo.

[1] http://rpc-repo.rackspace.com/pools/mosquitto/mosquitto-1.2.3-py2-none-any.whl

PR #983, opened 2016-04-12
Build Beaver package from git

hughsaunders:
Mosquitto was renamed to paho-mqtt and the mosquitto package was removed
from pypi. Beaver requires mosquitto until the latest commit which
changes the requirement to paho-mqtt. Beaver has not released a new
version to pypi yet, so we need to build beaver from git until they do.

Related: #982


git-harry:
No longer required.

Issue #973, opened 2016-04-11
OSA [1563448] - Change to linuxbridge agent defaults affects upgrades


git-harry:
##### What version (commit) do you encounter this issue on?
master
##### What type of issue is this (bug, feature, documentation, regression)?
bug
##### What is the expected behavior?
See https://bugs.launchpad.net/openstack-ansible/+bug/1563448
##### What is actually happening?
See https://bugs.launchpad.net/openstack-ansible/+bug/1563448
##### How can we reproduce this issue?
See https://bugs.launchpad.net/openstack-ansible/+bug/1563448
##### How can we verify your issue is addressed?
See https://bugs.launchpad.net/openstack-ansible/+bug/1563448
##### What is your evaluation on the importance of this issue, and why?
This issue is impacting upgrades and has the potential to cause additional downtime.
###### Remove branches (lines) you do not wish to target.
- [x] master https://github.com/rcbops/rpc-openstack/pull/1375
- [x] mitaka-13.0 https://github.com/rcbops/rpc-openstack/pull/1377
- [x] liberty-12.2 https://github.com/rcbops/rpc-openstack/pull/1379

git-harry:
I think the best solution for r12.1.0 is to override the variable on upgrades to set [vxlan]/l2_population = True. This minimises the blocking effect of this issue on the release

For a future release, r13.1 at the latest, we need to create an automated upgrade path for changing this variable to the recommended default of False.

prometheanfire:
this is going to be done in doc, making a note here

tomc603:
I'm going to write up what situations Support may need to change this in and what will need to be changed. I'll pass it by @busterswt before sending to Docs to make sure it looks sane.

b3rnard0:
@tomc603 How's this coming along? This is priority for r12.2.0 docs. Adding @erikmwilson as he'll be adding this to r12.2.0 release notes.

tomc603:
I've sent over my notes to @erikmwilson, and he has noted he'll look them over tomorrow. If he needs, I'll gladly dig around more or add info where needed.

b3rnard0:
@tomc603 What's the status on the documentation for this issue?

tomc603:
@b3rnard0 I've sent everything over to @erikmwilson, and it seems to have passed a sniff test by a couple people. I'm going to mark this closed.

git-harry:
[...]

PR #985, opened 2016-04-12
add override to enable l2pop

prometheanfire:
Upstream OSA disabled l2pop in liberty and later.  This will cause
existing customers to have failed guest networking.  A migration
to vxlan multicast will eventually be needed, but for the moment we
can override and turn on l2pop.

related-bug: #973

### DIFF
--- a/rpcd/etc/openstack_deploy/user_variables.yml
+++ b/rpcd/etc/openstack_deploy/user_variables.yml
@@ -29,6 +29,9 @@ neutron_db_max_overflow: 60
 neutron_db_max_pool_size: 120
 neutron_db_pool_timeout: 60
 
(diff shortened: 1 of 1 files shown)


prometheanfire:
recheck_upgrade

git-harry:
👎 I don't think this should be applied to all deployments, only upgrades. We've established that we want to eventually set this to False for all customers and that there is an impact to connectivity to change this setting so I don't see the benefit of imposing this migration on new deployments.

tomc603:
👎 

Given the fact that we know what needs to be done during a migration away from l2pop, I don't think we should be doing this at all in master. Instead we should test the migration mechanism that I believe Denton already has put together. Adding this patch only carries the problem forward and could potentially be accidentally carried into a release.

prometheanfire:
closing, doing this in doc

PR #987, opened 2016-04-12
Optionally apply host security hardening

major:
This patch allows deployers to optionally apply host security
hardening configurations to RPC-O deployments. It also enables
security hardening for all AIO builds.

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -13,6 +13,7 @@ export DEPLOY_TEMPEST=${DEPLOY_TEMPEST:-"no"}
 export DEPLOY_CEILOMETER="no"
 export DEPLOY_CEPH=${DEPLOY_CEPH:-"no"}
 export DEPLOY_SWIFT=${DEPLOY_SWIFT:-"yes"}
+export DEPLOY_HARDENING=${DEPLOY_HARDENING:-"no"}
 export FORKS=${FORKS:-$(grep -c ^processor /proc/cpuinfo)}
 export ANSIBLE_PARAMETERS=${ANSIBLE_PARAMETERS:-""}
 export ANSIBLE_FORCE_COLOR=${ANSIBLE_FORCE_COLOR:-"true"}
@@ -125,6 +126,14 @@ if [[ "${DEPLOY_OA}" == "yes" ]]; then
   # setup the haproxy load balancer
   if [[ "${DEPLOY_HAPROXY}" == "yes" ]]; then
     run_ansible haproxy-install.yml
+  fi
+
+  # Apply host security hardening with openstack-ansible-security
+  if [[ "$DEPLOY_HARDENING" == "yes" ]]
+  then
+    echo "apply_security_hardening: true" >> $RPCD_VARS
+  else
+    echo "apply_security_hardening: false" >> $RPCD_VARS
   fi
 
   # setup the hosts and build the basic containers


major:
Flake8 fixes are proposed upstream in https://review.openstack.org/304815.

scripts/deploy.sh
```diff
@@ -70,6 +70,8 @@ if [[ "${DEPLOY_AIO}" == "yes" ]]; then
     sed -i "s/maas_notification_plan: .*/maas_notification_plan: npTechnicalContactsEmail/" $RPCD_VARS
     # set network speed for vms
     echo "net_max_speed: 1000" >>$RPCD_VARS
+    # Enable host security hardening for AIO builds
+    sed -i "s/apply_security_hardening: .*/apply_security_hardening: true/" $RPCD_VARS
```
git-harry: We need to be able to choose whether or not this is enabled. AIO builds are used by gating and so this would interfere with testing.

scripts/deploy.sh
```diff
@@ -130,6 +132,12 @@ if [[ "${DEPLOY_OA}" == "yes" ]]; then
   # setup the hosts and build the basic containers
   run_ansible setup-hosts.yml
 
+  # Apply security hardening if apply_security_hardening is true
+  # (By default, apply_security_hardening is false and this will be skipped)
```
git-harry: I think this should use the existing pattern for controlling execution by defining a variable that is used by this script, existing examples include DEPLOY_MAAS and DEPLOY_CEPH.

git-harry:
👎 see comments on diff regarding controlling execution of the playbook.

scripts/deploy.sh
```diff
@@ -48,8 +49,13 @@ if [[ "${DEPLOY_AIO}" == "yes" ]]; then
   if [ -n "${DATA_DISK_DEVICE}" ]; then
     export BOOTSTRAP_OPTS="${BOOTSTRAP_OPTS} bootstrap_host_data_disk_device=${DATA_DISK_DEVICE}"
   fi
+
   # force the deployment of haproxy for an AIO
   export DEPLOY_HAPROXY="yes"
+
+  # Enable host security hardening for AIO builds
+  export DEPLOY_HARDENING="yes"
```
git-harry: I think this should be removed, I see no reason to force an AIO to deploy this playbook. As it stands the feature is optional and so the AIO should be able to take advantage of that configurability.

major: I can certainly understand that, but we are currently testing upstream Mitaka/Newton OSA AIO's with security hardening enabled. I've also proposed that for OSA's Liberty release as well.  The role should not impact a running RPC-O deployment, and if it does, we've found a bug.

Are there any specific concerns about the content of the role and applying those configurations to an AIO?

git-harry: Well this feature makes a whole load of changes that, as I understand it, will only be used by a limited number of customers. So we'll essentially be testing an edge case deployment as in all our standard gate job with this setting. In additional anyone wanting to do any AIO testing without this hardening stuff now has to manually edit the script to make that happen.

The only things that should be hardcoded in the AIO section are those things that are required to make the AIO work, e.g. haproxy.

git-harry: My understanding is that we need to carry this playbook because OSA have rejected backporting this feature to Liberty. If it's being backported we don't need this patch.

git-harry:
Note, this functionality may get backported upstream [1] which will mean it is not necessary for RPCO to carry the playbook.

[1] https://review.openstack.org/#/c/302385/

git-harry:
https://review.openstack.org/#/c/302385/ has now merged and so this pull request should be updated to make use of the upstream functionality.

major:
I guess I'll need to wait on Jesse to make a new tag for the security role upstream first. Then we will need to bump the OSA SHA here.

git-harry:
@major we don't need a tag in the security role, the code is already in OSA Liberty so there is nothing stopping us using it. Checked with @odyssey4me and he confirms we should be fine to pull it in.

scripts/deploy.sh
```diff
@@ -127,6 +128,12 @@ if [[ "${DEPLOY_OA}" == "yes" ]]; then
     run_ansible haproxy-install.yml
   fi
 
+  # Apply security hardening
+  if [[ "$DEPLOY_HARDENING" == "yes" ]]
+  then
```
git-harry: Can we set this to false if $DEPLOY_HARDENING != "yes"? Just so no one gets caught out.

git-harry: if [[ "$DEPLOY_HARDENING" == "yes" ]]
then
 set it to true
else
 set it to false
fi

d34dh0r53:
recheck_ceph

d34dh0r53:
Ceph is failing with:

```
16:04:44 TASK: [ceph.ceph-common | generate ceph configuration file] ******************* 
16:04:44 fatal: [jrpcaioiad-e7b_ceph_mon_container-8a415f1c] => module config_template not found in configured module paths
16:04:44 fatal: [jrpcaioiad-e7b_ceph_mon_container-5516f77b] => module config_template not found in configured module paths
16:04:44 fatal: [jrpcaioiad-e7b_ceph_mon_container-814df850] => module config_template not found in configured module paths
16:04:44 
16:04:44 FATAL: all hosts have already failed -- aborting
16:04:44 
16:04:44 PLAY RECAP ******************************************************************** 
16:04:44            to retry, use: --limit @/jenkins/ceph-all.retry
16:04:44 
16:04:44 jrpcaioiad-e7b_ceph_mon_container-5516f77b : ok=37   changed=4    unreachable=1    failed=0   
16:04:44 jrpcaioiad-e7b_ceph_mon_container-814df850 : ok=37   changed=4    unreachable=1    failed=0   
16:04:44 jrpcaioiad-e7b_ceph_mon_container-8a415f1c : ok=37   changed=4    unreachable=1    failed=0   
```

sigmavirus24:
This is failing on Travis because it cannot find the `pip_lock_down` role. I don't understand how that's related to this PR though. Master seems to be passing as well

d34dh0r53:
recheck_upgrade

git-harry:
@major I'm closing this pull request and replacing it with [1] so increase the chances of getting this merged today. Your commit was referencing a commit on the OSA mitaka branch which was causing our gating to fail. You made the pull request from a branch on your own repo which meant I was unable to update it, the contributing guidelines requests creating the branch in this repo [2]. I've also tweaked the script so that, if run multiple times, multiple var lines aren't created. The other thing is I've created an issue for this pull request, which is the current expected practice [3].

[1] https://github.com/rcbops/rpc-openstack/pull/1000
[2] https://github.com/rcbops/rpc-openstack/blob/master/CONTRIBUTING.md#fixing-a-bug
[3] https://github.com/rcbops/rpc-openstack/issues/999

Issue #995, opened 2016-04-14
Set galera_wait_timeout to 3600 to match SQLAlchemy


BjoernT:
##### What version (commit) do you encounter this issue on?

All

##### What type of issue is this (bug, feature, documentation, regression)?

Bug

##### What is the expected behavior?

Unnecessary sql connections should not stay open for 8 hours when SQL Alchemy actually recycles the connection after an hour

##### What is actually happening?

Unused SQL Alchemy connections stay open for 8 hours causing a high floor if idling sessions
 
##### How can we reproduce this issue?

If the idling sessions (sleeping) get dropped after an hour

##### If this is a large issue, what can be done immediately to alleviate the issue while working on a longer-term solution?

Minor issue

##### What is your evaluation on the importance of this issue, and why?

-

###### Remove branches (lines) you do not wish to target.
- [x] Master
- [x] liberty-12.2
- [ ] kilo

prometheanfire:
https://review.openstack.org/307231

prometheanfire:
Upstream merged master, @BjoernT you want to backport upstream as well?

git-harry:
This has been addressed upstream for Mitaka and Liberty but will need overriding in Kilo if required in that release.

BjoernT:
Being addressed upstream at https://review.openstack.org/#/c/307231/. Closing here

PR #996, opened 2016-04-14
Set galera_wait_timeout override to 3600

BjoernT:
This Galera setting uses the same timeout for SQl connection
as configured inside SQL Alchemy (pool_recycle) and prevent possibly
SQL session build up due to a large default setting (8 Hours)

Closes-Bug: #995


git-harry:
If this value is generally more appropriate, maybe this should be changed upstream.

odyssey4me:
@git-harry Agreed.

BjoernT:
Yeah commiting upstream would be OK. See  https://review.openstack.org/307231

git-harry:
Closing because this is being addressed in OSA.

Issue #991, opened 2016-04-13
repo-pip-setup doesnt run on haproxy hosts, causes beaver.yml play to fail


jacobwagner:
##### What version (commit) do you encounter this issue on?
11.1.6, kilo
##### What type of issue is this (bug, feature, documentation, regression)?
bug
##### What is the expected behavior?
all proper pip.conf files get laid down on all hosts that will require rpc specific pip packages
##### What is actually happening?
haproxy host is not getting rpc specific repo links in pip.conf
##### How can we reproduce this issue?
install a kilo release on a lab that has an haproxy host only hosting ha proxy, then install logging on lab
##### What is your evaluation on the importance of this issue, and why?
low, logging plays will not finish until it is fixed
###### Remove branches (lines) you do not wish to target.
- [x] kilo

Only runs on hosts:all_containters
https://github.com/rcbops/rpc-openstack/blob/kilo/rpcd/playbooks/repo-pip-setup.yml#L20

Runs on all
https://github.com/rcbops/rpc-openstack/blob/kilo/rpcd/playbooks/beaver.yml#L17

Currently, hosts in openstack_inventory doesnt contain the ha_proxy host

```
    "hosts": {
        "children": [
            "metering-compute_hosts",
            "storage_hosts",
            "metering-infra_hosts",
            "os-infra_hosts",
            "osds_hosts",
            "network_hosts",
            "log_hosts",
            "infra_hosts",
            "swift-proxy_hosts",
            "compute_hosts",
            "swift_hosts",
            "identity_hosts",
            "repo-infra_hosts",
            "shared-infra_hosts",
            "storage-infra_hosts",
            "mons_hosts"
        ],
        "hosts": []
    },
```

prometheanfire:
since haproxy isn't supported (to my knowlege) I'd reset the importance to low.

jacobwagner:
@prometheanfire done, good point

git-harry:
Considering haproxy is not supported and this won't affect AIO's I'm not sure there's necessarily something to fix. @jacobwagner why has this issue now appeared for you? Are you now performing additional forms of testing on a Kilo build?

prometheanfire:
If it's not supported then we should remove it from use entirely.

prometheanfire:
it's over

PR #997, opened 2016-04-14
Deploy beaver only to hosts where pip is laid down

prometheanfire:
Deploy beaver only to hosts where pip is laid down

Beaver depends on the pip.conf gile being laid down.  Therefore we
should only deploy beaver to the same places pip.conf is laid down.

https://github.com/rcbops/rpc-openstack/blob/0c26acf4766c716295ed6a4e4db107512fc5841d/rpcd/playbooks/beaver.yml#L17

related bug: #991

### DIFF
--- a/rpcd/playbooks/beaver.yml
+++ b/rpcd/playbooks/beaver.yml
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 - name: Setup beaver log shipper
-  hosts: all
+  hosts: hosts:all_containers
   max_fail_percentage: 20
   user: root
   roles:


sigmavirus24:
:shipit:

odyssey4me:
👎 Putting a pip.conf on all containers and hosts will results in pip install failures in unexpected ways by the various OSA tasks as they are designed to trust the pip.conf for the source that has all the files it needs. For each hosts/container the required pip conf is different (eg: the repo containers need access to pypi, whereas other containers do not).

I would recommend that the pip installation actions for beaver (and any other RPC-O tasks) be specific about what pip sources are trusted.

git-harry:
Looking at the list of log monitors, it doesn't appear that beaver would capture any haproxy specific logs. Maybe it would be better to change the beaver playbook to 'hosts:all_containers'?
The commit message doesn't provide a great deal of detail. The commit message title should start with a capital letter. There is no reference to the issue that is being addressed.

prometheanfire:
I'm not sure all containers is right either, because then it wouldn't check things like cinder-volume or nova-compute which are bare metal (I think).

prometheanfire:
I'll fix the commit message when we decide on the course of action, no reason to change it just to have to change something else.

stevelle:
@odyssey4me is right about the issues here.  Do we need a new inv group "beaver_hosts" or a role that is "beaver_clients" or the like?

prometheanfire:
good question, what do you think?

stevelle:
I'd say inv group is a good start. let's look at a solution using that first.

prometheanfire:
Would changing the target for where beaver gets installed be a better fix for this, to match where pip goes?  Instead of changing where pip goes to match beaver.

stevelle:
I dunno. Maybe @d34dh0r53 can help.

odyssey4me:
I don't think you need new groups, although you could do something like that. I think it would make more sense to include extra options in the pip install task to force it to use the local repo. Something like:

```
  --no-index
  --find-links "{{ openstack_repo_url }}/os-releases/{{ openstack_release }}"
  --trusted-host "{{ openstack_repo_url | netloc_no_port }}"
```

prometheanfire:
Updated, now beaver only gets laid down on hosts where pip is laid down.

prometheanfire:
14:24:08 /usr/sbin/xenstore-read fail

recheck_ceph
recheck_swift

prometheanfire:
recheck_ceph
recheck_swift

prometheanfire:
recheck_ceph

    15:30:46 sudo: lxc-attach: command not found

mancdaz:
Won't this change mean that beaver now doesn't go on compute hosts, and therefore no nova-compute logs will get shipped to logstash?

prometheanfire:
I'll need to look at an inventory to be sure, but I think we still would ship logs.  all_containers isn't actually just containers I think.

mancdaz:
Ah yeah I think you're right. In that case I can see how this change would help the issue, but I think the commit message and [less importantly] PR description need a bit of work.

rpcd/playbooks/beaver.yml
```diff
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 - name: Setup beaver log shipper
-  hosts: all
+  hosts: hosts:all_containers
   max_fail_percentage: 20
```
mancdaz: should this just be 
```hosts: all_containers```
?

prometheanfire: It matches where the pip.conf gets laid down.  I'd rather hit more than less.

prometheanfire:
I'll update the commit/description later tonight or tomorrow

prometheanfire:
recheck_maas

prometheanfire:
oh, didn't need to do maas, ah well

done

mancdaz:
@prometheanfire a little of the problem description in the commit message would be useful for those trolling history. From the contributing guidelines:

_"Do not assume the reviewer understands what the original problem was"_

Other than that, I now understand the host pattern so once the comment message is fleshed out I'll be happy.

prometheanfire:
@mancdaz updated

mancdaz:
@prometheanfire personally I'd go with even more verbosity "...this caused beaver installs to fail because the host didn't have the correct pip configuration to point at the rpc pip repo, where the beaver packages are located. This patch addresses the problem by ensuring that we only install beaver on hosts that have the correct pip.conf, and can therefore access the beaver package from the repo"

From contributing:

_"The commit message must contain all the information required to fully understand & review the patch for correctness. Less is not more. More is More"_

prometheanfire:
I thought I did do more, I'm just not a very verbose person.

d34dh0r53:
@prometheanfire `-vvvv`

mancdaz:
👍  thanks!

d34dh0r53:
:+1:

Issue #999, opened 2016-04-15
Add security hardening functionality


git-harry:
##### What version (commit) do you encounter this issue on?
master
##### What type of issue is this (bug, feature, documentation, regression)?
feature
##### What is the expected behavior?
Liberty now includes the ability to use openstack-ansible-security, this is a role to apply security best practices host configurations. RPCO needs updating to be able to selectively deploy this capability.
##### What is actually happening?
Currently the openstack-ansible submodule does not include to relevant commit and deploy.sh has no ability to control the use of this role.
##### How can we verify your issue is addressed?
Once updated the role will apply the security best practices it defines, this will be observed in the logs and config.
##### What is your evaluation on the importance of this issue, and why?
High, this is a requirement for 12.2.0
###### Remove branches (lines) you do not wish to target.
- [x] Master

prometheanfire:
merged

PR #1000, opened 2016-04-15
Add host security hardening functionality

git-harry:
openstack-ansible Liberty has backported the ability to use the role
openstack-ansible-security [1].

This commit updates the openstack-ansible submodule to include the new
security functionality. A new variable, DEPLOY_HARDENING, is added to
deploy.sh to determine whether this new functionality is applied to a
deployment. It is used to set the var apply_security_hardening in
user_variables.yml which controls whether or not the playbook
security-hardening.yml is run as part of setup-hosts.yml.

[1] https://git.openstack.org/openstack/openstack-ansible-security

Closes-issue: https://github.com/rcbops/rpc-openstack/issues/999

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -13,6 +13,7 @@ export DEPLOY_TEMPEST=${DEPLOY_TEMPEST:-"no"}
 export DEPLOY_CEILOMETER="no"
 export DEPLOY_CEPH=${DEPLOY_CEPH:-"no"}
 export DEPLOY_SWIFT=${DEPLOY_SWIFT:-"yes"}
+export DEPLOY_HARDENING=${DEPLOY_HARDENING:-"no"}
 export FORKS=${FORKS:-$(grep -c ^processor /proc/cpuinfo)}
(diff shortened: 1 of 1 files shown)


git-harry:
recheck_swift

git-harry:
recheck_swift
recheck_ceph
recheck_upgrade

scripts/deploy.sh
```diff
@@ -103,6 +104,24 @@ if [[ "${DEPLOY_AIO}" == "yes" ]]; then
   rm -f /etc/openstack_deploy/conf.d/ceilometer.yml
 fi
 
+# Apply host security hardening with openstack-ansible-security
+# The is applied as part of setup-hosts.yml
+if [[ "$DEPLOY_HARDENING" == "yes" ]]
```
sigmavirus24: I think by this point we have ansible bootstrapped, yes? Why not simply use `lineinfile` to manage this section? I think it would be more readable, understandable, and maintainable that way.

sigmavirus24:
I'm :+1: on this as a first pass, but I really think a tiny invocation with lineinfile would be better going forward.

d34dh0r53:
:+1: good on this as a first pass as well.

PR #1009, opened 2016-04-20
Update openstack-ansible SHA

git-harry:
Update the OSA SHA to that used by r12.1.rc3 which is the latest rc tag
of the next minor release.


git-harry:
recheck_ceph

mancdaz:
👍 based on conversations with @git-harry, it makes sense to bump the OSA sha in the 12.0 branch only up to where 12.1 is currently pointing, else the 12.0 branch could get ahead of the 12.1 branch whilst 12.1 is still in rc phase.

sigmavirus24:
This seems reasonable to me as well.

PR #1010, opened 2016-04-20

git-harry:

related bug #967

(cherry picked from commit 68621a5b8375a386f33f82a82a5760f0da5d1ba4)
(cherry picked from commit 56a272d037719ff73a2c79d493b71d95211035cf)


git-harry:
recheck_swift

mattt416:
The double `cherry picked from` is a bit confusing, but otherwise LGTM.

mancdaz:
👍 as this is a cherry-pick, and I originally also approved of the master commit.

Issue #1011, opened 2016-04-20
Elasticsearch URL double quoted


BjoernT:
##### What version (commit) do you encounter this issue on?

r12.0.0

##### What type of issue is this (bug, feature, documentation, regression)?

Bug

##### What is the expected behavior?

Remove one level of quoting inside the template

##### What is actually happening?

double quoting

##### How can we reproduce this issue?
---

##### How can we verify your issue is addressed?

Kibana works through a tunnel

##### If this is a large issue, what can be done immediately to alleviate the issue while working on a longer-term solution?

Minor

##### What is your evaluation on the importance of this issue, and why?
-

###### Remove branches (lines) you do not wish to target.
- [x] Master
- [x] liberty-12.2
- [x] liberty-12.1
- [x] kilo

PR #1012, opened 2016-04-20
Removing one level of elasticsearch_url quoting

BjoernT:
The default value for elasticsearch_url already includes the correct
quotes so there is no need for additional quoting inside the config.js
kibana template.

Closes-Bug: #1011


prometheanfire:
lgtm, @d34dh0r53 ?

:+1:

prometheanfire:
@BjoernT the tense of the top line of the commit message should likely be 'remove' instead of 'removing'.

d34dh0r53:
:+1:

Issue #1028, opened 2016-04-25
[TRACKING] Beaver repo_pip_lockdown dep causes repo rebuild to fail


hughsaunders:
##### What version (commit) do you encounter this issue on?
Upgrades within liberty (eg r12.2.0rc1 --> master)

##### What type of issue is this (bug, feature, documentation, regression)?
bug

##### What is the expected behavior?
The repo server does not have its pip conf locked down to itself, preventing future repo builds from failing. This also means that Minor upgrades will be successful.

##### What is actually happening?
The beaver role depends on pip_lock_down and is applied to all hosts. This includes the repo server which causes the repo server's pip configuration to be locked down. This causes future repo builds to fail as no external package sources are available.

##### How can we reproduce this issue?
Deploy RPC liberty, then re-run the repo build playbook, alternatively deploy a rpc liberty version than attempt to upgrade to a newer liberty version. 

##### How can we verify your issue is addressed?
Upgrades and re-runs of repo-build are successful  after a full RPC liberty deploy.

##### What is your evaluation on the importance of this issue, and why?
Medium - it needs to be fixed, impact is mostly on gates at the moment, the impact for customers is low as ops can manually remove the pip.conf from the repo servers and continue.

###### Remove branches (lines) you do not wish to target.
- [ ] Master
- [ ] liberty-12.2
- [ ] liberty-12.1

hughsaunders:
There are two things to fix here:

[ ] Remove the pip_lock_down dep from the beaver role
[ ] Add a cleanup step to deploy.sh

hughsaunders:
@git-harry has proposed a fix for this upstream in two patches:
- 1 https://review.openstack.org/#/c/310475/ - Supply environment variable to repo build tasks that overrides pip lock down and allows external repo access.
- 2e https://review.openstack.org/#/c/310479/ - Add pip_lock_down dep back into repo_server role. 

RPC PR no longer required.

git-harry:
There is also https://review.openstack.org/#/c/310690/

prometheanfire:
we might need to make sure this is backported upstream

prometheanfire:
needs backport to mitaka and liberty

git-harry:
The upstream changes proposed above impact gating and there is no guarantee the work will be backported to liberty.

This project needs to address the issue locally pending the outcome of the upstream fix.

prometheanfire:
I think only  Remove the pip_lock_down dep from the beaver role is needed now

[...]

PR #1029, opened 2016-04-25
Fix pip_lock_down on repo servers

hughsaunders:
Issue: https://github.com/rcbops/rpc-openstack/issues/1028

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -133,6 +133,13 @@ openstack-ansible -i "localhost," patcher.yml
 
 # begin the openstack installation
 if [[ "${DEPLOY_OA}" == "yes" ]]; then
+
+  # This deploy script is also used for minor upgrades (within an openstack release)
+  # Some versions of liberty deploy pip lockdown to the repo server, in order for an
+  # upgrade to suceed the pip config must be removed so that repo builds
+  # have access to external repos.
+  ansible repo_all -m file -a 'name=/root/.pip state=absent' 2>/dev/null ||:
+
   cd ${OA_DIR}/playbooks/
 
   # ensure that the ELK containers aren't created if they're not
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "openstack-ansible"]
 	path = openstack-ansible
-	url = https://git.openstack.org/openstack/openstack-ansible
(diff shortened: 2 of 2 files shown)


scripts/deploy.sh
```diff
@@ -133,6 +133,13 @@ openstack-ansible -i "localhost," patcher.yml
 
 # begin the openstack installation
 if [[ "${DEPLOY_OA}" == "yes" ]]; then
+
+  # This deploy script is also used for minor upgrades (within an openstack release)
+  # Some versions of liberty deploy pip lockdown to the repo server, in order for an
+  # upgrade to suceed the pip config must be removed so that repo builds
+  # have access to external repos.
+  ansible repo_all -m file -a 'name=/root/.pip state=absent'
```
git-harry: This causes the initial deployment to fail because the containers don't exist.
I think it would be worth putting a note saying when this should be removed. Do we still work on the assumption that support deploy all minors? If so this would only need to be in 12.2, otherwise it needs to be here till 13.1.

hughsaunders: Ah yes, I forgot to ||: that.

hughsaunders: Done

hughsaunders:
Hosts which have run pip_lock_down can only install beaver <=36.2.0 (due to missing paho-mqtt dep?) hosts which have not run pip_lock_down can only install beaver>=36.2.0 (due to mosquitto dep in older versions). Have added Beaver 36.2.0 to rpc-repo in an attempt to allow both locked and unlocked hosts to install 36.2.0.

recheck_all

git-harry:
@hughsaunders I don't like the beaver package being forced to change by this. I wonder if it would be better to address this in OSA by applying the pip_lock_down role the the repo servers and then overriding it for repo building pip tasks, e.g. https://github.com/git-harry/os-ansible-deployment/commit/1746506c39da8fc951a4c4b26359d4af4cc1781b

git-harry:
recheck_upgrade

git-harry:
@hughsaunders this seems to work. I've made a couple of reviews to try and get this in upstream. It would be helpful if you would review them.
https://review.openstack.org/#/c/310479/
https://review.openstack.org/#/c/310475/

git-harry:
Reopened for additional testing

Issue #1026, opened 2016-04-22
openstack-ansible-security role is missing when upgrading from 12.1 -> 12.2


git-harry:
##### What version (commit) do you encounter this issue on?
12.2
##### What type of issue is this (bug, feature, documentation, regression)?
bug
##### What is the expected behavior?
Running the deploy.sh script in 12.2 on an existing 12.1 deployment should result in the script exiting after successfully running the playbooks.
##### What is actually happening?
The 12.2 version of deploy.sh fails with 
```
23:57:54 + run_ansible setup-hosts.yml
23:57:54 + openstack-ansible --forks 4 setup-hosts.yml
23:57:54 Variable files: "-e @/etc/openstack_deploy/user_apt_conf_files.yml -e @/etc/openstack_deploy/user_extras_secrets.yml -e @/etc/openstack_deploy/user_extras_variables.yml -e @/etc/openstack_deploy/user_secrets.yml -e @/etc/openstack_deploy/user_variables.yml "
23:57:55 ERROR: cannot find role in /jenkins/workspace/RPC-AIO/openstack-ansible/playbooks/roles/openstack-ansible-security or /jenkins/workspace/RPC-AIO/openstack-ansible/playbooks/openstack-ansible-security or /etc/ansible/roles/openstack-ansible-security
```
This is taken from http://jenkins.propter.net/job/RPC-AIO/4361/consoleFull
##### How can we reproduce this issue?
Deploy an AIO from r12.1.0rc3
Checkout liberty-12.2 and run deploy.sh
##### How can we verify your issue is addressed?
Running 12.2 deploy.sh on an existing 12.1 deployment will cause openstack-ansible-security to be downloaded.
##### If this is a large issue, what can be done immediately to alleviate the issue while working on a longer-term solution?
To mitigate this, the script openstack-ansible/scripts/bootstrap-ansible.sh can be run prior to running deploy.sh.
##### What is your evaluation on the importance of this issue, and why?
High, this impacts the 12.2 release
###### Remove branches (lines) you do not wish to target.
- [x] Master
- [x] liberty-12.2
- [x] liberty-12.1

PR #1032, opened 2016-04-26
Bootstrap Ansible whenever deploy.sh is run

git-harry:
openstack-ansible/scripts/bootstrap-ansible.sh is used to setup Ansible
which includes downloading the roles required by openstack-ansible
(OSA).

deploy.sh is updated so that bootstrap-ansible.sh runs whenever
deploy.sh does. This resolves an issue where the role
openstack-ansible-security is not getting downloaded because OSA
introduced it as part of a patch release.

This change should also ensure that all the OSA required roles are at
the correct versions as well as Ansible itself.

Issue: https://github.com/rcbops/rpc-openstack/issues/1026
(cherry picked from commit c756edd0a78ccc1ad737fb5b8d834c2cec522100)

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -32,7 +32,7 @@ function run_ansible {
 cd ${OA_DIR}
 
 # bootstrap ansible and install galaxy roles (needed whether AIO or multinode)
-which openstack-ansible || ./scripts/bootstrap-ansible.sh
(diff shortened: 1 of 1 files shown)


prometheanfire:
simple backport :+1:

tomc603:
Same as before, this looks fine. 👍

PR #1034, opened 2016-04-26
Include commit message checks from hacking

sigmavirus24:
Hacking 0.9 was the last version that included commit title checks.
After that, they were removed from hacking and so we need to vendor
these into our own project.

### DIFF
--- a/hacking/rpco_checks.py
+++ b/hacking/rpco_checks.py
@@ -1,0 +1,111 @@
+#  Licensed under the Apache License, Version 2.0 (the "License"); you may
+#  not use this file except in compliance with the License. You may obtain
+#  a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#  WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#  License for the specific language governing permissions and limitations
+#  under the License.
+"""Contain all custom extensions to hacking and flake8 for RPCO."""
+
+import os
+import re
+import subprocess
+
+from hacking import core
+
+
+class GitCheck(core.GlobalCheck):
+    """Base-class for Git related checks."""
+
+    def _get_commit_title(self):
+        # Check if we're inside a git checkout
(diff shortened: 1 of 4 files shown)


hacking/rpco_checks.py
```diff
@@ -0,0 +1,117 @@
...
+    a 72 character limit
+
+    H802 Title limited to 72 chars
+    """
+    name = "GitCheckCommitTitleLength"
+
+    def run_once(self):
+        title = self._get_commit_title()
+
+        if title and len(title) > 72:
```
git-harry: Why have this at 72 if the limit is 50?

git-harry:
This could do with an issue being created.

sigmavirus24:
recheck

git-harry:
recheck_all

tox.ini
```diff
@@ -23,5 +24,5 @@ commands =
 #     F403 'from ansible.module_utils.basic import *' used; unable to detect undefined names
 #     H303  No wildcard (*) import.
 # Excluding our upstream submodule, and our vendored f5 configuration script.
-ignore=F403,H303
+ignore=F403,H303,R
```
git-harry: What does R refer to?

sigmavirus24: `R` is the prefix for our checks. This is bit of a hack to avoid printing out those messages more than once per run. By default we ignore the checks we add for the case where we find all python files and run the default checks against them. Then we run flake8 a second time in serial just adding the `R` checks so that those messages are printed once and only once. It's a hack but it works well.

git-harry: Ah, okay that makes sense. I think it would be worth adding something along those lines to the comment above it, ',R' looks more like a typo.

I've tested the patch and it all seems good so other than the above I think it's good to go.

Also, currently gating is broken by https://github.com/rcbops/rpc-openstack/issues/1037 so the checks will fail.

sigmavirus24: Done

git-harry:
👍

prometheanfire:
reckeck_all

prometheanfire:
recheck_all

git-harry:
recheck_all

sigmavirus24:
Everything is passing and we have 1 👍 ... anyone for a second 👍 ?

Issue #1025, opened 2016-04-22
Memcached monitoring should be using curr_connections instead of total_connections


JCallicoat:
Memcached monitoring should be using curr_connections, which is the current active connection count, instead of total_connections, which is the lifetime total connections, and should alert when curr_connection is close to the configured max in memcached_connections.

##### What version (commit) do you encounter this issue on?

r11.1.5 (and all others)

##### What type of issue is this (bug, feature, documentation, regression)?

Feature (and bug?) / Bug (and feature?)

##### What is the expected behavior?

MaaS memcached plugin checks the currently active memcached connections.

##### What is actually happening?

MaaS memcached plugin checks the total lifetime memcached connections.

##### How can we reproduce this issue?

Run the plugin and look at the metric

##### How can we verify your issue is addressed?

Run the plugin and look at the metric

##### If this is a large issue, what can be done immediately to alleviate the issue while working on a longer-term solution?

N/A

##### What is your evaluation on the importance of this issue, and why?

Medium. Customers are impacted when the memcached connection limit is hit. Alerts are generated because memcached cannot open a new connection to provide stats to the plugin.

##### What branches / milestones do you believe this issue should target?

All supported

###### Remove branches (lines) you do not wish to target.
- [x] Master
- [x] liberty-12.2
- [x] liberty-12.1
- [x] kilo

Reference issue #1023

prometheanfire:
#1023 as a refrence

prometheanfire:
so, changing the metric is easy, do we need to create a new alarm as well?

BjoernT:
Are we going to backport this ?

prometheanfire:
yes, need master to merge first though

prometheanfire:
merged, closing :D

PR #1038, opened 2016-05-02
Change memcached monitoring to monitor curr_conns

prometheanfire:
Memcached previously monitored total connections, this is does not
tell us useful information, so I changed it to monitor current
connections and also alarm (WARNING) against reaching 90% of the
maximum active connection limit.

Related: #1025

### DIFF
--- a/rpcd/playbooks/roles/rpc_maas/templates/memcached_status.yaml.j2
+++ b/rpcd/playbooks/roles/rpc_maas/templates/memcached_status.yaml.j2
@@ -15,3 +15,11 @@ alarms      :
             if (metric["memcache_api_local_status"] != 1) {
                 return new AlarmStatus(CRITICAL, "memcache unavailable");
             }
+    memcache_curr_connections :
+        label                   : memcache_curr_connections--{{ ansible_hostname }}
+        notification_plan_id    : "{{ maas_notification_plan }}"
+        criteria                : |
+            :set consecutiveCount={{ maas_alarm_local_consecutive_count }}
+            if (metric["memcache_curr_connections"] > {{ ((memcached_connections*0.9)|round|int) }}) {
+                return new AlarmStatus(WARNING, "memcached connection count is >= 90% of max");
+            }
--- a/maas/plugins/README.md
+++ b/maas/plugins/README.md
@@ -275,7 +275,7 @@ IP address of memcached server
     metric memcache_total_items uint64 563324 items
     metric memcache_get_hits uint64 4543534 hits
     metric memcache_get_misses uint64 2346565 misses
-    metric memcache_total_connections uint64 42565 connections
+    metric memcache_curr_connections uint64 42565 connections
 
 ***
 #### horizon_check.py
--- a/maas/plugins/memcached_status.py
+++ b/maas/plugins/memcached_status.py
@@ -31,7 +31,7 @@ VERSIONS = ['1.4.14 (Ubuntu)', '1.4.15']
 MEMCACHE_METRICS = {'total_items': 'items',
                     'get_hits': 'cache_hits',
                     'get_misses': 'cache_misses',
-                    'total_connections': 'connections'}
+                    'curr_connections': 'connections'}
 
 
 def item_stats(host, port):
--- a/maas/testing/tests/integration/maas_local
+++ b/maas/testing/tests/integration/maas_local
@@ -876,6 +876,15 @@
                         "disabled": false,
                         "metadata": null,
                         "scheduled_suppressions": []
+                    },
+                    "memcache_curr_connections--CONTROLLER_memcached_container_UID": {
+                        "active_suppressions": [],
+                        "confd_hash": null,
+                        "confd_name": null,
+                        "criteria": ":set consecutiveCount=3 if (metric[\"memcache_curr_connections\"] > 922) { return new AlarmStatus(WARNING, \"memcached connection count is >= 90% of max\"); }",
+                        "disabled": false,
+                        "metadata": null,
+                        "scheduled_suppressions": []
                     }
                 },
                 "confd_hash": null,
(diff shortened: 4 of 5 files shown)


prometheanfire:
check_maas

claco:
When we move from `memcache_total_connections` to `memcache_curr_connections`, I assume the conf.d style of changes to the maas agent will delete the old alarm? IF not, we should doc that human steps need to be taken.

Also, why `memcache_curr_connections` and not `memcache_current_connections`?

prometheanfire:
My maas knowlege is limited, so no clue if it'll remove automatically now.

Also, if people care about the name so much they should update the issue, which is the spec for this work :P

prometheanfire:
recheck_all
recheck_maas

prometheanfire:
recheck_all
recheck_maas

prometheanfire:
recheck_all
recheck_maas

prometheanfire:
recheck_maas

prometheanfire:
@git-harry @hughsaunders 
maas testing broken?

`22:03:10 + raxmon-entities-list --debug
22:03:10 No username/API key or auth token/API URL provided!
22:03:10 You need to either put credentials in ~/.raxrc or pass them to the command using --username/--api-key or --auth-token/--api-url options`

git-harry:
@prometheanfire that line can be ignored, it was added for debugging gating.

It looks like there is a genuine failure here, the patch does not appear to work

21:59:08 fatal: [jrpcaioiad-751_memcached_container-0c4e6290 -> jrpcaioiad-751] => {'msg': "AnsibleUndefinedVariable: One or more undefined variables: 'memcached_connections' is undefined", 'failed': True}
21:59:08 fatal: [jrpcaioiad-751_memcached_container-0c4e6290 -> jrpcaioiad-751] => {'msg': 'One or more items failed.', 'failed': True, 'changed': False, 'results': [{'skipped': True, 'changed': False}, {'skipped': True, 'changed': False}, {'msg': "AnsibleUndefinedVariable: One or more undefined variables: 'memcached_connections' is undefined", 'failed': True}]}

prometheanfire:
do I need to move this somewhere then?

openstack-ansible/playbooks/roles/memcached_server/defaults/main.yml:memcached_connections: 1024

prometheanfire:
recheck_maas

prometheanfire:
recheck_maas

prometheanfire:
forgot to make the maas alarm a separate alarm, included it in the other alarm, fixed now (two alarms)

sigmavirus24:
> forgot to make the maas alarm a separate alarm, included it in the other alarm, fixed now (two alarms)

It looks like you didn't push this though

prometheanfire:
no, I pushed that through

rpcd/playbooks/roles/rpc_maas/templates/memcached_status.yaml.j2
```diff
@@ -15,3 +15,11 @@ alarms      :
             if (metric["memcache_api_local_status"] != 1) {
                 return new AlarmStatus(CRITICAL, "memcache unavailable");
             }
+    memcache_curr_connections :
```
sigmavirus24: Shouldn't this be a separate conf.d template?

prometheanfire: I don't think so, you can have multiple alarms off of one check (I thought)

prometheanfire:
fixed it, was doing the math outside of ansible, which maas doesn't support

prometheanfire:
recheck_maas

prometheanfire:
recheck_maas

not the best at intrepreting this output, but I think it was in warning state?

`failed: [jrpcaioiad-206] => {"attempts": 10, "changed": false, "failed": true, "failed_when_result": true, "rc": 1, "stdout_lines": ["--- 2016-05-11 00:46:12.765940 ---", "Checks and Alarms with failures:", " - Entity:enS8k2IeKz:jrpcaioiad-206", "   - Check:agent.plugin:chTY6qsOqS:memcached_status--jrpcaioiad-206_memcached_container-a27087cc: Completed:True Status:okay", "     ! Alarm:al83y9xOx8:memcache_curr_connections--jrpcaioiad-206_memcached_container-a27087cc:WARNING"]}`

prometheanfire:
same thing http://jenkins.propter.net/job/RPC-AIO/5600/console

git-harry:
I believe it's telling you that the alarm has fired. The expectation is that the deployment should finish with all alarms in the okay state. If default values are not appropriate for an AIO they should be adjusted when building an AIO.

prometheanfire:
What's odd is that it was OK when I was testing on on an AIO...

prometheanfire:
Also, given the nature of the alarm and that it's for memcache, which I don't think we change anywhere, I don't think this is AIO related.  I wonder if it is just in a warning state at first.

prometheanfire:
recheck_maas

prometheanfire:
well, increased the limit and still getting the warning

rpcd/playbooks/roles/rpc_maas/templates/memcached_status.yaml.j2
sigmavirus24: I misread 

> forgot to make the maas alarm a separate alarm, included it in the other alarm, fixed now (two alarms)

This is fine

prometheanfire:
recheck_maas

prometheanfire:
my problem was an order of operations thing in jinja, and casting

    if (metric["memcache_curr_connections"] > 0) {

1024*0.9|int is 0, because int just truncates and 0.9|int is 0
(1024*0.9) to generate 921.6, then round it correctly to 922.0, then convert to int (922)

d34dh0r53:
:+1:

tomc603:
I feel like the next bug filed should be for a configurable threshold. Either way 👍

prometheanfire:
the threshold is configurable, but for testing it's static

Issue #536, opened 2015-10-31
disable innobackupex --version-check


ihavegerms:
When rebuilding a failed galera container, the joiner attempts to pull updates from the donor but the donor fails due to what appears to be a permission issue with the percona-version-check in /tmp. After modifying the permissions to mysql:mysql, the transfer succeeds and the container can then join the cluster.

A few others have ran in to this same issue as well. Is it necessary that we keep this enabled? 

- [x] kilo

JCallicoat:
Does this need more information? It is still an issue when version checking is enabled for Galera.

SATroll:
This is still an issue.

claco:
Should we be disabling version check, or should we be changing the location instead? (--tmpdir)

BjoernT:
I think this is still a valid issue. If the removal of the version check fixes this issue, it debatable. I seem to recall a percona fix around that. I guess we are open to all ways of fixing this issue

prometheanfire:
Are you getting this issue when running a play or running something manually?  Do you hit this when simply rebuilding the container instead of trying to rejoin an existing contianer?

BjoernT:
Usually we get this issue after deploying the cluster, at least I did not encounter it during play run. Suddenly replication dies with this error after running galera for few days/weeks

prometheanfire:
So why is this check running in the first place?

JCallicoat:
It is enabled by default as of xtrabackup 2.1.6.

https://www.percona.com/doc/percona-xtrabackup/2.1/release-notes/2.1/2.1.6.html

I believe we'd need to add the following to one of the mysql config files if we want to disable it.

```
[xtrabackup]
no-version-check=true
```

prometheanfire:
if we set this value it'll likely need to go into this file

playbooks/roles/galera_server/templates/cluster.cnf.j2

we may not need to set it based on what @BjoernT for newer versions of galera (they fix this possibly)  so maybe kilo only?

prometheanfire:
I think before we submit this bug upstream (because that's where the fix would go) we need to understand why this is happening.  I'm hesitant to disable something when we don't know the root cause.

JCallicoat:
Upstream closed as Won't Fix at https://bugs.launchpad.net/openstack-ansible/+bug/1511910 since it's a known upstream bug with the package versions we use.

I think we were looking for an RPC-specific band aid.

For reference, here's an older bug mentioning the same issue and work arounds:

[...]

PR #1039, opened 2016-05-02
Disable xtrabackup version check

prometheanfire:
Galera upstream has known bug in mariadb 5.5 where the permissiions
on the version check file in /tmp will be reset, causing errors.
Disable the check for kilo, where the error exists.  This is fixed
in liberty where we use maria 10.

Closes: #536

### DIFF
--- a/rpcd/etc/openstack_deploy/user_variables.yml
+++ b/rpcd/etc/openstack_deploy/user_variables.yml
@@ -55,6 +55,9 @@ keystone_token_driver: "keystone.token.persistence.backends.sql.Token"
 
 # Galera overrides
 galera_cluster_name: rpc_galera_cluster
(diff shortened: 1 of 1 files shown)


prometheanfire:
recheck_ceph
recheck_swift

tomc603:
recheck_ceph

prometheanfire:
recheck_ceph
recheck_swift

git-harry:
recheck_ceph
recheck_swift

git-harry:
recheck_ceph

tomc603:
Disabling version check until we get a newer, better Galera package in a newer, better RPC-O relying on a newer OSA version. 👍

d34dh0r53:
👍

Issue #1046, opened 2016-05-03
handle_internal_only_routers behavior changes


BjoernT:
##### What version (commit) do you encounter this issue on?

r12.x

##### What type of issue is this (bug, feature, documentation, regression)?

Introduced in OSA we silently changed another neutron behavior from default (handle_internal_only_routers)  see https://bugs.launchpad.net/openstack-ansible/+bug/1577868 and https://bugs.launchpad.net/neutron/+bug/1572390

##### What is the expected behavior?

Routers do start regardless of external network connectivity

##### What is actually happening?

Router do not start when no external network is attached resulting in not working setups

##### How can we reproduce this issue?

Create L3 agent with no external network attached and reboot infra node

##### How can we verify your issue is addressed?

Router becomes operational and is assigned to a neutron agent

##### If this is a large issue, what can be done immediately to alleviate the issue while working on a longer-term solution?

Minor implementation

##### What is your evaluation on the importance of this issue, and why?
High prio

###### Remove branches (lines) you do not wish to target.
- [x] Master https://github.com/rcbops/rpc-openstack/pull/1047
- [x] liberty-12.1 https://github.com/rcbops/rpc-openstack/pull/1194
- [x] liberty-12.2 https://github.com/rcbops/rpc-openstack/pull/1195

PR #1047, opened 2016-05-03
Enable handle_internal_only_routers by default

BjoernT:
The default value for handle_internal_only_routers changed from True to
False in OSA upstream, causing a unexpected behavior.
RPC-O enables handle_internal_only_routers in order allow Neutron
router without having to attach external networks. This is a valid
customer use case.

Closes-Bug: #1046

### DIFF
--- a/rpcd/etc/openstack_deploy/user_variables.yml
+++ b/rpcd/etc/openstack_deploy/user_variables.yml
@@ -90,7 +90,13 @@ neutron_neutron_conf_overrides:
   nova:
     endpoint_type: internal
 
-# Diasble port security binding for liberty. See https://bugs.launchpad.net/neutron/+bug/1509312
+# Disable port security binding for liberty. See https://bugs.launchpad.net/neutron/+bug/1509312
 neutron_ml2_conf_ini_overrides:
     ml2:
         extension_drivers: ''
+
+# Enable handle_internal_only_routers based on expected behavior since Juno
+# See https://bugs.launchpad.net/openstack-ansible/+bug/1577868
+neutron_l3_agent_ini_overrides:
+  DEFAULT:
+    handle_internal_only_routers: True


BjoernT:
recheck

prometheanfire:
recheck_all

prometheanfire:
recheck_swift

BjoernT:
recheck_upgrade

mancdaz:
Hi @BjoernT - what is the upgrade impact here?

We need to understand what the impact is for upgrades to mitaka if we merge this in to master.
If there is some kind of impact, I'd be very wary about backporting to any of the liberty branches.

BjoernT:
There is no impact to upgrades. It retains the default configuration in neutron. It was just plain wrong that OSA changed the value without knowing what it even does

mancdaz:
Hi @BjoernT I mean upgrade impact to any deployments that may have already got the overridden 'False' setting.

BjoernT:
OK, the impact is unknown since we do not always know how customer organize their router.
In worst case, they reboot a neutron agent container and the networks won't come up, at least if they use internal only neutron routers. Hence I would back port it where we can

mancdaz:
@BjoernT if I'm understanding this issue correctly, if any customers are already deployed with the setting as False, then they wouldn't be able to create routers that are not attached to external networks. So is it fair to say that if we then set it to True, there should be no affect to those customers since they never could create those routers in the first place? 

If you're agreed then I will +1 this patch.

BjoernT:
Yes correct, this is a non risk change. On the flip side, not setting it to true will result in issue for customers who are using internal neutron router, which is a OK configuration in RPC. And most people would not even be able to debug this because the error message Neutron reports is absolutely wrong, which why we have a different neutron bug open to fix this so everyone will know what the problem is

BjoernT:
👍

mancdaz:
OK on your confirmation, this makes sense to me 👍 

Note we still need another dev team +1

d34dh0r53:
:+1:

Issue #1037, opened 2016-04-29
Paramiko v2.0 causes bootstrap-ansible.sh to fail


git-harry:
##### What version (commit) do you encounter this issue on?
master
##### What type of issue is this (bug, feature, documentation, regression)?
bug
##### What is the expected behavior?
bootstrap-ansible.sh should install ansible and all its dependencies.
##### What is actually happening?
bootstrap-ansible.sh fails because the new version of paramiko requires additional libraries to support cryptography.

http://jenkins.propter.net/job/RPC-AIO/4825/console
##### How can we reproduce this issue?
Attempt to deploy an AIO.
##### How can we verify your issue is addressed?
bootstrap-ansible.sh should exit after successfully installing ansible.
##### If this is a large issue, what can be done immediately to alleviate the issue while working on a longer-term solution?
Run the following prior to bootstrap-ansible.sh
```
apt-get install build-essential libssl-dev libffi-dev
```
##### What is your evaluation on the importance of this issue, and why?
High, this impacts all branches and all new deployments.
###### Remove branches (lines) you do not wish to target.
- [x] Master
- [x] liberty-12.1
- [x] liberty-12.0
- [x] kilo

git-harry:
There are open reviews to address this in OSA

https://review.openstack.org/#/c/311187/1
https://review.openstack.org/#/c/311193/1

prometheanfire:
we will need to leave this open so we can revert #1051 when 11.2.15 is released with the upstream fix

git-harry:
The required OSA SHA has merged in all Liberty branches.

prometheanfire:
guess that means we can revert?

git-harry:
Yes, the changes to deploy.sh can be reverted.

PR #1045, opened 2016-05-03
Update OSA SHA to bring in zero weight swift fix

prometheanfire:
Kilo is the first to get a release/tag with the fix in it (from OSA)
OSA Liberty is having requirements issues.

Related: #878

PR #1051, opened 2016-05-04
Install packages required by Paramiko 2.0.0

git-harry:
Paramiko recently released version 2.0.0, this is installed by
bootstrap-ansible.sh and causes the script to fail due to missing
dependencies. This is addressed in openstack-ansible kilo but is not
included in the latest tag. deploy.sh has been modified to install the
missing packages. This is required until the next openstack-ansible tag
is released.

Issue: https://github.com/rcbops/rpc-openstack/issues/1037

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -26,6 +26,13 @@ function run_ansible {
   openstack-ansible ${ANSIBLE_PARAMETERS} --forks ${FORKS} $@
 }
 
+# https://github.com/rcbops/rpc-openstack/issues/1037
+# The latest openstack-ansible kilo tag, 11.2.14, fails due to a new version of
+# paramiko (2.0.0). The upstream fix is not in a tagged release. Installing the
+# following packages resolves the issue and can be removed once the openstack-ansible
+# SHA contains 75890e2372a66a8a9be24bc147179b6f08f90e76
+apt-get update && apt-get install -y build-essential libssl-dev libffi-dev python-dev
+
 # begin the bootstrap process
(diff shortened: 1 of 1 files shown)


tomc603:
Initial 👍  pending results of gate

scripts/deploy.sh
```diff
@@ -26,6 +26,13 @@ function run_ansible {
   openstack-ansible ${ANSIBLE_PARAMETERS} --forks ${FORKS} $@
 }
 
+# https://github.com/rcbops/rpc-openstack/issues/1037
+# The latest openstack-ansible kilo tag, 11.2.14, fails due to a new version of
+# paramiko (2.0.0). The upstream fix is not in a tagged release. Installing the
+# following packages resolves the issue and can be removed once the openstack-ansible
+# SHA contains 75890e2372a66a8a9be24bc147179b6f08f90e76
+apt-get update && apt-get install -y build-essential libssl-dev libffi-dev python-dev
+
```
prometheanfire: to be more clear this fix will be in OSA when 11.2.15 is tagged

prometheanfire:
:+1: an issue should be created to track the removal of that section targeted for a couple of weeks from now.

git-harry:
@tomc603 @prometheanfire I removed the SHA bump from this patch because the OSA tag doesn't include the required commit so that can be handled separately by https://github.com/rcbops/rpc-openstack/pull/1045

prometheanfire:
that works for me, even better, once 11.2.15 is out we can just git revert this

mattt416:
👍

PR #1065, opened 2016-05-10
Unlock repo server pip during deploy

git-harry:
The deploy script is also used for minor upgrades (within an openstack release)
Some versions of liberty deploy pip lockdown to the repo server, in order for
an upgrade to suceed the pip config must be removed so that repo builds have
access to external repos.

Partial-Bug: #1028
(cherry picked from commit 610cd8c34addcd24fd325d8c3bf1381ee1d28417)


sigmavirus24:
👍 looks like a straightforward backport

d34dh0r53:
:+1: yes it does

PR #1072, opened 2016-05-12
[WIP] Use correct ceph role names

### LINKED ISSUE #1071: Incorrect ceph role is being pulled down

12.1.1

Bug

Ceph should deploy, it does not.

Ceph does not deploy.  It throws an error when deploying MONs stating:

```
fatal: [ceph-aio_ceph_mon_container-26e67c07] => error while evaluating conditional: cephx and is_ceph_infernalis
```

Deploy 12.1.1 w/ DEPLOY_CEPH=yes

The deploy of ceph will fail with the above error.

As a deployer, you can:

```
# cd /opt/rpc-openstack/rpcd/playbooks/roles
# rm -rf ceph.ceph-common
# ln -s ceph-common/ ceph.ceph-common
```

High

This issue is due to the galaxy roles for ceph-ansible updating, but the issue is actually due to our configuration.  Originally, the role was called ceph-common, but it is now called ceph.ceph-common.  A dependency has been added to all the roles on ceph.ceph-common which means we install both ceph-common and ceph.ceph-common (installed automatically as a dependency).  When ceph is deployed, it uses ceph.ceph-common which is newer than the ceph-common we pull down and hence the failure.

This was actually an issue before, but the ceph-ansible galaxy roles were updated today which brought master of ceph.ceph-common much further ahead than the ceph-common we were pinning. 

- [x] Master
- [x] liberty-12.2
- [x] liberty-12.1

mattt416:
Closes #1071

### DIFF
--- a/ansible-role-requirements.yml
+++ b/ansible-role-requirements.yml
@@ -1,9 +1,9 @@
-- name: ceph-common
+- name: ceph.ceph-common
   src: https://github.com/ceph/ansible-ceph-common.git
   version: 1c623611a1d43b4c9116d1a0c21f0bdbdd87a0e7
-- name: ceph-mon
+- name: ceph.ceph-mon
   src: https://github.com/ceph/ansible-ceph-mon.git
   version: 01d3d6f0b06125b33b1e25745c06fa1d7137f7e9
-- name: ceph-osd
+- name: ceph.ceph-osd
   src: https://github.com/ceph/ansible-ceph-osd.git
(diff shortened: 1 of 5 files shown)


mattt416:
This should do it -- but we should do more testing before this merges.  Leaving WIP for now.

git-harry:
@mattt416 .gitignore needs modifying. Also, don't forget to update the commit message to remove WIP and to explain why this change is required.

git-harry:
Closing in favour of https://github.com/rcbops/rpc-openstack/pull/1077

PR #1077, opened 2016-05-13
Update Ceph roles used

### LINKED ISSUE #1071: Incorrect ceph role is being pulled down

git-harry:
Update the names of the Ceph roles used to correctly reference them with
their Ansible Galaxy names.

The roles ceph-mon and ceph-osd have a dependency of ceph.ceph-common.
This was originally ceph-common but was changed to allow the roles to be
uploaded to Ansible Galaxy. ansible-role-requirements.yml was specifying
ceph-common and not ceph.ceph-common. This meant that installing
ceph-mon and ceph-osd caused ceph.ceph-common to be downloaded and the
version used was HEAD of the ceph-common master branch instead of the
version specified in ansible-role-requirements.yml. This issue was not
detected for two reasons, firstly because the newer versions of
ceph-common worked with the older versions of ceph-mon and ceph-osd.
Secondly the Ceph role repos are generated using git subtree from a
single shared repo [1] and that process was broken which meant that
changes to the roles were not being published.

To address the above issue this patch updates the role names and SHAs
used.

A clean up task is added to remove the roles downloaded using the
previous names.

[1] https://github.com/ceph/ceph-ansible

Closes #1071

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -54,6 +54,6 @@ docs/_build/
 target/
 
 # generated files
-rpcd/playbooks/roles/ceph-common
-rpcd/playbooks/roles/ceph-mon
-rpcd/playbooks/roles/ceph-osd
+rpcd/playbooks/roles/ceph.ceph-common
+rpcd/playbooks/roles/ceph.ceph-mon
+rpcd/playbooks/roles/ceph.ceph-osd
--- a/ansible-role-requirements.yml
+++ b/ansible-role-requirements.yml
@@ -1,9 +1,9 @@
-- name: ceph-common
+- name: ceph.ceph-common
   src: https://github.com/ceph/ansible-ceph-common.git
   version: 1c623611a1d43b4c9116d1a0c21f0bdbdd87a0e7
-- name: ceph-mon
+- name: ceph.ceph-mon
   src: https://github.com/ceph/ansible-ceph-mon.git
   version: 01d3d6f0b06125b33b1e25745c06fa1d7137f7e9
-- name: ceph-osd
+- name: ceph.ceph-osd
   src: https://github.com/ceph/ansible-ceph-osd.git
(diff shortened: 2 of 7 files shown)


git-harry:
@mattt416 take a look at this when you get a chance. I've updated .gitignore and modified deploy.sh to remove the old roles. I've also updated the commit message.

prometheanfire:
this looks good, waiting on testing to finish, tentative :+1:

sigmavirus24:
👍 pending passing AIO tests.

prometheanfire:
recheck_upgrade

mattt416:
Thanks @git-harry ... this LGTM!

Issue #1073, opened 2016-05-12
Use Kibana version >= 4


galstrom21:
##### What version (commit) do you encounter this issue on?
12.1.1

##### What type of issue is this (bug, feature, documentation, regression)?
enhancement

##### What is the expected behavior?
An updated kibana would remove the need for an enduser's browser to connect directly with elasticsearch. As well as adding the ability to start using the elasticsearch beats dashboards.

##### What is your evaluation on the importance of this issue, and why?
Medium

The new version of Kibana will help with operator adoption of the ELK into their workflow, as it is now accessible over an SSH tunnel. This is also a precursor to being able to use ElasticSearch beats for additional data collection from the environment.

###### Remove branches (lines) you do not wish to target.
- [ ] Master
- [ ] liberty-12.2

git-harry:
This issue was closed because there is already an effort underway to address concerns related to the current logging solution.

PR #1084, opened 2016-05-16
Update Kibana to version 4.5.X

galstrom21:
In version 4+ of Kibana, Kibana now runs as its own service. Apache is
now configured to proxypass to the kibana application. This means that
the end user's browser does not need to connect directly to
ElasticSearch any longer. This change also moves to installing Kibana
from the elastic.co package repository instead of from a tarball.

This version of Kibana is dependent on ElasticSearch version >= 2.3.
Support Matrix URL
https://www.elastic.co/support/matrix#show_compatibility

Connects #1073
Connects rcbops/u-suk-dev#176


galstrom21:
@d34dh0r53 This is the kibana v4 change.. but it is dependent on ES >= 2.3 work you are doing

rpcd/playbooks/roles/kibana/tasks/kibana_post_install.yml
```diff
@@ -40,8 +74,8 @@
   template:
     src: "{{ item }}"
     dest: "/etc/apache2/sites-available/{{ item }}"
-    owner: "{{ kibana_system_user }}"
-    group: "{{ kibana_system_group }}"
+    owner: "root"
+    group: "root"
```
d34dh0r53: Shouldn't these be the new `{{ kibana_service_user }}` and `{{ kibana_service_group }}` variables defined in `main.yml`?

rpcd/playbooks/roles/kibana/tasks/kibana_post_install.yml
```diff
@@ -62,25 +96,13 @@
...
-  tags:
-    - kibana-post-install
-
 - name: Link Kibana Site
   file:
     state: "link"
     src: "/etc/apache2/sites-available/000-kibana.conf"
     dest: "/etc/apache2/sites-enabled/000-kibana.conf"
-    owner: "{{ kibana_system_user }}"
-    group: "{{ kibana_system_group }}"
```
d34dh0r53: See comment above

galstrom21:
@d34dh0r53 So all of the installed apache2 files (/etc/apache2/*) are owned by root:root. So that template does not need to be owned by the kibana_system_user:kibana_system_group

rpcd/playbooks/roles/kibana/tasks/kibana_post_install.yml
galstrom21: @d34dh0r53 So all of the installed apache2 files (/etc/apache2/*) are owned by root:root. So that template does not need to be owned by the kibana_system_user:kibana_system_group

galstrom21:
@d34dh0r53 kibana_service_user and kibana_service_group are only used in the templates/kibana file (template for /etc/default/kibana) for controlling the startup of the kibana service.

prometheanfire:
passing is a good start though

prometheanfire:
I think those apache configs being root is fine, this is apache config, not kibana config

git-harry:
As mentioned in the issue, there is currently an effort underway to review the ELK solution. Until the spike is has finished the project will not be merging any pull requests related to ELK.

galstrom21:
@git-harry understood.. was just wanting to put this up for @d34dh0r53 and the spike.

hughsaunders:
recheck_swift

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Liberty-PR: Fail

hughsaunders:
recheck_swift

hughsaunders:
recheck_ceph

d34dh0r53:
recheck_swift

d34dh0r53:
recheck_ceph

sigmavirus24:
Is this still something we aren't going to merge? I believe we've settled on sticking to ELK for now and that should clear the "status-dont-merge" label.

d34dh0r53:
@sigmavirus24 yeah, we can clear that flag.  It was there because this patch was made while we were still in liberty.

rpcd/playbooks/roles/kibana/tasks/kibana_post_install.yml
sigmavirus24: Or rather, shouldn't these be owned by the Apache user?

rpcd/playbooks/roles/kibana/tasks/kibana_pre_install.yml
```diff
@@ -13,50 +13,40 @@
...
-    state: latest
-    update_cache: yes
-    cache_valid_time: 600
-  register: install_apt_packages
-  until: install_apt_packages|success
+- name: Add kibana apt-keys
+  apt_key:
+    url: "{{ item.url }}"
+    state: "{{ item.state }}"
+  with_items: kibana_apt_keys
```
sigmavirus24: Last time I tried this, we needed to switch it to `"{{ kibana_apt_keys }"`.

rpcd/playbooks/roles/kibana/tasks/kibana_pre_install.yml
```diff
@@ -13,50 +13,40 @@
...
+  when: kibana_apt_keys is defined
+  register: add_kibana_keys_url
+  until: add_kibana_keys_url|success
```
sigmavirus24: We really shouldn't need to retry adding the URL

rpcd/playbooks/roles/kibana/tasks/kibana_pre_install.yml
```diff
@@ -13,50 +13,40 @@
...
-    group: "{{ kibana_system_group }}"
-    groups: adm
-    home: "/var/lib/{{ kibana_system_user }}"
-    system: "yes"
-    createhome: "yes"
-  when: kibana_system_group is defined and kibana_system_user is defined
+- name: Add kibana apt repositories
+  apt_repository:
+    repo: "{{ item.repo }}"
+    state: "{{ item.state }}"
```
sigmavirus24: Also probably makes sense to add `update_cache: yes`. Also probably don't need retries here.

rpcd/playbooks/roles/kibana/tasks/kibana_pre_install.yml
```diff
@@ -13,50 +13,40 @@
...
-  file:
-    state: directory
-    path: "{{item.name}}"
-    mode: "{{item.mode}}"
-    owner: "{{kibana_system_user}}"
-    group: "{{kibana_system_group}}"
-  with_items: kibana_directories
-  tags:
-    - kibana-pre-install
+#- name: Create kibana directories
```
sigmavirus24: Why not delete this instead of commenting it out?

rpcd/playbooks/roles/kibana/templates/000-kibana.conf
```diff
@@ -21,25 +21,19 @@
...
-    <Directory {{ kibana_root }}>
-      Options Indexes FollowSymLinks MultiViews
-      AllowOverride AuthConfig
-      AuthType Basic
-      AuthName Kibana
-      AuthUserFile /etc/apache2/users
-      Require user kibana
-    </Directory>
+    ProxyRequests Off
+    ProxyPass / http://127.0.0.1:{{ kibana_app_port }}
```
sigmavirus24: Does this require `proxy` and `proxy_http` Apache modules? Do we actually install those? I didn't think we did.

d34dh0r53:
@galstrom21 can you address @sigmavirus24's comments, or would you like me to take this over?

rpcd/playbooks/roles/kibana/templates/000-kibana.conf
galstrom21: They are already part of the kibana_apache_modules list:

https://github.com/galstrom21/rpc-openstack/blob/53ee2bfd3260e6306cbcd10a2a87a52c0b57860a/rpcd/playbooks/roles/kibana/defaults/main.yml#L32-L33

rpcd/playbooks/roles/kibana/tasks/kibana_pre_install.yml
galstrom21: deleting in update to PR

rpcd/playbooks/roles/kibana/tasks/kibana_pre_install.yml
galstrom21: Fixed in update to PR

rpcd/playbooks/roles/kibana/tasks/kibana_pre_install.yml
galstrom21: Same question as the item below.

rpcd/playbooks/roles/kibana/tasks/kibana_pre_install.yml
galstrom21: Looking through the rest of the roles in this repo, none of them have `with_items: {{ var_name }}`. They are all using the same pattern.

rpcd/playbooks/roles/kibana/tasks/kibana_pre_install.yml
sigmavirus24: No, I'm not certain we should remove them, but I'm also not certain that we need the retries. @d34dh0r53 thoughts?

rpcd/playbooks/roles/kibana/tasks/kibana_pre_install.yml
sigmavirus24: [So](https://github.com/rcbops/rpc-openstack/blob/c68e257057371e2ac68d0f78d74422af9f5e1944/rpcd/playbooks/roles/horizon_extensions/tasks/main.yml#L27), we used to write it the way I described. I'm glad this syntax works though and is now the de facto standard here. (The last time I tried to convert us to it, it was shot down.) I suspect Ansible 2.x has fixed the problems we had with it.

👍

rpcd/playbooks/roles/kibana/tasks/kibana_pre_install.yml
galstrom21: hm.. i somehow lost the item below.

The retry pattern is on every 'add apt-keys' and 'add apt-repo' action in every role in the repository.

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

andymcc:
recheck_ceph

sigmavirus24:
I know Harry is opposed to squash merging commits, but this is passing, looks good and I'd rather not have to fight with our excessively flakey gate tests.

Here's my 👍  and a vote that we squash merge instead of making @galstrom21 squash & rebase his commit.

sigmavirus24:
Looks like my comment triggered two new builds, so I guess it doesn't matter at this point. We'll be fighting the gate all day. =/

d34dh0r53:
recheck_ceph

d34dh0r53:
recheck_swift

andymcc:
Should we not squash merge then? I'm happy with the PR - I imagine there is work to be done in total to get all the logging bits to go properly - plus side this works (in my testing at least) on greenfields and as an upgrade from the previous incarnation of kibana.

The end configuration bits need some work from what I can see - there is a card to setup kibana dashboards etc. My concern is that we're going from a working --> not working (until we do the dashboards/config), although you could argue the current logging solution is not really working.

If we're happy with that then I'm happy for this to merge.

rpcd/playbooks/roles/kibana/tasks/kibana_post_install.yml
prometheanfire: I'm kinda inclined to agree that these probably shouldn't be root, though being root isn't the worst thing for an apache config file.

rpcd/playbooks/roles/kibana/tasks/kibana_post_install.yml
```diff
@@ -110,13 +132,14 @@
...
-  template:
-    src: "{{ item }}"
-    dest: "/opt/kibana/app/dashboards/{{ item }}"
-    owner: "root"
-    group: "root"
-  with_items:
-    - Event-Dashboard.json
-  tags:
-    - kibana-post-install
+# TODO(shep): this needs to be ported to kibana-4
```
prometheanfire: kibana-5?

rpcd/playbooks/roles/kibana/tasks/kibana_pre_install.yml
prometheanfire: I'd rather not have/need retry where they are not needed (here and elsewhere) Though it doesn't really matter.

rpcd/playbooks/roles/kibana/tasks/kibana_pre_install.yml
```diff
@@ -13,50 +13,30 @@
...
-  with_items: kibana_directories
+- name: Add kibana apt repositories
+  apt_repository:
+    repo: "{{ item.repo }}"
+    state: "{{ item.state }}"
+    update_cache: yes
+  with_items: kibana_apt_repos
+  register: add_kibana_repos
+  until: add_kibana_repos|success
+  retries: 5
```
prometheanfire: same with the retries

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Liberty-PR: Fail

d34dh0r53:
recheck_ceph

d34dh0r53:
recheck_swift

rpcd/playbooks/roles/kibana/tasks/kibana_install.yml
```diff
@@ -13,27 +13,22 @@
...
+  apt:
+    pkg: "{{ item }}"
+    state: latest
+    update_cache: yes
+    cache_valid_time: 600
+  register: install_apt_packages
+  until: install_apt_packages|success
+  retries: 5
+  delay: 2
+  with_items: kibana_apt_packages
```
evrardjp: bare variables are deprecated and should be replaced by "{{ kibana_apt_packages }}"

rpcd/playbooks/roles/kibana/tasks/kibana_post_install.yml
```diff
@@ -14,6 +14,40 @@
...
 
 
+- name: Deploy Kibana configuration files
+  template:
+    src: "{{ item }}"
+    dest: "/opt/kibana/config/{{ item }}"
+    owner: "root"
+    group: "root"
+  with_items:
+    - kibana.yml
```
evrardjp: No need for a with_items here.

rpcd/playbooks/roles/kibana/tasks/kibana_post_install.yml
```diff
@@ -14,6 +14,40 @@
...
+  tags:
+    - kibana-post-install
+
+- name: Deploy Kibana service configuration file
+  template:
+    src: "{{ item }}"
+    dest: "/etc/default/{{ item }}"
+    owner: "root"
+    group: "root"
+  with_items:
```
evrardjp: no need for a with_items here

rpcd/playbooks/roles/kibana/tasks/kibana_post_install.yml
evrardjp: I'd be enclined to use apache user. I don't really have a problem of root:root, because I guess 644 would be the default chmod and would be fine for apache to read the file.

rpcd/playbooks/roles/kibana/tasks/kibana_pre_install.yml
evrardjp: It's not because none of them are good that we should stick with doing the wrong thing.

Bare variables are deprecated. We should enclose the variables with "{{ }}"

evrardjp:
👎 See my comments in-line.

I'd also like to know why elastic's shield was not used/thought of when securing ES/Kibana, and use Kibana directly behind the LB -- instead of having a apache proxy passing with basic auth (which doesn't secure the ES, if I understood correctly what's stated above).

I'll move to +1 when I'll be convinced of the reasons.

galstrom21:
@evrardjp The main reason I did not look at shield is due to it being licensed software (https://www.elastic.co/subscriptions).

evrardjp:
Fair enough!

sigmavirus24:
So I have an AIO with this, #1245 and #1246 running and it seems to be working. Logstash is shipping things to elasticsearch and kibana is accessible and showing the items in queries out of elasticsearch.

So a 👍 having tested everything.

sigmavirus24:
So this should be rebased down to a single commit per our [contributing guidelines](https://github.com/rcbops/rpc-openstack/blob/master/CONTRIBUTING.md#pull-requests).

andymcc:
Agree with @sigmavirus24 - if we can get this rebased into a single commit we can get this merged in.
I was also able to complete an install with #1245 and #1246 working successfully so happy for this to merge after that.

andymcc:
Thanks @galstrom21. its 1 commit so 👍 
Merge pending gate.

sigmavirus24:
👍  once the gate passes as well. Thanks @galstrom21

prometheanfire:
:+1: once passing

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Issue #1085, opened 2016-05-17
The disk_utilisation check should create a WARNING rather than CRITICAL alert


JCallicoat:
##### What version (commit) do you encounter this issue on?

All

##### What type of issue is this (bug, feature, documentation, regression)?

Bug

##### What is the expected behavior?

When VMs backed by local disks are I/O intensive we should receive a WARNING alert

##### What is actually happening?

When VMs backed by local disks are I/O intensive we receive a CRITICAL alert

##### How can we reproduce this issue?

Run dd on a bunch of instances to trigger the check

##### How can we verify your issue is addressed?

Run dd on a bunch of instances to trigger the check and verify a WARNING alert is generated

##### What is your evaluation on the importance of this issue, and why?

Medium. It impacts customers with high I/O instances by generating tickets from CRITICAL alerts

##### What branches / milestones do you believe this issue should target?

Kilo
Liberty

###### Remove branches (lines) you do not wish to target.
- [ ] Master
- [ ] liberty-12.2
- [ ] liberty-12.1
- [ ] kilo

JCallicoat:
Closing in favor of PR #1090

mancdaz:
wait - this is a bug report not a PR. #1090 may address this bug report, but does not replace it.

d34dh0r53:
Closing in favor of #1091

mancdaz:
Lolz I'm going to keep saving this bug.

#1091 is an issue about some checks missing a CRITICAL alert, meaning tickets don't get generated.

This issue is actually reporting that the disk_utilisation check should have *only* WARNING and *not* CRITICAL.  There was some discussion about this in PR #1090, where it was decided that PR #1085 should address #1090 only, and that a separate discussion/PR would address this issue.

/cc @kstev

npawelek:
I think this issue is more around the noise that these checks generate for support on a daily basis. Instead of changing these to WARNING in order to not generate a ticket, why not tune them properly? The issue we see in support is that these alert and then clear repeatedly for customers that run hot on their compute nodes. I would think that a longer period, say 30-90 minutes with a consecutive count of 3 would provide a better overall experience for support with less repeated alerts and tickets generated.

PR #1090, opened 2016-05-18
Ensure critical threshold

kstev:
Checks without a CRITICAL threshold will never be seen by 90% of Rackspace customers so add CRITICAL threshold to various alarms.

Addresses bug #1091

### DIFF
--- a/rpcd/playbooks/roles/rpc_maas/defaults/main.yml
+++ b/rpcd/playbooks/roles/rpc_maas/defaults/main.yml
@@ -242,10 +242,12 @@ cinder_volumes_vg_critical_threshold: 90.0
 cinder_vg_name: "{{ cinder_backends['lvm']['volume_group'] }}"
 
 # MaaS CPU idle threshold
-cpu_idle_percent_avg_threshold: 10.0
+cpu_idle_percent_avg_warning_threshold: 10.0
+cpu_idle_percent_avg_critical_threshold: 1.0
 
 # MaaS Memory in use threshold
-memory_used_percentage_threshold: 95.0
+memory_used_percentage_warning_threshold: 90.0
+memory_used_percentage_critical_threshold: 99.0
 
 # Set the threshold for the "Max Channels per Connection" Rabbitmq alarm
 rabbitmq_max_channels_per_con_threshold: 10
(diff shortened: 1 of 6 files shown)


prometheanfire:
This seems fine, we'll probably need better commit messages (more than just titles) and a bug

https://github.com/rcbops/rpc-openstack/blob/master/CONTRIBUTING.md#commits

mancdaz:
@kstev also please squash these two commits into one

prometheanfire:
@kstev please don't squash them into one commit.

* Mixing two unrelated functional changes in the same commit.

from https://github.com/rcbops/rpc-openstack/blob/master/CONTRIBUTING.md#commits

mancdaz:
@prometheanfire the second commit is a typo fix to something introduced in the first commit, so they can be squashed

prometheanfire:
ah, ok then

kstev:
Will try to address bug 1085 with this as well. That okay?

prometheanfire:
Ya, it looks like that's what this addresses.

mancdaz:
@kstev so the commit title needs to be something like 'ensure all alerts have both warning/critical thresholds'

#1085 looks like it's asking for only warning which I don't necessarily agree with

tomc603:
This PR seems to address #1091 but not #1085. The latter is simply pointing out that "CRITICAL" is not a correct status for high disk utilization.

This PR may well be related to 1085, though, in that it at least reports a "WARNING" status. However, it will still report a CRITICAL status, and provides no method to not specify a critical level or to skip warning of critical status.

BjoernT:
👎 
Core does not understand warning thresholds hence this pull looses the use. We would never see warning. For the disk utilization we only have to fix my old bugs, not report vitrtual scsi device like iscsi, so we only monitor the actual physical disks and not so much instance activity

JCallicoat:
WARNINGs can still be polled via the MaaS API and email alerts can be configured for WARNING.

Not creating CORE tickets for metrics that are non-CRITICAL is a benefit.

For compute hosts with qcow images on local disk backing the instances, we get CRITICAL alerts for /dev/sda when there is high instance disk I/O for example.

BjoernT:
@JCallicoat sure but 99% of customers do not look in the dashboard so use is still limited.

mancdaz:
@kstev I'd change this PR to just address #1091. The discussion around #1085 should happen in that issue, and once a decision is made a separate PR can be made to address that.

Also, you're failing on flake8 for the commit title:

```hacking/rpco_checks.py:1:1: R802  git commit title ('Add warning and critical thresholds for various alarms') should be under 50 chars```

kstev:
I think I have something that will address #1085. Will target in another PR.

mancdaz:
👎  @kstev you'll need to have the CRITICAL alerts first, otherwise we'll never see them - WARNING will always match first

kstev:
recheck 61f57a5

kstev:
recheck RPC-AIO-Ceph

mancdaz:
@kstev looks like you'll need to rebase this commit. Jenkins is reporting:

"GitHub pull request #1090 of commit 61f57a51778a2c1b8d59a86e29f763608eea5153, has merge conflicts."

mancdaz:
👍

d34dh0r53:
:+1: @BjoernT are you okay with this as it stands?

PR #1101, opened 2016-05-25
[WIP] Run ceph roles serially

mattt416:
This commit uses ansible's `serial` to deploy MONs and OSDs serially,
which should prevent masses of services restarting at the same time.

In order for MONs to bootstrap one by one, we need to set `mon initial
members` in ceph.conf.

Additionally, we separate out the tasks in ceph-mon.yml so these can
be targetted at a different host group (with the end goal of these
tasks only being run once).

NOTE: Bootstrapping OSDs serially may be slow, so we can introduce a
      variable override that can be passed in on the command line to
      disable `serial: 1`.

### DIFF
--- a/rpcd/etc/openstack_deploy/user_rpco_variables_defaults.yml
+++ b/rpcd/etc/openstack_deploy/user_rpco_variables_defaults.yml
@@ -136,6 +136,11 @@ secure_cluster: true
 # List of secure flags to set on for a pool (options for the list are nodelete, nopgchange, nosizechange - prevents deletion, pg from changing and size from changing respectively).
 secure_cluster_flags:
   - nodelete
+# This is necessary if we want to bootstrap MONs serially, failing to set this
+# will result in additional MONs failing to bootstrap due to lack of quorum.
+ceph_conf_overrides:
+  mon:
+    mon initial members: "{{ hostvars[groups['mons'][0]]['ansible_hostname'] }}"
 
 # We need to override the OpenStack upper constraint for Mitaka which is version 1.9.0, We should be able to remove this in Newton as the upper constraint moves to 2.3.0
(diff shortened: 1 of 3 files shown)


mattt416:
Hi @BjoernT , what are your thoughts here?  This will cause the initial deploy and upgrades to run a fair bit slower, but will hopefully prevent the cluster from going into a bad state (albeit temporarily) while an upgrade takes place.

BjoernT:
👍  I would not oppose slightly slower upgrade in favor of less race conditions

prometheanfire:
recheck_upgrade

mattt416:
There is an upstream WIP at https://github.com/ceph/ceph-ansible/pull/815 to add `mon initial members` to ceph.conf, so we can remove this override once that lands and we bump to a later SHA.

mancdaz:
recheck_upgrade

mancdaz:
recheck_all

Apsu:
RPC-AIO-Liberty-PR: Pass

Apsu:
RPC-AIO-Liberty-PR: Pass

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Ceph-Liberty-PR: Pass

Apsu:
RPC-AIO-Liberty-PR: Fail

mattt416:
recheck_swift

d34dh0r53:
Looks like the gate checks are stuck and this will need a rebase.

major:
Added status-blocked since this needs a rebase. Need help on that, @mattt416?

mattt416:
Closing this for now, will revisit this at a later date.

Issue #1100, opened 2016-05-24
horizon extension fails to update/install


BjoernT:
##### What version (commit) do you encounter this issue on?

r12.1.1

##### What type of issue is this (bug, feature, documentation, regression)?

Bug

##### What is the expected behavior?

Installs without errors

##### What is actually happening?
```
TASK: [horizon_extensions | Install python dependencies] **********************
failed: [Infra01_horizon_container-8ed2a89b] => {"cmd": "/usr/local/bin/pip install --force-reinstall -r /opt/rackspace/horizon-extensions/requirements.txt", "failed": true}
msg: stdout: Ignoring indexes: https://pypi.python.org/simple
Collecting markdown (from -r /opt/rackspace/horizon-extensions/requirements.txt (line 1))

:stderr: DEPRECATION: --allow-all-external has been deprecated and will be removed in the future. Due to changes in the repository protocol, it no longer has any effect.
  Could not find a version that satisfies the requirement markdown (from -r /opt/rackspace/horizon-extensions/requirements.txt (line 1)) (from versions: )
No matching distribution found for markdown (from -r /opt/rackspace/horizon-extensions/requirements.txt (line 1))

failed: [infra02_horizon_container-755bd245] => {"cmd": "/usr/local/bin/pip install --force-reinstall -r /opt/rackspace/horizon-extensions/requirements.txt", "failed": true}
msg: stdout: Ignoring indexes: https://pypi.python.org/simple
Collecting markdown (from -r /opt/rackspace/horizon-extensions/requirements.txt (line 1))

:stderr: DEPRECATION: --allow-all-external has been deprecated and will be removed in the future. Due to changes in the repository protocol, it no longer has any effect.
  Could not find a version that satisfies the requirement markdown (from -r /opt/rackspace/horizon-extensions/requirements.txt (line 1)) (from versions: )
No matching distribution found for markdown (from -r /opt/rackspace/horizon-extensions/requirements.txt (line 1))

failed: [infra03_horizon_container-e5e8181e] => {"cmd": "/usr/local/bin/pip install --force-reinstall -r /opt/rackspace/horizon-extensions/requirements.txt", "failed": true}
msg: stdout: Ignoring indexes: https://pypi.python.org/simple
Collecting markdown (from -r /opt/rackspace/horizon-extensions/requirements.txt (line 1))

[...]

PR #1104, opened 2016-05-27
Alter pip behavior of horizon_extensions play

d34dh0r53:
This fix alters the pip behavior of the `horizon_extensions` play to
bring it to liberty standards.  Since we are no longer running the
`repo-build` play from within the rpcd directory the existing
requirements file is no longer being used by the OSA `repo-build`
play.  This commit adds the `horizon_extensions_pip_packages` variable
to the defaults and uses the updated pip form for installation.

Closes-bug: #1100

### DIFF
--- a/rpcd/playbooks/roles/horizon_extensions/defaults/main.yml
+++ b/rpcd/playbooks/roles/horizon_extensions/defaults/main.yml
@@ -16,5 +16,5 @@
 horizon_extensions_git_repo: https://github.com/rcbops/horizon-extensions
 horizon_extensions_git_install_branch: master
 horizon_extensions_dest_dir: /opt/rackspace/horizon-extensions
-
-horizon_extensions_pip_requirements_file: "/opt/rackspace/horizon-extensions/requirements.txt"
+horizon_extensions_pip_packages:
+   - "markdown"
--- a/rpcd/playbooks/roles/horizon_extensions/tasks/main.yml
+++ b/rpcd/playbooks/roles/horizon_extensions/tasks/main.yml
@@ -38,10 +38,19 @@
   become: true
   notify: Restart apache2
 
-- name: Install python dependencies
+- name: Install pip packages
   pip:
-    requirements: "{{ horizon_extensions_pip_requirements_file }}"
+    name: "{{ item }}"
+    state: present
     extra_args: "{{ pip_install_options | default('') }}"
+  register: install_pip_packages
+  until: install_pip_packages|success
+  retries: 5
+  delay: 2
+  with_items: horizon_extensions_pip_packages
+  tags:
+    - horizon-extensions-install
+    - horizon-extensions-pip-packages
 
 - name: "Create /etc/rackspace if it doesn't exist"
   file:


d34dh0r53:
This is failing ansible linting, I think due to incorrect test-requirements.  I tried bumping to 2.3.9 but that didn't work. @hughsaunders can you take a look when you get a chance?

d34dh0r53:
cc: @git-harry

git-harry:
@d34dh0r53 I have raised a bug for the ansible-lint failure https://github.com/rcbops/rpc-openstack/issues/1105 and will look to get it addressed shortly.

Regarding the issue being addressed here, did you determine why we haven't seen this before? Regarding the fix, did you consider having the repo server build a wheel for horizon-extensions? I'm not 100% sure how the repo process works but it might take care of the dependencies as part of building the wheel which would remove the need to duplicate the list of packages in RPCO.

evrardjp:
👍 but not tested.

git-harry:
@d34dh0r53 This issue never got 👍 from two reviewers. While I understand the pressure to get things committed quickly, we need to make sure pull requests receive the appropriate amount of scrutiny and so I would prefer you not to merge your own commits before that process is complete.

My comments were never addressed, I would appreciate feedback. In addition, I now realise the most obvious question I didn't ask is, why is this not blocking the testing we do [1]? It would be useful to understand so that we can be sure to catch similar issues in the future.

[1] http://jenkins.propter.net:443/

git-harry:
It seems on an AIO markdown gets installed by 'python setup.py install'. Below is an extract of the output from running that command.
```
Processing dependencies for horizon-extensions==1.0.0
Searching for Markdown
Reading https://pypi.python.org/simple/Markdown/
Best match: Markdown 2.6.6
Downloading https://pypi.python.org/packages/9b/53/4492f2888408a2462fd7f364028b6c708f3ecaa52a028587d7dd729f40b4/Markdown-2.6.6.tar.gz#md5=d6737edcdb9a9aceaef0bf28455673be
Processing Markdown-2.6.6.tar.gz
Writing /tmp/easy_install-vMw0Hy/Markdown-2.6.6/setup.cfg
Running Markdown-2.6.6/setup.py -q bdist_egg --dist-dir /tmp/easy_install-vMw0Hy/Markdown-2.6.6/egg-dist-tmp-yptxqi
Moving Markdown-2.6.6-py2.7.egg to /usr/local/lib/python2.7/dist-packages
Adding Markdown 2.6.6 to easy-install.pth file
Installing markdown_py script to /usr/local/bin

Installed /usr/local/lib/python2.7/dist-packages/Markdown-2.6.6-py2.7.egg
Finished processing dependencies for horizon-extensions==1.0.1
```
This means that markdown is already installed before the task 'Install pip packages' runs.

odyssey4me:
I might be wrong, but with the right ``pkg_locations`` included in https://github.com/rcbops/rpc-openstack/blob/master/rpcd/etc/openstack_deploy/user_variables.yml#L73-L78 it should not be necessary to do https://github.com/rcbops/rpc-openstack/blob/master/rpcd/playbooks/roles/horizon_extensions/tasks/main.yml#L28-L39 at all because the repo server will pull the git repo and do the wheel build.

Instead the horizon-extensions package should just be in the ``horizon_extensions_pip_packages`` list.

d34dh0r53:
@git-harry wow, that's my bad totally.  I could have sworn that this had two :+1:'s I must have misread something.  Really sorry about that.

PR #1111, opened 2016-06-01
add question to issue template

mancdaz:
Adding a new question to the issue template to try and solicit more
actual useful infromation from the issue submitter


prometheanfire:
recheck_upgrade

:+1:  on pass

Seems like a good addition, but we need to be careful about asking too many questions and making it too hard to report issues.

mancdaz:
recheck_upgrade

d34dh0r53:
very much :+1:

PR #1114, opened 2016-06-03
Mitaka Greenfield that works*

evrardjp:
Here is a PR for mitaka, it's a greenfield deployment that works* for AIOs. 
Deployment goes through for horizon extensions, ELK and MaaS. 

This PR is not a proper one (because it can't merge and it isn't in a squashed commit form), it's just for not losing the work, and for everyone to have a view of the work to be done for greenfield, and get the first reviews/comments.

I'd be happy to bring these commits to a new branch when the branch will be cut.

*Some parts were not tested, like MaaS, Ceph... No change was done there, so everything should work (except maybe an impact of the change in the endpoints?).

### DIFF
--- a/rpcd/etc/openstack_deploy/user_variables.yml
+++ b/rpcd/etc/openstack_deploy/user_variables.yml
@@ -76,6 +76,7 @@ pkg_locations:
   - /etc/ansible/roles
   - /etc/openstack_deploy
   - /opt/rpc-openstack/rpcd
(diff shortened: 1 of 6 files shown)


evrardjp:
It works for me.

d34dh0r53:
recheck_ceph

d34dh0r53:
Failing on Holland:
`19:51:11 No handlers could be found for logger "holland.core.util.bootstrap"`

PR #1122, opened 2016-06-07
fix holland backups

### LINKED ISSUE #1121: Holland broken in Liberty

r12.1.0 (two separate environments both on same version)

Bug

The "holland bk" command works

So, there are a few issues here.

Firstly, for some reason the rpc_support database users / grants were not created in these environments, even though it appears the rpc_support playbook was run. Since we set `run_once: true` for the `Create rpc_support backup user` task, any subsequent runs will skip the task and the user will never be created. This may be more of an operational issue -- maybe the play was run with some set of tags limiting it or something like that (will follow up on my end with deployments team). But there is no harm in letting the task run every time, so the `run_once` should probably be removed.

----

Secondly, the holland bk command fails to find `innobackupx` unless the full path is given in the config file (/etc/holland/backupsets/rpc_support.conf), even though it's in /usr/bin and that is in the PATH.
```
# holland bk
[...snip...]
Backup failed: Failed to find innobackupex script
```
Setting `innobackupex = /usr/bin/innobackupex` resolves this error.

----

Lastly, after setting `innobackupx` another error pops up
```
# holland bk
[...snip...]
    ! xtrabackup: error: compressed and encrypted backups are incompatible with the
    ! 'tar' streaming format. Use --stream=xbstream instead.
```
Setting `stream = xbstream` allows holland to successfully complete the backup.

----

I suspect that some defaults changed in xtrabackup that now require these options to be set explicitly.

Run holland bk from a galera container

Verify backup is created at /var/backup/holland_backups/rpc_support/ in galera container

Support manually create DB user and grants (as needed), manually edit config file template (always needed until bug is fixed).

Very high. Running without any DB backups is bad.

- [x] Master (https://github.com/rcbops/rpc-openstack/pull/1122)
- [x] liberty-12.2 (https://github.com/rcbops/rpc-openstack/pull/1136)
- [x] liberty-12.1 (https://github.com/rcbops/rpc-openstack/pull/1137)
- [x] kilo (https://github.com/rcbops/rpc-openstack/pull/1160)

mancdaz:
This commit fixes holland backups with a few modifications:

- The db grants for the backup user are moved into a separate task that is now
  properly called for the first host in the galera groups (run_once was causing the task
  to be skipped because of the way the host list was evaluated)
- The stream type is set explicitly to xbstream away from the default of tar
- The path to the innobackupex binary is explicitly set
- The log directory is now set and created properly

Fixes #1121 

```
root@aio3_galera_container-8890e499:/# holland bk
Holland 1.0.10 started with pid 10168
--- Starting backup run ---
Creating backup path /var/backup/holland_backups/rpc_support/20160607_101644
* Generating mysql option file: /var/backup/holland_backups/rpc_support/20160607_101644/my.cnf
  + Added !include /etc/mysql/my.cnf
  + Added !include /root/.my.cnf
  + Added [client] section with credentials from [mysql:client] section
Estimated Backup Size: 2.36GB
Starting backup[rpc_support/20160607_101644] via plugin xtrabackup
Executing: /usr/bin/innobackupex --defaults-file=/var/backup/holland_backups/rpc_support/20160607_101644/my.cnf --stream=xbstream --tmpdir=/var/backup/holland_backups/rpc_support/20160607_101644 --no-timestamp /var/backup/holland_backups/rpc_support/20160607_101644
  > /var/backup/holland_backups/rpc_support/20160607_101644/backup.xb 2 > /var/backup/holland_backups/rpc_support/20160607_101644/xtrabackup.log
Skipping --prepare/--apply-logs since backup is streamed
Final on-disk backup size 7.41MB
0.31% of estimated size 2.36GB
Backup completed in 2.32 seconds
No backups purged
Released lock /etc/holland/backupsets/rpc_support.conf
--- Ending backup run ---
```

### DIFF
--- a/rpcd/playbooks/roles/rpc_support/tasks/holland_config.yml
+++ b/rpcd/playbooks/roles/rpc_support/tasks/holland_config.yml
@@ -13,35 +13,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-- name: Create rpc_support backup user
-  mysql_user:
-    name: "{{ item.name }}"
-    host: "{{ item.host }}"
-    password: "{{ item.password }}"
-    priv: "{{ item.priv }}"
-    state: "{{ item.state }}"
-  with_items:
-    - name: "rpc_support"
-      host: "%"
-      password: "{{ rpc_support_holland_password }}"
-      priv: "*.*:ALL"
-      state: present
-    - name: "rpc_support"
-      host: "localhost"
-      password: "{{ rpc_support_holland_password }}"
-      priv: "*.*:ALL"
-      state: present
-  run_once: true
-  tags:
-    - holland_sql_user
-    - holland_all
-
 - name: Create supporting holland directories
   file:
     state: directory
     path: "{{ item }}"
   with_items:
     - "/var/backup/holland_backups"
+    - "/var/log/holland"
     - "/etc/holland"
   tags:
     - holland_dir
--- a/rpcd/playbooks/roles/rpc_support/tasks/holland_db_setup.yml
+++ b/rpcd/playbooks/roles/rpc_support/tasks/holland_db_setup.yml
@@ -1,0 +1,36 @@
+---
+# Copyright 2014, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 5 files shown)


mancdaz:
recheck_upgrade

mancdaz:
@JCallicoat if you get a chance, see if you can test/review this patch and ensure it fixes your issue.

hughsaunders:
recheck_upgrade

hughsaunders:
👍

d34dh0r53:
:+1: but I would prefer to have @JCallicoat test this on something that is known broken before merging.

JCallicoat:
:+1: Verified the patch works in an environment where I originally encountered the issue.

PR #1136, opened 2016-06-09
fix holland backups

mancdaz:
This commit fixes holland backups with a few modifications:

- The db grants for the backup user are moved into a separate task that is now
  properly called for the first host in the galera groups (run_once was causing the task
  to be skipped because of the way the host list was evaluated)
- The stream type is set explicitly to xbstream away from the default of tar
- The path to the innobackupex binary is explicitly set
- The log directory is now set and created properly

Partial Issue: #1121 
(cherry picked from commit 349a61b1e41f9a67e7887dbfa4572a94556a3f46)


mancdaz:
recheck_upgrade

hughsaunders:
👍

Issue #1098, opened 2016-05-22
Cinder service check failed


BjoernT:
##### What version (commit) do you encounter this issue on?

r11.1.9

##### What type of issue is this (bug, feature, documentation, regression)?

Bug

##### What is the expected behavior?

cinder service check will locate cinder-scheduler services on bare metal along with proper reporting of the cinder-volume_status metric when one or many backends are down

##### What is actually happening?

```
Print service array:
{u'status': u'enabled', u'binary': u'cinder-scheduler', u'zone': u'nova', u'host': u'12345-cinder01', u'updated_at': u'2016-05-22T20:59:33.000000', u'state': u'up', u'disabled_reason': None}
{u'status': u'enabled', u'binary': u'cinder-volume', u'zone': u'nova', u'host': u'12345-cinder01@lvm', u'updated_at': u'2016-05-22T20:59:32.000000', u'state': u'up', u'disabled_reason': None}
{u'status': u'enabled', u'binary': u'cinder-backup', u'zone': u'nova', u'host': u'12345-cinder01', u'updated_at': u'2016-05-22T20:59:33.000000', u'state': u'up', u'disabled_reason': None}

Plugin output:
status okay
metric cinder-volume-lvm_status uint32 1
metric cinder-backup_status uint32 1
```
##### How can we reproduce this issue?

Install maas on a cinder bare metal which also runs the cinder-scheduler

##### How can we verify your issue is addressed?

No more errors

##### If this is a large issue, what can be done immediately to alleviate the issue while working on a longer-term solution?

Prio 2

##### What is your evaluation on the importance of this issue, and why?

Customer plugins report false alarms now

###### Remove branches (lines) you do not wish to target.
- [x] Master (#1241)
- [x] liberty-12.2 (#1250)
- [x] kilo

andymcc:
I've submitted a fix that takes into consideration the feedback from the previous PR - since we hadn't had movement in 4 weeks - if we can get that reviewed and merged we can close this out (post backports).

andymcc:
Will need to adjust the actual monitors to match the metrics.

andymcc:
OK passing tests -ready for reviews!

sigmavirus24:
@alextricity25 why did you close this?

andymcc:
ok we're all done - backports are ready. closing this out!

PR #1151, opened 2016-06-12
Report status for non @backend cinder services

BjoernT:
This fix reports service states for cinder service names
for all services beyond cinder backends like LVM.
Additionally it adds a cinder-volume_status status check
in order to trigger a MAAS alarm since the alarm is only configured
to look at the cinder-volume_status service.

Connects #1098

### DIFF
--- a/maas/plugins/cinder_service_check.py
+++ b/maas/plugins/cinder_service_check.py
@@ -86,18 +86,25 @@ def check(auth_ref, args):
             if '@' in service['host']:
                 [host, backend] = service['host'].split('@')
                 name = '%s-%s_status' % (service['binary'], backend)
-                metric_bool(name, service_is_up)
+            else:
+                name = '%s_status' % service['binary']
 
-        name = '%s_status' % service['binary']
-        metric_bool(name, all_services_are_up)
+            metric_bool(name, service_is_up)
+
+        metric_bool('cinder-volume_status', all_services_are_up)
     else:
+        all_services_are_up = True
+
         for service in services:
             service_is_up = True
             if service['status'] == 'enabled' and service['state'] != 'up':
                 service_is_up = False
+                all_services_are_up = False
 
             name = '%s_on_host_%s' % (service['binary'], service['host'])
             metric_bool(name, service_is_up)
+
+        metric_bool('cinder-volume_status', all_services_are_up)
 
 
 def main(args):


mancdaz:
@BjoernT this could possibly be worked to address https://github.com/rcbops/rpc-openstack/issues/1171 as well

BjoernT:
Actually it will. My PR was changed to address the missing cinder-volume_status metric in a multiple volume backend scenario.  I would say #1171 is a dupe to this bug here

sigmavirus24:
The code looks fine but I don't have anything set-up with multiple cinder-volume backing stores. The rest *looks* fine running this against an AIO.

maas/plugins/cinder_service_check.py
```diff
@@ -86,19 +86,26 @@ def check(auth_ref, args):
...
                 name = '%s-%s_status' % (service['binary'], backend)
-                metric_bool(name, service_is_up)
+            else:
+                name = '%s_status' % service['binary']
 
-        name = '%s_status' % service['binary']
-        metric_bool(name, all_services_are_up)
+            metric_bool(name, service_is_up)
+
+        metric_bool('cinder-volume_status', all_services_are_up)
```
git-harry: This line assumes all services are cinder-volume services. I have an AIO with the following Cinder services with cinder-scheduler currently down:

```
# cinder service-list
+------------------+-------------------------------------------+------+---------+-------+----------------------------+-----------------+
|      Binary      |                    Host                   | Zone |  Status | State |         Updated_at         | Disabled Reason |
+------------------+-------------------------------------------+------+---------+-------+----------------------------+-----------------+
|  cinder-backup   |                   test1                   | nova | enabled |   up  | 2016-07-05T10:40:17.000000 |        -        |
| cinder-scheduler | test1_cinder_scheduler_container-6b2fcb78 | nova | enabled |  down | 2016-07-05T10:38:59.000000 |        -        |
|  cinder-volume   |                 test1@lvm                 | nova | enabled |   up  | 2016-07-05T10:40:11.000000 |        -        |
+------------------+-------------------------------------------+------+---------+-------+----------------------------+-----------------+

```

For the cinder-scheduler check the current version of the plugin gives the following output showing that the scheduler is down
```
# python /usr/lib/rackspace-monitoring-agent/plugins/cinder_service_check.py test1_cinder_api_container-9b5cadb7 --host test1_cinder_scheduler_container-6b2fcb78
status okay
metric cinder-scheduler_status uint32 0
```

This change is adding a cinder-volume metric to the cinder-scheduler check as well as reporting the incorrect status of the cinder-volume service.
```
# python /usr/lib/rackspace-monitoring-agent/plugins/cinder_service_check.py test1_cinder_api_container-9b5cadb7 --host test1_cinder_scheduler_container-6b2fcb78
status okay
metric cinder-scheduler_status uint32 0
metric cinder-volume_status uint32 0
```

maas/plugins/cinder_service_check.py
```diff
@@ -86,19 +86,26 @@ def check(auth_ref, args):
...
         for service in services:
             service_is_up = True
             if service['status'] == 'enabled' and service['state'] != 'up':
                 service_is_up = False
+                all_services_are_up = False
 
             name = '%s_on_host_%s' % (service['binary'], service['host'])
             metric_bool(name, service_is_up)
 
+        metric_bool('cinder-volume_status', all_services_are_up)
```
git-harry: Similar to above, this reports a cinder-volume status metric that actually represents the status of all services, not just cinder-volume.

git-harry:
👎  see inline comments.

Apsu:
RPC-AIO-Liberty-PR: Fail

maas/plugins/cinder_service_check.py
BjoernT: The cinder-volume_status is only for the MAAS alarm, so it'll trigger. The individual servers will be still marked as down in the check but maas only looks at this one metric

git-harry: - We can always create more alarms if necessary
- This change means if, for example, cinder-scheduler is down cinder-volume is reported as down
- There is now an additional volume-related metric added to all cinder service checks and not just the volume ones.

git-harry:
recheck_maas

git-harry:
whoops, I meant

recheck_swift

maas/plugins/cinder_service_check.py
mancdaz: I agree - we need to ensure that the metric matches what is being reported. Marking 'cinder-volume_status' as down when it could be any of the cinder services is very bad practice, and has the potential to cause great confusion and delays in troubleshooting.

mancdaz:
👎 based on @git-harry inline comments and my own observations

andymcc:
I can't adjust this PR - but we haven't had feedback/changes in 4 weeks so I submitted a separate PR: #1241 
If/when that merges I think we can go ahead and close this issue out.

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Liberty-PR: Fail

andymcc:
Closing this, we can move comments over to the new PR #1241

Issue #1088, opened 2016-05-18
RPC Logo is not in Horizon


brainerd123:
##### What version (commit) do you encounter this issue on?
12.1.1

##### What type of issue is this (bug, feature, documentation, regression)?'

Bug

##### What is the expected behavior?

We should see the RPC Logo in Horizon, not the native Openstack Logo

##### What is actually happening?

We are seeing the native OpenStack logo in Liberty.  We upgraded this environment from 11.1.2 to 12.1.1 and noticed we lost the RPC Logo.

##### How can we reproduce this issue?

https://cloud1.osic.org

##### How can we verify your issue is addressed?

click this link for the OSIC horizon dashboard: https://cloud1.osic.org 

##### If this is a large issue, what can be done immediately to alleviate the issue while working on a longer-term solution?

Not an urgent matter...more branding

##### What is your evaluation on the importance of this issue, and why?

Not super high priority, but has cosmetic and branding implications

##### What branches / milestones do you believe this issue should target?

12.1.1 RPC
###### The list below is what is currently supported for backports

NA
###### Remove branches (lines) you do not wish to target.

- [x] Master
- [ ] liberty-12.2

d34dh0r53:
Were the `rpc_support` plays re-run after the upgrade?

d34dh0r53:
Confirmed that the `rpc_support` play was run, we should look into this.

BjoernT:
It sense a race condition issue here, if you rerun os-horizon-install that seem to revert the settings, if they are part of rpc-support. Something similar your be how we patch the nova spice console

mancdaz:
@evrardjp if you're going to backport this to liberty, then some notes will need to be added to ensure the deployer copies the appropriate settings from ```rpcd/etc/user_variables.yml``` into ```/etc/openstack_deploy/user_variables.yml``` for this to take effect.

evrardjp:
Ok, I'll make sure the doc will be included. Thanks for the reminder!

b3rnard0:
This being a priority 2 issue, I'm moving this back to the backlog and we can reprioritize in the future.

evrardjp:
Move this to need review when WIP is lower.

git-harry:
@evrardjp is there any reason to keep this open now you have created https://github.com/rackerlabs/docs-rpc/issues/619

git-harry:
Closing because now handled by docs team https://github.com/rackerlabs/docs-rpc/issues/619

PR #1155, opened 2016-06-13
Use OSA code for RPC branding

evrardjp:
This commit leverages OSA work instead of reinventing the wheel in RPC.
It uses OSA's os_horizon role "horizon_custom_uploads" variable for
determining which files to upload where.

Issue: https://github.com/rcbops/rpc-openstack/issues/1088


prometheanfire:
I like this one :+1:

mancdaz:
Tested, and works as expected. Also ensured that the settings remain regardless of when the rpc/osa playbooks are run (which is what the complaint in the referenced issue is).

👍

Issue #1124, opened 2016-06-07
Constant pip issues on host level due to bad pip links config handing inside pip_lockdown_role


BjoernT:
##### What version (commit) do you encounter this issue on?

r12.1.2

##### What type of issue is this (bug, feature, documentation, regression)?

Bug

##### What is the expected behavior?

pip configuration is 100% managed during RPC upgrade, no more stale links

##### What is actually happening?

```
TASK: [rpc_support | Install pccommon pip dependencies] ***********************
failed: [controller2] => (item=python-openstackclient) => {"attempts": 2, "cmd": "/usr/local/bin/pip install --force-reinstall python-openstackclient", "failed": true, "item": "python-openstackclient"}
msg: Task failed as maximum retries was encountered
failed: [controller3] => (item=python-openstackclient) => {"attempts": 2, "cmd": "/usr/local/bin/pip install --force-reinstall python-openstackclient", "failed": true, "item": "python-openstackclient"}
msg: Task failed as maximum retries was encountered
failed: [controller1] => (item=python-openstackclient) => {"attempts": 2, "cmd": "/usr/local/bin/pip install --force-reinstall python-openstackclient", "failed": true, "item": "python-openstackclient"}
msg: Task failed as maximum retries was encountered

FATAL: all hosts have already failed -- aborting

PLAY RECAP ********************************************************************
rpc_support | Add rpc_support key to nova ------------------------------ 10.87s
rpc_support | Check for rpc_support keypair in nova -------------------- 10.84s
rpc_support | Install pccommon pip dependencies ------------------------- 5.83s
rpc_support | Add HWRaid apt keys --------------------------------------- 5.81s
rpc_support | Set history format file ----------------------------------- 1.36s
rpc_support | Host kernel sysctl tuning --------------------------------- 1.35s
rpc_support | Install HWRaid packages ----------------------------------- 0.68s
rpc_support | Install packages required for support --------------------- 0.68s
rpc_support | Install host packages required for support ---------------- 0.67s
rpc_support | Drop rpc-release file ------------------------------------- 0.35s
           to retry, use: --limit @/root/rpc-support.retry

controller1                : ok=24   changed=1    unreachable=0    failed=1
controller2                : ok=24   changed=1    unreachable=0    failed=1
controller3                : ok=24   changed=1    unreachable=0    failed=1
controller3_utility_container-841d3a0b : ok=16   changed=2    unreachable=0    failed=1

Ouch :

[...]

PR #1159, opened 2016-06-14
Ensure all hosts have a pip.conf

mancdaz:
Previously we were relying on some other side effects to get a
pip.conf file to be put onto the hosts:

- compute hosts got it because of the nova role
- infra hosts sometimes got it because of beaver/other roles

This commit ensures that we explicitly install a pip.conf for
the version that we are deploying by adding a playbook to run
the pip-lockdown role. Additionally, we remove any lingering
link files from previous versions to ensure that pip.conf
contains only links for the new version being installed.

Connects #1124

### DIFF
--- a/rpcd/playbooks/pip-lockdown.yml
+++ b/rpcd/playbooks/pip-lockdown.yml
@@ -1,0 +1,35 @@
+---
+# Copyright 2016, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# The purpose here is to allow for the environment to update/build the
+# python wheel files from the CURRENT release, as set in the rpc_release
+# variable.
+- name: Clean up pip release files from previous versions
+  hosts: hosts:all_containers
+  user: root
+  gather_facts: False
+  tasks:
+    - name: remove pip release directory
+      file:
+        path: /root/.pip/links.d/
+        state: absent
+      tags:
+        - pip-link-cleanup
+
+- name: Place pip.conf on hosts
+  hosts: hosts
+  user: root
+  roles:
+    - { role: "pip_lock_down" }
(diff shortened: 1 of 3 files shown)


rpcd/playbooks/pip-lockdown.yml
```diff
@@ -0,0 +1,33 @@
...
+    file:
+      path: /root/.pip/links.d/
+      state: absent
+      tags:
+        - pip-link-cleanup
+
+- name: Place pip.conf on hosts
+  hosts: hosts
+  user: root
+  pre_tasks:
```
git-harry: This line should be removed

rpcd/playbooks/pip-lockdown.yml
```diff
@@ -0,0 +1,33 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# The purpose here is to allow for the environment to update/build the
+# python wheel files from the CURRENT release, as set in the rpc_release
+# variable.
+- name: Clean up pip release files from previous versions
+  hosts: hosts:all_containers
+  user: root
+    file:
```
git-harry: There should be 'tasks:' above this line and the file task should be part of a list e.g.

```
  tasks:
    - file:
```

scripts/deploy.sh
```diff
@@ -157,6 +157,9 @@ if [[ "${DEPLOY_OA}" == "yes" ]]; then
     run_ansible haproxy-install.yml
   fi
 
+  # ensure correct pip.conf
+  openstack-ansible pip-lockdown.yml
+
```
git-harry: Checks failed because the playbook isn't found.
```
pushd ${RPCD_DIR}/playbooks
  openstack-ansible pip-lockdown.yml
popd
```

mancdaz:
recheck_upgrade

git-harry:
recheck_upgrade

git-harry:
recheck_upgrade

mancdaz:
recheck_upgrade

git-harry:
👍  the code makes sense, it would be nice to see this tested on a multinode deployment.

rpcd/playbooks/pip-lockdown.yml
```diff
@@ -0,0 +1,35 @@
...
+  tasks:
+    - name: remove pip release directory
+      file:
+        path: /root/.pip/links.d/
+        state: absent
+      tags:
+        - pip-link-cleanup
+
+- name: Place pip.conf on hosts
+  hosts: hosts
```
sigmavirus24: We're not targeting `all_containers` here too because a different playbook (or playbooks) in OSA will handle this for us, yes?

mancdaz: Yes

sigmavirus24:
👍

Issue #1107, opened 2016-05-31
The nova_cloud_stats.py plugin is reporting wrong number of vcpus


OchiengEd:
##### What version (commit) do you encounter this issue on?
All

##### What type of issue is this (bug, feature, documentation, regression)?
Bug

##### What is the expected behavior?
The nova_cloud_stats.py plugin should return the number of vcpus in the environment.

##### What is actually happening?
The plugin returns the number of physical cores available across the hypervisors which does not factor in the cpu allocation ratio. This is causing false positive alerts for high vcpu usage..

##### How can we reproduce this issue?

```
# python /usr/lib/rackspace-monitoring-agent/plugins/nova_cloud_stats.py | egrep 'vcpu|status'
status okay
metric cloud_resource_used_vcpus uint32 222 vcpu
metric cloud_resource_total_vcpus uint32 224 vcpu
# 
```

##### How can we verify your issue is addressed?
Compare the number of physical cores versus total vcpus reported. Total vcpus should be higher unless cpu allocation ratio of 1 to 1 is used.

##### If this is a large issue, what can be done immediately to alleviate the issue while working on a longer-term solution?
Disable the check to prevent unnecessary alerting

##### What is your evaluation on the importance of this issue, and why?
Medium. Knowing the amount of vcpu resources available can help in capacity planning before resources actually run out.

##### What branches / milestones do you believe this issue should target?
Kilo
Liberty

###### Remove branches (lines) you do not wish to target.
- [x] Master
- [x] liberty-12.2 (https://github.com/rcbops/rpc-openstack/pull/1239)
- [x] kilo (https://github.com/rcbops/rpc-openstack/pull/1240)

mancdaz:
Hi @OchiengEd 

Is the request here that the plugin should factor in the cpu allocation ratio? (e.g 200 cores with allocation ratio of 2 should show 400 vcpu). If so, this was discussed at the time of implementation and deemed outside the scope of this plugin.

OchiengEd:
Yes @mancdaz ,

If this is by design like you say and there are no desires to change how this works, perhaps we should stop giving alerts for high vcpu usage once vcpus used reaches 90% of the physical cores. Which in essence implies we will not have any alarms setup to generate alerts for vcpu usage in private cloud environments.

BjoernT:
[...]

PR #1162, opened 2016-06-15
Add cpu/mem multipliers to nova_cloud_stats plugin

mancdaz:
The nova API reports 'total_vcpus' as the total number of physical cores
available on the hypervisors that it can see. However, because of the
setting cpu_allocation_ratio it's possible to see more 'used_vcpus' than
'total_vcpus' (actual physical cores in the machines).

The same is true of memory and mem_allocation_ratio.

Whilst the nova API does not have the capability of factoring in the
*_allocation_ratios across all hypervisors when reporting 'total_*' numbers,
we can fake it by passing a multiplier to the plugin using numbers that match
the nova conf settings. The default for the plugin remains to use the real
numbers as reported by the nova API.

It's accepted that this does not cover the case where different hypervisors
may have different allocation ratios, but should cover 95% of use cases where
the allocation_ratios are the same across all hypervisors.

Connects #1107

### DIFF
--- a/maas/plugins/nova_cloud_stats.py
+++ b/maas/plugins/nova_cloud_stats.py
@@ -82,8 +82,13 @@ def check(auth_ref, args):
         stats = nova.hypervisor_stats.statistics()
         cloud_stats = collections.defaultdict(dict)
         for metric_name, vals in stats_mapping.iteritems():
+            multiplier = 1
+            if metric_name == 'total_vcpus':
+                multiplier = args.cpu_allocation_ratio
+            elif metric_name == 'total_memory':
+                multiplier = args.mem_allocation_ratio
             cloud_stats[metric_name]['value'] = \
-                getattr(stats, vals['stat_name'])
+                (getattr(stats, vals['stat_name']) * multiplier)
             cloud_stats[metric_name]['unit'] = \
                 vals['unit']
             cloud_stats[metric_name]['type'] = \
@@ -106,6 +111,20 @@ if __name__ == "__main__":
     with print_output():
         parser = argparse.ArgumentParser(
             description='Check Nova hypervisor stats')
+        parser.add_argument('--cpu',
+                            type=float,
+                            default=1.0,
+                            required=False,
+                            action='store',
+                            dest='cpu_allocation_ratio',
+                            help='cpu allocation ratio')
+        parser.add_argument('--mem',
+                            type=float,
+                            default=1.0,
+                            required=False,
+                            action='store',
+                            dest='mem_allocation_ratio',
+                            help='mem allocation ratio')
         parser.add_argument('ip', nargs='?',
                             type=ipaddr.IPv4Address,
                             help='Nova API IP address')
--- a/rpcd/etc/openstack_deploy/user_extras_variables.yml
+++ b/rpcd/etc/openstack_deploy/user_extras_variables.yml
@@ -79,6 +79,10 @@ maas_filesystem_critical_threshold: 90.0
 #    warning_threshold: 80.0
 #    critical_threshold: 90.0
 
+# overrides for the nova_cloud_stats  maas plugin
+cloud_resource_cpu_allocation_ratio: "{{ nova_cpu_allocation_ratio }}"
+cloud_resource_mem_allocation_ratio: "{{ nova_ram_allocation_ratio }}"
+
 # For an AIO it's recommended to set the following to limit the expected RAM
 # usage of elasticsearch.
 # elasticsearch_heap_size_mb: 1024
--- a/rpcd/playbooks/roles/rpc_maas/defaults/main.yml
+++ b/rpcd/playbooks/roles/rpc_maas/defaults/main.yml
@@ -301,6 +301,8 @@ cloud_resource_warning_vcpus: 80.0
 cloud_resource_critical_vcpus: 90.0
 cloud_resource_warning_disk_space: 80.0
 cloud_resource_critical_disk_space: 90.0
+cloud_resource_cpu_allocation_ratio: 2.0
+cloud_resource_mem_allocation_ratio: 1.0
 
 # List of checks, by type - variable:
 hp_checks_list:
--- a/releasenotes/notes/nova-allocation-ratios-869b6d3cabbf914d.yaml
+++ b/releasenotes/notes/nova-allocation-ratios-869b6d3cabbf914d.yaml
@@ -1,0 +1,10 @@
+---
+fixes:
+  - Previously, the nova_cloud_stats maas plugin was
+    incorrectly reporting total cpu and total memory
+    amounts available across all hypervisors, as it was
+    not taking into account the allocation_ratios that
+    are set at the hypervisor level. Now, it attempts
+    to correctly scale those values by passing in a
+    multiplier based on the allocation ratios set in
+    the config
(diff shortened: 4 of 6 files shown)


mancdaz:
@BjoernT @OchiengEd here you go, let me know what you think

rpcd/playbooks/roles/rpc_maas/defaults/main.yml
```diff
@@ -297,6 +297,8 @@ cloud_resource_warning_vcpus: 80.0
 cloud_resource_critical_vcpus: 90.0
 cloud_resource_warning_disk_space: 80.0
 cloud_resource_critical_disk_space: 90.0
+cloud_resource_cpu_allocation_ratio: 2
+cloud_resource_mem_allocation_ratio: 1
```
git-harry: What happens if these ratios are not the defaults? At a minimum this needs documenting somewhere. Maybe it would be better to set these equal to the ratio vars used by os_nova and then set them in user_variables. This would prevent anyone getting caught out.

maas/plugins/nova_cloud_stats.py
```diff
@@ -82,8 +82,13 @@ def check(auth_ref, args):
         stats = nova.hypervisor_stats.statistics()
         cloud_stats = collections.defaultdict(dict)
         for metric_name, vals in stats_mapping.iteritems():
+            multiplier = 1
+            if metric_name == 'total_vcpus':
+                multiplier = args.cpu_allocation_ratio
+            elif metric_name == 'total_memory':
```
git-harry: Why was the same not done for 'used_memory' or 'free_memory'?

mancdaz: Because used_memory and free_memory are the actual real values, and need to be for this whole thing to work. Only the totals need to be artificially multiplied to match the allocation_ratios

rpcd/playbooks/roles/rpc_maas/defaults/main.yml
mancdaz: Yeah as discussed in the issue, this is not ideal but other than referring to vars from another role (which I think is what you're suggesting) I couldn't think of another way.

git-harry:
In addition to inline comments this should probably have a release note.

Apsu:
RPC-AIO-Liberty-PR: Fail

Apsu:
RPC-AIO-Liberty-PR: Fail

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

rpcd/etc/openstack_deploy/user_extras_variables.yml
```diff
@@ -79,6 +79,10 @@ maas_filesystem_critical_threshold: 90.0
 #    warning_threshold: 80.0
 #    critical_threshold: 90.0
 
+# overrides for the nova_cloud_resource maas plugin
```
git-harry: There is no plugin with that name

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Liberty-PR: Fail

evrardjp:
Good fix itself.

It shows we are doing it wrong in OSA (and therefore RPC). IMO, the override should only be defined when needed on the host (default should be 0.0): the scheduler should have the details about the overcommit per aggregate (if I recall that's how it was done before, you define the nova nova_cpu_allocation_ratio to 0.0 on your nodes).

My point is: we should have a better approach on how to reliably discover the cpu_allocation_ratio on the node. This should be a node local fact, that we consume in the maas script. OSA should do that properly. This deserves a "whishlist" type of bug in OSA.

Else I'd say good, but not tested.

mancdaz:
Genuine failure with the patch. I'll investigate and rework

```
12:22:18 Checks and Alarms with failures:
12:22:18  - Entity:enopz4h4H8:jrpcaioiad-82d
12:22:18    ! Check:agent.plugin:chEQMNz4Kh:nova_cloud_stats_check--jrpcaioiad-82d-nova-api-os-compute-container-83b3fd5b: Completed:False Status:Plugin exited with non-zero status code (code=2)
12:22:18      ! Alarm:alCUxhMGtv:nova_cloud_vcpu_status--jrpcaioiad-82d-nova-api-os-compute-container-83b3fd5b:CRITICAL
12:22:18      ! Alarm:alnBt1JI9b:nova_cloud_memory_status--jrpcaioiad-82d-nova-api-os-compute-container-83b3fd5b:CRITICAL
12:22:18      ! Alarm:alzmeG3zko:nova_cloud_disk_status--jrpcaioiad-82d-nova-api-os-compute-container-83b3fd5b:CRITICAL
12:22:18 
12:22:18 msg: Task failed as maximum retries was encountered
12:22:18 
12:22:18 FATAL: all hosts have already failed -- aborting
```

maas/plugins/nova_cloud_stats.py
```diff
@@ -106,6 +111,20 @@ def main(args):
     with print_output():
         parser = argparse.ArgumentParser(
             description='Check Nova hypervisor stats')
+        parser.add_argument('--cpu',
+                            type=int,
```
git-harry: I think this is the cause of the failure. You're supplying floats but requiring ints.

mancdaz: argh, I only switched to floats in the vars file to match the nova options

rpcd/playbooks/roles/rpc_maas/defaults/main.yml
```diff
@@ -301,6 +301,8 @@ cloud_resource_warning_vcpus: 80.0
 cloud_resource_critical_vcpus: 90.0
 cloud_resource_warning_disk_space: 80.0
 cloud_resource_critical_disk_space: 90.0
+cloud_resource_cpu_allocation_ratio: 2.0
```
sigmavirus24: If someone does change `nova_*_allocation_ratio` will they also know to change these variables that are just slightly differently named (e.g., `nova_ram_allocation_ratio` -> `cloud_resource_mem_allocation_ratio`)?

Relatedly, do we really want to force support to set *4* variables so that monitoring does the right thing?

maas/plugins/nova_cloud_stats.py
```diff
@@ -106,6 +111,20 @@ def main(args):
     with print_output():
         parser = argparse.ArgumentParser(
             description='Check Nova hypervisor stats')
+        parser.add_argument('--cpu',
```
sigmavirus24: These flag names could be more descriptive for people who test these scripts on the CLI.

rpcd/playbooks/roles/rpc_maas/defaults/main.yml
mancdaz: No, the way this PR is now structured is that only the nova vars need to be set for an rpco deployment. This in ```user_extras_variables.yml```:

```
cloud_resource_cpu_allocation_ratio: "{{ nova_cpu_allocation_ratio }}"
cloud_resource_mem_allocation_ratio: "{{ nova_ram_allocation_ratio }}"
```

and this in ```user_variables.yml```

```
nova_cpu_allocation_ratio: 2.0
nova_ram_allocation_ratio: 1.0
```

allow that to happen

sigmavirus24: I totally missed the `user_extras_variables` change... Sorry about that.

sigmavirus24:
The flag names (and help descriptions) are unhelpful for anyone trying to debug the issue, but otherwise this looks like it solves the problem. 👍

git-harry:
👍

PR #1165, opened 2016-06-15
Move user variables/secrets files to new location

prometheanfire:
Here we use a new filestructure so that deployers and devs will
not step on eachother's feet when changing variables around.  The
new user variables file layout looks like this.

```
user_osa_secrets.yml
user_osa_variables_defaults.yml
user_osa_variables_overrides.yml
user_rpco_secrets.yml
user_rpco_variables_defaults.yml
user_rpco_variables_overrides.yml
```

This patch does the following

1. Combine rpcd/etc/openstack_deploy/user_extras_variables.yml and
   rpcd/etc/openstack_deploy/user_variables.yml into a new file
   named rpcd/etc/openstack_deploy/user_rpco_variables_defaults.yml.
2. Delete rpcd/etc/openstack_deploy/user_extras_variables.yml and
3. Rename (move) rpcd/etc/openstack_deploy/user_extras_secrets.yml
   to rpcd/etc/openstack_deploy/user_rpco_secrets.yml.
4. Update install script to move OSA user/secrets files to the
   correct location.

Connects https://github.com/rcbops/u-suk-dev/issues/185

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,10 +39,10 @@ as well as adding a Rackspace tab and a Solutions tab, which provides
 Heat templates for commonly deployed applications.
 * `kibana.yml` - Setup Kibana on the Kibana hosts for the logging dashboard.
 * `logstash.yml` - deploys a logstash host. If this play is used, be sure to
-uncomment the related block in user_extra_variables.yml before this play is
-run and then rerun the appropriate plays in openstack-ansible after this
-play to ensure that rsyslog ships logs to logstash. See steps 11 - 13 below
-for more.
+copy the related block in user_rpco_variables_defaults.yml to
+user_rpco_variables_overrides before this play is run and then rerun the
+appropriate plays in openstack-ansible after this play to ensure that rsyslog
+ships logs to logstash. See steps 2-4 and 3-2 below for more.
 * `repo-build.yml` - scans throug the YAML files in the source tree and builds
 any packages or git sources into wheels and deploys them to the local repo
 server(s).
@@ -70,13 +70,14 @@ alarm configured for it.
 2. Unless doing an AIO build, prepare the openstack-ansible configuration.
   1. recursively copy the openstack-ansible-deployment configuration files:
      `cp -R openstack-ansible/etc/openstack_deploy /etc/openstack_deploy`
-  2. merge /etc/openstack_deploy/user_variables.yml with rpcd/etc/openstack_deploy/user_variables.yml:
+  2. move OSA variables to the correct locations.
 
      ```
-     scripts/update-yaml.py /etc/openstack_deploy/user_variables.yml rpcd/etc/openstack_deploy/user_variables.yml
+     rm /etc/openstack_deploy/user_variables.yml  # unused, nothing set
+     mv /etc/openstack_deploy/user_secrets.yml /etc/openstack_deploy/user_osa_secrets.yml
      ```
   3. copy the RPC configuration files:
-     1. `cp rpcd/etc/openstack_deploy/user_extras_*.yml /etc/openstack_deploy`
+     1. `cp rpcd/etc/openstack_deploy/user_*_defaults.yml /etc/openstack_deploy`
      2. `cp rpcd/etc/openstack_deploy/env.d/* /etc/openstack_deploy/env.d`
   4. If the ELK stack is not going to be used, remove the container
      configurations from the environment:
@@ -90,7 +91,8 @@ alarm configured for it.
   1. If building AIO, set `DEPLOY_AIO=yes` before running
   2. If building without the ELK stack, set `DEPLOY_ELK=no` before running
 4. If you want MaaS working with AIO, do the following:
-  1. edit `/etc/openstack_deploy/user_extras_variables.yml` to add credentials
+  1. edit `/etc/openstack_deploy/user_rpco_variables_overrides.yml` to add
+     credentials
   2. run the MaaS setup plays:
      `cd /opt/rpc-openstack/rpcd/playbooks && openstack-ansible setup-maas.yml`
   3. run the MaaS verify play:
--- a/releasenotes/notes/master-96b4f6d34b87fe3d.yaml
+++ b/releasenotes/notes/master-96b4f6d34b87fe3d.yaml
@@ -1,0 +1,21 @@
+---
+features:
+  - Variables and secrets are now managed in a manner that
+    allows downstream deployers to not have to worry about
+    their settings being overwritten.  They are provided
+    with overrides files that they alone manage after we
+    lay them down.  We have also taken this time to rename
+    our user_* files to separate out what specifically the
+    variables override, whether it be upstream splitting
+    the OSA relevant changed of the local RPC-O changes or
+    local RPCO changes.
+deprecations:
+  - We are deprecating the use of user_variables and
+    user_secrets files other than the ones found below.
+
+    user_osa_secrets.yml
+    user_osa_variables_defaults.yml
+    user_osa_variables_overrides.yml
+    user_rpco_secrets.yml
+    user_rpco_variables_defaults.yml
+    user_rpco_variables_overrides.yml
--- a/rpcd/etc/openstack_deploy/user_osa_variables_defaults.yml
+++ b/rpcd/etc/openstack_deploy/user_osa_variables_defaults.yml
@@ -1,5 +1,5 @@
 ---
-# Copyright 2015, Rackspace US, Inc.
+# Copyright 2016, Rackspace US, Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -12,6 +12,11 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+## Warning
+#
+# This file is controlled by the devs and will be overwritten on upgrade
+# If you wish to override the variables set here, use /etc/openstack_deploy/user_osa_variables_overrides.yml
 
 # SQLAlchemy/Olso Thread Pool Settings
 rpc_conn_pool_size: 180
@@ -122,3 +127,11 @@ horizon_custom_uploads:
 # Use RPC python package index
 repo_build_pip_extra_indexes:
   - "https://rpc-repo.rackspace.com/pools"
+
+## Host security hardening
+# The openstack-ansible-security role provides security hardening for hosts
+# by applying security configurations from the STIG. Hardening is disabled by
+# default, but an option to opt-in is available by setting the following
+# variable to 'true'.
+# Docs: http://docs.openstack.org/developer/openstack-ansible-security/
+# apply_security_hardening: true
--- a/rpcd/etc/openstack_deploy/user_rpco_variables_defaults.yml
+++ b/rpcd/etc/openstack_deploy/user_rpco_variables_defaults.yml
@@ -1,5 +1,5 @@
 ---
-# Copyright 2014, Rackspace US, Inc.
+# Copyright 2016, Rackspace US, Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -12,6 +12,11 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+## Warning
+#
+# This file is controlled by the devs and will be overwritten on upgrade
+# If you wish to override the variables set here, use /etc/openstack_deploy/user_rpco_variables_overrides.yml
 
 ## Rackspace Cloud Details
 rackspace_cloud_auth_url: https://identity.api.rackspacecloud.com/v2.0
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -21,8 +21,8 @@ export BOOTSTRAP_OPTS=${BOOTSTRAP_OPTS:-""}
 
 OA_DIR='/opt/rpc-openstack/openstack-ansible'
 RPCD_DIR='/opt/rpc-openstack/rpcd'
-RPCD_VARS='/etc/openstack_deploy/user_extras_variables.yml'
-RPCD_SECRETS='/etc/openstack_deploy/user_extras_secrets.yml'
+RPCD_VARS='/etc/openstack_deploy/user_rpco_variables_defaults.yml'
+RPCD_SECRETS='/etc/openstack_deploy/user_rpco_secrets.yml'
 
 function run_ansible {
   openstack-ansible ${ANSIBLE_PARAMETERS} --forks ${FORKS} $@
@@ -56,12 +56,13 @@ if [[ "${DEPLOY_AIO}" == "yes" ]]; then
   export DEPLOY_HAPROXY="yes"
   if [[ ! -d /etc/openstack_deploy/ ]]; then
     ./scripts/bootstrap-aio.sh
+    # move OSA variables file to AIO location.
+    mv /etc/openstack_deploy/user_variables.yml /etc/openstack_deploy/user_osa_aio_variables.yml
     pushd ${RPCD_DIR}
       for filename in $(find etc/openstack_deploy/ -type f -iname '*.yml'); do
         if [[ ! -a "/${filename}" ]]; then
           cp "${filename}" "/${filename}";
         fi
-        ../scripts/update-yaml.py "/${filename}" "${filename}";
       done
     popd
     # ensure that the elasticsearch JVM heap size is limited
@@ -90,9 +91,14 @@ if [[ "${DEPLOY_AIO}" == "yes" ]]; then
       echo "osd_directory: true" | tee -a $RPCD_VARS
       echo "osd_directories:" | tee -a $RPCD_VARS
       echo "  - /var/lib/ceph/osd/mydir1" | tee -a $RPCD_VARS
-      sed -i "s/glance_default_store:.*/glance_default_store: rbd/" /etc/openstack_deploy/user_variables.yml
-      echo "nova_libvirt_images_rbd_pool: vms" | tee -a /etc/openstack_deploy/user_variables.yml
-      echo "cinder_ceph_client_uuid:"  | tee -a /etc/openstack_deploy/user_secrets.yml
+      echo "glance_default_store: rbd" | tee -a /etc/openstack_deploy/user_osa_variables_defaults.yml
+      echo "nova_libvirt_images_rbd_pool: vms" | tee -a $RPCD_VARS
+    else
+      if [[ "$DEPLOY_SWIFT" == "yes" ]]; then
+        echo "glance_default_store: swift" | tee -a /etc/openstack_deploy/user_osa_variables_defaults.yml
+      else
+        echo "glance_default_store: file" | tee -a /etc/openstack_deploy/user_osa_variables_defaults.yml
+      fi
     fi
     # set the ansible inventory hostname to the host's name
     sed -i "s/aio1/$(hostname)/" /etc/openstack_deploy/openstack_user_config.yml
@@ -109,31 +115,41 @@ if [[ "${DEPLOY_AIO}" == "yes" ]]; then
   rm -f /etc/openstack_deploy/conf.d/ceilometer.yml
 fi
 
+# move OSA secrets to correct locations
+mv /etc/openstack_deploy/user_secrets.yml /etc/openstack_deploy/user_osa_secrets.yml
+
 # Apply host security hardening with openstack-ansible-security
 # The is applied as part of setup-hosts.yml
 if [[ "$DEPLOY_HARDENING" == "yes" ]]; then
-  if grep -q '^apply_security_hardening:' /etc/openstack_deploy/user_variables.yml
-  then
-    sed -i "s/^apply_security_hardening:.*/apply_security_hardening: true/" /etc/openstack_deploy/user_variables.yml
+  if grep -q '^apply_security_hardening:' /etc/openstack_deploy/user_osa_variables_defaults.yml; then
+    sed -i "s/^apply_security_hardening:.*/apply_security_hardening: true/" /etc/openstack_deploy/user_osa_variables_defaults.yml
   else
-    echo "apply_security_hardening: true" >> /etc/openstack_deploy/user_variables.yml
+    echo "apply_security_hardening: true" >> /etc/openstack_deploy/user_osa_variables_defaults.yml
   fi
 else
-  if grep -q '^apply_security_hardening:' /etc/openstack_deploy/user_variables.yml
-  then
-    sed -i "s/^apply_security_hardening:.*/apply_security_hardening: false/" /etc/openstack_deploy/user_variables.yml
+  if grep -q '^apply_security_hardening:' /etc/openstack_deploy/user_osa_variables_defaults.yml; then
+    sed -i "s/^apply_security_hardening:.*/apply_security_hardening: false/" /etc/openstack_deploy/user_osa_variables_defaults.yml
   else
-    echo "apply_security_hardening: false" >> /etc/openstack_deploy/user_variables.yml
+    echo "apply_security_hardening: false" >> /etc/openstack_deploy/user_osa_variables_defaults.yml
   fi
 fi
 
 # ensure all needed passwords and tokens are generated
-./scripts/pw-token-gen.py --file /etc/openstack_deploy/user_secrets.yml
+./scripts/pw-token-gen.py --file /etc/openstack_deploy/user_osa_secrets.yml
 ./scripts/pw-token-gen.py --file $RPCD_SECRETS
 
 # Apply any patched files.
 cd ${RPCD_DIR}/playbooks
 openstack-ansible -i "localhost," patcher.yml
+
+# set permissions and lay down overrides files
+chmod 0440 /etc/openstack_deploy/user_*_defaults.yml
+if [[ ! -f /etc/openstack_deploy/user_osa_variables_overrides.yml ]]; then
+  cp "${RPCD_DIR}"/etc/openstack_deploy/user_osa_variables_overrides.yml /etc/openstack_deploy/user_osa_variables_overrides.yml
+fi
+if [[ ! -f /etc/openstack_deploy/user_rpco_variables_overrides.yml ]]; then
+  cp "${RPCD_DIR}"/etc/openstack_deploy/user_rpco_variables_overrides.yml /etc/openstack_deploy/user_rpco_variables_overrides.yml
+fi
 
 # begin the openstack installation
 if [[ "${DEPLOY_OA}" == "yes" ]]; then
--- a/rpcd/etc/openstack_deploy/user_osa_variables_overrides.yml
+++ b/rpcd/etc/openstack_deploy/user_osa_variables_overrides.yml
@@ -1,5 +1,5 @@
 ---
-# Copyright 2014, Rackspace US, Inc.
+# Copyright 2016, Rackspace US, Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -13,8 +13,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-maas_keystone_password:
-rpc_support_holland_password:
-kibana_password:
-maas_rabbitmq_password:
-fsid_uuid:
+## Using this file
+#
+# This file is laid down only once and is used for overriding variables based on the file name.
+# This file will not be overwritten by upgrades.
+
+# this variable is here because vars files need to have at least ONE variable defined.
+this_file_is_reserved_for_support: True
--- a/rpcd/etc/openstack_deploy/user_rpco_secrets.yml
+++ b/rpcd/etc/openstack_deploy/user_rpco_secrets.yml
@@ -1,0 +1,20 @@
+---
+# Copyright 2016, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+maas_keystone_password:
+rpc_support_holland_password:
+kibana_password:
+maas_rabbitmq_password:
+fsid_uuid:
--- a/rpcd/etc/openstack_deploy/user_rpco_variables_overrides.yml
+++ b/rpcd/etc/openstack_deploy/user_rpco_variables_overrides.yml
@@ -1,0 +1,22 @@
+---
+# Copyright 2016, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+## Using this file
+#
+# This file is laid down only once and is used for overriding variables based on the file name.
+# This file will not be overwritten by upgrades.
+
+# this variable is here because vars files need to have at least ONE variable defined.
+this_file_is_reserved_for_support: True


prometheanfire:
```
18:24:16 failed: [jrpcaioiad-ed5] => {"changed": true, "cmd": "virsh secret-set-value --secret --base64 AQA85mJXuUxgMRAALrzU5BL5cftCszZ/S/A5KA==", "delta": "0:00:00.012142", "end": "2016-06-16 18:24:16.529965", "rc": 1, "start": "2016-06-16 18:24:16.517823", "warnings": []}
18:24:16 stderr: error: failed to get secret '--base64'
18:24:16 error: uuidstr in virSecretLookupByUUIDString must be a valid UUID
```

prometheanfire:
recheck_ceph

prometheanfire:
recheck_swift

failed on maas

```
18:38:43 failed: [jrpcaioiad-497] => {"failed": true}
18:38:43 msg: Zero entities with the label jrpcaioiad-497 exist. Entities should be created as part of the hardware provisioning process, if missing, please consult the internal documentation for associating one with the device.
```

prometheanfire:
same two problems

prometheanfire:
recheck_ceph
recheck_swift

prometheanfire:
just making a note that this passed all but the commit looks janked up, so working on that

prometheanfire:
recheck_ceph

same error as before, so...

prometheanfire:
recheck_swift

prometheanfire:
recheck_all

prometheanfire:
recheck_all

prometheanfire:
recheck_all

prometheanfire:
recheck_all

prometheanfire:
recheck_all

prometheanfire:
recheck_all

prometheanfire:
setting blocked til @hughsaunders has time to work with me on gating

git-harry:
The description of the new file layout does not match the description of the changes made or the changes themselves.

The new file layout has user_variables.yml renamed to user_osa_variables_defaults.yml and user_extras_variables.yml renamed to user_rpco_variables_defaults.yml. What the commit has actually done is combine both user_variables.yml and user_extras_variables.yml into user_rpco_variables_defaults.yml while also keeping user_extras_variables.yml. In addition the deploy script seems to be adding some vars to user_rpco_variables_defaults.yml and some to user_osa_variables_defaults.yml.

prometheanfire:
user_variables and user_extra_variables from rpc-o should become user_rpco_variables_defaults.yml because they are OUR (rpc-o's) overrides.  The user_variables from osa moves to user_osa_variables_defaults.

as far as removing user_extra_secrets from rpco I forgot to do that it looks like, will amend in a sec

git-harry:
@prometheanfire, the way it is currently set up user_variables.yml contains OSA variables that RPCO is overriding. user_extras_variables.yml contains RPCO variables, e.g. MaaS or ELK, that are being overridden/set. If you combine them in user_rpco_variables_defaults.yml what is the point in user_osa_variables_defaults.yml because OSA doesn't set anything overrides by default.

In addition, regarding the changes deploy.sh I can't work out how you've decided which variables you are putting in user_rpco_variables_defaults.yml and which you are putting in user_osa_variables_defaults.yml.

prometheanfire:
What should I do with osa's user_variables then? just delete it?

prometheanfire:
swift failed on this.

```
18:39:02 failed: [jrpcaioiad-5a9] => {"failed": true}
18:39:02 msg: Zero entities with the label jrpcaioiad-5a9 exist. Entities should be created as part of the hardware provisioning process, if missing, please consult the internal documentation for associating one with the device.```

**Comment by git-harry** [comment]
The following is what was originally agreed:

```
user_osa_secrets_defaults.yml
user_osa_secrets_overrides.yml
user_osa_variables_defaults.yml
user_osa_variables_overrides.yml
user_rpco_secrets_defaults.yml
user_rpco_secrets_overrides.yml
user_rpco_variables_defaults.yml
user_rpco_variables_overrides.yml
```

If the vars for the two parts of the project are combined, you would only need to have something like:

```
user_secrets_defaults.yml
user_secrets_overrides.yml
user_variables_defaults.yml
user_variables_overrides.yml
```

Currently your patch is a mixture of the two approaches. If you want to combine the vars please update the patch so that:
1. it actually does that
2. the commit message and pull request descriptions describe that
3. you explain why you want to do that

**Comment by hughsaunders** [comment]
The "Zero entities with label" failure is unlikely to be related to this patch. I think the problem is "Task Failed: Install cinder server / ceph_client / Set nova secret value in libvirt"

**Comment by prometheanfire** [comment]
@git-harry the reason it's a combined method is because we (rpc-o) override osa configs already.  This means we have three sources for that `user_osa_variables_*.yml` level instead of two.  Support, Devs, and Upstream.  Since we sed on that already (to enable hardening for example) perhaps we should move those variables we sed into our (rpc-o's) osa variables file.  At that point we wouldn't need to copy upstream OSA's variables file and would just have two sources (Devs and Support) for `user_osa_variables_*.yml`

File: rpcd/etc/openstack_deploy/user_rpco_variables_defaults.yml [hunk 1]
```diff
@@ -1,5 +1,5 @@
 ---
-# Copyright 2014, Rackspace US, Inc.
+# Copyright 2015, Rackspace US, Inc.
```
evrardjp: When we are at changing years, I'd use 2016. Because we are in 2016, right? :p

prometheanfire: ya, I noticed that last time I amended but forgot to change.  I think it came from rpcd/etc/openstack_deploy/user_variables.yml though, where it was 2015

scripts/deploy.sh
```diff
@@ -56,12 +56,14 @@ if [[ "${DEPLOY_AIO}" == "yes" ]]; then
   export DEPLOY_HAPROXY="yes"
   if [[ ! -d /etc/openstack_deploy/ ]]; then
     ./scripts/bootstrap-aio.sh
+    # move OSA conf files to correct locations
+    mv /etc/openstack_deploy/user_variables.yml /etc/openstack_deploy/user_osa_variables_defaults.yml
```
hughsaunders: I think its confusing to rename files from OSA. It will be difficult to trace where user_osa_variables_defaults came from because its not in the OSA or RPC repos.

scripts/deploy.sh
```diff
@@ -90,9 +92,9 @@ if [[ "${DEPLOY_AIO}" == "yes" ]]; then
       echo "osd_directory: true" | tee -a $RPCD_VARS
       echo "osd_directories:" | tee -a $RPCD_VARS
       echo "  - /var/lib/ceph/osd/mydir1" | tee -a $RPCD_VARS
-      sed -i "s/glance_default_store:.*/glance_default_store: rbd/" /etc/openstack_deploy/user_variables.yml
-      echo "nova_libvirt_images_rbd_pool: vms" | tee -a /etc/openstack_deploy/user_variables.yml
-      echo "cinder_ceph_client_uuid:"  | tee -a /etc/openstack_deploy/user_secrets.yml
+      sed -i "s/glance_default_store:.*/glance_default_store: rbd/" /etc/openstack_deploy/user_osa_variables_defaults.yml
+      echo "nova_libvirt_images_rbd_pool: vms" | tee -a /etc/openstack_deploy/user_rpco_variables_defaults.yml
+      echo "cinder_ceph_client_uuid:"  | tee -a /etc/openstack_deploy/user_rpco_variables_defaults.yml
```
hughsaunders: cinder_ceph_client_uuid was being appended to user_secrets and is now being appended to defaults. I think this line should be removed as cinder_ceph_client_uuid is in OSA/user_secrets already. This is probably causing the problem as the cinder_ceph_client_uuid is specified with a null value (to be replaced by the pwgen script) but pwgen is not run against the defaults file. The defaults file is then specified after the secrets file by openstack-ansible, so null from user_rpco_variables_defaults takes precedent over the generated secret from user_osa_secrets_defaults.yml

scripts/deploy.sh
prometheanfire: we have to rename it because of the way we are doing precedence (by name).

user_variables overrides user_rpc_variables.

scripts/deploy.sh
prometheanfire: oh, typo then for that, will submit update to pr to test

hughsaunders:
recheck_ceph recheck_swift

hughsaunders:
recheck_ceph recheck_swift

prometheanfire:
recheck_ceph recheck_swift

b3rnard0:
@nrb Please provide feedback on this issue when you have a chance.

prometheanfire:
recheck_ceph

nrb:
This looks like a reasonable solution to me, especially given the constraints around OSA's Mitaka branch right now.

There's ongoing discussion about how variables might be managed alongside reworked inventory upstream, but my best guess is those would be in Newton at the earliest, but far more likely Ocata or P, given the lateness of the current dev cycle and the scope of changes.

https://review.openstack.org/#/c/332595/ aims to tackle a similar problem from OSA's side, but with the env.d structure rather than the variable files. This makes OSA upgrades easier, but it doesn't necessarily solve for RPC-specific environment files. I'm open to suggestions on that front, but I'm not sure this kind of thing would be accepted as a backport to Mitaka.

evrardjp:
I agree with @git-harry on the variable filenames. 

This code is different than the initial plan.

If we are having another approach, let me propose this one:

    user_osa_secrets.yml                #This one will be overriden anyway
    user_rpco_secrets.yml               #This one will be overriden anyway
    user_osa_variables_defaults.yml     #This one gathers Rackspace sane defaults
    user_osa_variables_overrides.yml    #This one gathers deployer's OSA changes
    user_rpco_variables_defaults.yml    #This one should be empty or non existing
    user_rpco_variables_overrides.yml   #This one gathers deployer's rpc-o changes

if sensitive details are added into these configuration files, we could even think of having: 

    user_sensitive_details.yml  

This one could be used for highly secure environments: it helps support (by allowing the creation of automatic tools that send the info from other variables) and it can be vault encrypted.

But let's forget the last one, that can come later... The thing is, we want to seggregate as much as possible, because we have tools doing the merging easily/automatically. We don't have tools for unmerging/recovering the source of the info. 

Merging directly ourselves was a mistake in the past and should (IMO) be avoided. The initally proposed pattern followed that.

prometheanfire:
I'm fine with altering this, we do need to keep in mind that each level of precedence needs to override the others based on it's file names.

It seems like osa isn't providing any defaults anymore because they've been integrated, which is good.

I'm not sure what to do about secrets, not having overrides for it makes sense I guess, but I'm still not totally convinced.  Could just be tired.

As far as sensitive details go, they can already do that.  We can lay it out in docs but I don't think we should lay the file down.

I do really really want to avoid having to merge anything at any point down the line.  Even for regular maintenance if possible (we just overwrite the entire defaults file).

In short, I'm fine with @evrardjp's proposal.  I think his suggestion also follows what was being said by others @git-harry and @hughsaunders

evrardjp:
Keep in mind I'm not pushing anything, just explaining a view. 

I fully agree the deployers can already do that ```user_sensitive_details.yml``` system, I'm just unsure if they do it or not... I think this kind of variable thing is more a communication issue than a code issue itself.

prometheanfire:
I need to talk to support next time I'm in to ask their opinion about how we divide up variables.

Ya, that's just a deployer's decision, we can doc it, but they'd create/modify the file.

prometheanfire:
recheck_ceph

failed the following tempest tests, maybe there's another file I'm missing somewhere?

```
tempest.api.volume.admin.test_volumes_backup.VolumesBackupsV1Test.test_volume_backup_create_get_detailed_list_restore_delete[id-a66eb488-8ee1-47d4-8e9f-575a095728c6]
tempest.api.volume.admin.test_volumes_backup.VolumesBackupsV2Test.test_volume_backup_create_get_detailed_list_restore_delete[id-a66eb488-8ee1-47d4-8e9f-575a095728c6]```

perhaps I need to readd the 

`echo "cinder_ceph_client_uuid:"  | tee -a /etc/openstack_deploy/user_secrets.yml` line?  One of the earlier reviews said I could remove it.

**Comment by prometheanfire** [comment]
that uuid var is in the osa secrets file though

**Comment by prometheanfire** [comment]
same tempest fail

`tempest.exceptions.VolumeBackupException: Volume backup 8cfca59c-54ec-4397-b456-5d6e39cff94d failed and is in ERROR status`

http://jenkins.propter.net/job/RPC-AIO/7869/console

**Comment by prometheanfire** [comment]
@BjoernT would the layout/location of the vars mentioned in https://github.com/rcbops/rpc-openstack/pull/1165#issuecomment-227811126 work for you or do you have any suggestions as to how you'd like to see it laid out?  The only requirements is that they be named user_*.yml and that it's ordered by name where the user_z.yml would override user_a.yml

**Comment by prometheanfire** [comment]
recheck_all

**Comment by Apsu** [comment]
RPC-AIO-Ceph-Liberty-PR: Fail

**Comment by Apsu** [comment]
RPC-AIO-Liberty-PR: Fail

**Comment by Apsu** [comment]
RPC-AIO-Liberty-PR: Pass

**Comment by Apsu** [comment]
RPC-AIO-Ceph-Liberty-PR: Fail

**Comment by prometheanfire** [comment]
`tempest.exceptions.VolumeBackupException: Volume backup 1459bf72-5f29-4a8d-b06a-7a5cf5aa6b2e failed and is in ERROR status`

so that's still neat

File: scripts/deploy.sh [hunk 4]
```diff
@@ -90,9 +92,8 @@ if [[ "${DEPLOY_AIO}" == "yes" ]]; then
       echo "osd_directory: true" | tee -a $RPCD_VARS
       echo "osd_directories:" | tee -a $RPCD_VARS
       echo "  - /var/lib/ceph/osd/mydir1" | tee -a $RPCD_VARS
-      sed -i "s/glance_default_store:.*/glance_default_store: rbd/" /etc/openstack_deploy/user_variables.yml
-      echo "nova_libvirt_images_rbd_pool: vms" | tee -a /etc/openstack_deploy/user_variables.yml
-      echo "cinder_ceph_client_uuid:"  | tee -a /etc/openstack_deploy/user_secrets.yml
```
evrardjp: Why has this line disappeared?

evrardjp:
@BjoernT Could you please give feedback on the variables?

https://github.com/rcbops/rpc-openstack/pull/1165#issuecomment-227811126 or https://github.com/rcbops/rpc-openstack/pull/1165#issuecomment-227669946 are different approaches, that this commit hasn't taken into consideration (yet) waiting for your feedback.

Thank you in advance!

scripts/deploy.sh
prometheanfire: The cinder_client_uuid line disapeared because it already exists in another location (in openstack-ansible itself).  This was suggested for removal by @hughsaunders iirc.

prometheanfire:
@evrardjp I talked to @BjoernT Friday and he stated that your separation method is preferred.

prometheanfire:
@evrardjp the only thing that concerns me is that your method will still require us to combine files, which I thought we were trying to avoid.

evrardjp:
@prometheanfire The only parts that need to combine are passwords in my case.
It's probably a better idea in the long run to split them too.

Reasons:

1. Deployers are able to diff the files based on first column of the secret file to see the introduced secrets
2. This allows to run playbooks without having to care if the variable will be there or not: the minimum set of secrets being already generated in the "defaults secrets"
3. Upgrades are easier: We just move our ```user_secrets.yml``` to the secret overrides.

BjoernT:
Yes that is correct. The filenames are secondary to me as long as we have a predictable way of how they are get loaded into ansible

prometheanfire:
Ah, right, since osa has everything in vars and their variables file is entirely commented out we can just not use it.  The ONLY thing we need to keep in mind with that is that the aio script currently seds some stuff in it.  This means we'll need to move those variables to our file and sed that instead or alter the method we use to add/change variables that targeted the osa user_defaults file.

What do you mean by moving our secrets to secret overrides?  we should NEVER touch the overrides as that's the highest priority and reserved for support.

Secrets is a harder use case to have, because copying into place and then populating the secrets is the 'standard'.  Meaning that we can't overwrite the file on each upgrade.  This means that even if we have the files separate between osa and rpc-o we'd still need to inspect it and munge it on each upgrade.

If, as @evrardjp suggests, support manually manages the secret file(s), then we just lay it down once and let them manage it from then on, diffing as needed should still work.  My current suggestion is to have a separate secrets file (as we already do), I'm not sure where @evrardjp sees this combining the secrets file.

Apsu:
RPC-AIO-Liberty-PR: Fail

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Liberty-PR: Fail

Apsu:
RPC-AIO-Liberty-PR: Fail

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Liberty-PR: Fail

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

prometheanfire:
Swift failed on this.

`fatal: [jrpcaioiad-9d7_glance_container-4df9ddb8] => error while evaluating conditional: (not is_metal | bool) and (glance_default_store == "file") and (glance_nfs_client is not defined)`

ceph failed on this.

`19:57:29     tempest.lib.exceptions.SSHTimeout: Connection to the 10.1.13.9 via SSH timed out.`

so I think a recheck is in order, at least for ceph, might as well do swift as well

recheck_all

Apsu:
RPC-AIO-Liberty-PR: Fail

prometheanfire:
same error on swift, it looks like that value has to be defined and isn't being defined

`22:23:38 fatal: [jrpcaioiad-d8_glance_container-5f25a0b6] => error while evaluating conditional: (not is_metal | bool) and (glance_default_store == "file") and (glance_nfs_client is not defined)`

I'm going to try defining the variable at all times to see if that helps, prefer ceph, then swift, then file.

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

prometheanfire:
same ceph fail though

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Liberty-PR: Fail

Apsu:
RPC-AIO-Liberty-PR: Fail

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Liberty-PR: Fail

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

prometheanfire:
looks like they both have the same fail now, so, progress

evrardjp:
Here are my comments:

First, it looks far better, and it's easier to follow.

Second, should we create the following dummy files ?

    user_osa_variables_overrides.yml
    user_osa_secrets_overrides.yml
    user_rpco_variables_overrides.yml
    user_rpco_secrets_overrides.yml    
We could just add --- and the license text in them.

Then, I'm confused by commit message (the code seems more clear to me): "Combine ```rpcd/etc/openstack_deploy/user_extras_variables.yml``` and ```rpcd/etc/openstack_deploy/user_variables.yml``` into a new file named ```rpcd/etc/openstack_deploy/user_rpco_variables_defaults.yml```.". We are effectively uncombining them, every different part of the product is now segregated.

Finally, we are not forced to remove the ```/etc/openstack_deploy/user_variables.yml```, we could rename it: ```user_osa_variables_aiodefaults.yml```.
This makes possible to **not** have to fully follow what's happening in osa's bootstrap aio process and maintain our list of variables seggregated as possible. Being only relevant for AIOs, I don't see the problem. If we want to improve the process later, and make OSA boostrap-aio have overrides and custom file names (matching what we have decided), it's still possible. That's something I could add in my todo list.

Do you want me to adapt your patchset?

prometheanfire:
The commit message needs work, hasn't been changed since a week ago.

We cannot keep OSA's user_variables renamed as `user_osa_variables_aiodefaults.yml` AND move RPC-O's OSA overrides to the same location.

prometheanfire:
Oh, and as far as keeping empty files around I'd like to avoid that, those are support's file's let them manage it.  Any time we touch them we violate that boundary we are trying to set up.

evrardjp:
You're right on the empty file.
The aio defaults was just a way to ensure we have all the variables into the file, (so keep the same process as it is, just mv to aiodefaults instead of rm-ing the file).

prometheanfire:
Oh, a new file name for aio defaults?  Only for use in the AIO?  That could work.  I did think of that, but instead changed the seds to be echos so we didn't need the file.  Since the file is fully commented out I'm not sure why we'd need it.

rpcd/etc/openstack_deploy/user_osa_variables_defaults.yml
```diff
@@ -116,3 +116,11 @@ horizon_custom_uploads:
   favicon:
     src: "{{ rackspace_static_files_folder }}/favicon.ico"
     dest: img/favicon.ico
+
```
git-harry: Why has this security hardening section been added?

prometheanfire: We were previously seding it in the variables file from OSA, however we are not copying that over as it is entirely commented out and we can't copy OSA's variables and RPC-O's OSA variables to the same location, so I copied this portion.  The other portions were easier to change by changing the sed's to echo, but this had extra logic and I wanted to keep the comments around it in the variables file.

Apsu:
RPC-AIO-Liberty-PR: Fail

Apsu:
RPC-AIO-Liberty-PR: Fail

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

rpcd/etc/openstack_deploy/user_rpco_secrets_defaults.yml
```diff
@@ -1,5 +1,5 @@
 ---
-# Copyright 2014, Rackspace US, Inc.
+# Copyright 2016, Rackspace US, Inc.
```
git-harry: Why is the date being changed?

scripts/deploy.sh
```diff
@@ -56,12 +56,15 @@ if [[ "${DEPLOY_AIO}" == "yes" ]]; then
   export DEPLOY_HAPROXY="yes"
   if [[ ! -d /etc/openstack_deploy/ ]]; then
     ./scripts/bootstrap-aio.sh
+    # remove unused variables file
+    rm /etc/openstack_deploy/user_variables.yml
```
git-harry: Removing this file means that the OSA AIO defaults, currently being used by RPCO AIO's, are no longer set. The vars are defined in the OSA file 'tests/roles/bootstrap-host/templates/user_variables.aio.yml.j2'. How have you taken this into account?

scripts/deploy.sh
```diff
@@ -112,24 +120,23 @@ fi
...
-    echo "apply_security_hardening: false" >> /etc/openstack_deploy/user_variables.yml
+    echo "apply_security_hardening: false" >> /etc/openstack_deploy/user_osa_variables_defaults.yml
   fi
 fi
 
 # ensure all needed passwords and tokens are generated
-./scripts/pw-token-gen.py --file /etc/openstack_deploy/user_secrets.yml
+./scripts/pw-token-gen.py --file /etc/openstack_deploy/user_osa_secrets_defaults.yml
 ./scripts/pw-token-gen.py --file $RPCD_SECRETS
+# users need to make sure to move these files to the override location so they are not overwriten on upgrade/redeploy
```
git-harry: What would overwrite them?

scripts/deploy.sh
```diff
@@ -112,24 +120,23 @@ fi
...
   else
-    echo "apply_security_hardening: false" >> /etc/openstack_deploy/user_variables.yml
+    echo "apply_security_hardening: false" >> /etc/openstack_deploy/user_osa_variables_defaults.yml
   fi
 fi
 
 # ensure all needed passwords and tokens are generated
-./scripts/pw-token-gen.py --file /etc/openstack_deploy/user_secrets.yml
+./scripts/pw-token-gen.py --file /etc/openstack_deploy/user_osa_secrets_defaults.yml
 ./scripts/pw-token-gen.py --file $RPCD_SECRETS
```
git-harry: What about user_osa_secrets_overides and user_rpco_secrets_overides?

git-harry:
I think you should create all the overrides files as part of the setup. That will prevent the files from being created with the wrong names and make it easier for support because they are already there. There may also be a benefit to adding a description at the top of all the variables files to explain their purpose.

Have you considered making the defaults files read-only to reduce the chances that they are mistakenly modified?

rpcd/etc/openstack_deploy/user_rpco_secrets_defaults.yml
prometheanfire: Earlier people (I think jp) were telling me to update it.

scripts/deploy.sh
prometheanfire: That would explain the fails, I'll probably need to still install it to a new location and change what points to it.

scripts/deploy.sh
prometheanfire: line 62 for instance (but that's only an AIO)

scripts/deploy.sh
prometheanfire: we don't touch overrides, they are not ours to touch

prometheanfire:
Making it readonly is a good idea.  Adding a description to the files makes sense too.  I don't like laying the overrides files down myself but since it seems to be a much desired thing I will.

scripts/deploy.sh
prometheanfire: That file is laid down no mater what in the aio, I can still remove the original.  So this isn't actually a problem.

```
- name: Set the user_variables
  config_template:
    src: user_variables.aio.yml.j2
    dest: /etc/openstack_deploy/user_variables.yml
    config_overrides: "{{ user_variables_overrides | default({}) }}"
    config_type: yaml
```

prometheanfire: though I remove it after it's laid down, so still need to move it

scripts/deploy.sh
git-harry: pw-token-gen.py is used to set a password for a variable defined in a file. If this script isn't run against the overrides files it doesn't help anyone because now support just have to remember to do it manually.

Apsu:
RPC-AIO-Liberty-PR: Fail

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Liberty-PR: Fail

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Liberty-PR: Fail

prometheanfire:
Can't have empty variables files

```
13:38:54 + openstack-ansible --forks 4 haproxy-install.yml
13:38:54 Variable files: "-e @/etc/openstack_deploy/user_conf_files.yml -e @/etc/openstack_deploy/user_extras_variables.yml -e @/etc/openstack_deploy/user_osa_aio_variables.yml -e @/etc/openstack_deploy/user_osa_secrets_defaults.yml -e @/etc/openstack_deploy/user_osa_secrets_overrides.yml -e @/etc/openstack_deploy/user_osa_variables_defaults.yml -e @/etc/openstack_deploy/user_osa_variables_overrides.yml -e @/etc/openstack_deploy/user_rpco_secrets_defaults.yml -e @/etc/openstack_deploy/user_rpco_secrets_overrides.yml -e @/etc/openstack_deploy/user_rpco_variables_defaults.yml -e @/etc/openstack_deploy/user_rpco_variables_overrides.yml "
13:38:55 ERROR: failed to combine variables, expected dicts but got a 'dict' and a 'NoneType'
```

git-harry:
That's disappointing

prometheanfire:
we could add a noop variable, but that's it

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

prometheanfire:
Not totally sure this is the cause, but could be helped by https://github.com/rcbops/jenkins-rpc/pull/604

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

prometheanfire:
@git-harry added the noop variables for you.

ceph still failing on something...

http://jenkins.propter.net/job/RPC-AIO/8191/consoleFull

As close as I can tell it's this.

```
17:57:40 Captured traceback:
17:57:40 ~~~~~~~~~~~~~~~~~~~
17:57:40     Traceback (most recent call last):
17:57:40       File "/opt/tempest_c2c90574a78c7704f06e8fac279c4f1d11365993/tempest/api/volume/admin/test_volumes_backup.py", line 57, in test_volume_backup_create_get_detailed_list_restore_delete
17:57:40         'available')
17:57:40       File "/opt/tempest_c2c90574a78c7704f06e8fac279c4f1d11365993/tempest/services/volume/base/base_backups_client.py", line 95, in wait_for_backup_status
17:57:40         raise exceptions.VolumeBackupException(backup_id=backup_id)
17:57:40     tempest.exceptions.VolumeBackupException: Volume backup c2c4c923-0847-4a1d-ab19-73ffd33d2a90 failed and is in ERROR status

atm, I'm waiting for the gating change to make it in...
```

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

rpcd/etc/openstack_deploy/user_osa_variables_defaults.yml
```diff
@@ -13,6 +13,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+## Warning
+#
+# This file is controled by the devs and will be overwritten on upgrade
```
git-harry: controlled

rpcd/etc/openstack_deploy/user_rpco_secrets_defaults.yml
```diff
@@ -13,6 +13,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+## Warning
+#
+# This file is controled by the devs and can be overwritten on upgrade
```
git-harry: controled

scripts/deploy.sh
```diff
@@ -21,8 +21,31 @@ export BOOTSTRAP_OPTS=${BOOTSTRAP_OPTS:-""}
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+## Using this file
+#
+# This file is laid down only once and is used for overriding variables based on the file name.
+# This file will not be overwritten by upgrades.
+
+noopvar: foobar'
```
git-harry: If everyone's happy with the idea of a dummy var. It might be worth adding a comment explaining its purpose, possibly linking to an upstream bug (assuming OSA accept that it shouldn't fail if an empty file is provided).

scripts/deploy.sh
```diff
@@ -112,29 +143,42 @@ fi
...
-./scripts/pw-token-gen.py --file /etc/openstack_deploy/user_secrets.yml
+./scripts/pw-token-gen.py --file /etc/openstack_deploy/user_osa_secrets_defaults.yml
 ./scripts/pw-token-gen.py --file $RPCD_SECRETS
 
 # Apply any patched files.
 cd ${RPCD_DIR}/playbooks
 openstack-ansible -i "localhost," patcher.yml
 
+# set permissions and lay down overrides files
+chown 0440 /etc/openstack_deploy/user_*_defaults.yml
```
git-harry: The secrets files can't be read-only unless their permissions are changed before and after pw-token-gen is run.

rpcd/etc/openstack_deploy/user_rpco_secrets_defaults.yml
prometheanfire: first it was controlled then it was controled? :P

scripts/deploy.sh
prometheanfire: ya, I'll add a comment

scripts/deploy.sh
prometheanfire: I don't know if they want us touching their file at all.

git-harry: My feeling is that it is fine, as long as we don't modify the var names listed there. @sigmavirus24 @evrardjp what are your thoughts?

prometheanfire:
@BjoernT we want to lay down the overrides files so that support knows which file names to use, for the secret versions of those overrides files is it ok if we populate secrests or should we leave it alone after we lay it down (as they are support's files)?

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

BjoernT:
@prometheanfire I'm not sure why we should overwrite secrets in the first place. If tokens are missing we can create them one time but then never touch them. I do not fully understand your question...

prometheanfire:
@BjoernT the way the password generator function works is to only provide passwords for unset variables in the secrets files, so we could do that.  Would you be fine with us doing that in an automated way each upgrade?  We wouldn't add variables to the override files ourselves, just provide passwords for unset variables support would add.

prometheanfire:
recheck_ceph

this is the problem, `backup_driver = cinder.backup.drivers.swift` for some reason, think we need another override in gate pinging @git-harry @hughsaunders 

I've added backup_driver =  cinder.backup.drivers.ceph to the overrides in the ceph mitaka gate

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

prometheanfire:
still fail, but at least it doesn't think it's swift now

prometheanfire:
@evrardjp mind checking http://jenkins.propter.net/job/RPC-AIO/8290/consoleFull everything 'should' be working...

prometheanfire:
wait a tic, wat

    23:17:59             Body: {"backup": {"status": "error", "object_count": 0, "container": null, "name": "tempest-Backup-1777915259", "links": [{"href": "http://172.29.236.100:8776/v1/126a7878ff5b4bcaba55aeb69b7a1991/backups/888b4e09-8ff0-43b9-a2ac-be2867970b8f", "rel": "self"}, {"href": "http://172.29.236.100:8776/126a7878ff5b4bcaba55aeb69b7a1991/backups/888b4e09-8ff0-43b9-a2ac-be2867970b8f", "rel": "bookmark"}], "availability_zone": "nova", "created_at": "2016-06-30T23:12:54.000000", "description": null, "updated_at": "2016-06-30T23:12:55.000000", "data_timestamp": "2016-06-30T23:12:54.000000", "has_dependent_backups": false, "snapshot_id": null, "volume_id": "200d0b0c-1c8f-4975-855b-e65eb78563c9", "fail_reason": "Backup driver reported an error: Could not determine which Swift endpoint to use. This can either be set in the service catalog or with the cinder.conf config option 'backup_swift_url'.", "is_incremental": false, "id": "888b4e09-8ff0-43b9-a2ac-be2867970b8f", "size": 1}}

prometheanfire:
running locally, more wat

```
2016-07-01 03:19:29.743 18953 ERROR oslo_messaging.rpc.dispatcher [req-3b48ecfd-e2de-4395-9393-351712560290 8dc6e704ac4348408201207f186dc3ed edbd25f6f7b44285a8bef216104ba924 - - -] Exception during message handling: error creating image
2016-07-01 03:19:29.743 18953 ERROR oslo_messaging.rpc.dispatcher Traceback (most recent call last):
2016-07-01 03:19:29.743 18953 ERROR oslo_messaging.rpc.dispatcher   File "/openstack/venvs/cinder-13.1.4/lib/python2.7/site-packages/oslo_messaging/rpc/dispatcher.py", line 138, in _dispatch_and_reply
2016-07-01 03:19:29.743 18953 ERROR oslo_messaging.rpc.dispatcher     incoming.message))
2016-07-01 03:19:29.743 18953 ERROR oslo_messaging.rpc.dispatcher   File "/openstack/venvs/cinder-13.1.4/lib/python2.7/site-packages/oslo_messaging/rpc/dispatcher.py", line 185, in _dispatch
2016-07-01 03:19:29.743 18953 ERROR oslo_messaging.rpc.dispatcher     return self._do_dispatch(endpoint, method, ctxt, args)
2016-07-01 03:19:29.743 18953 ERROR oslo_messaging.rpc.dispatcher   File "/openstack/venvs/cinder-13.1.4/lib/python2.7/site-packages/oslo_messaging/rpc/dispatcher.py", line 127, in _do_dispatch
2016-07-01 03:19:29.743 18953 ERROR oslo_messaging.rpc.dispatcher     result = func(ctxt, **new_args)
2016-07-01 03:19:29.743 18953 ERROR oslo_messaging.rpc.dispatcher   File "/openstack/venvs/cinder-13.1.4/lib/python2.7/site-packages/cinder/backup/manager.py", line 311, in create_backup
2016-07-01 03:19:29.743 18953 ERROR oslo_messaging.rpc.dispatcher     self._update_backup_error(backup, context, six.text_type(err))
2016-07-01 03:19:29.743 18953 ERROR oslo_messaging.rpc.dispatcher   File "/openstack/venvs/cinder-13.1.4/lib/python2.7/site-packages/oslo_utils/excutils.py", line 220, in __exit__
2016-07-01 03:19:29.743 18953 ERROR oslo_messaging.rpc.dispatcher     self.force_reraise()
2016-07-01 03:19:29.743 18953 ERROR oslo_messaging.rpc.dispatcher   File "/openstack/venvs/cinder-13.1.4/lib/python2.7/site-packages/oslo_utils/excutils.py", line 196, in force_reraise
2016-07-01 03:19:29.743 18953 ERROR oslo_messaging.rpc.dispatcher     six.reraise(self.type_, self.value, self.tb)
2016-07-01 03:19:29.743 18953 ERROR oslo_messaging.rpc.dispatcher   File "/openstack/venvs/cinder-13.1.4/lib/python2.7/site-packages/cinder/backup/manager.py", line 305, in create_backup
2016-07-01 03:19:29.743 18953 ERROR oslo_messaging.rpc.dispatcher     self._run_backup(context, backup, volume)
2016-07-01 03:19:29.743 18953 ERROR oslo_messaging.rpc.dispatcher   File "/openstack/venvs/cinder-13.1.4/lib/python2.7/site-packages/cinder/backup/manager.py", line 353, in _run_backup
2016-07-01 03:19:29.743 18953 ERROR oslo_messaging.rpc.dispatcher     backup_service.backup(backup, device_path)
2016-07-01 03:19:29.743 18953 ERROR oslo_messaging.rpc.dispatcher   File "/openstack/venvs/cinder-13.1.4/lib/python2.7/site-packages/cinder/backup/drivers/ceph.py", line 872, in backup
2016-07-01 03:19:29.743 18953 ERROR oslo_messaging.rpc.dispatcher     volume_name, length)
2016-07-01 03:19:29.743 18953 ERROR oslo_messaging.rpc.dispatcher   File "/openstack/venvs/cinder-13.1.4/lib/python2.7/site-packages/cinder/backup/drivers/ceph.py", line 708, in _full_backup
2016-07-01 03:19:29.743 18953 ERROR oslo_messaging.rpc.dispatcher     stripe_count=self.rbd_stripe_count)
2016-07-01 03:19:29.743 18953 ERROR oslo_messaging.rpc.dispatcher   File "/openstack/venvs/cinder-13.1.4/lib/python2.7/site-packages/rbd.py", line 222, in create
2016-07-01 03:19:29.743 18953 ERROR oslo_messaging.rpc.dispatcher     raise make_ex(ret, 'error creating image')
2016-07-01 03:19:29.743 18953 ERROR oslo_messaging.rpc.dispatcher PermissionError: error creating image
```

evrardjp:
recheck_ceph
Rechecking, the latest gating issue seems not only linked to this PR:
screen shot 2016-07-01 at 9 36 22 am

If it continues that way I'll add an issue into this repo.
In the meantime, I'll spin up a new AIO with ceph to see what's going on.

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

git-harry:
Speaking with @evrardjp and @odyssey4me it seems the OSA secrets file should already contain the keys for all the possible secret vars OSA is configured to use. Assuming RPCO is the same that means that there should never be a use for a secrets overrides file. This suggests the following would be more appropriate:

```
user_osa_secrets_defaults.yml - this is openstack-ansible/etc/openstack_deploy/user_secrets.yml
user_osa_variables_defaults.yml - this is rpcd/etc/openstack_deploy/user_variables.yml
user_osa_variables_overrides.yml - this is for support to set any deployment-specific overrides
user_rpco_secrets_defaults.yml - this is rpcd/etc/openstack_deploy/user_extras_secrets.yml
user_rpco_variables_defaults.yml - this is rpcd/etc/openstack_deploy/user_extras_variables.yml
user_rpco_variables_overrides.yml- this is for support to set any deployment-specific overrides
```
The secrets files would effectively be completely managed by deploy.sh - it would copy them across and run pw-token-gen.py on them.

git-harry:
recheck_ceph - the Jenkins config looked incorrect, I have modified it.

evrardjp:
@git-harry @prometheanfire does support want to override these files? If yes, there is still a use case for the overrides. Cf https://github.com/rcbops/rpc-openstack/pull/1165#issuecomment-228758281 . Else I'd say remove them. For now it seems working even with ceph on my AIO.

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

evrardjp:
recheck_ceph

git-harry:
@evrardjp there is an issue with jenkins, I'm currently looking into it. Until that's fixed rechecks won't help

evrardjp:
I thought you fixed it.

git-harry:
@evrardjp there is another issue.

prometheanfire:
I think support still does want to override passwords, I'll try to double confirm today.

git-harry:
@prometheanfire what do you mean by override passwords?

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

git-harry:
recheck_ceph

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

prometheanfire:
recheck_ceph

I think it's unrelated http://jenkins.propter.net/job/RPC-AIO/8331/console

```
16:50:01 Captured traceback:
16:50:01 ~~~~~~~~~~~~~~~~~~~
16:50:01     Traceback (most recent call last):
16:50:01       File "/opt/tempest_c2c90574a78c7704f06e8fac279c4f1d11365993/tempest/scenario/manager.py", line 108, in delete_wrapper
16:50:01         delete_thing(*args, **kwargs)
16:50:01       File "/opt/tempest_c2c90574a78c7704f06e8fac279c4f1d11365993/tempest/lib/services/compute/floating_ips_client.py", line 62, in delete_floating_ip
16:50:01         resp, body = self.delete(url)
16:50:01       File "/opt/tempest_c2c90574a78c7704f06e8fac279c4f1d11365993/tempest/lib/common/rest_client.py", line 290, in delete
16:50:01         return self.request('DELETE', url, extra_headers, headers, body)
16:50:01       File "/opt/tempest_c2c90574a78c7704f06e8fac279c4f1d11365993/tempest/lib/services/compute/base_compute_client.py", line 53, in request
16:50:01         method, url, extra_headers, headers, body)
16:50:01       File "/opt/tempest_c2c90574a78c7704f06e8fac279c4f1d11365993/tempest/lib/common/rest_client.py", line 645, in request
16:50:01         resp, resp_body)
16:50:01       File "/opt/tempest_c2c90574a78c7704f06e8fac279c4f1d11365993/tempest/lib/common/rest_client.py", line 764, in _error_checker
16:50:01         message=message)
16:50:01     tempest.lib.exceptions.ServerFault: Got server fault
16:50:01     Details: Unexpected API Error. Please report this at http://bugs.launchpad.net/nova/ and attach the Nova API log if possible.
16:50:01     <class 'neutronclient.common.exceptions.PortNotFoundClient'>
```

sigmavirus24:
:+1:

README.md
```diff
@@ -70,13 +70,14 @@ alarm configured for it.
...
+  2. move OSA variables to the correct locations.
 
      ```
-     scripts/update-yaml.py /etc/openstack_deploy/user_variables.yml rpcd/etc/openstack_deploy/user_variables.yml
+     rm /etc/openstack_deploy/user_variables.yml  # unused, nothing set
+     mv /etc/openstack_deploy/user_secrets.yml /etc/openstack_deploy/user_osa_secrets_defaults.yml
      ```
   3. copy the RPC configuration files:
-     1. `cp rpcd/etc/openstack_deploy/user_extras_*.yml /etc/openstack_deploy`
+     1. `cp rpcd/etc/openstack_deploy/user_*_defaults.yml /etc/openstack_deploy`
```
evrardjp: For a fresh (greenfield) installation, also copy the standard overrides files:
`cp rpcd/etc/openstack_deploy/user_*_overrides.yml /etc/openstack_deploy`

README.md
```diff
@@ -39,10 +39,10 @@ as well as adding a Rackspace tab and a Solutions tab, which provides
...
 * `kibana.yml` - Setup Kibana on the Kibana hosts for the logging dashboard.
 * `logstash.yml` - deploys a logstash host. If this play is used, be sure to
-uncomment the related block in user_extra_variables.yml before this play is
-run and then rerun the appropriate plays in openstack-ansible after this
-play to ensure that rsyslog ships logs to logstash. See steps 11 - 13 below
-for more.
+copy the related block in user_rpco_variables_defaults.yml to
+user_rpco_variables_overrides before this play is run and then rerun the
+appropriate plays in openstack-ansible after this play to ensure that rsyslog
+ships logs to logstash. See steps 11 - 13 below for more.
```
evrardjp: Could we change steps 11 - 13 by a name and what these steps do? This way we don't have a reference issue if we shuffle the doc steps.

releasenotes/notes/master-96b4f6d34b87fe3d.yaml
```diff
@@ -0,0 +1,22 @@
+---
+features:
+  - Variables and secrets are now managed in a manner that
+    allows downstream deployers to not have to worry about
+    their settings being overwritten.  They are provided
+    with overrides files that they alone manage after we
+    lay them down.  We have also taken this time to rename
+    our user_* files to separate out what specifically the
+    variables override, whether it be upstream OSA changes
```
evrardjp: "splitting the OSA relevant changed of the local RPC-O changes" instead of "OSA changes" (because it's not an osa change, it's an RPC change on OSA scope)

releasenotes/notes/master-96b4f6d34b87fe3d.yaml
```diff
@@ -0,0 +1,22 @@
...
+    or local RPCO changes.
+deprecations:
+  - We are depricating the use of user_variables and
```
evrardjp: nit: typo.
deprecating

scripts/deploy.sh
```diff
@@ -109,32 +115,48 @@ if [[ "${DEPLOY_AIO}" == "yes" ]]; then
...
 ./scripts/pw-token-gen.py --file $RPCD_SECRETS
 
 # Apply any patched files.
 cd ${RPCD_DIR}/playbooks
 openstack-ansible -i "localhost," patcher.yml
 
+# set permissions and lay down overrides files
+chmod 0440 /etc/openstack_deploy/user_*_defaults.yml
+if [[ ! -f /etc/openstack_deploy/user_osa_secrets_overrides.yml ]]; then
+  mv "${RPCD_DIR}"/etc/openstack_deploy/user_osa_secrets_overrides.yml /etc/openstack_deploy/user_osa_secrets_overrides.yml
```
evrardjp: I can't remember the reason we defined a mv instead of a copy?

evrardjp:
See my comments in-line.
Not giving a +1 yet, but the code doesn't deserve a -1 either.

scripts/deploy.sh
```diff
@@ -90,9 +91,14 @@ if [[ "${DEPLOY_AIO}" == "yes" ]]; then
       echo "osd_directory: true" | tee -a $RPCD_VARS
       echo "osd_directories:" | tee -a $RPCD_VARS
       echo "  - /var/lib/ceph/osd/mydir1" | tee -a $RPCD_VARS
-      sed -i "s/glance_default_store:.*/glance_default_store: rbd/" /etc/openstack_deploy/user_variables.yml
-      echo "nova_libvirt_images_rbd_pool: vms" | tee -a /etc/openstack_deploy/user_variables.yml
-      echo "cinder_ceph_client_uuid:"  | tee -a /etc/openstack_deploy/user_secrets.yml
+      echo "glance_default_store: rbd" | tee -a /etc/openstack_deploy/user_osa_variables_defaults.yml
```
git-harry: I would have thought it would make more sense to send these (lines 94 to 100) to the overrides files instead of modifying the defaults files.

scripts/deploy.sh
```diff
@@ -56,12 +56,13 @@ if [[ "${DEPLOY_AIO}" == "yes" ]]; then
   export DEPLOY_HAPROXY="yes"
   if [[ ! -d /etc/openstack_deploy/ ]]; then
     ./scripts/bootstrap-aio.sh
+    # move OSA variables file to AIO location.
```
git-harry: A detailed comment would be helpful here, explaining why there is a need for an additional file and regarding the required ordering.

git-harry:
@prometheanfire can you give me an example of how user_osa_secrets_overrides.yml would be used.

README.md
prometheanfire: The reason that wasn't included is because it is laid down by the deploy.sh script.  Should I remove it from there and add a manual instruction here?

README.md
prometheanfire: I have no idea what that was referencing, I don't know exactly what to call it, but I did change it to reference 2-4 and 3-2 instead (which I think is right).

releasenotes/notes/master-96b4f6d34b87fe3d.yaml
prometheanfire: ok

releasenotes/notes/master-96b4f6d34b87fe3d.yaml
prometheanfire: fixed

scripts/deploy.sh
prometheanfire: no reason I can think of, will change to cp

scripts/deploy.sh
prometheanfire: we are not support, we don't get to put anything in the overrides file

prometheanfire:
@git-harry ask @evrardjp about the secrets overrides, I'm not sure why.

scripts/deploy.sh
prometheanfire: That will be part of the official documentation, leaving it in a comment is the wrong place for docs.

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

prometheanfire:
recheck_ceph

scripts/deploy.sh
git-harry: This is for an AIO, support do not deploy customers using the AIO code. This code is essentially providing the function of support by determining a set of vars that should be set for the deployment, because it's an AIO and because it's always the same it can be automated.

scripts/deploy.sh
git-harry: What do you mean by official documentation? If you're referring to the documentation managed by the docs team, I am not aware of them ever documenting anything related to AIOs.

The point about having a better comment here is because the name of this file is important and if it's changed it has the potential the change the precedence. It breaks the new format of having defaults and overrides files and it isn't necessarily obvious why it's required. The idea of the comment is to prevent someone changing this section because they don't fully appreciate the context.

scripts/deploy.sh
prometheanfire: Then we should have another file for that, like we do for gating.  user_zzz_foobarbaz.yml

scripts/deploy.sh
prometheanfire: AIO location was incorrect, maybe 'rpc-o managed location' would be better?

scripts/deploy.sh
git-harry: I disagree, we have too many files as it is. I can just about except the aio and gating files because the allow it was the easiest way to keep the existing variable precedence without having the merge variables together but here it makes no sense to me. What is the point in creating an additional file location. This is for an AIO. These values are used to override the defaults. The contract with support is not to modify the overrides files on customer deployments. The overrides file is exactly the place these changes should go. If I'm working with an AIO I should have a set up that is as close to a production deployment as possible.

prometheanfire: down below you criticize me about this file being used as an example that's looked to by support, we should treat it as if it is a deployment, meaning we shouldn't put it in overrides.

hughsaunders:
👍  without the secrets overrides this makes more sense.

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

evrardjp:
Seems good at first sight.

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

prometheanfire:
recheck_ceph

```
Instance ce597881-0117-4af1-97b8-215dc480c912 could not be found.
```

prometheanfire:
so, I need another :+1: for this to merge

cloader89:
Looks good to me! Tests passed.

cloader89:
👍

sigmavirus24:
YOLO MERGE!

PR #1169, opened 2016-06-16
Make holland use 127.0.0.1 for backups

### LINKED ISSUE #348: Holland backup needs to backup localhost
Note: This is the Kilo+ location for this bug report. For Juno, see: https://bugs.launchpad.net/openstack-ansible/+bug/1479486

The current template for roles/rpc_support_holland/templates/holland-xtrabackup.conf.j2 has the host set to {{ internal_vip_address }}. This can cause issue if it is routed to a remote node from where holland is ran. The host should be set to localhost so we do not backup a remote node.

i.e. set host = 127.0.0.1 in /etc/holland/backupsets/rpc_support.conf

- [ ] Reproduce the issue. (N/A)
- [ ] Create a PR for fix and/or docs.
- [ ] Incorporate feedback from pull request reviews.
- [ ]  What is the gating for this on DEV and/or QE? Add related gating if appropriate. Also make sure current gating passes with this fix.
- [ ] Reach out to reporter to determine if the fix is resolving the issue.
- [ ] What are the upgrade impacts (if any).
- [ ] What are the backports are needed (if any).

mancdaz:
Currently holland runs in a galera container and connects to the load
balancer VIP. This commit ensures that it backs up from the local mysql
instance meaning backups can still happen even if the load balancer is
unavailable.

Resolves #348

### DIFF
--- a/releasenotes/notes/xtrabackup-localhost-8d1562d924ffd6e5.yaml
+++ b/releasenotes/notes/xtrabackup-localhost-8d1562d924ffd6e5.yaml
@@ -1,0 +1,3 @@
+---
+other:
+  - The holland configuration for xtrabackup (the tool for performing mariadb database backups) is now set to perform backups against localhost in each container, as opposed to the load balanced IP address. This ensures that each of the containers is actually backing up it's own copy of the databases, which can be helpful if there are inconsistencies discovered between cluster members. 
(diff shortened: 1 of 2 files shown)


git-harry:
The change looks fine but I think a release note is appropriate.

Apsu:
RPC-AIO-Liberty-PR: Fail

Apsu:
RPC-AIO-Liberty-PR: Fail

evrardjp:
recheck_swift -- tempest neutron issue

git-harry:
recheck_all - retesting now holland is tested by jenkins

git-harry:
👍 assuming tests pass

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

mancdaz:
recheck_ceph

prometheanfire:
:+1: ezpz, good notes

prometheanfire:
why did my :+1: retrigger tests?

Issue #1167, opened 2016-06-16
Current AIO configurations drop instance traffic to heat metadata url


alextricity25:
##### What version (commit) do you encounter this issue on?
master
##### What type of issue is this (bug, feature, documentation, regression)?
bug
##### What is the expected behavior?
Instances should be able to communicate to the heat metadata url.

##### What is actually happening?
Spinning up a heat template that utilizes the software-deployment resource will fail. This is because the resource requires the instance to communicate to the heat metadata url, but can't due to incorrect packet checksums that appear on an AIO. It is essentially the same problem as:
https://github.com/rcbops/rpc-openstack/blob/master/scripts/deploy.sh#L179-L183

##### What were you attempting to do when you encountered this error (give as much detail as possible)? 
I was doing some testing by creating basic heat stack that spins up a server and uses software-deployments/configs to run a trivial script.

##### How can we reproduce this issue?
Use the following heat template to create a heat stack on an AIO. Note a private network name 'test-network' must be created.
```
heat_template_version: 2015-04-30

description: Simple template to deploy a single compute instance

parameters:
  keyname:
    type: string
    label: Salt-Master Keyname.
    description: The name of the key that should be used to ssh into the Salt-Master.

  image:
    type: string
    label: Image Name
    description: Name of the existing image to use for every vm spun up by this stack. The image should have heat-config and heat-config-script baked in.

  flavor:
    type: string
    label: Flavor
    description: Flavor of all vms. This may be changed in the future to allow different flavors for the different vms in this stack.
    default: m1.small

resources:
  config-salt-master:
    type: OS::Heat::SoftwareConfig
    properties:
      group: script
      outputs:
        - name: result
          type: Json
      config: |
        #!/bin/bash
        # Install salt master
        set -ex
        logdir=/var/log/heat-deployments
        prefix=${logdir}/test-stack-log
        mkdir -p $logdir
        exec &> >(tee -a ${prefix}.log)
        [ -e ${prefix}.ran ] && exit 0

        echo "This is a test heat template"

  # Software deployment for salt-master. Deploys the salt-master softrware
  # configuration.
  # Also lists the git repos necessary to make the drupal stack work.
  deploy-salt-master:
    type: OS::Heat::SoftwareDeployment
    properties:
      config:
        get_resource: config-salt-master
      server:
        get_resource: master

  master:
    type: OS::Nova::Server
    properties:
      key_name:
        get_param: keyname
      image:
        get_param: image
      flavor:
[...]

PR #1170, opened 2016-06-16
Adds iptables checksum fix for port 8000

alextricity25:
Port 8000 is the heat metadata url that some heat
resources use for running heat-related hooks inside
an instance. Without this rule communication between
the instance and the service will fail.

Connects #1167


prometheanfire:
:+1: ezpz fix

cloader89:
LGTM 👍

Issue #1113, opened 2016-06-03
Plugin rabbitmq_status.py not properly excluding notifications queues on returned metrics.


hybridpollo:
Greetings,

I don't submit these very often,  it seems that this maybe the right place for this, otherwise then any guidance is always welcomed.   The maas plugin rabbitmq_status.py looks to not be properly excluding notification queues in the metrics reported.    Its possible this has to do with rabbitmq now using vhosts per each openstack component starting in 12.x.x.x implementations, but I will include as much  information as I can below.

##### What version (commit) do you encounter this issue on?

```
root@xxxxxx-infra01:/opt/rpc-openstack/rpcd# git describe --tags
r12.1.1
```

##### What type of issue is this (bug, feature, documentation, regression)?
Bug

##### What is the expected behavior?
rabbitmq_status.py properly excluding notification queues in the metrics.

##### What is actually happening?
MaaS alerts for rabbit are triggering with the following alarm, when all queues are at 0 except the notification.* queues:

REASON: RabbitMQ sum of queued messages excluding notifications queues is reaching configured limit. Currently above 100

##### How can we reproduce this issue?
Here is an example rabbitmq node with a growing notifications queue on the /ceilometer vhost.

```
root@xxxxxx-Infra01_rabbit_mq_container-db618bcd:/var/log/rabbitmq# for x in $(rabbitmqctl list_vhosts | awk '$1 !~ '/Listing'/ { print }') ; do echo "Backed up queues for vhost $x" ; rabbitmqctl list_queues -p $x | awk '$2 > 1 { print }' ; echo " " ; done
Backed up queues for vhost /
Listing queues ...
 
Backed up queues for vhost /ceilometer
Listing queues ...
notifications.sample	212
 
Backed up queues for vhost /cinder
Listing queues ...
 
Backed up queues for vhost /glance
Listing queues ..
 
Backed up queues for vhost /heat
Listing queues ..
 
Backed up queues for vhost /keystone
Listing queues ..
 
Backed up queues for vhost /neutron
Listing queues ..
 
Backed up queues for vhost /nova
Listing queues ..
```

Metrics returned by the rabbitmq_status.py plugin:

```
root@xxxxxx-infra01:/usr/lib/rackspace-monitoring-agent/plugins# ./rabbitmq_status.py -H 172.29.238.65 -n xxxxxx-Infra01_rabbit_mq_container-db618bcd -U maas_user -p xxxxxxxxxxxxxx
status okay
[...]

PR #1175, opened 2016-06-20
Give maas rabbit user perms on all vhosts

mancdaz:
Previously the maas user only had permisions on the vhost "/" which is
no longer used by the openstack services, as they each have their own
vhost.

No modification to the plugin is necessary, as the same api calls will
return data from all the vhosts the user has access to.

Related: #1113
(cherry picked from commit 6427f195ce6d5fcb5dd712638a1e3b63f3322090)


hughsaunders:
👍

git-harry:
👍

git-harry:
recheck_swift

PR #1174, opened 2016-06-20
OSA SHA bump to Mitaka

mancdaz:
This is using the current head of stable/mitaka for OSA, and includes the minimum change necessary to make rpc-o master switch over to using OSA mitaka.

Connects https://github.com/rcbops/u-suk-dev/issues/160

PR #1179, opened 2016-06-21
Enable security hardening by default

major:
This patch enables security hardening by default for all deployments.
Deployers can opt-out of this change by setting the
``DEPLOY_HARDENING`` environment variable to ``no`` prior to running
``deploy.sh``.

Release notes are included.

Connects to https://github.com/rcbops/u-suk-dev/issues/159


git-harry:
👎 This setting is currently controlled by [deploy.sh][1] and so that should be the only thing you need to change.

[1]: https://github.com/rcbops/rpc-openstack/blob/master/scripts/deploy.sh#L16

major:
Darn -- I totally forgot that we made the change there. I'll get this fixed up.

major:
Just rebased onto latest master to test against Mitaka. This brings in #1174.

tomc603:
Currently testing with AIO

git-harry:
👍

evrardjp:
I trust @Apsu bot for testing this functionality.
I think this is fine, and makes sense to enable on mitaka. 👍

Issue #1180, opened 2016-06-21
Increase maas per plugin metric limit 30 > 50


mancdaz:
##### What version (commit) do you encounter this issue on?

All versions
##### What type of issue is this (bug, feature, documentation, regression)?

Feature
##### What is the requested Feature?

The maas team have updated the maas system to allow 50 metrics per agent plugin. We currently artificially limit to the old value of 30 metrics to ensure we didn't cause errors.  We should update maas_common.py to reflect the new 50 metric limit. New docs referencing the new limit:

https://developer.rackspace.com/docs/rackspace-monitoring/v1/developer-guide/#agent-plugin
##### What branches / milestones do you believe this issue should target?
- [x] Master
- [ ] liberty-12.2
- [ ] liberty-12.1
- [ ] kilo
##### Does your change require documentation?
- [x] No documentation required for this change

alextricity25:
This might be a good candidate for impacts-onboarding if we need to look for more.

PR #1181, opened 2016-06-21
increase maas plugin metric limit to 50

mancdaz:
The maas team have increased the global limit of metrics allowed per
maas agent plugin from 30 to 50. This commit increases the artificial
limit in maas_common to match.

reference: https://developer.rackspace.com/docs/rackspace-monitoring/v1/developer-guide/#id76

Connects #1180

### DIFF
--- a/maas/plugins/maas_common.py
+++ b/maas/plugins/maas_common.py
@@ -531,8 +531,8 @@ METRICS = []
 
 def metric(name, metric_type, value, unit=None):
     global METRICS
-    if len(METRICS) > 29:
-        status_err('Maximum of 30 metrics per check')
+    if len(METRICS) > 49:
+        status_err('Maximum of 50 metrics per check')
     metric_line = 'metric %s %s %s' % (name, metric_type, value)
     if unit is not None:
         metric_line = ' '.join((metric_line, unit))
--- a/releasenotes/notes/maas-metric-limit-increase-fc17cbec85a7952c.yaml
+++ b/releasenotes/notes/maas-metric-limit-increase-fc17cbec85a7952c.yaml
@@ -1,0 +1,5 @@
+---
+other:
+  - The number of metrics that maas plugins can emit has
+    been increased from 30 to 50, in line with new
+    limits in the maas API.


maas/plugins/maas_common.py
```diff
@@ -531,7 +531,7 @@ def status_ok(message=None, force_print=False):
 
 def metric(name, metric_type, value, unit=None):
     global METRICS
-    if len(METRICS) > 29:
+    if len(METRICS) > 49:
```
claco: I assume there is some historical math going on here? As in, the limit, minus the `status` line first?

tomc603:
👎 The math is incorrect, and the error message in the `if` clause needs to be updated.

`len(['a', 'b', 'c'])` is 3, not 2. This if statement seems to be off by one, unless there's a metric limit of 49 because one metric is already consumed somewhere. If that's the case, I don't see it documented anywhere in the source.

mancdaz:
@tomc603 @claco the METRICS list is added to each time a metric is gathered by the plugin. If there is attempt made to push another metric into that list when the list is already at capacity (>49 is considered at, or past capacity), then the attempt to push another metric on will fail. This isn't a test of 'is it at capacity', it's a test of 'will adding another metric push it past capacity'.

evrardjp:
For those looking at the value, it's more specifically there: 
https://developer.rackspace.com/docs/rackspace-monitoring/v1/developer-guide/#id76

maas/plugins/maas_common.py
```diff
@@ -531,7 +531,7 @@ def status_ok(message=None, force_print=False):
...
         status_err('Maximum of 30 metrics per check')
```
evrardjp: I'd replace 30 metrics by 50. And maybe use a variable.

evrardjp:
👎 see comment in-line.

git-harry:
I don't think it's a good idea to use a goo.gl link, it's not clear what it links to and there is no guarantee that it will be there in a year's time. The first line of a commit message should start with a capital letter.

Did you consider adding a release note? Not sure if one is required, this commit is not addressing something specific by itself, but maybe worth a mention?

On the topic of metric counting, for those that are confused by it, '> 49' is equivalent to '>= 50'

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

evrardjp:
👍 Concerns expressed by the group seemed to be solved.

git-harry:
👍

PR #1185, opened 2016-06-22
Make ansible playbook runs unbuffered

evrardjp:
This makes sure the content of the ansible runs are directly
sent to stdout, instead of waiting the end of the playbook.

The need appeared when debugging verify-maas playbook, which can
take a while, and didn't output anything (because jenkins job
timed out before the end of the run).

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -3,6 +3,7 @@
 set -e -u -x
 set -o pipefail
 
+export PYTHONUNBUFFERED=1
 export ADMIN_PASSWORD=${ADMIN_PASSWORD:-"secrete"}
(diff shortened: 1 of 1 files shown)


mancdaz:
👍

git-harry:
👎 this is addressed by https://github.com/rcbops/jenkins-rpc/pull/597/files in jenkins. For inclusion in RPCO we need an issue to start with.

prometheanfire:
If this is only for gating and we have it fixed in gating I think we don't need this.

:-1:

Apsu:
RPC-AIO-Liberty-PR: Fail

git-harry:
@evrardjp if this is something we should still look to add, create an issue so we can properly track it.

PR #1186, opened 2016-06-22
Accept cinder's HTTP 300 return code

evrardjp:
According to [1], we should accept 300 as a normal return code for
cinder checks. This commit adds this support.

[1]: http://developer.openstack.org/api-ref-blockstorage-v2.html

### DIFF
--- a/rpcd/playbooks/roles/rpc_maas/templates/lb_api_check_cinder.yaml.j2
+++ b/rpcd/playbooks/roles/rpc_maas/templates/lb_api_check_cinder.yaml.j2
@@ -17,6 +17,6 @@ alarms            :
         label               : lb_api_alarm_cinder
         criteria            : |
             :set consecutiveCount={{ maas_alarm_local_consecutive_count }}
-            if (metric['code'] != '200') {
+            if (metric['code'] != '200' && metric['code'] != '300') {
                 return new AlarmStatus(CRITICAL, 'API unavailable.');
             }


mancdaz:
I'd say this should be added to #1174 since this is new mitaka behaviour

git-harry:
👎
I'd like to understand why we now see 300. I think the alarms should be specific and so don't want us allowing multiple response codes unless there is a clear reason to do it. I appreciate that the API says that both are allowed but they will be generated under different circumstances and so the project should understand what those are and whether or not they are both applicable, whether it should be configurable or whether we should always use the same one. This seems like something that should be done after #1174 is merged.

I don't mind something that is required for #1174 being done in a separate patch but it needs to:
- address the issue and not be a work around
- properly show the relationship by referencing the appropriate issue
- as with any commit clearly explain why the change is now required
- it needs its own issue if it is suitable for backport to Liberty
- update any Mitaka cards that relate to this change so that the change can be reviewing in context of the whole feature

The purpose of #1174 is to contain all the changes needed for the checks to pass so that the Mitaka project is no longer serialised on that one commit. It is okay to include workarounds in #1174 as long as the appropriate feature card is updated to track that the change needs reviewing as part of the feature update. This change currently doesn't seem to address the why, and so I would prefer to roll it in to #1174 and document it needs further investigation.

evrardjp:
The workaround is now in #1174 .

This should be further investigated.

git-harry:
@evrardjp now this is in #1174 I'm closing this one.

PR #1196, opened 2016-06-24
Use rpc python package index

evrardjp:
This adds rpc specific package index on top of pypi.
The bump of OSA to mitaka changed the way to use PyPI, and the variable
we should use from now on is ```repo_build_pip_extra_indexes```.

Connects rcbops/u-suk-dev#171

### DIFF
--- a/releasenotes/notes/rpc_package_index-a03d4d131c16d710.yaml
+++ b/releasenotes/notes/rpc_package_index-a03d4d131c16d710.yaml
@@ -1,0 +1,5 @@
+---
+other:
+  - A new user variable (list type) named
+    ``repo_build_pip_extra_indexes`` was introduced
+    to fetch pip packages from rpc repository.
--- a/rpcd/etc/openstack_deploy/user_variables.yml
+++ b/rpcd/etc/openstack_deploy/user_variables.yml
@@ -116,3 +116,7 @@ horizon_custom_uploads:
   favicon:
     src: "{{ rackspace_static_files_folder }}/favicon.ico"
     dest: img/favicon.ico
+
+# Use RPC python package index
+repo_build_pip_extra_indexes:
+  - "https://rpc-repo.rackspace.com/pools"


sigmavirus24:
So I think @git-harry will murder me if I don't point out that "That's it.\nUsing both is better" is not much of an explanation of *why* this change is happening. Could you update your commit message to explain the purpose behind this?

evrardjp:
True. Updating.

Apsu:
RPC-AIO-Ceph-Liberty-PR: Pass

Apsu:
RPC-AIO-Liberty-PR: Pass

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

d34dh0r53:
recheck_ceph

sigmavirus24:
Should probably add a documentation note that we're doing this.

sigmavirus24:
Or at least we should add a release note (now that I'm thinking about it)

odyssey4me:
The default index for pip is pypi, so only the RPC extra index is required as this list gets passed to pip as an extra index and the default is left alone, thus already achieving the goal of using both.

evrardjp:
@sigmavirus24 I would be happy to know what you see for the documentation.

My understanding would be to add a "upgrade" release note, mentioning the need to use ```repo_build_pip_extra_indexes``` for deployers that want to configure their own index.

Is that what you meant?

git-harry:
In Liberty the playbook repo-build.yml sets the following:
```
repo_build_pip_default_index: "https://rpc-repo.rackspace.com/pools"
repo_build_pip_extra_index: "https://pypi.python.org/simple"
```

In Mitaka the playbook repo-build.yml no longer sets these vars and the role defaults are commented out.

Why set extra instead of default? Does it make any difference?

Upgrades are out of scope so, if there's any upgrade work, we need to log that on the appropriate upgrade card.

I'm not sure a release note is necessary, to all intents and purposes this var is internal.

evrardjp:
The indexes are not sorted, they are justed used, and the "best relevant" version of the package that needs to be installed will be fetched from these indexes.

rpcd/etc/openstack_deploy/user_variables.yml
```diff
@@ -116,3 +116,7 @@ horizon_custom_uploads:
   favicon:
     src: "{{ rackspace_static_files_folder }}/favicon.ico"
     dest: img/favicon.ico
+
+# Use RPC python package index
+repo_build_pip_extra_indexes:
+    - "https://rpc-repo.rackspace.com/pools"
```
sigmavirus24: Should this be 4 spaces or 2?

evrardjp:
It could even 0. Yaml accepts everything all of these. 
I will switch to 2 and add a release note.

Off-Topic: I blame my editor that forgot that .yml and .yaml are the same kind of files.

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

odyssey4me:
To add to the comments by @evrardjp - pip has a preconfigured default of using pypi, so it makes sense that the change is simply the addition of an extra index as the change is more minimal. Just adding the extra index makes the most sense to keep the amount of change minimal and has no real difference in practical value versus changing both the default and extra index.

releasenotes/notes/rpc_package_index-a03d4d131c16d710.yaml
```diff
@@ -0,0 +1,12 @@
...
+  - The ``repo_build_pip_extra_index`` should be
+    replaced by a new variable (list type) named
+    ``repo_build_pip_extra_indexes``. This is required
+    if the deployer uses a python package index in addition
+    to the standard PyPI.
+
+deprecations:
+  - The ``repo_build_pip_extra_index`` variable is now
+    deprecated, and the deployer should use the list
+    ``repo_build_pip_extra_indexes`` instead.
```
git-harry: I'm not sure these notes are useful as currently written, they do not feel very specific to RPCO. For a standard RPCO deployment, the requirement to set repo_build_pip_extra_indexes so that rpc-repo is used, is completely internal to the project. I'm not sure we should account for the potential that this var, or some other related var, has been overidden because ultimately support can use anything OSA puts at their disposal. I think there is a reasonable case to say that a release note should be added because the var is being added to user_variables but as it stands the release notes do not seem to address that point.

prometheanfire: ya, I'd say the upgrade/deprecation is not needed here and is covered by OSA.  A simple note stating we are using the variable should be sufficient.

git-harry: Yes, I think I agree that having a note to explain that we are now overriding the variable and why should be sufficient.

git-harry:
👍

alextricity25:
:+1:

Issue #950, opened 2016-04-02
Script update-yaml.py fails to re-create base configuration file with overrides


mrhillsman:
##### What version (commit) do you encounter this issue on?

r12.0.0

##### What type of issue is this (bug, feature, documentation, regression)?

Bug

##### What is the expected behavior?

Merging of base configuration file ```/etc/openstack_deploy/user_variables.yml``` with overrides file ```/opt/rpc-openstack/rpcd/etc/openstack_deploy/user_variables.yml``` using python script ```/opt/rpc-openstack/scripts/update-yaml.py``` into a new file ```/etc/openstack_deploy/user_variables.yml``` which contains the content of both the base configuration and the overrides file.

##### What is actually happening?

An error occurs:

```
root@infra01:/opt/rpc-openstack# /opt/rpc-openstack/scripts/update-yaml.py /etc/openstack_deploy/user_variables.yml /opt/rpc-openstack/rpcd/etc/openstack_deploy/user_variables.yml 
Traceback (most recent call last):
  File "/opt/rpc-openstack/scripts/update-yaml.py", line 52, in <module>
    config = dict(base.items() + overrides.items())
AttributeError: 'NoneType' object has no attribute 'items'
```

##### How can we reproduce this issue?

SSH into deployment host
```
# cd /opt
# git clone -b r12.0.0 --recursive http://github.com/rcbops/rpc-openstack
# cd rpc-openstack/openstack-ansible
# cp -r etc/openstack_deploy /etc/
# scripts/bootstrap-ansible.sh
# cd /opt/rpc-openstack
# scripts/update-yaml.py /etc/openstack_deploy/user_variables.yml /opt/rpc-openstack/rpcd/etc/openstack_deploy/user_variables.yml
```
##### How can we verify your issue is addressed?

Without modifying the base configuration file ```/etc/openstack_deploy/user_variables.yml``` the file ```/opt/rpc-openstack/scripts/update-yaml.py``` successfully creates ```/etc/openstack_deploy/user_variables.yml``` with the base configuration and overrides in a new ```/etc/openstack_deploy/user_variables.yml``` file

##### If this is a large issue, what can be done immediately to alleviate the issue while working on a longer-term solution?
```shell
diff --git a/scripts/update-yaml.py b/scripts/update-yaml.py
index ad47d50..19eb17c 100755
--- a/scripts/update-yaml.py
+++ b/scripts/update-yaml.py
@@ -32,6 +32,7 @@ def get_config(path):
[...]

PR #1198, opened 2016-06-27
Account for empty yaml file in update-yaml.py

mancdaz:
If reading a yaml file with no entries, yaml.safe_read() produces a
NoneType object. This commit fixes the issue by re-ordering to ensure
we convert to NoneType before the code to handle NoneType is executed.

Connects #950 

Co-Authored-By: codebauss@gmail.com

### DIFF
--- a/scripts/update-yaml.py
+++ b/scripts/update-yaml.py
@@ -37,12 +37,13 @@ def get_config(path):
             data = None
         else:
             raise e
+    else:
+        data = yaml.safe_load(data)
 
     if data is None:
         return {}
     else:
-        # assume config is a dict
-        return yaml.safe_load(data)
+        return data
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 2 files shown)


Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Liberty-PR: Fail

scripts/update-yaml.py
```diff
@@ -37,12 +37,14 @@ def get_config(path):
             data = None
         else:
             raise e
+    else:
+        data = yaml.safe_load(data)
 
     if data is None:
         return {}
     else:
         # assume config is a dict
```
git-harry: This comment no longer makes sense here.

mancdaz: Not sure I agree...we do in fact assume we are returning a dict, else we get the error that this issue was raised for.
Having said that I don't care that much one way or another, so I'm happy to just remove it.

git-harry:
👍

evrardjp:
Not tested but looks good to me. 👍

Issue #1147, opened 2016-06-10
RabbitMQ MaaS checks failing Unauthorized


xeregin:
##### What version (commit) do you encounter this issue on?
12.2rc3

##### What type of issue is this (bug, feature, documentation, regression)?
Bug

##### What is the expected behavior?
The Python plugin should authenticate to the API and make a request.

##### What is actually happening?
The Python plugin is failing to authenticate to the API, receiving a 401 error.

##### What were you attempting to do when you encountered this error (give as much detail as possible)?
Configured and deployed MaaS. The RabbitMQ checks all failed because the password in user_extras_secrets did not match the actual RabbitMQ password. This happened on both labs that we upgraded.

##### How can we reproduce this issue?
I believe that the RabbitMQ checks were working properly preupgrade. The labs were on 12.2rc2 and the issue came about after the upgrade to rc3.

##### How can we verify your issue is addressed?
By verifying that the 24 RabbitMQ alarms are passing (8 per infra node)

##### If this is a large issue, what can be done immediately to alleviate the issue while working on a longer-term solution?
This is not a large issue and definitely not a blocker. A workaround is to jump into the rabbit container on an infra node and type the following command:
```
rabbitmqctl change_password maas_user <password from secrets file>
```

##### What is your evaluation on the importance of this issue, and why?
Not a high priority, because there is a very basic workaround.

###### Remove branches (lines) you do not wish to target.
- [ ] Master
- [ ] liberty-12.2

##### Does your change require documentation?
Unsure, but I believe:
- [x] No documentation required for this change

Tasks:

- [ ] Reproduce the issue. 
- [x] Create a PR for fix and/or docs.
- [ ] Incorporate feedback from pull request reviews.
- [ ]  What is the gating for this on DEV and/or QE? Add related gating if appropriate. Also make sure current gating passes with this fix.
- [ ] Reach out to reporter to determine if the fix is resolving the issue.
- [ ] What are the upgrade impacts (if any).
- [ ] What are the backports are needed (if any).

hughsaunders:
Did the password in the u_e_secrets file change during the upgrade? or was it removed?

git-harry:
As Hugh suggested, it looks like you removed the secrets file or 'maas_rabbitmq_password' from the file. To address this we would need to add 'force="yes"' [1] to the user setup task [2].

[1] http://docs.ansible.com/ansible/rabbitmq_user_module.html
[...]

PR #1200, opened 2016-06-28
Force the maas_rabbitmq_password to set up in RabbitMQ

cloader89:
This change allows for the rabbitmq_user.yml file to force and
grab the correct password from user_extras_secret.yml file. This
was done to ensure that RabbitMQ can authenticate properly.

Connects #1147

### DIFF
--- a/rpcd/playbooks/roles/rpc_maas/tasks/rabbitmq_user.yml
+++ b/rpcd/playbooks/roles/rpc_maas/tasks/rabbitmq_user.yml
@@ -23,6 +23,7 @@
     write_priv: ".*"
     tags: "administrator"
     state: "present"
+    force: "yes"
   tags:
     - rabbitmq-user
 


Apsu:
RPC-AIO-Liberty-PR: Fail

cloader89:
tempest.lib.exceptions.SSHTimeout: Connection to the 172.29.248.7 via SSH timed out.

ameserole:
recheck_swift

evrardjp:
Could explain why we should need that? 

The only reason I should see that happening is when a password is redefined. In that case, we should maybe have a different playbook. Because currently this is basically destroying and recreating the user for no apparent reason.

Forcing consistency is not a problem for me, but I'm just making sure we are not destroying idempotency for wrong reasons.

git-harry:
👍 I agree with @evrardjp that ideally we shouldn't be recreating it just for the sake of it however it's still idempotent because a user with the same name and password will always be recreated if the if the same name and password are supplied. I guess we could have a task to test that the credentials work and then only create/modify the user if they don't.

@evrardjp I don't think a separate playbook makes sense, it seems like it would be adding unnecessary complexity.

mancdaz:
👎 I'm generally not a fan of seeing yellow when nothing needs to change. I'd rather see something like @git-harry suggested and have a test to see first if we need to do anything.  That is, if we really feel like we need to account for what is a potentially unexplained issue in the first place.

sigmavirus24:
> That is, if we really feel like we need to account for what is a potentially unexplained issue in the first place.

Right. What I want to know is *what* changes Rabbit's password if the secrets file with Rabbit's password doesn't change.

git-harry:
So it strikes me that there are two things here:
 - the root cause analysis of the issue
 - the inability of the rpc_maas role to handle a change in the RabbitMQ user password

Now given that we have never seen this issue anywhere else, nor are we likely to be able to reproduce it given the available information, it seems that the simplest answer is probably the right one i.e. the secrets file was replaced/modified and that caused the two to become out of sync.

So while it's possible that making this change could potentially mask some other issue it also strikes me as reasonable that the role should have the ability to update a password if changed in the config.

cloader89:
I ran some test analysis, and ran the AIO without the force: "yes" and changing the password manually. It did not work. I can dig deeper into this issue and report back about why it is changing.

cloader89:
Or I can add some logic to see if it's changed.

evrardjp:
I don't think we should have this new feature "check if the pass has changed, and force change it".
If it was changed, there was a reason and should probably be evaluated.

I think this ```force: yes``` should not be added, because "changed" should happen when something really have to change.

I still believe a change password tool (playbook/script) should be created to change passwords overall (not only for this rabbit user) as part of an ops tooling, but this definitely not part of the scope of this question.

git-harry:
@evrardjp to me that is akin to suggesting we should create a change config tool to modify a config file. If we are going to have two sets of code, one to deploy and one to modify, we need to be explicit about that. As it stands, I believe the general expectation is that if a var is modified on an existing deployment that modification will be realised by the playbooks when they are re-run.

evrardjp:
then ```force: yes``` makes sense.

cloader89:
I have updated the bug report.

mancdaz:
The issue has been closed, so are we saying that this pr can be closed since we were unable to reproduce the issue?

cloader89:
yes

PR #1207, opened 2016-06-30
Install horizon_extensions from pip wheel instead of source

mancdaz:
In order to properly address the issue of horizon_extensions not building properly, we are backporting the change to the way we install it (via a pip package rather than installing from source). In order to cherry-pick that patch cleanly, I have also needed to include the patch to brand horizon using osa code, as opposed to rpc-specific kludging. Both are appropriate backports, and I have kept them as separate commits for obvious reasons.

Partial-Issue: #1100


mancdaz:
12.1 branch is going away, since 12.2 has been tagged. Killing this PR

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:

PR #1218, opened 2016-07-06
Always install the latest version of a pip package

evrardjp:
When re-running playbooks, if a version of the pip package was changed,
the playbooks would just ignore the package change because the package
is already installed. This change forces to use the latest package
available.

Connects rcbops/u-suk-dev#218


evrardjp:
Not tested on my AIO, will let jenkinks do its job.

git-harry:
👍

prometheanfire:
:+1:

PR #1220, opened 2016-07-06
Issue template update

git-harry:
The issue template is used by GitHub when a new issue is created.
Currently its main purpose is to prompt the issue reporter in to
providing all the information required to triage and investigate the
issue.

The main purpose of this patch is to give the individual working on
fixing the issue better guidance as to the steps that need to be taken
to complete it. This includes requiring the the individual to keep the
description (the section populated by the template) up-to-date with the
progress on work that has been done, e.g. pull requests created or
decisions made. In addition this patch seeks to make a clearer
distinction between the responsibilities of the issue reporter and the
issue assignee.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,6 +1,11 @@
-##### What version (commit) do you encounter this issue on?
+#### Do NOT include any confidential information, this issue will be created on a public repo
+#### This section is completed by the issue reporter
+##### On which version(s) do you encounter this issue?
+###### Please provide the output to 'git describe --all' for  all deployments where you have confirmed it occurs.
 
-##### What type of issue is this (bug, feature, documentation, regression)?
+##### What type of issue is this? Please tick the appropriate box.
+- [ ] bug - existing functionality that does not work for you as documented
+- [ ] feature - new functionality that you would like to see added
 
 ##### What is the expected behavior?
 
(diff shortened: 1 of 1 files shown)


b3rnard0:
These updates are good. We had been using a similar list to track some of these issues for the onboarding team. These included the following requirements for the people doing the work so I'm wondering if we need a similar one to the revised bug template:

- [ ] Reproduce the issue. 
- [ ] Create a PR for fix and/or docs.
- [ ] Incorporate feedback from pull request reviews.
- [ ]  What is the gating for this on DEV and/or QE? Add related gating if appropriate. Also make sure current gating passes with this fix.
- [ ] Reach out to reporter to determine if the fix is resolving the issue.
- [ ] What are the upgrade impacts (if any).
- [ ] What are the backports are needed (if any).

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,6 +1,11 @@
-##### What version (commit) do you encounter this issue on?
+#### Do NOT include any confidential information, this issue will be created on a public repo
+#### This section is completed by the issue reporter
+##### On which version(s) do you encounter this issue?
+###### Please provide the output to 'git describe --all' for  all deployments where you have confirmed it occurs.
 
-##### What type of issue is this (bug, feature, documentation, regression)?
+##### What type of issue is this? Please tick the appropriate box.
+- [ ] bug
+- [ ] feature
```
claco: I might even be more explicit (unless we're parsing that in other tools or something).

  bug (existing functionality does not work as documented)
  feature (adding functionality)

Or some such thing. Give more clarity, and less chance ot checking the wrong thing.

mancdaz:
This looks like a good change to me, and I like the clear separation of concerns between submitter and fixer. Shame we have to track it all in an issue description, but given that's what we have to work with...
👍

hughsaunders:
👍  I agree with @claco's comment, otherwise all good.

.github/ISSUE_TEMPLATE.md
prometheanfire: ya, defining what a bug/feature are would be good

prometheanfire:
Other than that it looks like a good change :+1:

sigmavirus24:
So... with the move to Jira, what does that mean for the issue template?

git-harry:
@sigmavirus24 I'll let @b3rnard0 answer that, also #1229 is in the same boat.

git-harry:
Outcome of team meeting was agreement to merge this pull request.

Issue #1188, opened 2016-06-23
MaaS API Authentication fails are not reported correctly


hughsaunders:
##### What version (commit) do you encounter this issue on?
liberty / e828f30a68645bb489c439669facbad88288bbec

##### What type of issue is this (bug, feature, documentation, regression)?
bug

##### What is the expected behavior?
If MaaS credentials are missing or incorrect, the first task that attempts to use them should fail. 

##### What is actually happening?
MaaS credentials are used in a shell pipeline, where the last command in the pipeline is wc, which always succeeds. 

```
- name: Entity {{ entity_name }} count
  shell: "raxmon-entities-list | grep ' label={{ entity_name }} provider=Rackspace Monitoring ...>$' | wc -l"
  register: entity_count
```
[link](https://github.com/rcbops/rpc-openstack/blob/master/rpcd/playbooks/roles/rpc_maas/tasks/raxmon_agent_install.yml#L41-L48)

This leads to a subsequent failure:

```
15:34:20 failed: [jrpcaioiad-c34] => {"failed": true}
15:34:20 msg: Zero entities with the label jrpcaioiad-c34 exist. Entities should be created as part of the hardware provisioning process, if missing, please consult the internal documentation for associating one with the device.
```

This failure is misleading, as the problem is not the lack of entities, but that inability of the previous task to query the api for entities.

##### What were you attempting to do when you encountered this error (give as much detail as possible)? 

Gate job for [rename user variables patch](https://github.com/rcbops/rpc-openstack/pull/1165)

##### How can we reproduce this issue?

Remove rackspace_cloud_* from the variables and deploy RPC including MaaS

##### How can we verify your issue is addressed?

Ensure that if MaaS credentials are missing or incorrect, the resulting failure indicates an inability to authenticate to the MaaS API. 

##### What is your evaluation on the importance of this issue, and why?

Low.. doesn't actually cause a failure, but can make diagnosing other failures more difficult.

###### Remove branches (lines) you do not wish to target.
- [ ] Master

##### Does your change require documentation?
- [ ] Yes, and I submitted an issue to [Docs issues](https://github.com/rackerlabs/docs-rpc/issues "Docs issues")
- [ ] Yes, and I contributed a documentation patch to [docs-rpc](https://github.com/rackerlabs/docs-rpc "docs-rpc")
- [ ] No documentation required for this change

alextricity25:
This is a good candidate for impacts-onboarding if we end up having to look for more.

[...]

PR #1222, opened 2016-07-06
Report MaaS API and raxmon auth fails correctly

ameserole:
This commit verifies that raxmon-entities-list command runs
correctly by running it without piping it first then running it
with the pipe. Now when the command fails it ends the error message
with: InvalidCredsError: 'Invalid credentials with the provider.'
Previously the command was piped to another shell command that
would never fail which would cause a different error to be
thrown later on.

Connects #1188

### DIFF
--- a/rpcd/playbooks/roles/rpc_maas/tasks/raxmon_agent_install.yml
+++ b/rpcd/playbooks/roles/rpc_maas/tasks/raxmon_agent_install.yml
@@ -38,6 +38,11 @@
     group: "root"
     mode: "0600"
 
+- name: Verify raxmon-entities-list and MaaS API authentication does not fail
+  command: "raxmon-entities-list"
+  register: raxmon_output
+  failed_when: raxmon_output.rc != 0
+
 - name: Entity {{ entity_name }} count
   shell: "raxmon-entities-list | grep ' label={{ entity_name }} provider=Rackspace Monitoring ...>$' | wc -l"
   register: entity_count


Apsu:
RPC-AIO-Liberty-PR: Fail

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

ameserole:
`tempest.lib.exceptions.ServerFault: Got server fault`
recheck_all

ameserole:
@hughsaunders What is your opinion on this solution for https://github.com/rcbops/rpc-openstack/issues/1188 ?

ameserole:
@hughsaunders In the original bug #1188 when it failed the playbook kept on running. Is that still supposed to happen in the bug fix when it fails or is it supposed to stop the playbook completely? The current fix I have posted marks the play as failed, prints the error, and keeps running.

rpcd/playbooks/roles/rpc_maas/tasks/raxmon_agent_install.yml
```diff
@@ -38,6 +38,11 @@
     group: "root"
     mode: "0600"
 
+- name: Verify raxmon-entities-list does not fail
```
hughsaunders: Would be good to indicate that this task is for testing the MaaS credentials / ability to authenticate against the MaaS API.

rpcd/playbooks/roles/rpc_maas/tasks/raxmon_agent_install.yml
```diff
@@ -38,6 +38,11 @@
     group: "root"
     mode: "0600"
 
+- name: Verify raxmon-entities-list and MaaS API authentication does not fail
+  shell: "raxmon-entities-list"
+  register: raxmon_output
+  failed_when: raxmon_output.stderr != ""
```
hughsaunders: raxmon-entities-list exists 0 on success and 1 on failure so you could check raxmon_output.rc instead of stderr.

hughsaunders:
@ameserole hard to fail early on this without setting "any_errors_fatal: true" in the setup-maas playbook, however that wouldn't be great for production. This patch is a good step forwards: better error message without impacting production.

rpcd/playbooks/roles/rpc_maas/tasks/raxmon_agent_install.yml
```diff
@@ -38,6 +38,11 @@
     group: "root"
     mode: "0600"
 
+- name: Verify raxmon-entities-list and MaaS API authentication does not fail
+  shell: "raxmon-entities-list"
+  register: raxmon_output
+  failed_when: raxmon_output.rc == 1
```
hughsaunders: For next time.. != 0 is safer, as different failures may have different codes.

hughsaunders:
👍

rpcd/playbooks/roles/rpc_maas/tasks/raxmon_agent_install.yml
tomc603: I'd actually suggest we fix this now rather than later when it bites us.

rpcd/playbooks/roles/rpc_maas/tasks/raxmon_agent_install.yml
```diff
@@ -38,6 +38,11 @@
     group: "root"
     mode: "0600"
 
+- name: Verify raxmon-entities-list and MaaS API authentication does not fail
+  shell: "raxmon-entities-list"
```
git-harry: command is probably better than shell here

ameserole: This command is run through shell in the next few plays so I was using shell for consistency.

git-harry: They are done with shell because they require the use of pipes. [Best practice](http://docs.ansible.com/ansible/shell_module.html#notes) is to use command where possible.

As an aside they are tasks not plays. Plays are found in a playbook. Tasks are found in a tasks file.

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

ameserole:
`19:21:01     tempest.lib.exceptions.ServerFault: Got server fault
19:21:01     Details: Unexpected API Error. Please report this at http://bugs.launchpad.net/nova/ and attach the Nova API log if possible.
19:21:01     <class 'neutronclient.common.exceptions.PortNotFoundClient `
recheck_ceph

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

ameserole:
`failed: [jrpcaioiad-a02_horizon_container-a91fdb68] => (item={'key_name': 'mariadb', 'hash_id': '0xcbcb082a1bb943db', 'keyserver': 'hkp://keyserver.ubuntu.com:80', 'fallback_keyserver': 'hkp://p80.pool.sks-keyservers.net:80'}) => {"attempts": 5, "cmd": "apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv CBCB082A1BB943DB", "failed": true, "item": {"fallback_keyserver": "hkp://p80.pool.sks-keyservers.net:80", "hash_id": "0xcbcb082a1bb943db", "key_name": "mariadb", "keyserver": "hkp://keyserver.ubuntu.com:80"}, "rc": 2}
20:21:24 stderr: gpg: requesting key 1BB943DB from hkp server keyserver.ubuntu.com`

recheck_ceph

tomc603:
Ok. :+1:

PR #1229, opened 2016-07-08
Update guidelines for issue triage and workflow

git-harry:
This change updates the contributing guidelines to define the workflow
by which issues are managed. This includes detailing the steps that take
an issue from the backlog to being done as well as providing a more
complete picture of the process for triaging issues.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -26,26 +26,156 @@
   * Security  - non-upgrade-impacting
   * Bug fixes - non-upgrade-impacting
 
-## Working in the repository
+# Issue workflow
 
-### Filing a bug report
+### Definitions
+#### Issues
+Issues are work items, they may be bugs or enhancements. They are managed using [Waffle](https://waffle.io/rcbops/u-suk-dev?source=rcbops%2Frpc-openstack) and stored in [GitHub](https://github.com/rcbops/rpc-openstack/issues). Issues that are logged in other places, for example on Launchpad for OSA, are not within scope unless they have been reported on this project by the creation of a new issue to track the upstream item.
 
-File a bug or requested feature against ```rpc-openstack```. Please include the following information:
+#### Waffle board
+The [Waffle board](https://waffle.io/rcbops/u-suk-dev?source=rcbops%2Frpc-openstack) is a Kanban board consisting of a series of columns used to manage the progress of issues. This board is the only place to view or update the current work commitments.
 
-* The branch (or tag) which you encountered this issue.
-* Steps to reproduce the issue.
-* Any output or logs from your environment.
+#### WIP (Work In Progress) limits
+Each column on the board has a WIP limit that is used to control the amount of new work to which the engineering team commits. The WIP limits will be listed in the column titles on the Waffle board. No column can receive new work if it violates the WIP limit unless there is an explicit exception made for it as defined in the processes below.
 
-### Fixing a bug
+#### Stakeholders
+* reporter - the individual that created the issue.
+* owner(s) - the individual(s) assigned to manage the issue through to its closure.
+* reviewer(s) - the individual(s) who review a pull request and provide feedback.
 
+#### Labels
+Labels are used to categorise issues to aid with tracking and prioritisation.
 
-1. After the bug is triaged and prioritised, make the required changes. New features, breaking changes and other patches of note must include a release note generated using the `reno tool`. Please see `Release Notes` for more information.
-2. Push changes directly to a branch on ```rpc-openstack``` in the ```rcbops``` github namespace (rather than a developers own fork). A pull request is then made from this branch to the ```master``` branch.
-3. Unless a bug is specific to a release branch (e.g. ```liberty-12.1```), commit fixes to the ```master``` branch before any potential backports.
-4. Bugs meeting backport criteria are backported to the release branches (e.g. ```liberty-12.1```) as appropriate.
-5. Github markdown is used to update the original issue with a checklist for tracking which branches have fixes merged.
-6. Each time a PR is merged, the associated branch is deleted from ```rpc-openstack```.
-7. When all PRs are completed the issue is then closed.
+|Type                     |Label name              |Description                                                                                                                                                   |Examples                                                                                      |
+|-------------------------|------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------|
+|Issue type               |type-bug                |Defects affecting the RPCO project.                                                                                                                           |                                                                                              |
+|Issue type               |type-enhancement-large  |Significant new features that require tracking by the product team.                                                                                           |Adding a new OpenStack service.                                                               |
+|Issue type               |type-enhancement-medium |Enhancements that impact other teams, such as requiring QE to implement significant new testing. These issues need to be factored into specific minor release.|                                                                                              |
+|Issue type               |type-enhancement-small  |Enhancements can reasonably be added to any minor release, these should in general not change default behaviour only add to it.                               |Additional MaaS plugin metrics or new logstash filters.                                       |
+|Issue type               |type-undecided          |Used when there is not enough information provided to determine the type.                                                                                     |                                                                                              |
+|Team affected            |impacts-qe              |Identifies issues that affect the QE team.                                                                                                                    |                                                                                              |
+|Team affected            |impacts-support         |Identifies issues that affect the Support team.                                                                                                               |                                                                                              |
+|Team affected            |impacts-onboarding      |Identifies issus that are suitable activities for new team members or individuals.                                                                            |                                                                                              |
+|Team affected            |impacts-training        |Identifies issues that affect the Training group.                                                                                                             |                                                                                              |
+|Team affected            |impacts-rpcr            |Identifies issues that affect the Rackspace-supported Red Hat OpenStack Platform team.                                                                        |                                                                                              |
+|Priority                 |prio-expedited          |Expedited issue, any issue that causes a fundamental failure in the product, especially if there is no workaround or it affects all deployments.              |Customers down, inoperable service, deployment failures of core functionality, blocked gating.|
+|Priority                 |prio-1                  |Priority 1 issues are fundamental failures where there is a workaround or only a subset of users are affected.                                                |Incorrect default configurations.                                                             |
+|Priority                 |prio-2                  |Priority 2 issues affect non-standard configurations or do not directly affect the customer experience or non-core features.                                  |Individual MaaS plugin failures.                                                              |
+|Priority                 |prio-3                  |Priority 3 issue are low impact, they are cosmetic or cause limited tangible impact.                                                                          |Typographical errors.                                                                         |
+|Priority                 |prio-undecided          |Unprioritised issue contain insufficient information on which to determine a priority.                                                                        |Issues that do not clearly describe what is broken or the impact it is having.                |
+|Workflow status (Waffle) |status-approved         |An issue that has been moved into the column Approved.                                                                                                        |                                                                                              |
+|Workflow status (Waffle) |status-doing            |An issue that has been moved into the column Doing.                                                                                                           |                                                                                              |
+|Workflow status (Waffle) |status-needs-review     |An issue that has been moved into the column Needs-review.                                                                                                    |                                                                                              |
+|Workflow status          |status-needs-information|The issue requires further information to enable it to be completed.                                                                                          |                                                                                              |
+|Workflow status          |status-blocked          |Indicates an issue is blocked on another item of work. The description should be updated with a link to the blocking issue.                                   |                                                                                              |
+|Workflow status          |status-duplicate        |Any issue that duplicates an existing issue.                                                                                                                  |                                                                                              |
+|Workflow status          |status-invalid          |Any issue that is not within scope of the project.                                                                                                            |                                                                                              |
+|Workflow status          |status-wontfix          |Any issue that it is determined will not be fixed.                                                                                                            |                                                                                              |
+|Tracking upstream project|tracking-upstream       |For tracking issues with upstream projects that are affecting the RPCO project.                                                                               |                                                                                              |
+|Product feature          |feature-gating          |Gating-related work.                                                                                                                                          |                                                                                              |
+|Product feature          |feature-upgrades        |Upgrades-related work.                                                                                                                                        |                                                                                              |
+|Product feature          |feature-monitoring      |Monitoring-related work.                                                                                                                                      |                                                                                              |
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -26,26 +26,156 @@
...
+|Product feature          |feature-upgrades        |Upgrades-related work.                                                                                                                                        |                                                                                              |
+|Product feature          |feature-monitoring      |Monitoring-related work.                                                                                                                                      |                                                                                              |
+|Product feature          |feature-osa             |OSA-related work.                                                                                                                                             |                                                                                              |
+
+## Pre-Backlog -> Backlog
```
hughsaunders: What is meant by pre-backlog? This seems to be the only mention.

CONTRIBUTING.md
```diff
@@ -26,26 +26,156 @@
...
+
+Each issue is given a priority and that priority is used to help decide the order in which issues are worked. It is important to remember that there is a finite resource available to fix issues. The process of prioritisation is used to focus the available resources on the most important tasks identified.
+
+Issues are labelled with a priority as defined in the [labels table](#Labels).
+
+#### Expedited issues
+Expedited issues are allowed to exceed the WIP limit on any column. These issues are critical to fix and must be prioritised above all other work. They are identified with the label 'prio-expedited'. Due to the negative impact expedited issues have on WIP it is important they are used appropriately.
+
+#### Enhancements
+Issues labeled as 'type-enhancement-large' or 'type-enhancement-medium' filter into planning for major and minor releases and should never be moved to Approved.
```
hughsaunders: If these are never in approved, does that mean they will be directly assigned to team members and jump from backlog to doing?

CONTRIBUTING.md
```diff
@@ -26,26 +26,156 @@
...
+
+Issues labeled as 'type-enhancement-small' should also be given a priority, in general they should be labeled as 'prio-3' but a higher priority can be given if the enhancement helps to prevent customer downtime.
+
+#### Classifying issues
+Issues are categorised to simplify the process of identifying the highest priority issues and allow different types of work to be tracked. The steps taken to correctly label an issue are:
+* Label issue type
+  * Assign the appropriate issue-type label.
+  * If the issue is determined to be 'type-undecided', also add the labels 'prio-undecided' and 'status-needs-information'. Add a comment detailing the required information.
+  * If the issue is determined to be 'type-enhancement-large', the issue should be logged as an idea on the [product roadmap](https://trello.com/b/OdRe9RxX/rpc-engineering-roadmap) and closed here with a link to the roadmap card.
+  * If the issue is determined to be 'type-enhancement-medium', the issue should be added to the [existing-feature backlog](https://waffle.io/rcbops/u-suk-dev?source=rcbops%2Fu-suk-dev) and closed here with a link to the new card.
```
hughsaunders: Backlog may be clearer than "existing-feature backlog", assuming there aren't multiple backlogs.

CONTRIBUTING.md
```diff
@@ -26,26 +26,156 @@
...
+
+##### New issues discovered while working an issue
+If a new bug or feature requirement is found while working on an assigned issue, that new work item should be logged as a separate issue in Backlog.
+
+If the new issue blocks the existing issue, you should assign yourself to the new issue and move it to Doing. The existing issue should be updated with all the work that has been done, the label 'status-blocked' added and a link to the blocking issue should be included in the issue description. In this situation the WIP limit can be exceeded. Similar to situations where the WIP limit prevents new work being started the team needs to work to reduce the WIP so that new tasks can be started.
+
+If the new issue does not prevent the resolution of your current task it will be treated as any new issue.
+
+##### Creating a pull request
+If an issue affects multiple branches the owner should always start by creating a pull request for the newest branch, generally this means master. Only one pull request should ever be in progress for a task in Doing.
```
hughsaunders: bold "**Only one pull request should ever be in progress for a task in Doing.**"

CONTRIBUTING.md
git-harry: 'Pre-backlog' refers to the state prior to the backlog i.e. before an issue is logged. It is not represented in Waffle because it represents the the point when an issue has not been logged.

CONTRIBUTING.md
git-harry: They are tracked on other boards and so they are closed.

CONTRIBUTING.md
git-harry: This is referring to the backlog for u-suk-dev not the one for rpc-openstack.

prometheanfire: Might be useful to add a note here then so we don't have to inspect the link.

CONTRIBUTING.md
```diff
@@ -26,26 +26,156 @@
...
+* owner(s) - the individual(s) assigned to manage the issue through to its closure.
+* reviewer(s) - the individual(s) who review a pull request and provide feedback.
+
+#### Labels
+Labels are used to categorise issues to aid with tracking and prioritisation.
+
+|Type                     |Label name              |Description                                                                                                                                                   |Examples                                                                                      |
+|-------------------------|------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------|
+|Issue type               |type-bug                |Defects affecting the RPCO project.                                                                                                                           |                                                                                              |
```
prometheanfire: We don't care about small vs large bugs?

CONTRIBUTING.md
```diff
@@ -26,26 +26,156 @@
...
+
+## Pre-Backlog -> Backlog
+### Adding an issue to the backlog
+Click [here](https://github.com/rcbops/rpc-openstack/issues/new) to load the form used when creating a new issue in the backlog. The form is generated from [this template](.github/ISSUE_TEMPLATE.md) which includes a set of questions for the reporter and a checklist to be completed by the owner as the issue is worked. Currently Waffle does not support the use of an issue template and so issues should not be created directly in Waffle.
+
+## Backlog
+All issues that are are not approved to be worked are listed in Backlog, this includes new issues and those that have been triaged but not yet committed to.
+
+### Issue triage
+Issue triage is an ongoing process that should happen on a daily basis, the engineering team will triage new issues and monitor unprioritised issues so that new commitments are always targetting the most import work.
```
prometheanfire: I thought we just did this twice weekly.

CONTRIBUTING.md
```diff
@@ -26,26 +26,156 @@
...
+
+## Doing
+#### Fixing an issue
+The issue description contains a list of the steps required to close an issue, this is added by the [issue template](.github/ISSUE_TEMPLATE.md). As each step is completed, update the description.
+
+##### New issues discovered while working an issue
+If a new bug or feature requirement is found while working on an assigned issue, that new work item should be logged as a separate issue in Backlog.
+
+If the new issue blocks the existing issue, you should assign yourself to the new issue and move it to Doing. The existing issue should be updated with all the work that has been done, the label 'status-blocked' added and a link to the blocking issue should be included in the issue description. In this situation the WIP limit can be exceeded. Similar to situations where the WIP limit prevents new work being started the team needs to work to reduce the WIP so that new tasks can be started.
```
prometheanfire: how is this not counted against the wip limit (to a casual observer)?

CONTRIBUTING.md
prometheanfire: Does this mean that once master is merged we can't merge to all other branches at the same time?  Wasn't this standard behavior (and isn't it standard behavior in OSA)?

CONTRIBUTING.md
```diff
@@ -26,26 +26,156 @@
...
+If an issue affects multiple branches the owner should always start by creating a pull request for the newest branch, generally this means master. Only one pull request should ever be in progress for a task in Doing.
+
+To ensure that issues and pull request are correctly associated, reference the issue at the end of the commit message, e.g.:
+    Connected https://github.com/rcbops/rpc-openstack/issues/0
+
+## Doing -> Needs review
+A card can only be moved from Doing to Needs-review if Needs-review is below its WIP limit. If that condition is met and you are ready for feedback on your pull request, move the card (this should be a connected pair of the issue and pull request) to the bottom of Needs-review.
+
+## Needs review
+Reviewers should always review the tasks in the order the are listed in Needs-review. If owners need to make changes to the pull request, for example base on feedback from reviewers, the task should remain in Needs-review while that is done. The reviewer merging the commit should delete the issue branch (using the link on the pull request page) and tick the issue checklist for the fixed branch.
```
prometheanfire: Who manages this order?

git-harry: In the previous step you'll see it says new cards go to the bottom of the list so it's based on the order they are added.

prometheanfire: Ok, the reason I asked is because when I drag a card over I just drag it without caring where it drops, sometimes it's top, sometimes not.

git-harry: This process will require individuals to modify their current working practices.

CONTRIBUTING.md
git-harry: It does count towards the WIP limit, so if the limit is reached/exceeded this blocks bring new cards across.

CONTRIBUTING.md
git-harry: Currently there is one weekly meeting. I would like to see more emphasis on triaging issues as they come in.

CONTRIBUTING.md
git-harry: If this is in comparison the how enhancements have been broken up, that relates to the fact that internal processes track different types of enhancement and so they need tracking somewhere else. All bugs are here and so there is no need to size them.

CONTRIBUTING.md
git-harry: Yes, one pull request at a time. Currently we don't define how this is done. Upstream is there own thing. This process is not set in stone so it will evolve over time. I made this choice to try and keep the board simple, i.e. each card represents one item of work, but am not wedded to the idea so we can change it if we find it doesn't work for us.

CONTRIBUTING.md
prometheanfire: Sure, just calling it out :D

CONTRIBUTING.md
prometheanfire: ok, I'd push for that (multiple active backports allowed)

CONTRIBUTING.md
prometheanfire: ack

CONTRIBUTING.md
prometheanfire: I'd like that as well, would we have meetings/standups every morning for this?

CONTRIBUTING.md
prometheanfire: ack

CONTRIBUTING.md
b3rnard0: From my POV, the only time that card order matters is in the Approval column. That's when are following a FIFO queueing approach. After that that step, we will focus on the cards that are the oldest to newest.

git-harry: How are you defining oldest to newest? I haven't seen any way to sort columns based on some property of the cards.

The idea behind always adding new cards to the bottom of Needs-review is so that the card at the top of the column is always the oldest (the card that has been in that column the longest). This reduces the chances that cards, that are difficult to review, will sit there for a long time because the will block other reviews.

CONTRIBUTING.md
git-harry: I'll rework this a bit. It's not as clear as it should be.

CONTRIBUTING.md
git-harry: Apologies, my previous comment is inaccurate, I think I had a previous iteration of this document in my head when I responded to you. Backports are done while the card is in Needs-review. What I was attempting to say in this section is that the fix and all its backports shouldn't be done at the same time because pull requests can get out of sync and it just ends up being more effort than waiting for the initial fix to merge.

prometheanfire: Then line 159 probably needs to me amended to allow for multiple backports after the primary (typically master) has merged.

git-harry: This section on pull requests relates to when the issue is in Doing. Backports are done when the issue is in Needs-review.

CONTRIBUTING.md
b3rnard0: @prometheanfire What are you concerns regarding small/large bugs? 

Once we adopt a WIP limit approach and start to use the metrics data, the size of the bug doesn't really matter. My expectation is that when committing to a bug, that we look at the work required to fix the bug and ask whether we can complete it within our cycle time. If our cycle time for most bugs is about 5 days and if the bug can be completed within those 5 days, great, size doesn't matter. However, if the bug appears to be larger than the cycle time of 5 days, than size does matter. If we feel a bug will take longer to fix than our historical average, then we can look into breaking the bug out or even doing a mini-spike to determine if there's a way that we can do it within the cycle time.

The other scenario would be that if we commit to a bug that we thought would take only 5 days but it's on day 6, then obviously we need to look at it as a team because the bug is an outlier that needs to be resolved. So that's a swarming or pair programming opportunity.

CONTRIBUTING.md
b3rnard0: I would clarify that when a team member is blocked, we really shouldn't commit to or pull in new work. The emphasis by the team should be to help out that individual to get unblocked.

CONTRIBUTING.md
b3rnard0: Waffle doesn't have any way to sort from oldest to newest. We would have to rely on GitHub for that sort of sorting and by manually making sure we are not putting new work on top of a column list.

CONTRIBUTING.md
prometheanfire: no questions, @git-harry answered it

CONTRIBUTING.md
git-harry: @b3rnard0 My thinking was that the WIP limits should guide this. So if the column is at the WIP limit, this prevents individuals starting new tasks and so they should look to help with existing tasks. If no WIP limit is reached, does it matter? In another comment you mention we need to be aware of cycle time and so in that case maybe that would be another reason to assist on the blockage. By default though I wonder if it is necessary - I may just need to read more kanban.

b3rnard0: Yes, that's correct -- the WIP should guide the team on what to focus on as well as trying to keep our cycle time consistent. My point is that if someone is drowning, why would they pull in new work when they can help their teammate? It's just like the kanban simulation. The decision points are fairly clear about what to focus on as they are based on data points (cycle time). Ultimately, it's is about committing to delivering value to our stakeholders and customers and keeping that consistent. If we delay that by pulling more work, we are making the wrong decisions. This is more philosophizing than actual comments about the updated triage process. :-)

git-harry:
Outcome of team meeting was agreement to merge this pull request.

PR #1231, opened 2016-07-11
Exclude remote Swift regions for Beaver install.

byronmccollum:
The Beaver install has a dependency on pip lockdown, causing the
pip.conf files in the remote Swift regions to be reconfigured to point
to the repo in a different region. This can result in hanging when
attempting to install packages on the remote Swift regions, as the
management network is not routed between regions.

This change alters the target hosts of the Beaver install to exclude
any remote Swift regions.

### DIFF
--- a/rpcd/playbooks/beaver.yml
+++ b/rpcd/playbooks/beaver.yml
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 - name: Setup beaver log shipper
-  hosts: all
+  hosts: all_containers:hosts
   max_fail_percentage: 20
   user: root
   roles:


Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

byronmccollum:
recheck

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

byronmccollum:
recheck

ameserole:
Beaver was removed from master with this commit https://github.com/rcbops/rpc-openstack/commit/2d7bce0ae13399b59ac98833b066ce9248f1fb00

ameserole:
@byronmccollum I haven't really looked into the issue you are trying to fix yet. I was just pointing out that the reason you are getting a merge conflict and gate test failures is due to the fact that the file you changed no longer exists in master. 
Also, the reason travis is failing is due to the period at the end of your commit title.

d34dh0r53:
I think we can close this as beaver is no longer a thing and this is against master.  We can revisit the issue should it still be a problem with filebeat.

PR #1240, opened 2016-07-18
Add cpu/mem multipliers to cloud stats plugin

mancdaz:
The nova API reports 'total_vcpus' as the total number of physical cores
available on the hypervisors that it can see. However, because of the
setting cpu_allocation_ratio it's possible to see more 'used_vcpus' than
'total_vcpus' (actual physical cores in the machines).

The same is true of memory and mem_allocation_ratio.

Whilst the nova API does not have the capability of factoring in the
*_allocation_ratios across all hypervisors when reporting 'total_*' numbers,
we can fake it by passing a multiplier to the plugin using numbers that match
the nova conf settings. The default for the plugin remains to use the real
numbers as reported by the nova API.

It's accepted that this does not cover the case where different hypervisors
may have different allocation ratios, but should cover 95% of use cases where
the allocation_ratios are the same across all hypervisors.

Connects #1107 
(cherry picked from commit 3cbd3da8e13035cba890a7d3e17c48b669b68f48)

### DIFF
--- a/maas/plugins/nova_cloud_stats.py
+++ b/maas/plugins/nova_cloud_stats.py
@@ -79,8 +79,13 @@ def check(args):
         stats = nova.hypervisor_stats.statistics()
         cloud_stats = collections.defaultdict(dict)
         for metric_name, vals in stats_mapping.iteritems():
+            multiplier = 1
+            if metric_name == 'total_vcpus':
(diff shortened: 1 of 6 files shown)


hughsaunders:
👍

Issue #1164, opened 2016-06-15
Monitoring for MariaDB status needed


BjoernT:
##### What version (commit) do you encounter this issue on?

All

##### What type of issue is this (bug, feature, documentation, regression)?

Feature

##### What is the expected behavior?

No more "my number of session hit max_connection setting" situations

##### What is actually happening?

MariaDB and it's health is virtually unmonitored

##### What were you attempting to do when you encountered this error (give as much detail as possible)? 

We need to monitor important status parameters like

- Current Number of DB sessions in relation to max_connection setting
- Max_used_connections in relation to max_connection setting
- Open_files < open file limit
- Innodb_row_lock_time_avg < 5000 ?
- Innodb_deadlocks = 0
- Aborted_clients < 1/s
- Aborted_connects < 1/s
- Access_denied_errors = 0

##### How can we reproduce this issue?

-

##### How can we verify your issue is addressed?

Above checks are implemented

##### If this is a large issue, what can be done immediately to alleviate the issue while working on a longer-term solution?
-

##### What is your evaluation on the importance of this issue, and why?

-

###### Remove branches (lines) you do not wish to target.
- [x] Master
- [x] liberty-12.2

##### Does your change require documentation?
- [ ] Yes, and I submitted an issue to [Docs issues](https://github.com/rackerlabs/docs-rpc/issues "Docs issues")
- [ ] Yes, and I contributed a documentation patch to [docs-rpc](https://github.com/rackerlabs/docs-rpc "docs-rpc")
- [x] No documentation required for this change

git-harry:
@BjoernT there is an [existing database plugin][1] and [alarms][2] are raised for some of the metrics it generates, are you aware of this? On the face of it, this setup seems to cover some of your request. How do the connection alarms that exist relate to your description of the expected behaviour?

[1]: https://github.com/rcbops/rpc-openstack/blob/master/maas/plugins/galera_check.py
[2]: https://github.com/rcbops/rpc-openstack/blob/master/rpcd/playbooks/roles/rpc_maas/templates/galera_check.yaml.j2

sigmavirus24:
In today's bug triage, we discussed that we might need to add some more metrics/alarms based on @BjoernT's original report.

We need more information around:

- thresholds for alarms
- which metrics aren't presently being provided by the current galera plugin
- whether the customer in question even had MaaS enabled

BjoernT:
Added thresholds for the last two alarms

claco:
[...]

PR #1242, opened 2016-07-18
Add more checks for the galera MaaS plugins

alextricity25:
This commit will add some more robust checks for the galera
plugin. We also create some alarms for the corresponding checks.
The criteria for these alarms include(when the alarms will trigger):

* Open_files > open file limit
* Innodb_row_lock_time_avg > Innodb_row_lock_time_max
* Innodb_deadlocks != 0
* Aborted_clients > mysql_aborted_clients_rate_warning_threshold [1]
* Aborted_connects > mysql_aborted_connects_rate_warning_threshold [1]
* access_denied_errors" >  mysql_access_denied_errors_rate_warning_threshold [1]
* percentage("mysql_max_seen_connections, mysql_max_configured_connections) > mysql_max_seen_connections_critical_threshold

[1] The rate function modifier is used here.
    By definition:
    Rate as defined by this formula where V=value, and T=time:
    (V1 - V0) / (T1 - T0).
    This is more of a ratio as to how many clients/connects are
    being aborted every {{ maas_check_period }} seconds (default is 60).

Connects #1164

### DIFF
--- a/maas/plugins/README.md
+++ b/maas/plugins/README.md
@@ -335,18 +335,25 @@ connects to an individual member of a galera cluster and checks various statuses
 
 ##### Example Output:
 
-    metric wsrep_replicated_bytes int64 320737952 bytes
-    metric wsrep_received_bytes int64 33470 bytes
-    metric wsrep_commit_window double 1.000000
+    metric wsrep_replicated_bytes int64 1307 bytes
+    metric wsrep_received_bytes int64 299878933 bytes
+    metric wsrep_commit_window_size double 1.000143 sequence_delta
     metric wsrep_cluster_size int64 3 nodes
-    metric queries_per_second int64 5792715 qps
-    metric wsrep_cluster_state_uuid string 67e41d08-165d-11e4-9d87-7e94ef43b302
-    metric wsrep_cluster_status string primary
-    metric wsrep_local_state_uuid string 67e41d08-165d-11e4-9d87-7e94ef43b302
-    metric wsrep_local_state_comment string synced
+    metric queries_per_second int64 380289 qps
+    metric wsrep_cluster_state_uuid string 6a3b85c0-4b07-11e6-8778-3f5baf70ad5c
+    metric wsrep_cluster_status string Primary
+    metric wsrep_local_state_uuid string 6a3b85c0-4b07-11e6-8778-3f5baf70ad5c
+    metric wsrep_local_state_comment string Synced
     metric mysql_max_configured_connections int64 800 connections
     metric mysql_current_connections int64 1 connections
     metric mysql_max_seen_connections int64 2 connections
+    metric num_of_open_files int64 14 files
+    metric open_files_limit int64 65535 files
+    metric innodb_row_lock_time_avg int64 0 milliseconds
+    metric innodb_deadlocks int64 0 deadlocks
+    metric access_denied_errors int64 0 access_denied_errors
+    metric aborted_clients int64 0 aborted_clients
+    metric aborted_connects int64 0 aborted_connects
 
 ***
 #### conntrack_count.py
--- a/maas/plugins/galera_check.py
+++ b/maas/plugins/galera_check.py
@@ -90,6 +90,20 @@ def print_metrics(replica_status):
            replica_status['Threads_connected'], 'connections')
     metric('mysql_max_seen_connections', 'int64',
            replica_status['Max_used_connections'], 'connections')
+    metric('num_of_open_files', 'int64',
+           replica_status['Open_files'], 'files')
+    metric('open_files_limit', 'int64',
+           replica_status['open_files_limit'], 'files')
+    metric('innodb_row_lock_time_avg', 'int64',
+           replica_status['Innodb_row_lock_time_avg'], 'milliseconds')
+    metric('innodb_deadlocks', 'int64',
+           replica_status['Innodb_deadlocks'], 'deadlocks')
+    metric('access_denied_errors', 'int64',
+           replica_status['Access_denied_errors'], 'access_denied_errors')
+    metric('aborted_clients', 'int64',
+           replica_status['Aborted_clients'], 'aborted_clients')
+    metric('aborted_connects', 'int64',
+           replica_status['Aborted_connects'], 'aborted_connects')
 
 
 def main():
--- a/rpcd/playbooks/roles/rpc_maas/defaults/main.yml
+++ b/rpcd/playbooks/roles/rpc_maas/defaults/main.yml
@@ -236,6 +236,36 @@ maas_filesystem_critical_threshold: 90.0
 mysql_connection_warning_threshold: 80
 mysql_connection_critical_threshold: 90
 
+# Maas mysql access denied thresholds
+# e.g If number of access denied errors inbetween checks
+# exceeds 20, the alarm will go into CRITICAL status.
+mysql_access_denied_errors_rate_warning_threshold: 10
+mysql_access_denied_errors_rate_critical_threshold: 20
+
+# Maas mysql aborted clients rate threshold
+# The rate is derived on a per-check basis.
+# In other words, this threshold defines how many
+# aborted_clients per check are needed to trigger
+# an alarm.
+mysql_aborted_clients_rate_warning_threshold: 1
+mysql_aborted_clients_rate_critical_threshold: 2
+
+# Maas mysql aborted connects rate threshold
+# The rate is derived on a per-check basis.
+# In other words, this threshold defines how many
+# aborted_connects per check are needed to trigger
+# an alarm
+mysql_aborted_connects_rate_warning_threshold: 1
+mysql_aborted_connects_rate_critical_threshold: 2
+
+# Maas mysql open file percentage threshold
+mysql_open_files_percentage_warning_threshold: 90
+mysql_open_files_percentage_critical_threshold: 95
+
+# Maas InnoDB row lock time avg thresholds, in milliseconds.
+innodb_row_lock_time_avg_warning_threshold: 2000
+innodb_row_lock_time_avg_critical_threshold: 10000
+
 # cinder-volumes volume group thresholds
 cinder_volumes_vg_warning_threshold: 80.0
 cinder_volumes_vg_critical_threshold: 90.0
--- a/rpcd/playbooks/roles/rpc_maas/templates/galera_check.yaml.j2
+++ b/rpcd/playbooks/roles/rpc_maas/templates/galera_check.yaml.j2
@@ -34,3 +34,66 @@ alarms      :
             if (percentage(metric["mysql_current_connections"], metric["mysql_max_configured_connections"]) > {{ mysql_connection_warning_threshold }} ) {
                 return new AlarmStatus(WARNING, "Mysql connectons has exceeded {{ mysql_connection_warning_threshold }}% of configured maximum");
             }
+    open_file_size_limit_reached :
+        label                   : open_file_size_limit_reached--{{ ansible_hostname }}
+        notification_plan_id    : "{{ maas_notification_plan }}"
+        criteria                : |
+            :set consecutiveCount={{ maas_alarm_local_consecutive_count }}
+            if (percentage(metric["num_of_open_files"], metric["open_files_limit"]) > {{ mysql_open_files_percentage_critical_threshold }}) {
+                return new AlarmStatus(CRITICAL, "Number of open files has exceeded {{ mysql_open_files_percentage_critical_threshold }}% of open file limit");
+            }
+            if (percentage(metric["num_of_open_files"], metric["open_files_limit"]) > {{ mysql_open_files_percentage_warning_threshold }}) {
+                return new AlarmStatus(WARNING, "Number of open files has exceeded {{ mysql_open_files_percentage_warning_threshold }}% of open file limit");
+            }
+    innodb_row_lock_time_avg :
+        label                   : innodb_row_lock_time_avg--{{ ansible_hostname }}
+        notification_plan_id    : "{{ maas_notification_plan }}"
+        criteria                : |
+            :set consecutiveCount={{ maas_alarm_local_consecutive_count }}
+            if (metric["innodb_row_lock_time_avg"] > {{ innodb_row_lock_time_avg_critical_threshold }}) {
+                return new AlarmStatus(CRITICAL, "InnoDB row lock time average has exceeded {{ innodb_row_lock_time_avg_critical_threshold }} milliseconds");
+            }
+            if (metric["innodb_row_lock_time_avg"] > {{ innodb_row_lock_time_avg_warning_threshold }}) {
+                return new AlarmStatus(WARNING, "InnoDB row lock time average has exceeded {{ innodb_row_lock_time_avg_warning_threshold }} milliseconds");
+            }
+    innodb_deadlocks :
+        label                   : innodb_deadlocks--{{ ansible_hostname }}
+        notification_plan_id    : "{{ maas_notification_plan }}"
+        criteria                : |
+            :set consecutiveCount={{ maas_alarm_local_consecutive_count }}
+            if (metric["innodb_deadlocks"] != 0) {
+                return new AlarmStatus(CRITICAL, "InnoDB has experienced a deadlock");
+            }
+    access_denied_errors :
+        label                   : access_denied_errors--{{ ansible_hostname }}
+        notification_plan_id    : "{{ maas_notification_plan }}"
+        criteria                : |
+            :set consecutiveCount={{ maas_alarm_local_consecutive_count }}
+            if (rate(metric["access_denied_errors"]) > {{ mysql_access_denied_errors_rate_warning_threshold }}) {
+                return new AlarmStatus(WARNING, "Access denied errors rate is greater than {{ mysql_access_denied_errors_rate_warning_threshold }} per {{ maas_check_period }} seconds");
+            }
+            if (rate(metric["access_denied_errors"]) > {{ mysql_access_denied_errors_rate_critical_threshold }}) {
+                return new AlarmStatus(CRITICAL, "Access denied errors rate is greater than {{ mysql_access_denied_errors_rate_critical_threshold }} per {{ maas_check_period }} seconds");
+            }
+    aborted_clients :
+        label                   : aborted_clients--{{ ansible_hostname }}
+        notification_plan_id    : "{{ maas_notification_plan }}"
+        criteria                : |
+            :set consecutiveCount={{ maas_alarm_local_consecutive_count }}
+            if (rate(metric["aborted_clients"]) > {{ mysql_aborted_clients_rate_warning_threshold }}) {
+                return new AlarmStatus(WARNING, "Aborted clients rate is greater than {{ mysql_aborted_clients_rate_warning_threshold }} per {{ maas_check_period }} seconds");
+            }
+            if (rate(metric["aborted_clients"]) > {{ mysql_aborted_clients_rate_critical_threshold }}) {
+                return new AlarmStatus(CRITICAL, "Aborted clients rate is greater than {{ mysql_aborted_clients_rate_critical_threshold }} per {{ maas_check_period }} seconds");
+            }
+    aborted_connects :
+        label                   : aborted_connects--{{ ansible_hostname }}
+        notification_plan_id    : "{{ maas_notification_plan }}"
+        criteria                : |
+            :set consecutiveCount={{ maas_alarm_local_consecutive_count }}
+            if (rate(metric["aborted_connects"]) > {{ mysql_aborted_connects_rate_warning_threshold }}) {
+                return new AlarmStatus(WARNING, "Aborted connects rate is greater than {{ mysql_aborted_connects_rate_warning_threshold }} per {{ maas_check_period }} seconds");
+            }
+            if (rate(metric["aborted_connects"]) > {{ mysql_aborted_connects_rate_critical_threshold }}) {
+                return new AlarmStatus(CRITICAL, "Aborted connects rate is greater than {{ mysql_aborted_connects_rate_critical_threshold }} per {{ maas_check_period }} seconds");
+            }
--- a/releasenotes/notes/add-more-galera-checks-032c765829e922a2.yaml
+++ b/releasenotes/notes/add-more-galera-checks-032c765829e922a2.yaml
@@ -1,0 +1,26 @@
+---
+features:
+  - |
+    The galera_check.py plugin now provides the following
+    metrics.
+    * num_of_open_files
+    * open_files_limit
+    * innodb_row_lock_time_avg
+    * innodb_deadlocks
+    * access_denied_errors
+    * aborted_clients
+    * aborted_connects
+  - |
+    | New alarms have been created with the following criteria.
+    | open_file_size_limit_reached:
+    | metric["num_of_open_files"] > metric["open_files_limit"]
+    | innodb_row_lock_time_avg:
+    | metric["innodb_row_lock_time_avg"] > {{ innodb_row_lock_time_avg_critical_threshold }}
+    | innodb_deadlocks:
+    | metric["innodb_deadlocks"] != 0
+    | access_denied_errors:
+    | rate(metric["access_denied_errors"]) > {{ mysql_access_denied_errors_rate_warning_threshold }}
+    | aborted_clients:
+    | rate(metric["aborted_clients"]) > 1
+    | aborted_connects:
+    | rate(metric["aborted_connects"]) > 1


alextricity25:
TODO items:
Add alarm for aborted clients and aborted connects.
Do research on what the threshold for innodb_row_lock_time_avg should be.

Apsu:
RPC-AIO-Liberty-PR: Fail

Apsu:
RPC-AIO-Liberty-PR: Fail

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Liberty-PR: Fail

Apsu:
RPC-AIO-Liberty-PR: Fail

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Liberty-PR: Fail

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Liberty-PR: Fail

alextricity25:
recheck_swift

Apsu:
RPC-AIO-Liberty-PR: Fail

Apsu:
RPC-AIO-Liberty-PR: Fail

alextricity25:
recheck_swift

Apsu:
RPC-AIO-Liberty-PR: Fail

alextricity25:
tempest.scenario.test_server_basic_ops.TestServerBasicOps.test_server_basic_ops[compute,id-7fff3fb3-91d8-4fd0-bd7d-0204f1f180ba,network,smoke]

recheck_swift

maas/plugins/README.md
```diff
@@ -335,18 +335,27 @@ connects to an individual member of a galera cluster and checks various statuses
...
+    metric wsrep_replicated_bytes int64 0 bytes
+    metric wsrep_received_bytes int64 226600043 bytes
+    metric wsrep_commit_window_size double 1.000161 sequence_delta
     metric wsrep_cluster_size int64 3 nodes
-    metric queries_per_second int64 5792715 qps
-    metric wsrep_cluster_state_uuid string 67e41d08-165d-11e4-9d87-7e94ef43b302
-    metric wsrep_cluster_status string primary
-    metric wsrep_local_state_uuid string 67e41d08-165d-11e4-9d87-7e94ef43b302
-    metric wsrep_local_state_comment string synced
+    metric queries_per_second int64 297004 qps
```
mancdaz: it's worth noting that the number that the plugin pulls from mysql is actually a cumulative total of queries since last daemon restart, so this metric is misnamed. We should fix the metric name, and then in the future if we ever need to monitor/alarm on qps we can use the 'rate' function in the alarm language to decipher actual qps.

maas/plugins/galera_check.py
```diff
@@ -90,6 +90,22 @@ def print_metrics(replica_status):
...
+    metric('innodb_row_lock_time_max', 'int64',
+           replica_status['Innodb_row_lock_time_max'], 'milliseconds')
+    metric('innodb_deadlocks', 'int64',
+           replica_status['Innodb_deadlocks'], 'deadlocks')
+    metric('access_denied_errors', 'int64',
+           replica_status['Access_denied_errors'], 'access_denied_errors')
+    metric('aborted_clients', 'int64',
+           replica_status['Aborted_clients'], 'aborted_clients')
+    metric('aborted_connects', 'int64',
+           replica_status['Aborted_connects'], 'aborted_connects')
```
mancdaz: I wonder, with this many metrics, whether we might be best creating a mapping dict, mapping the parameter name obtained from lysql, to the metric name, along with the units.  A bit like here:

https://github.com/rcbops/rpc-openstack/blob/009ae7d2b4751d52d012987a111ad206ba935502/maas/plugins/nova_cloud_stats.py#L32-L60

I'm asking the question, rather than saying we definitely should, just because I'm actually not sure how much better one approach or the other is.

maas/plugins/galera_check.py
```diff
@@ -90,6 +90,22 @@ def print_metrics(replica_status):
            replica_status['Threads_connected'], 'connections')
     metric('mysql_max_seen_connections', 'int64',
            replica_status['Max_used_connections'], 'connections')
+    metric('num_of_open_files', 'int64',
```
mancdaz: I wonder, with this many metrics, whether we might be best creating a mapping dict, mapping the parameter name obtained from lysql, to the metric name, along with the units. A bit like here:

https://github.com/rcbops/rpc-openstack/blob/009ae7d2b4751d52d012987a111ad206ba935502/maas/plugins/nova_cloud_stats.py#L32-L60

I'm asking the question, rather than saying we definitely should, just because I'm actually not sure how much better one approach or the other is.

rpcd/playbooks/roles/rpc_maas/templates/galera_check.yaml.j2
```diff
@@ -34,3 +34,54 @@ alarms      :
...
+            :set consecutiveCount={{ maas_alarm_local_consecutive_count }}
+            if (metric["num_of_open_files"] > metric["open_files_limit"]) {
+                return new AlarmStatus(CRITICAL, "Mysql open files has exceeded open files limit");
+            }
+    innodb_row_lock_time_avg :
+        label                   : innodb_row_lock_time_avg--{{ ansible_hostname }}
+        notification_plan_id    : "{{ maas_notification_plan }}"
+        criteria                : |
+            :set consecutiveCount={{ maas_alarm_local_consecutive_count }}
+            if (metric["innodb_row_lock_time_avg"] > metric["innodb_row_lock_time_max"]) {
```
andymcc: is this even possible? I think this should be innodb_row_lock_time_avg > var_threshold perhaps

andymcc:
One additional note. we're missing one check that was requested:

Max_used_connections in relation to max_connection setting

Currently we check current number of connections vs max connection setting, but not max_used_connections.

rpcd/playbooks/roles/rpc_maas/templates/galera_check.yaml.j2
alextricity25: You're right. I was thinking this criteria could catch any cases where it took an unusual amount of time getting the lock, but I do realize now that innodb_row_lock_time_max is maximum time in milliseconds spent getting an InnoDB row lock, which could be a value just *slightly* above the average at any given time.

I like your suggestion, but we still have to decided on a good threshold to set by default.

andymcc: 2 options I can think of: 
- We could assign a value arbitrarily (since we have no real data to base a threshold decision off) - if it's a variable we can change it later/adjust the default (I'd defer to support for a decent idea).
- We could consider the rate of change as more important than the actual value itself - e.g. if the average lock time is increasing steadily then this is more indicative of an issue than the average lock time itself being high. We do this already in the rabbitmq checks: https://github.com/rcbops/rpc-openstack/blob/master/rpcd/playbooks/roles/rpc_maas/templates/rabbitmq_status.yaml.j2#L76 again we'd probably need supports input since this is a request directly from them.

alextricity25:
blocked until I can figure out why https://github.com/rcbops/rpc-openstack/pull/1242/files#diff-50fff0dae21040cc684fefc936d5f01bR42 keep putting that alarm in the CRITICAL state

alextricity25:
Turns out it was just a typo on my end.
I've removed the status-blocked/dont-merge and re-added the status-needs-review labels.

maas/plugins/galera_check.py
alextricity25: I honestly don't think this is needed. All of the mysql status variables are already loaded into a list here:
https://github.com/rcbops/rpc-openstack/blob/master/maas/plugins/galera_check.py#L111-L112

maas/plugins/README.md
alextricity25: Done.

rpcd/playbooks/roles/rpc_maas/templates/galera_check.yaml.j2
alextricity25: I set two thresholds for now:
https://github.com/rcbops/rpc-openstack/pull/1242/files#diff-1d0825392161b9f338fa4db5ddd23713R244

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Liberty-PR: Fail

Apsu:
RPC-AIO-Liberty-PR: Fail

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Liberty-PR: Fail

alextricity25:
19:17:22     tempest.lib.exceptions.ServerFault: Got server fault
19:17:22     Details: Unexpected API Error. Please report this at http://bugs.launchpad.net/nova/ and attach the Nova API log if possible.
19:17:22     <class 'neutronclient.common.exceptions.PortNotFoundClient'>

recheck_swift
recheck_ceph

Apsu:
RPC-AIO-Liberty-PR: Fail

alextricity25:
tempest.scenario.test_server_basic_ops.TestServerBasicOps.test_server_basic_ops[compute,id-7fff3fb3-91d8-4fd0-bd7d-0204f1f180ba,network,smoke]

recheck_swift

Apsu:
RPC-AIO-Liberty-PR: Fail

alextricity25:
tempest.scenario.test_server_basic_ops.TestServerBasicOps.test_server_basic_ops[compute,id-7fff3fb3-91d8-4fd0-bd7d-0204f1f180ba,network,smoke]

recheck_swift

rpcd/playbooks/roles/rpc_maas/templates/galera_check.yaml.j2
```diff
@@ -34,3 +34,65 @@ alarms      :
             if (percentage(metric["mysql_current_connections"], metric["mysql_max_configured_connections"]) > {{ mysql_connection_warning_threshold }} ) {
                 return new AlarmStatus(WARNING, "Mysql connectons has exceeded {{ mysql_connection_warning_threshold }}% of configured maximum");
             }
+    mysql_max_used_connections :
+        label                   : mysql_max_used_connections--{{ ansible_hostname }}
+        notification_plan_id    : "{{ maas_notification_plan }}"
+        criteria                : |
+            :set consecutiveCount={{ maas_alarm_local_consecutive_count }}
+            if (metric["mysql_max_seen_connections"] > metric["mysql_max_configured_connections"]) {
```
andymcc: I think this is similar to the previous one in that it isn't really possible for max_seen connections to be higher than max_connections, at best it can be the same.

What we probably want is more like the "percentage_used_mysql_connections" check which uses percentage - e.g. seen_connections as a percentage of max_connections, vs a threshold that is defined as a var (e.g. 95%).
That would probably be quite conservative since max_seen_connections never goes down until you flush status.

alextricity25: Thanks @andymcc  for the input. I was going off what Bjoern said in one of the bug report comments:
"Basically current number of sessions , max_used_connections < max_connections limit".
According to the documentation:
Max_used_connections -  The maximum number of connections that have been in use simultaneously since the server started. 
max_connections - The maximum permitted number of simultaneous client connections

Also in the documentation it mentions the possiblity of a "Too many connections error"[1], which I presume is what we are trying to monitor, right @BjoernT?

[1] http://dev.mysql.com/doc/refman/5.7/en/too-many-connections.html

Thoughts?

alextricity25: Still I think you're right that Max_used_connections can never be higher than max_connections. 
I'll do a percentage and check against a threshold then submit another commit shortly.

BjoernT: Well it can if someone is lowering max_connections for a reason. It is not guaranteed that max_used_connections are never higher than max_connections

andymcc: @BjoernT yes, of course, but are you saying you want it to be a straight comparison? Its a support request, so if that is what you want Alex's patch does that - it seems less useful if the only criteria for alarming is if somebody has lowered the max_connections var and not done a flush status.

alextricity25: @BjoernT, @andymcc brings in a good point, but if this is what you are requesting, then I can keep the PR the way it is. Let me know what you decide on and I'll take action.
Thanks.

alextricity25:
Waiting for @BjoernT's feedback. Other than that, this PR is ready for review

maas/plugins/galera_check.py
```diff
@@ -74,8 +74,8 @@ def print_metrics(replica_status):
            replica_status['wsrep_commit_window'], 'sequence_delta')
     metric('wsrep_cluster_size', 'int64',
            replica_status['wsrep_cluster_size'], 'nodes')
-    metric('queries_per_second', 'int64',
-           replica_status['Queries'], 'qps')
+    metric('queries', 'int64',
```
sigmavirus24: Removing this metric might affect some of the people who use our MaaS metrics. @mancdaz do we usually straight off just rename something or do we report two metrics for backwards compatibility?

claco: I'd say we should be true to what we would do for incompat changes. 'Deprecate then Delete' cycle. So provide both now, and document the change for consumers. file a bug for Newton, remove the old one then.

alextricity25: @sigmavirus24 @claco 
That's a good point. I honestly didn't consider anyone was using these metrics outside of what we have in place in the alarm's criteria. But even if they are using "queries_per_second" and I leave it in there for backwards compatibility, it still is kind of a misleading name( the metric reports number of queries total, not queries_per_second). I'm afraid that the mis-interpretation of the value might cause issues for anyone trying to utilize the metric. The sooner we make it unavailable, the better IMO.
What do you guys think about keeping the name change(and leaving queries_per_second out), then adding a release note documenting the change?

BjoernT:
In regards to max_used_connections we just want to get an alarm when it is close to max_connections. We don't have to report necessarily percentage metrics, I would probably prefer numbers so we can see how the numbers changed once we increase max_connections over a history.

In regards to queries metric, yes we should fix that and convert it into a queries per second. 
Reporting overall number make no sense IMHO. Just highlight it in the release notes

sigmavirus24:
> In regards to queries metric, yes we should fix that and convert it into a queries per second.
Reporting overall number make no sense IMHO. Just highlight it in the release notes

We can't actually report a per-second thing with the plugins. We don't get that metric from Galera and the plugin is run at the intervals that the monitoring agent is configured to run it. That can, however, be created in the intelligence portal. Further, support isn't the only consumer of the metrics, so I would *rather* have a deprecation cycle for the metric as @claco described.

Apsu:
RPC-AIO-Liberty-PR: Fail

Apsu:
RPC-AIO-Liberty-PR: Fail

alextricity25:
I've opened a separate bug for the queries_per_second issue so we don't mix things up and for tracking purposes.
https://github.com/rcbops/rpc-openstack/issues/1254
PR for this fix to come..

b3rnard0:
@alextricity25 Do we need to triage/prioritize 1254? How severe is it and does it block you completing #1242?

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

alextricity25:
@b3rnard0 #1254 Won't block this one, but it does need to be prioritized and triaged.

mancdaz:
If you alarm on a mysql status variable that is essentially a high tide mark, that alarm will never clear. 
Better to alarm on current connection average over a number of polls (this can be expressed in the alarm language).

BjoernT:
If we can separate alarm from metric we can leave the alarm at percentage, the actual metric is preferred to be reported as counter rather than percentage

sigmavirus24:
> If we can separate alarm from metric we can leave the alarm at percentage, the actual metric is preferred to be reported as counter rather than percentage

I believe that's what was being proposed this entire time.

alextricity25:
I've updated the PR  so the alarm for max_seen_connections compares against a percentage threshold, using max_connections to derive the percentage [1]
The max_seen_connections(maps to the max_used_connections mysql status variable) metric will stay, so it can still be monitored via the RAX intelligence dashboard as requested by @BjoernT.
@sigmavirus24 @andymcc @BjoernT Could you guys take another look to review when you get the chance? Thanks!

[1] https://github.com/rcbops/rpc-openstack/pull/1242/files#diff-50fff0dae21040cc684fefc936d5f01bR42

mancdaz:
I'm still not convinced that using a high tide mark to alert on is a good thing. Once that goes in to alert, it will not clear without manual intervention (by either restarting mariadb, or logging in and manually resetting that status variable in mysql).  I would argue that you would be best using the alarm language to calculate the average current connections over the last n polls, and alert when that average breaches a percentage of max_connections.

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Liberty-PR: Fail

Apsu:
RPC-AIO-Liberty-PR: Fail

andymcc:
I agree with @mancdaz - @BjoernT is that your intention for support to have to take manual action for the max_seen_connections alert every time, since once it alerts it will never clear until a flush status or some other manual action is taken (e.g. service restart).

This seems like a poor alert mechanic - it seems more sensible for it to clear automatically if it has been seen and dealt with, and would really generate an alert in the same way as the max_connections vs current_connections (except it would never clear) - since the max_connections vs current_connections alarm is already setup I'm not sure what this would add.

Saying that - if support deem this to be important I don't have an issue approving the PR.

BjoernT:
This is not an issue for us, if max_used_connections hits to close to max_connections we have to do something anyway. Either rising the limit or flush the status if it was a problem. I'm ok with the alarm being critical in that case

andymcc:
Gate looks to be failing because the access denied_errors is already above 0 after build, so it starts off in error - maas-verify fails.

Unless we want to clear the status of all those vars before setting up MaaS (which we should definitely not do at all), we'll need a new approach to that (e.g. setting it a lot higher, which also seems like a very bad approach).

@BjoernT what issues are we aiming to detect by seeing the access_denied_errors? Surely this implies some service is unable to access mysql correctly - so we should be seeing issues with the service rather than relying on mysql access_denied errors.

BjoernT:
Access denied should always be 0 if not then this is a hint that a service is misconfigured.
I would not say that we can expect a service error in this case, this check would just add another layer of monitoring to monitor the database rather than just the service side

andymcc:
@BjoernT the essence of the problem is that the counter doesn't start at 0, so it is already in alert - failing the gate (and alerting straight away).

Additionally flush status won't reset that - in both cases support will need to make changes that aren't related to the error. E.g. error causes max connections/denied errors to generate - support resolves the issue, but alert remains. Ticket doesn't clear, and if nothing further is done then either no further alerts will happen (if a subsequent issue occurs), or it will continue to create alerts.

I'm not sure I understand what this gives additional to what is already setup (max_connection check will do the exact same except it will clear when the issue is resolved, and access_denied is a flat counter that could fire as a result of pretty much any misconfigured client connection - plus it lets you know of a client side server issue, which seems to be best found on the client itself).

BjoernT:
Why doesn't it start as 0 it does for me, it should be 0. Do we know why it is not 0 in gating ?

mancdaz:
@BjoernT somebody trying to access mysql from a client and maybe mistyping a password is going to increase this counter. Surely this is not a sensible thing to be alerting on?

I think anywhere we're trying to alert on high tide marks or cumulative counters is simply doing it wrong.

BjoernT:
Yes that's why I asked to monitor rates not counter so we know if counter is growing. If denied counter is already at >0 then we can take it off if we are not able to monitor rates.

andymcc:
A rate would be ok - @alextricity25 can you change that to be a rate of increase of the access_denied_errors value.

I think we should just remove the max_connections_seen alarm though.

alextricity25:
Made new commit:
* Added release note
* changed access_denied_errors alarm criteria to use the rate function modifier
* added thresholds for the access_denied_errors alarm in defaults/main.yml
* removed the mysql_seen_connections alarm

Apsu:
RPC-AIO-Liberty-PR: Fail

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Liberty-PR: Fail

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Liberty-PR: Fail

alextricity25:
recheck_all

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Liberty-PR: Fail

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

alextricity25:
18:18:00 Captured traceback:
18:18:00 ~~~~~~~~~~~~~~~~~~~
18:18:00     Traceback (most recent call last):
18:18:00       File "/opt/tempest_c2c90574a78c7704f06e8fac279c4f1d11365993/tempest/scenario/manager.py", line 108, in delete_wrapper
18:18:00         delete_thing(*args, **kwargs)
18:18:00       File "/opt/tempest_c2c90574a78c7704f06e8fac279c4f1d11365993/tempest/lib/services/compute/floating_ips_client.py", line 62, in delete_floating_ip
18:18:00         resp, body = self.delete(url)
18:18:00       File "/opt/tempest_c2c90574a78c7704f06e8fac279c4f1d11365993/tempest/lib/common/rest_client.py", line 290, in delete
18:18:00         return self.request('DELETE', url, extra_headers, headers, body)
18:18:00       File "/opt/tempest_c2c90574a78c7704f06e8fac279c4f1d11365993/tempest/lib/services/compute/base_compute_client.py", line 53, in request
18:18:00         method, url, extra_headers, headers, body)
18:18:00       File "/opt/tempest_c2c90574a78c7704f06e8fac279c4f1d11365993/tempest/lib/common/rest_client.py", line 645, in request
18:18:00         resp, resp_body)
18:18:00       File "/opt/tempest_c2c90574a78c7704f06e8fac279c4f1d11365993/tempest/lib/common/rest_client.py", line 764, in _error_checker
18:18:00         message=message)
18:18:00     tempest.lib.exceptions.ServerFault: Got server fault
18:18:00     Details: Unexpected API Error. Please report this at http://bugs.launchpad.net/nova/ and attach the Nova API log if possible.
18:18:00     <class 'neutronclient.common.exceptions.PortNotFoundClient'>

recheck_ceph
recheck_swift

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

ameserole:
19:26:43 TASK: [Verify Checks & Alarms are registered] ********************************* 
19:45:18 failed: [jrpcaioiad-123] => {"attempts": 8, "changed": false, "cmd": ". /openstack/venvs/maas-UNKNOWN_RELEASE/bin/activate\n /opt/rpc-openstack/scripts/rpc-maas-tool.py verify-created --entity jrpcaioiad-123", "delta": "0:00:44.090162", "end": "2016-07-25 19:45:17.061452", "failed": true, "failed_when_result": true, "rc": 1, "start": "2016-07-25 19:44:32.971290", "stdout_lines": ["--- 2016-07-25 19:45:17.047960 ---", "The following checks have config files but are not registered with the maas api: ['lb_api_check_nova--unspecified_lb'] ", "The following alarms are present in config files but are not registered with the maas api: ['lb_api_alarm_nova'] "], "warnings": []}
19:45:18 stdout: --- 2016-07-25 19:45:17.047960 ---
19:45:18 The following checks have config files but are not registered with the maas api: ['lb_api_check_nova--unspecified_lb'] 
19:45:18 The following alarms are present in config files but are not registered with the maas api: ['lb_api_alarm_nova'] 
19:45:18 msg: Task failed as maximum retries was encountered
19:45:18 
19:45:18 FATAL: all hosts have already failed -- aborting

alextricity25:
recheck_ceph

rpcd/playbooks/roles/rpc_maas/templates/galera_check.yaml.j2
```diff
@@ -34,3 +34,57 @@ alarms      :
             if (percentage(metric["mysql_current_connections"], metric["mysql_max_configured_connections"]) > {{ mysql_connection_warning_threshold }} ) {
                 return new AlarmStatus(WARNING, "Mysql connectons has exceeded {{ mysql_connection_warning_threshold }}% of configured maximum");
             }
+    open_file_size_limit_reached :
+        label                   : open_file_size_limit_reached--{{ ansible_hostname }}
+        notification_plan_id    : "{{ maas_notification_plan }}"
+        criteria                : |
+            :set consecutiveCount={{ maas_alarm_local_consecutive_count }}
+            if (metric["num_of_open_files"] > metric["open_files_limit"]) {
```
andymcc: This is another one that I don't think can be possible - sorry I should've picked it up earlier.
Can we change this to a percentage metric?

rpcd/playbooks/roles/rpc_maas/templates/galera_check.yaml.j2
```diff
@@ -34,3 +34,57 @@ alarms      :
...
+            }
+            if (rate(metric["access_denied_errors"]) > {{ mysql_access_denied_errors_rate_critical_threshold }}) {
+                return new AlarmStatus(CRITICAL, "Access denied errors rate is greater than {{ mysql_access_denied_errors_rate_critical_threshold }}");
+            }
+    aborted_clients :
+        label                   : aborted_clients--{{ ansible_hostname }}
+        notification_plan_id    : "{{ maas_notification_plan }}"
+        criteria                : |
+            :set consecutiveCount={{ maas_alarm_local_consecutive_count }}
+            if (rate(metric["aborted_clients"]) > 1) {
```
andymcc: I think we should make these variables.
1 Sounds fine, however if it is too much/too little a variable will allow support to edit this on the fly.

Additionally, I can't remember exactly how the rate is calculated, e.g. is it per second or is it based on a change since last poll (which would mean its per minute) - which might be worth looking into!

(Same for the below check)

andymcc:
I know this has gone on for a bit - but 2 more changes! Hopefully the last 2.

rpcd/playbooks/roles/rpc_maas/templates/galera_check.yaml.j2
alextricity25: Good catch!

rpcd/playbooks/roles/rpc_maas/templates/galera_check.yaml.j2
alextricity25: According to the MaaS docs:
Rate as defined by this formula where V=value, and T=time:
(V1 - V0) / (T1 - T0).
To me, this seems to be a rate derived on a per check basis (per minute, by default). Should be fine, IMO.

alextricity25:
Updated commit:
- Added threshold variables for aborted_connects and aborted_clients
- Changed the open_files alarm criteria to check against a percentage and a defined threshold.

andymcc:
👍  LGTM - assuming the gate passes.

Thanks for making those changes!

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Liberty-PR: Fail

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Liberty-PR: Fail

ameserole:
18:13:12 TASK: [Verify Checks & Alarms are registered] ********************************* 
18:30:42 failed: [jrpcaioiad-740] => {"attempts": 8, "changed": false, "cmd": ". /openstack/venvs/maas-UNKNOWN_RELEASE/bin/activate\n /opt/rpc-openstack/scripts/rpc-maas-tool.py verify-created --entity jrpcaioiad-740", "delta": "0:01:04.209792", "end": "2016-07-26 18:30:41.391258", "failed": true, "failed_when_result": true, "rc": 1, "start": "2016-07-26 18:29:37.181466", "stdout_lines": ["--- 2016-07-26 18:30:41.374708 ---", "The following alarms are present in config files but are not registered with the maas api: ['aborted_connects--jrpcaioiad-740-galera-container-7e4b42f3'] "], "warnings": []}
18:30:42 stdout: --- 2016-07-26 18:30:41.374708 ---
18:30:42 The following alarms are present in config files but are not registered with the maas api: ['aborted_connects--jrpcaioiad-740-galera-container-7e4b42f3'] 
18:30:42 msg: Task failed as maximum retries was encountered
18:30:42 
18:30:42 FATAL: all hosts have already failed -- aborting

recheck_all

Also going to clean up some of the Apsu bot comments

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Liberty-PR: Fail

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

alextricity25:
recheck_ceph

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

alextricity25:
tempest.scenario.test_server_basic_ops.TestServerBasicOps.test_server_basic_ops[compute,id-7fff3fb3-91d8-4fd0-bd7d-0204f1f180ba,network,smoke]
recheck_ceph

andymcc:
👍 passing gate now.

rpcd/playbooks/roles/rpc_maas/defaults/main.yml
```diff
@@ -236,6 +236,36 @@ maas_filesystem_critical_threshold: 90.0
...
+# In other words, this threshold defines how many
+# aborted_connects per check are needed to trigger
+# an alarm
+mysql_aborted_connects_rate_warning_threshold: 1
+mysql_aborted_connects_rate_critical_threshold: 2
+
+# Maas mysql open file percentage threshold
+mysql_open_files_precentage_warning_threshold: 90
+mysql_open_files_precentage_critical_threshold: 95
+
```
mancdaz: precentage?

rpcd/playbooks/roles/rpc_maas/templates/galera_check.yaml.j2
```diff
@@ -34,3 +34,66 @@ alarms      :
...
                 return new AlarmStatus(WARNING, "Mysql connectons has exceeded {{ mysql_connection_warning_threshold }}% of configured maximum");
             }
+    open_file_size_limit_reached :
+        label                   : open_file_size_limit_reached--{{ ansible_hostname }}
+        notification_plan_id    : "{{ maas_notification_plan }}"
+        criteria                : |
+            :set consecutiveCount={{ maas_alarm_local_consecutive_count }}
+            if (percentage(metric["num_of_open_files"], metric["open_files_limit"]) > {{ mysql_open_files_precentage_critical_threshold }}) {
+                return new AlarmStatus(CRITICAL, "Number of open files has exceeded {{ mysql_open_files_precentage_critical_threshold }}% of open file limit");
+            }
```
mancdaz: precentage?

rpcd/playbooks/roles/rpc_maas/templates/galera_check.yaml.j2
```diff
@@ -34,3 +34,66 @@ alarms      :
...
+                return new AlarmStatus(WARNING, "Number of open files has exceeded {{ mysql_open_files_precentage_warning_threshold }}% of open file limit");
+            }
+    innodb_row_lock_time_avg :
+        label                   : innodb_row_lock_time_avg--{{ ansible_hostname }}
+        notification_plan_id    : "{{ maas_notification_plan }}"
+        criteria                : |
+            :set consecutiveCount={{ maas_alarm_local_consecutive_count }}
+            if (metric["innodb_row_lock_time_avg"] > {{ innodb_row_lock_time_avg_critical_threshold }}) {
+                return new AlarmStatus(CRITICAL, "InnoDB row lock time average has exceeded {{ innodb_row_lock_time_avg_critical_threshold }}");
+            }
```
mancdaz: you need to add the units to the end of the message

rpcd/playbooks/roles/rpc_maas/templates/galera_check.yaml.j2
```diff
@@ -34,3 +34,66 @@ alarms      :
             if (percentage(metric["mysql_current_connections"], metric["mysql_max_configured_connections"]) > {{ mysql_connection_warning_threshold }} ) {
                 return new AlarmStatus(WARNING, "Mysql connectons has exceeded {{ mysql_connection_warning_threshold }}% of configured maximum");
             }
+    open_file_size_limit_reached :
+        label                   : open_file_size_limit_reached--{{ ansible_hostname }}
+        notification_plan_id    : "{{ maas_notification_plan }}"
+        criteria                : |
+            :set consecutiveCount={{ maas_alarm_local_consecutive_count }}
+            if (percentage(metric["num_of_open_files"], metric["open_files_limit"]) > {{ mysql_open_files_precentage_critical_threshold }}) {
+                return new AlarmStatus(CRITICAL, "Number of open files has exceeded {{ mysql_open_files_precentage_critical_threshold }}% of open file limit");
```
mancdaz: precentage

mancdaz: precentage

rpcd/playbooks/roles/rpc_maas/templates/galera_check.yaml.j2
```diff
@@ -34,3 +34,66 @@ alarms      :
...
+        label                   : innodb_row_lock_time_avg--{{ ansible_hostname }}
+        notification_plan_id    : "{{ maas_notification_plan }}"
+        criteria                : |
+            :set consecutiveCount={{ maas_alarm_local_consecutive_count }}
+            if (metric["innodb_row_lock_time_avg"] > {{ innodb_row_lock_time_avg_critical_threshold }}) {
+                return new AlarmStatus(CRITICAL, "InnoDB row lock time average has exceeded {{ innodb_row_lock_time_avg_critical_threshold }}");
+            }
+            if (metric["innodb_row_lock_time_avg"] > {{ innodb_row_lock_time_avg_warning_threshold }}) {
+                return new AlarmStatus(WARNING, "InnoDB row lock time average has exceeded {{ innodb_row_lock_time_avg_warning_threshold }}");
+            }
```
mancdaz: you need to add the units to the end of the message

rpcd/playbooks/roles/rpc_maas/templates/galera_check.yaml.j2
```diff
@@ -34,3 +34,66 @@ alarms      :
...
+        label                   : aborted_clients--{{ ansible_hostname }}
+        notification_plan_id    : "{{ maas_notification_plan }}"
+        criteria                : |
+            :set consecutiveCount={{ maas_alarm_local_consecutive_count }}
+            if (rate(metric["aborted_clients"]) > {{ mysql_aborted_clients_rate_warning_threshold }}) {
+                return new AlarmStatus(WARNING, "Aborted clients rate is greater than {{ mysql_aborted_clients_rate_warning_threshold }}");
+            }
+            if (rate(metric["aborted_clients"]) > {{ mysql_aborted_clients_rate_critical_threshold }}) {
+                return new AlarmStatus(CRITICAL, "Aborted clients rate is greater than {{ mysql_aborted_clients_rate_critical_threshold }}");
+            }
```
mancdaz: same

rpcd/playbooks/roles/rpc_maas/templates/galera_check.yaml.j2
```diff
@@ -34,3 +34,66 @@ alarms      :
...
+    aborted_connects :
+        label                   : aborted_connects--{{ ansible_hostname }}
+        notification_plan_id    : "{{ maas_notification_plan }}"
+        criteria                : |
+            :set consecutiveCount={{ maas_alarm_local_consecutive_count }}
+            if (rate(metric["aborted_connects"]) > {{ mysql_aborted_connects_rate_warning_threshold }}) {
+                return new AlarmStatus(WARNING, "Aborted connects rate is greater than {{ mysql_aborted_connects_rate_warning_threshold }}");
+            }
```
mancdaz: same

rpcd/playbooks/roles/rpc_maas/templates/galera_check.yaml.j2
```diff
@@ -34,3 +34,66 @@ alarms      :
...
+            if (rate(metric["aborted_connects"]) > {{ mysql_aborted_connects_rate_critical_threshold }}) {
+                return new AlarmStatus(CRITICAL, "Aborted connects rate is greater than {{ mysql_aborted_connects_rate_critical_threshold }} ");
+            }
```
mancdaz: same

rpcd/playbooks/roles/rpc_maas/templates/galera_check.yaml.j2
```diff
@@ -34,3 +34,66 @@ alarms      :
...
+                return new AlarmStatus(CRITICAL, "InnoDB has experienced a deadlock");
+            }
+    access_denied_errors :
+        label                   : access_denied_errors--{{ ansible_hostname }}
+        notification_plan_id    : "{{ maas_notification_plan }}"
+        criteria                : |
+            :set consecutiveCount={{ maas_alarm_local_consecutive_count }}
+            if (rate(metric["access_denied_errors"]) > {{ mysql_access_denied_errors_rate_warning_threshold }}) {
+                return new AlarmStatus(WARNING, "Access denied errors rate is greater than {{ mysql_access_denied_errors_rate_warning_threshold }}");
+            }
```
mancdaz: greater than x what? Need to clarify it's greater than x errors in poll period y

rpcd/playbooks/roles/rpc_maas/templates/galera_check.yaml.j2
```diff
@@ -34,3 +34,66 @@ alarms      :
...
+            if (rate(metric["access_denied_errors"]) > {{ mysql_access_denied_errors_rate_critical_threshold }}) {
+                return new AlarmStatus(CRITICAL, "Access denied errors rate is greater than {{ mysql_access_denied_errors_rate_critical_threshold }}");
+            }
```
mancdaz: same

rpcd/playbooks/roles/rpc_maas/templates/galera_check.yaml.j2
```diff
@@ -34,3 +34,66 @@ alarms      :
...
+    aborted_clients :
+        label                   : aborted_clients--{{ ansible_hostname }}
+        notification_plan_id    : "{{ maas_notification_plan }}"
+        criteria                : |
+            :set consecutiveCount={{ maas_alarm_local_consecutive_count }}
+            if (rate(metric["aborted_clients"]) > {{ mysql_aborted_clients_rate_warning_threshold }}) {
+                return new AlarmStatus(WARNING, "Aborted clients rate is greater than {{ mysql_aborted_clients_rate_warning_threshold }}");
+            }
```
mancdaz: same

mancdaz:
👎 based on comments inline

rpcd/playbooks/roles/rpc_maas/defaults/main.yml
alextricity25: Per @andymcc 's observation: 
> This is another one that I don't think can be possible - sorry I should've picked it up earlier.
   Can we change this to a percentage metric (with variable for threshold)?

rpcd/playbooks/roles/rpc_maas/templates/galera_check.yaml.j2
alextricity25: See above note

rpcd/playbooks/roles/rpc_maas/templates/galera_check.yaml.j2
alextricity25: I'm not sure what kind of units you want for this one, but I could add "per check" at the end of the sentence.

rpcd/playbooks/roles/rpc_maas/templates/galera_check.yaml.j2
alextricity25: same :)

rpcd/playbooks/roles/rpc_maas/templates/galera_check.yaml.j2
alextricity25: I'm not sure what kind of units you want for this one, but I could add "per check" at the end of the sentence and get rid of "rate"

rpcd/playbooks/roles/rpc_maas/defaults/main.yml
mancdaz: it's the spelling - you've got pre-centage everywhere

git-harry:
Removed labels because the connected issue is used to track these pieces of information.

alextricity25:
UPDATE: The latest commit makes the changes @mancdaz pointed out.

alextricity25:
This PR is ready for another look @mancdaz, @andymcc, @BjoernT.

cloader89:
The code looks good and no errors that I saw during AIO build. 👍

mancdaz:
👍

alextricity25:
liberty12-2 backport here: https://github.com/rcbops/rpc-openstack/pull/1269

PR #1245, opened 2016-07-19
Update elasticsearch to latest version

d34dh0r53:
This updates elasticsearch to it latest version to the current version for the Mitaka work. This also removes the site plugins that we were using in the legacy stack as they have been deprecated in this version of elasticsearch. I've removed the mappings file as it was causing errors and the new version of elasticsearch is doing a good job of dynamically configuring the mappings.

Connects rcbops/u-suk-dev#176


Apsu:
RPC-AIO-Liberty-PR: Fail

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Liberty-PR: Fail

d34dh0r53:
recheck_swift
Compute failure

rpcd/playbooks/roles/elasticsearch/tasks/elasticsearch_post_install.yml
```diff
@@ -17,8 +17,8 @@
   template:
     src: "{{ item }}"
     dest: "/etc/elasticsearch/{{ item }}"
-    owner: "root"
-    group: "root"
+    owner: "elasticsearch"
```
sigmavirus24: Should we convert these to role defaults too?

sigmavirus24:
One question, otherwise, LGTM.

rpcd/playbooks/roles/elasticsearch/tasks/elasticsearch_post_install.yml
prometheanfire: yes, though given we only consume here...

andymcc:
Greenfields works ok - are we concerned about upgrades at all? i believe those won't work due to plugins (which we can remove in the upgrade script easily enough), and (potentially more tricky) there are incorrect logstash filters that prevent elasticsearch from starting because of type mismatch between ip and string (I believe this is fields that are labeled as both ip and string in different places). The second one may require a crafty solution wrt dumping/reimporting the data.

That's probably ok - we have to sort out the filters etc anyway - but as with the kibana PR my concern is that we are relying on future tasks that HAVE to happen in order for this to work properly - we don't really have any gate checks that test any functionality so that these pass the gate is mostly irrelevant (they deploy fine).

Otherwise this LGTM.

sigmavirus24:
@andymcc this is the work for greenfields only. When we start working on 13.1 we'll start bothering about upgrades. :)

claco:
Question: How do we not lose track of this additional work in 13.1? Do we ensure we have a card on the board now for the milestone to 'review' this area of code?

sigmavirus24:
There were rcbops/u-suk-dev#181 and rcbops/u-suk-dev#182 and I just added rcbops/u-suk-dev#237 to cover this specifically.

d34dh0r53:
recheck_swift

Apsu:
RPC-AIO-Liberty-PR: Fail

prometheanfire:
I'm :+1: on this

rpcd/playbooks/roles/elasticsearch/tasks/elasticsearch_post_install.yml
mancdaz: For now.....but the correct pattern as @sigmavirus24 suggested is to convert to role defaults.

rpcd/playbooks/roles/elasticsearch/tasks/elasticsearch_install.yml
```diff
@@ -42,50 +42,6 @@
...
-  command: ./plugin -install {{ item }}
-  args:
-    chdir: /usr/share/elasticsearch/bin
-    creates: /usr/share/elasticsearch/plugins/bigdesk
-  with_items:
-    - lukas-vlcek/bigdesk/2.4.0
-  tags:
-    - elasticsearch-plugins
-    - elasticsearch-install
-
```
mancdaz: how come we don't need any of these plugins now?

rpcd/playbooks/roles/elasticsearch/tasks/elasticsearch_post_install.yml
```diff
@@ -48,21 +48,6 @@
...
-    group: "root"
-    mode: "0755"
-  tags:
-    - elasticsearch-post-install
-
-- name: Run ElasticSearch schema mapping script
-  shell: /opt/mapping.sh
-  tags:
-    - elasticsearch-post-install
-
```
mancdaz: why is this not needed now?

mancdaz:
The commit message needs a bit more detail explaining why we're removing a bunch of the install steps. I'm sure there's a perfectly good reason as to why they aren't needed in the new version of elasticsearch, but we need to explain in the commit.

d34dh0r53:
@mancdaz I have updated the commit message and the initial comment to reflect the fact that site plugins are deprecated in this version of elasticsearch.  The reason for the mapping removal is that the new version of elasticsearch seems to do a really good job of type mapping, but we may still need to add some custom mappings once the filter work begins.

rpcd/playbooks/roles/elasticsearch/tasks/elasticsearch_post_install.yml
d34dh0r53: This mapping file actually causes elasticsearch to not start, and it seems that the new version of elasticsearch does a really good job of dynamic mapping.  We may need to add some custom mappings once the filter work begins but for now it is not needed.

rpcd/playbooks/roles/elasticsearch/tasks/elasticsearch_install.yml
d34dh0r53: Site plugins are deprecated in `2.3.x` and will be removed in the next major.

Apsu:
RPC-AIO-Liberty-PR: Fail

d34dh0r53:
recheck_swift

rpcd/playbooks/roles/elasticsearch/tasks/elasticsearch_install.yml
evrardjp: Are the features not needed anymore or included in the default ES?

rpcd/playbooks/roles/elasticsearch/tasks/elasticsearch_post_install.yml
evrardjp: Agreed

evrardjp:
I think the commit message still need to explain the why - and how do we ensure feature continuity.

Except that, I'm good, but I haven't tested it.

d34dh0r53:
Commit message updated.

Apsu:
RPC-AIO-Liberty-PR: Fail

sigmavirus24:
So I have an AIO with this, #1245 and #1246 running and it seems to be working. Logstash is shipping things to elasticsearch and kibana is accessible and showing the items in queries out of elasticsearch.

So a 👍 having tested everything.

sigmavirus24:
recheck_swift

andymcc:
replicated what @sigmavirus24 has seen and this works with #1084 and #1246.
👍

Issue #1135, opened 2016-06-09
MaaS Horizon check fails when using custom branding


JCallicoat:
##### What version (commit) do you encounter this issue on?

r12.1.1

##### What type of issue is this (bug, feature, documentation, regression)?

Bug

##### What is the expected behavior?

The horizon check does not fail when using custom branding

##### What is actually happening?

The horizon check returns "could not load login page" when using custom branding.

##### What were you attempting to do when you encountered this error (give as much detail as possible)? 

Set `horizon_site_name` in user_variables.yml and run os-horizon-install.yml which will set `SITE_BRANDING` in /etc/horizon/local_settings.py, overriding the default `SITE_BRANDING = 'OpenStack Dashboard'` in /openstack_dashboard/settings.py.

The horizon check will do a regex search for 'openstack dashboard' in the response body and fail.

https://github.com/rcbops/rpc-openstack/blob/r12.1.1/maas/plugins/horizon_check.py#L61-L63

##### How can we reproduce this issue?

Follow steps above

##### How can we verify your issue is addressed?

Run horizon check and verify metrics are returned

##### What is your evaluation on the importance of this issue, and why?

Low impact, most customers do not use custom branding at this time

###### Remove branches (lines) you do not wish to target.
- [x] Master (#1255)
- [x] liberty-12.2 (#1264)
- [x] kilo (#1265)

b3rnard0:
- [ ] Reproduce the issue. (N/A)
- [ ] Create a PR for fix and/or docs.
- [ ] Incorporate feedback from pull request reviews.
- [ ]  What is the gating for this on DEV and/or QE? Add related gating if appropriate. Also make sure current gating passes with this fix.
- [ ] Reach out to reporter to determine if the fix is resolving the issue.
- [ ] What are the upgrade impacts (if any).
- [ ] What are the backports are needed (if any).

ameserole:
UPDATE:
Running into some issues getting an r12.1.1 environment set up to try and reproduce the bug. Tomorrow we will see if we can get an r12.1.1 lab set up. We will try and reproduce it on master as well.

ameserole:
UPDATE:
We have reproduced the issues and are now working on a solution.

ameserole:
UPDATE:
Submitted a work in progress pull request. The fix has not been fully tested yet.

alextricity25:
Update: The pull request is ready for review.

[...]

PR #1255, opened 2016-07-21
MaaS Horizon Site Name Check

ameserole:
This commit fixes a bug in the horizon_check.py plugin for MaaS.
Previously in the horizon check plugin it would search for the
horizon site name using a hardcoded value. This causes the
MaaS check to fail if the horizon site name is overridden in
the user variables file.

This is fixed setting the default value for horizon_site_name
in the rpc_maas main.yml that will get overridden if
horizon_site_name is set in user variables. That value is
passed to the horizon check plugin through the
horizon_local_check.yaml.j2 template and used by the horizon
check plugin as an argument for site name regex search.

Connects #1135

### DIFF
--- a/maas/plugins/horizon_check.py
+++ b/maas/plugins/horizon_check.py
@@ -59,7 +59,7 @@ def check(args):
         is_up = False
     else:
         if not (r.ok and
-                re.search('openstack dashboard', r.content, re.IGNORECASE)):
+                re.search(args.site_name_regexp, r.content, re.IGNORECASE)):
             status_err('could not load login page')
 
         splash_status_code = r.status_code
@@ -111,5 +111,9 @@ if __name__ == "__main__":
         parser.add_argument('ip',
                             type=ipaddr.IPv4Address,
                             help='horizon dashboard IP address')
+        parser.add_argument('site_name_regexp',
+                            type=str,
+                            default='openstack dashboard',
+                            help='Horizon Site Name')
         args = parser.parse_args()
         main(args)
--- a/releasenotes/notes/horizon-site-name-62c749ec2ab8b0b3.yaml
+++ b/releasenotes/notes/horizon-site-name-62c749ec2ab8b0b3.yaml
@@ -1,0 +1,7 @@
+---
+features:
+  - Added a new optional argument to the MaaS plugin 
+    horizon_check.py that takes in the name of the 
+    horizon dashboard that is used to find the login 
+    page. The default value is set to 
+    'openstack dashboard'. 
--- a/rpcd/playbooks/roles/rpc_maas/defaults/main.yml
+++ b/rpcd/playbooks/roles/rpc_maas/defaults/main.yml
@@ -454,3 +454,7 @@ maas_venv_bin: "{{ maas_venv }}/bin"
 # Set this to enable or disable installing in a venv
 maas_venv_enabled: true
 
+#
+# Default horizon site name
+#
+horizon_site_name: "openstack dashboard"
--- a/rpcd/playbooks/roles/rpc_maas/templates/horizon_local_check.yaml.j2
+++ b/rpcd/playbooks/roles/rpc_maas/templates/horizon_local_check.yaml.j2
@@ -5,7 +5,7 @@ period      : "{{ maas_check_period }}"
 timeout     : "{{ maas_check_timeout }}"
 details     :
     file    : run_plugin_in_venv.sh
-    args    : ["{{ maas_plugin_dir }}horizon_check.py", "{{ ansible_ssh_host }}"]
+    args    : ["{{ maas_plugin_dir }}horizon_check.py", "{{ ansible_ssh_host }}", "{{ horizon_site_name }}"]
 alarms      :
     horizon_local_status :
         label                   : "horizon_local_status--{{ ansible_hostname }}"
--- a/maas/plugins/README.md
+++ b/maas/plugins/README.md
@@ -285,6 +285,11 @@ Checks the status of the horizon dashboard. First checks that the login page is 
 
 ##### Mandatory Arguments:
 IP address of service to test
+
+#### Optional Arguments:
+
+site_name_regexp: name of the horizon dashboard used to find the login page (default: 'openstack dashboard')
+
 ##### Example Output:
 
     metric splash_status_code uint32 200


maas/plugins/horizon_check.py
```diff
@@ -111,5 +111,8 @@ def main(args):
         parser.add_argument('ip',
                             type=ipaddr.IPv4Address,
                             help='horizon dashboard IP address')
-        args = parser.parse_args()
+        parser.add_argument('site_name', 
+                            type=str, 
+                            help='Horizon Site Name')
+	args = parser.parse_args()
```
sigmavirus24: Need to correct the indentation here so it matches the statements above and below (need one more space).

rpcd/playbooks/roles/rpc_maas/defaults/main.yml
```diff
@@ -454,3 +454,7 @@ maas_venv_bin: "{{ maas_venv }}/bin"
 # Set this to enable or disable installing in a venv
 maas_venv_enabled: true
 
+#
+# Default horizon site name
+#
+horizon_site_name: "openstack dashboard"
```
sigmavirus24: I didn't think of this before, but it's a regular expression so maybe name it something like `horizon_site_name_regexp` to be very clear how it's used?

rpcd/playbooks/roles/rpc_maas/templates/horizon_local_check.yaml.j2
```diff
@@ -5,7 +5,7 @@ period      : "{{ maas_check_period }}"
 timeout     : "{{ maas_check_timeout }}"
 details     :
     file    : run_plugin_in_venv.sh
-    args    : ["{{ maas_plugin_dir }}horizon_check.py", "{{ ansible_ssh_host }}"]
+    args    : ["{{ maas_plugin_dir }}horizon_check.py", "{{ ansible_ssh_host }}", "'{{ horizon_site_name }}'"]
```
sigmavirus24: Good call using the single quotes here. 👍

maas/plugins/horizon_check.py
ameserole: @sigmavirus24 that wasn't showing up in vim haha. Thanks for pointing that out.

rpcd/playbooks/roles/rpc_maas/defaults/main.yml
ameserole: @sigmavirus24 I agree

ameserole: @sigmavirus24 I agree but It is supposed to be overridden if horizon_site_name is set in user variables though so I think it should probably stay the same.

sigmavirus24: Fair

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Liberty-PR: Fail

Apsu:
RPC-AIO-Liberty-PR: Fail

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Liberty-PR: Fail

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Liberty-PR: Fail

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Liberty-PR: Fail

ameserole:
@alextricity25 The default value gets set in [1] and is overridden here [2] if set elsewhere. The idea right now is to set the default in the main.yml for rpc_maas so that if it is set elsewhere it should override the default in rpc_maas and be passed as an argument to the horizon maas plugin.

[1] https://github.com/openstack/horizon/blob/master/openstack_dashboard/settings.py#L48
[2] https://github.com/openstack/openstack-ansible-os_horizon/blob/7a0b2665fea077428d9c61cb2c01e080525e9c6a/templates/horizon_local_settings.py.j2#L719-L721

alextricity25:
We noticed the the wrapper script, run_plugins_in_venv.sh, invokes the plugins with the bash arguments not delimited with quotes. This could cause issues with multi-word bash arguments. The solution here was to use $@ with double quotes like so:
python2.7 "$@". We are going to go ahead and include that in this patch.

sigmavirus24:
> We are going to go ahead and include that in this patch.

No. That should be a separate bug and a separate pull request which blocks this one.

alextricity25:
This is being blocked by PR #1259

alextricity25:
This is ready for review.

maas/plugins/horizon_check.py
```diff
@@ -111,5 +111,8 @@ def main(args):
         parser.add_argument('ip',
                             type=ipaddr.IPv4Address,
                             help='horizon dashboard IP address')
+        parser.add_argument('site_name_regexp',
```
nrb: I'm not seeing where this variable is getting a value in this PR. Is it meant to be there for overides, but not set by default?

rpcd/playbooks/roles/rpc_maas/templates/horizon_local_check.yaml.j2
```diff
@@ -5,7 +5,7 @@ period      : "{{ maas_check_period }}"
 timeout     : "{{ maas_check_timeout }}"
 details     :
     file    : run_plugin_in_venv.sh
-    args    : ["{{ maas_plugin_dir }}horizon_check.py", "{{ ansible_ssh_host }}"]
+    args    : ["{{ maas_plugin_dir }}horizon_check.py", "{{ ansible_ssh_host }}", "{{ horizon_site_name }}"]
```
sigmavirus24: @palendae this is where that argument is provided to the script

alextricity25:
I wonder if we need a release note outlining the addition of the new argument in the horizon_check.py plugin. IMO it'll definitely be worth mentioning it somewhere. Also I believe the documentation needs updating[1] 

[1] https://github.com/rcbops/rpc-openstack/tree/master/maas/plugins#horizon_checkpy

maas/plugins/horizon_check.py
```diff
@@ -111,5 +111,8 @@ def main(args):
...
+                            type=str,
+                            help='Horizon Site Name')
```
mancdaz: Is it maybe worth having a default value of 'openstack dashboard' for this arg and making it optional? This way anyone already using this plugin won't have to make any changes to continue using it in the same way they already are (e.g. rpc-r)

ameserole: @mancdaz the default is set here [1] and can be overridden in one of the user variables files.

[1] https://github.com/rcbops/rpc-openstack/blob/d2b5efdafbd52ebc49c975f97d2fd25e4a80d1c6/rpcd/playbooks/roles/rpc_maas/defaults/main.yml#L457-L460

mancdaz: @ameserole that's the default that will get used if someone is using our templates to configure checks/alarms. Not if someone is using the plugin directly. We used to hardcode the name inside the plugin, and we have removed that now. We should add it back in as a default to this new plugin arg.

evrardjp:
Not tested but the wiring seems ok for me.
Wondering if a weird name in the horizon_site_name containing a valid regexp will cause failures, because it's not escaped. It's probably fine 'though.

andymcc:
Tested - works.
Ready to merge, pending gate checks.

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

ameserole:
Added a release note to the PR

Apsu:
RPC-AIO-Liberty-PR: Fail

releasenotes/notes/horizon-site-name-62c749ec2ab8b0b3.yaml
```diff
@@ -0,0 +1,6 @@
+---
+features:
+  - Added a new optional argument to the MaaS plugin 
+    horizon_check.py that takes in the name of the 
+    horizon dashboard that is used to find the login 
+    page. The default value is set to openstack dashboard.
```
mancdaz: Can you please quote openstack dashboard like this:

```The default value is set to 'openstack dashboard'```

andymcc:
@mancdaz - feel free to merge if you're satisfied with the change.
Although if we're having gating issues re: MaaS this probably won't pass gate just yet.

andymcc:
gonna merge - has 2 +1's and @mancdaz's concern has been addressed.

PR #1264, opened 2016-07-25
MaaS Horizon Site Name Check

ameserole:
This commit fixes a bug in the horizon_check.py plugin for MaaS.
Previously in the horizon check plugin it would search for the
horizon site name using a hardcoded value. This causes the
MaaS check to fail if the horizon site name is overridden in
the user variables file.

This is fixed setting the default value for horizon_site_name
in the rpc_maas main.yml that will get overridden if
horizon_site_name is set in user variables. That value is
passed to the horizon check plugin through the
horizon_local_check.yaml.j2 template and used by the horizon
check plugin as an argument for site name regex search.

Connects #1135

(cherry picked from commit dfeccf2a332382a0d9c9b82c84b476529fcf99e8)

### DIFF
--- a/maas/plugins/README.md
+++ b/maas/plugins/README.md
@@ -285,6 +285,11 @@ Checks the status of the horizon dashboard. First checks that the login page is 
 
 ##### Mandatory Arguments:
 IP address of service to test
+
+#### Optional Arguments:
+
+site_name_regexp: name of the horizon dashboard used to find the login page (default: 'openstack dashboard')
+
 ##### Example Output:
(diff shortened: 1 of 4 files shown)


Apsu:

ameserole:
recheck_upgrade

git-harry:
@ameserole in future please don't create backports until the initial patch has been merged.

andymcc:
👍  backport is clean.

mancdaz:
👍

PR #1268, opened 2016-07-27
Update openstack-ansible SHA

git-harry:
This commit updates the openstack-ansible SHA as part of the work to
prepare for the release of r13.0.0.

Connected https://github.com/rcbops/u-suk-dev/issues/254


sigmavirus24:
@git-harry in order to review this, do you want us to set-up an AIO or something else with this PR? For simple updates like this, I'm not sure what we're expecting the rigor of review to be.

git-harry:
@sigmavirus24 the checks should test that the SHA doesn't introduce any major failures. Apart from confirming the SHA is a recent Mitaka one I don't think there is a lot to do here. If the commit message referenced issues that were resolved by the SHA update you'd want to confirm those, but in this case it doesn't.

evrardjp:
Including what I need. Thanks.  👍

hughsaunders:
No issues specifically linked, but here's the OSA Diff: https://gist.github.com/hughsaunders/7ad4dbc557e326dbc01c9a8e4d5da582

👍  SHA is most recently Mitaka

Issue #1211, opened 2016-06-30
tempest.scenario.test_server_basic_ops.TestServerBasicOps.test_server_basic_ops fails with PortNotFound


git-harry:
##### What version (commit) do you encounter this issue on?
master (when both Liberty and Mitaka) however it is more common since the move to Mitaka
##### What type of issue is this (bug, feature, documentation, regression)?
bug
##### What is the expected behavior?
Pull request and periodic jobs run to test the codebase. The final step of the job is to run tempest. These tempest tests should complete successfully.
##### What is actually happening?
The tempest test tempest.scenario.test_server_basic_ops.TestServerBasicOps.test_server_basic_ops fails.
```
07:58:09 ==============================
07:58:09 Failed 1 tests - output below:
07:58:09 ==============================
07:58:09 
07:58:09 tempest.scenario.test_server_basic_ops.TestServerBasicOps.test_server_basic_ops[compute,id-7fff3fb3-91d8-4fd0-bd7d-0204f1f180ba,network,smoke]
```

An extract of tempest's debugging information for [one of these failures](http://jenkins.propter.net/job/RPC-AIO/8244/consoleFull) is
```
07:58:09 Captured traceback:
07:58:09 ~~~~~~~~~~~~~~~~~~~
07:58:09     Traceback (most recent call last):
07:58:09       File "/opt/tempest_c2c90574a78c7704f06e8fac279c4f1d11365993/tempest/scenario/manager.py", line 108, in delete_wrapper
07:58:09         delete_thing(*args, **kwargs)
07:58:09       File "/opt/tempest_c2c90574a78c7704f06e8fac279c4f1d11365993/tempest/lib/services/compute/floating_ips_client.py", line 62, in delete_floating_ip
07:58:09         resp, body = self.delete(url)
07:58:09       File "/opt/tempest_c2c90574a78c7704f06e8fac279c4f1d11365993/tempest/lib/common/rest_client.py", line 290, in delete
07:58:09         return self.request('DELETE', url, extra_headers, headers, body)
07:58:09       File "/opt/tempest_c2c90574a78c7704f06e8fac279c4f1d11365993/tempest/lib/services/compute/base_compute_client.py", line 53, in request
07:58:09         method, url, extra_headers, headers, body)
07:58:09       File "/opt/tempest_c2c90574a78c7704f06e8fac279c4f1d11365993/tempest/lib/common/rest_client.py", line 645, in request
07:58:09         resp, resp_body)
[...]

Issue #1201, opened 2016-06-28
Failures starting mysql on automatic Kilo to Liberty upgrades


git-harry:
##### What version (commit) do you encounter this issue on?
Upgrades from 11.1.9, 11.1.10, 11.1.11 to 12.1 or 12.2 branches
##### What type of issue is this (bug, feature, documentation, regression)?
bug
##### What is the expected behavior?
When galera_server role (Liberty version) is run on an existing deployment as part of a Kilo to Liberty upgrade it should be able to successfully restart the mysql service as part of applying any changes.
##### What is actually happening?
The test upgrade jobs are intermittently timing out.
The [console log](http://jenkins.propter.net/job/RPC-AIO/8001/console) from one of these jobs has the following
```
00:31:08 TASK: [galera_server | Start MySQL] ******************************************* 
00:51:08 Build timed out (after 20 minutes). Marking the build as aborted.
00:51:08 Build was aborted
```

An extract from [galera_server_error.log](http://jenkins.propter.net/job/RPC-AIO/8001/artifact/archive/openstack/jrpcaioiad-249_galera_container-e18df3e1/galera_server_error.log) provides further insight
```
160626  0:31:11 [ERROR] WSREP: Address already in use
160626  0:31:11 [ERROR] WSREP: failed to open gcomm backend connection: 98: error while trying to listen 'tcp://0.0.0.0:4567?socket.non_blocking=1', asio error 'Address already in use': 98 (Address already in use)
	 at gcomm/src/asio_tcp.cpp:listen():777
160626  0:31:11 [ERROR] WSREP: gcs/src/gcs_core.cpp:gcs_core_open():208: Failed to open backend connection: -98 (Address already in use)
160626  0:31:11 [ERROR] WSREP: gcs/src/gcs.cpp:gcs_open():1379: Failed to open channel 'rpc_galera_cluster' at 'gcomm://172.29.239.31,172.29.238.255,172.29.238.253': -98 (Address already in use)
160626  0:31:11 [ERROR] WSREP: gcs connect failed: Address already in use
160626  0:31:11 [ERROR] WSREP: wsrep::connect(gcomm://172.29.239.31,172.29.238.255,172.29.238.253) failed: 7
160626  0:31:11 [ERROR] Aborting
```

To see the latest failures reported by jenkins go to the [build summary page](http://jenkins.propter.net:443/#) and search for "Build Timeout: Install galera server / galera_server / Start MySQL".
[...]

Issue #1119, opened 2016-06-06
Conflicting Tempest pip package versions


jpmontez:
##### What version (commit) do you encounter this issue on?
r12.2.0rc3

##### What type of issue is this (bug, feature, documentation, regression)?
Bug

##### What is the expected behavior?
The correct package version requirements and dependencies be installed.

##### What is actually happening?
The `oslo.log` package installed is too recent and isn't the version required in the `requirements.txt` file.

```
${PYTHON:-python} -m subunit.run discover -t ${OS_TOP_LEVEL:-./} ${OS_TEST_PATH:-./tempest/test_discover} --list 
--- import errors ---
Failed to import test module: tempest.test_discover.test_discover
Traceback (most recent call last):
  File "/opt/tempest_0e25a90793a17728ab235d21ee282db31add3d74/lib/python2.7/site-packages/unittest2/loader.py", line 456, in _find_test_path
    module = self._get_module_from_name(name)
  File "/opt/tempest_0e25a90793a17728ab235d21ee282db31add3d74/lib/python2.7/site-packages/unittest2/loader.py", line 395, in _get_module_from_name
    __import__(name)
  File "/opt/tempest_0e25a90793a17728ab235d21ee282db31add3d74/tempest/test_discover/test_discover.py", line 18, in <module>
    from tempest.test_discover import plugins
  File "/opt/tempest_0e25a90793a17728ab235d21ee282db31add3d74/tempest/test_discover/plugins.py", line 21, in <module>
    from tempest.lib.common.utils import misc
  File "/opt/tempest_0e25a90793a17728ab235d21ee282db31add3d74/tempest/lib/common/utils/misc.py", line 19, in <module>
    from oslo_log import log as logging
  File "/opt/tempest_0e25a90793a17728ab235d21ee282db31add3d74/lib/python2.7/site-packages/oslo_log/log.py", line 49, in <module>
    from oslo_log import _options
  File "/opt/tempest_0e25a90793a17728ab235d21ee282db31add3d74/lib/python2.7/site-packages/oslo_log/_options.py", line 38, in <module>
    help='If set to true, the logging level will be set to '
  File "/opt/tempest_0e25a90793a17728ab235d21ee282db31add3d74/lib/python2.7/site-packages/oslo_config/cfg.py", line 982, in __init__
    super(BoolOpt, self).__init__(name, type=types.Boolean(), **kwargs)
TypeError: __init__() got an unexpected keyword argument 'mutable'
```

[...]

PR #1272, opened 2016-08-01
Update openstack-ansible SHA

git-harry:
This commit updates the openstack-ansible SHA in preparation for the
next release.

openstack-ansible has changed the way Python package constraints are
applied, see 7055f795095d8e19c052371259ca77bf42c3095f, to address an
issue where they were being applied incorrectly. This change means the
version of elasticsearch specified in the elasticsearch role is no
longer honoured by the repo_build role. This commit sets the var
repo_build_upper_constraints_overrides to allow the requirements of the
role elasticsearch to be met.

The following issues are addressed by this change:
Connected https://github.com/rcbops/rpc-openstack/issues/1119
Connected https://github.com/rcbops/rpc-openstack/issues/1201
Connected https://github.com/rcbops/rpc-openstack/issues/1211

### DIFF
--- a/rpcd/etc/openstack_deploy/user_variables.yml
+++ b/rpcd/etc/openstack_deploy/user_variables.yml
@@ -79,6 +79,10 @@ pkg_locations:
   - /etc/openstack_deploy
   - /opt/rpc-openstack/rpcd
 
+# Override OpenStack upper constraints
+repo_build_upper_constraints_overrides:
+  - "elasticsearch<2.1.0"
+
 # NOTE(mancdaz) there may be a better place to override this
 horizon_venv_tag: "{{ openstack_release }}"
 


mancdaz:
Confirmed the following fixes are in the OSA SHA:

```
876b25 Provide ability to override upper-constraints
5cb4b9 Better control of mariadb restarts in k>l upgrades
bf8d8f Clean tempest dir before cloning.
a940f1 Fix config entry regression introduced in container restart patch
7055f7 Fix setting of upper_constraints fact
```

Which together address the 3 issues referenced above.

Also confirmed that providing an override for elasticsearch meant that the correct package version got built in the repo (i.e. was not overridden by global constraints):

```-rw-r--r-- 1 nginx www-data 58336 Aug  1 09:20 /var/www/repo/pools/elasticsearch/elasticsearch-2.0.0-py2.py3-none-any.whl```

Pending jenkins upgrade test completing successfully, I am 👍

evrardjp:
👍

Issue #1273, opened 2016-08-01
scripts/deploy.sh can't run twice on master


evrardjp:
##### On which version(s) do you encounter this issue?
master

##### What type of issue is this? Please tick the appropriate box.
- [x] bug - existing functionality that does not work for you as documented
- [ ] feature - new functionality that you would like to see added

##### What is the expected behavior?
Ability to use ./scripts/deploy.sh multiple times

##### What is actually happening?
Impossible to use ./scripts/deploy.sh multiple times, due to a failure in a mv line.

##### What were you attempting to do when you encountered this error (give as much detail as possible)? 
Ran the ./scripts/deploy.sh , it failed for an unrelated reason, then restarted the scripts/deploy.sh, which activated a different part of its code, by checking the existence of the folder /etc/openstack_deploy. (It wasn't existing before for, and it now exists after my first run)

##### How can we reproduce this issue?
Run the script twice.

##### How can we verify your issue is addressed?

You can run the script twice.

##### What is your evaluation on the importance of this issue, and why?
Moderate priority: High criticality (could impact gates for minor upgrades later, negative impact on deployer confidence, ...) - Low impact (master only - no customer affected) 

----------------------------------------------------------------------------------------------------------------------------
###### For each pull request, update the branch line to reference it.
- [ ] master
- [ ] liberty-12.2
- [ ] kilo

##### Upgrade impact. Consider the impact on moving between major, minor and patch releases.
- [ ] This change does not impact upgrades
- [ ] This change addresses a potential upgrade impact
- [ ] The upgrade impact is being tracked by another issue.

The issue tracking the upgrade impact is:

##### Testing impact
- [ ] An issue/pull-request has been create or exists to track the testing gap, and is referenced below
- [ ] This change includes additional testing
- [ ] This change is tested by the existing setup

Additional testing links:

##### Release note
- [ ] This change includes a release note.
- [ ] This change does not require a release note.

##### Does your change require documentation?
- [ ] Yes, and I submitted an issue to [Docs issues](https://github.com/rackerlabs/docs-rpc/issues "Docs issues")
- [ ] Yes, and I contributed a documentation patch to [docs-rpc](https://github.com/rackerlabs/docs-rpc "docs-rpc")
- [ ] No documentation required for this change

Additional documentation links (please add links if you have ticked yes):

evrardjp:
Probably a 
`if test -f /etc/openstack_deploy/user_variables.yml; then`

on L62 and L121 would help.

[...]

PR #1276, opened 2016-08-01
make deploy.sh idempotent

mancdaz:
Because of some file 'mv' steps introduced in 23aaa8ba, the script was
no longer idempotent. This commit fixes that issue by doing a file test
first for each of the 'mv' steps introduced.

Connects #1273

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -118,7 +118,9 @@ if [[ "${DEPLOY_AIO}" == "yes" ]]; then
 fi
 
 # move OSA secrets to correct locations
-mv /etc/openstack_deploy/user_secrets.yml /etc/openstack_deploy/user_osa_secrets.yml
+if [[ ! -f /etc/openstack_deploy/user_osa_secrets.yml ]] && [[ -f /etc/openstack_deploy/user_secrets.yml ]]; then
+  mv /etc/openstack_deploy/user_secrets.yml /etc/openstack_deploy/user_osa_secrets.yml
+fi
 
 # Apply host security hardening with openstack-ansible-security
 # The is applied as part of setup-hosts.yml


evrardjp:
Making it idempotent is quite a programme, knowing that openstack-ansible calls aren't...
But removing the mv idempotency issue if fine for me :D

evrardjp:
I'm ok with it if it passes gates.

sigmavirus24:
👍 if it passes the gate. I've run into this several times and just kept copying the files back into place.

scripts/deploy.sh
```diff
@@ -59,7 +59,9 @@ if [[ "${DEPLOY_AIO}" == "yes" ]]; then
   if [[ ! -d /etc/openstack_deploy/ ]]; then
     ./scripts/bootstrap-aio.sh
     # move OSA variables file to AIO location.
-    mv /etc/openstack_deploy/user_variables.yml /etc/openstack_deploy/user_osa_aio_variables.yml
```
git-harry: I think it would be better to overwrite the file otherwise someone testing a change with an AIO might not get what they expected.

scripts/deploy.sh
```diff
@@ -118,7 +120,9 @@ if [[ "${DEPLOY_AIO}" == "yes" ]]; then
 fi
 
 # move OSA secrets to correct locations
-mv /etc/openstack_deploy/user_secrets.yml /etc/openstack_deploy/user_osa_secrets.yml
```
git-harry: There is the same concern here although the file can't be overwritten.

git-harry:
The [contributing guidelines](https://github.com/rcbops/rpc-openstack/blob/master/CONTRIBUTING.md#commits) include the following:
- capitalise first word of commit message title line.
- include reference to issue in commit message.

prometheanfire:
looks good

:+1:

evrardjp:
👍

PR #1277, opened 2016-08-01
Create run level links for Beaver

ameserole:
This commit fixes an issue where the beaver deamon would not
restart after each reboot because it is not linked inside of the
legacy init /etc/rcX.d directories. This issue would also cause
commands like 'update-rc.d -f beaver enable 2 3 4 5' to fail.

This fix is done by adding a task inside of
beaver_post_install.yml that makes sure the beaver service is
listed in the appropriate rcX.d folders.

Connects #934

PR #1279, opened 2016-08-01
Revert "Create run level links for Beaver"

ameserole:
Reverts rcbops/rpc-openstack#1277

### DIFF
--- a/rpcd/playbooks/roles/beaver/tasks/beaver_post_install.yml
+++ b/rpcd/playbooks/roles/beaver/tasks/beaver_post_install.yml
@@ -25,11 +25,3 @@
   tags:
     - beaver-post-install
     - beaver-conf
-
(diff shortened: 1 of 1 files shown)


Apsu:

git-harry:
@ameserole is this revert solely because https://github.com/rcbops/rpc-openstack/pull/1277 didn't receive the appropriate level of review i.e two +1's? If that's the case then hopefully we can avoid the need to revert the patch, @mancdaz reviewed https://github.com/rcbops/rpc-openstack/pull/1277 and has given it a +1 which means the merge can stand assuming you identified no other issues with it.

ameserole:
@git-harry Yes I made it just in case someone tested the original bug and found some issues with it.

git-harry:
The commit being reverted in the pull request does not need to be removed.

Issue #934, opened 2016-03-25
Beaver init script can not be enabled


BjoernT:
##### What version (commit) do you encounter this issue on?

11.1.5

##### What type of issue is this (bug, feature, documentation, regression)?

Bug

##### What is the expected behavior?

It automatically enables upon installation

##### What is actually happening?

update-rc.d -f  beaver enable 2 3 4 5

update-rc.d: warning:  start runlevel arguments (none) do not match beaver Default-Start values (2 3 4 5)
update-rc.d: warning:  stop runlevel arguments (none) do not match beaver Default-Stop values (0 1 6)
 System start/stop links for /etc/init.d/beaver do not exist.
 

##### How can we reproduce this issue?

See above

##### How can we verify your issue is addressed?

Beaver autostarts after reboot

##### If this is a large issue, what can be done immediately to alleviate the issue while working on a longer-term solution?

Manual workaround possible

##### What is your evaluation on the importance of this issue, and why?

Normal Prio

----------------------------------------------------------------------------------------------------------------------------
###### For each pull request, update the branch line to reference it.
- [x] liberty-12.2 #1277 
- [ ] kilo #1280 

##### Upgrade impact. Consider the impact on moving between major, minor and patch releases.
- [x] This change does not impact upgrades
- [ ] This change addresses a potential upgrade impact
- [ ] The upgrade impact is being tracked by another issue.

The issue tracking the upgrade impact is:

##### Testing impact
- [ ] An issue/pull-request has been create or exists to track the testing gap, and is referenced below
- [ ] This change includes additional testing
- [ ] This change is tested by the existing setup

Additional testing links:

##### Release note
- [ ] This change includes a release note.
- [ ] This change does not require a release note.

##### Does your change require documentation?
- [ ] Yes, and I submitted an issue to [Docs issues](https://github.com/rackerlabs/docs-rpc/issues "Docs issues")
- [ ] Yes, and I contributed a documentation patch to [docs-rpc](https://github.com/rackerlabs/docs-rpc "docs-rpc")
- [ ] No documentation required for this change

Additional documentation links (please add links if you have ticked yes):

BjoernT:
I''l convert the script to systemd then we should be fine

d34dh0r53:
@BjoernT Is this an RPC-R issue, or are you seeing it on RPC-O?

BjoernT:
This is RPC-O, didn't know if RPC-R consumes that role even

b3rnard0:
[...]

PR #1280, opened 2016-08-02
Create run level links for Beaver

ameserole:
This commit fixes an issue where the beaver deamon would not
restart after each reboot because it is not linked inside of the
legacy init /etc/rcX.d directories. This issue would also cause
commands like 'update-rc.d -f beaver enable 2 3 4 5' to fail.

This fix is done by adding a task inside of
beaver_post_install.yml that makes sure the beaver service is
listed in the appropriate rcX.d folders.

Connects https://github.com/rcbops/rpc-openstack/issues/934

### DIFF
--- a/rpcd/playbooks/roles/beaver/tasks/beaver_post_install.yml
+++ b/rpcd/playbooks/roles/beaver/tasks/beaver_post_install.yml
@@ -25,3 +25,11 @@
   tags:
     - beaver-post-install
     - beaver-conf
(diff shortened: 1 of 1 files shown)


ameserole:
Backport to kilo of #1277

mancdaz:
👍  pending checks

evrardjp:
not tested but seems legit. 👍

PR #1281, opened 2016-08-02
Enumerate log file targets and multiline config

sigmavirus24:
Using more than one prospector allows us to configure the openstack
services with multiline log parsing. This prevents us from applying
that same logic to non-OpenStack services (which will mess up log
shipping). It also allows us to define multiline rules for each of
the services we hope to ship to logstash with multiline log messages.

Filebeat will ignore directories/paths that do not actually exist on
the host so we can safely enumerate all of the services we want to
apply multiline parsing to in one prospector.

We *do* have to, however, ensure that paths do not overlap with the
prospectors. Filebeat explicitly calls this out as "undefined"
behaviour.

Included near the multline configurations are play.golang.org links so
others can experiment with what we used to test the multiline regular
expressions. Also, given the complexity and widespread use of our
OpenStack multiline expression, we've attempted to thoroughly document
it.

Connects rcbops/u-suk-dev#261

### DIFF
--- a/rpcd/playbooks/roles/filebeat/templates/filebeat.yml.j2
+++ b/rpcd/playbooks/roles/filebeat/templates/filebeat.yml.j2
@@ -1,2 +1,41 @@
 # {{ ansible_managed }}
-{{ filebeat_config_overrides | to_yaml }}
+filebeat:
+  prospectors:
+{% for service in filebeat_logging_paths %}
+    - paths: {{ service.paths }}
+{% if service.document_type is defined %}
+      document_type: {{ service.document_type }}
+{% endif %}
+      encoding: {{ service.encoding |default('utf-8') }}
+{% if service.tags is defined %}
+      fields_under_root: true
+{# NOTE(sigmavirus24): So... filebeats 5.0 will have proper tags support,
+ # at which point we can remove this hack. This little piece of trickery,
+ # however, needs explanation. In order to send tags so that logstash will
+ # process them and send them to ElasticSearch, we need to hide them in the
+ # fields attribute on a prospector below. In order to then send the tags as a
+ # top-level attribute, we set fields_under_root to be true. Then all is
+ # happy.
+ #}
+      fields:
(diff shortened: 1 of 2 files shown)


Apsu:
RPC-AIO-Liberty-PR: Fail

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Liberty-PR: Fail

d34dh0r53:
This is working acceptably in the lab, there may be some tweaking require but it is a very good start.  Well done @sigmavirus24 :+1:

evrardjp:
For now it looks nice. I'll re-test on my aio.

rpcd/playbooks/roles/filebeat/templates/filebeat.yml.j2
```diff
@@ -1,2 +1,42 @@
...
-{{ filebeat_config_overrides | to_yaml }}
+filebeat:
+  prospectors:
+{% for service in filebeat_logging_paths %}
+    - paths: {{ service.paths }}
+{% if service.document_type is defined %}
+      document_type: {{ service.document_type }}
+{% endif %}
+      encoding: {{ service.encoding |default('utf-8') }}
+      input_type: log
```
sigmavirus24: I like to be explicit rather than implicit @evrardjp. We don't have to specify this but the filebeat documentation doesn't describe what the default value for this might be.

Further, filebeat 5.0 is the next version which is breaking things with the underlying beats implementation but shouldn't affect filebeat (according to their developers in their IRC channel). I trust them, but want to make sure there's no implicit default that changes with this.

sigmavirus24:
@evrardjp I removed `input_type: log` per your request.

evrardjp:
The code seems fine and I see multiline content in my kibana. However, I didn't fully test it.
It looks generally ok.

If it passes gates, it has my 👍

PR #1283, opened 2016-08-02
Update f5-monitor.sh to monitor keystonev3

prometheanfire:
The f5-monitors need updating for mitaka, so we do that here.

Connects https://github.com/rcbops/u-suk-dev/issues/264

### DIFF
--- a/scripts/f5-monitor.sh
+++ b/scripts/f5-monitor.sh
@@ -81,7 +81,7 @@ get_token() {
       case $check_port in
           5000)
               #Keystone
-              check_url="$check_proto://$check_ip:$check_port/v2.0/"
+              check_url="$check_proto://$check_ip:$check_port/v3/"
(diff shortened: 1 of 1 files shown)


andymcc:
👍 change LGTM.

andymcc:
Oh 1 tiny change - can you change the commit message to have f5-monitor.sh rather than f5-moonitor.sh

evrardjp:
yup it boothers me too.
The rest is doing what it needs to be doing.

prometheanfire:
now with less moo

sigmavirus24:
👍

git-harry:
Did this change get validated in a hardware lab? I can find no comment anywhere confirming that happened.

prometheanfire:
yes, http://ip:port/v3

Issue #1149, opened 2016-06-10
Disable L3HA by default


BjoernT:
##### What version (commit) do you encounter this issue on?

r12.1.2

##### What type of issue is this (bug, feature, documentation, regression)?

Bug

##### What is the expected behavior?

L3HA working correctly

##### What is actually happening?

https://bugs.launchpad.net/neutron/+bug/1590845
Duplicate: https://bugs.launchpad.net/neutron/+bug/1591386

##### How can we reproduce this issue?

Spinning up multiple L3HA enabled router will result in nearly 100% with ports stuck in BUILD state
which may or may not pass traffic.

##### How can we verify your issue is addressed?

Once upstream bug is fixed

##### If this is a large issue, what can be done immediately to alleviate the issue while working on a longer-term solution?

Normal issue since it's just a config change 

##### What is your evaluation on the importance of this issue, and why?

It is prio 1 because neutron router created over Horizon can not be configured in a non L3HA mode

----------------------------------------------------------------------------------------------------------------------------
###### For each pull request, update the branch line to reference it.
- [x] master
- [ ] liberty-12.2

##### Upgrade impact. Consider the impact on moving between major, minor and patch releases.
- [ ] This change does not impact upgrades
- [ ] This change addresses a potential upgrade impact
- [ ] The upgrade impact is being tracked by another issue.

The issue tracking the upgrade impact is:

##### Testing impact
- [ ] An issue/pull-request has been create or exists to track the testing gap, and is referenced below
- [ ] This change includes additional testing
- [ ] This change is tested by the existing setup

Additional testing links:

##### Release note
- [x] This change includes a release note.
- [ ] This change does not require a release note.

##### Does your change require documentation?
- [ ] Yes, and I submitted an issue to [Docs issues](https://github.com/rackerlabs/docs-rpc/issues "Docs issues")
- [x] Yes, and I contributed a documentation patch to [docs-rpc](https://github.com/rackerlabs/docs-rpc "docs-rpc")
- [ ] No documentation required for this change

Additional documentation links (please add links if you have ticked yes):

BjoernT:
FYI, beyond the reported issue, Neutron has several other issues open.
Hence the recommendation to turn it off before it becomes a liability in customer environment.

https://bugs.launchpad.net/neutron/+bug/1523780 Race between HA router create and HA router delete
https://bugs.launchpad.net/neutron/+bug/1533455 Stale processes lives after a fanout deleting HA router RPC between L3 agents
[...]

PR #1285, opened 2016-08-02
Disable L3HA as default setting

alextricity25:
Based on many open Neutron issues around L3HA RPC-O currently
disables L3HA until the known issues like
https://bugs.launchpad.net/neutron/+bug/1590845
Neutron community.

The HA functionality will be still served by the neutron-ha-tool script.

Connects #1149
Docs Patch https://github.com/rackerlabs/docs-rpc/pull/496

### DIFF
--- a/rpcd/etc/openstack_deploy/user_osa_variables_defaults.yml
+++ b/rpcd/etc/openstack_deploy/user_osa_variables_defaults.yml
@@ -127,3 +127,9 @@ repo_build_pip_extra_indexes:
 # variable to 'true'.
 # Docs: http://docs.openstack.org/developer/openstack-ansible-security/
 # apply_security_hardening: true
+
+# Based on https://github.com/rcbops/rpc-openstack/issues/1149
+# L3HA has to be disabled until all major issues are fixed.
+neutron_neutron_conf_overrides:
+  DEFAULT:
+    l3_ha: False
--- a/releasenotes/notes/turn-off-l3ha-18274d19d54a6315.yaml
+++ b/releasenotes/notes/turn-off-l3ha-18274d19d54a6315.yaml
@@ -1,0 +1,20 @@
+---
+upgrade:
+  - |
+    If upgrading from an environment that has L3HA on,
+    please keep the following in mind:
+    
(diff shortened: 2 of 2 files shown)


git-harry:
👎 the variable is being added to the wrong file, it should be added to rpcd/etc/openstack_deploy/user_osa_variables_defaults.yml

Would also like to see opinions on whether or not this is appropriate for Mitaka. The referenced upstream bug is likely to be fixed in Mitaka but is only one of a number that were highlighted when this change was discussed for Liberty. Is this functionality still considered to be so broken that it should also be disabled in Mitaka?

alextricity25:
Sorry, for some reason I thought overrides that are specific to the RPCO product go in rpco_variables. Now I see that only variables *used* in the rpc playbooks go in rpc_variables.

Apsu:
RPC-AIO-Liberty-PR: Fail

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

alextricity25:
Going to also make a PR to docs before it's merged in.

alextricity25:
Docs patch is here:
https://github.com/rackerlabs/docs-rpc/pull/496

tomc603:
Looks good. Maybe once the docs PR merges we should reference the appropriate section here too? Certainly not required though, since the docs PR is linked in the comments above.

👍

rpcd/etc/openstack_deploy/user_osa_variables_defaults.yml
```diff
@@ -127,3 +127,9 @@ repo_build_pip_extra_indexes:
 # variable to 'true'.
 # Docs: http://docs.openstack.org/developer/openstack-ansible-security/
 # apply_security_hardening: true
+
+# Based on https://bugs.launchpad.net/neutron/+bug/1590845
```
git-harry: I think it's preferable to reference the RPCO issue instead of this neutron one, the RPCO issue states that this was one of a number or issues affecting l3ha and so I wouldn't want the impression to be given that once 1590845 is fixed it can be re-enabled.

PR #1282, opened 2016-08-02
Update SHA for openstack-ansible

andymcc:
This commit updates the openstack-ansible SHA as part of the work to
prepare for the release of r13.0.0

Connected https://github.com/rcbops/u-suk-dev/issues/267

PR #1286, opened 2016-08-02
Update versions of Elasticsearch pip packages

andymcc:
The elasticsearch pip package for the new version of Elasticsearch that
we are using should be updated to 2.3.0.

In order to utilise 2.3.0 we have to override the mitaka default upper
constraint for OpenStack which is version 1.9.0

Additionally, the elasticsearch-curator pip package should be updated to
keep up to date, and to ensure everything remains compatible.

Connected https://github.com/rcbops/u-suk-dev/issues/262

### DIFF
--- a/rpcd/etc/openstack_deploy/user_rpco_variables_defaults.yml
+++ b/rpcd/etc/openstack_deploy/user_rpco_variables_defaults.yml
@@ -136,3 +136,7 @@ secure_cluster: true
 # List of secure flags to set on for a pool (options for the list are nodelete, nopgchange, nosizechange - prevents deletion, pg from changing and size from changing respectively).
 secure_cluster_flags:
   - nodelete
+
+# We need to override the OpenStack upper constraint for Mitaka which is version 1.9.0, We should be able to remove this in Newton as the upper constraint moves to 2.3.0
+repo_build_upper_constraints_overrides:
+  - "elasticsearch==2.3.0"
--- a/rpcd/playbooks/roles/elasticsearch/defaults/main.yml
+++ b/rpcd/playbooks/roles/elasticsearch/defaults/main.yml
@@ -26,8 +26,8 @@ elasticsearch_apt_packages:
   - openjdk-7-jre-headless
 
 elasticsearch_pip_packages:
-  - elasticsearch<2.1.0
-  - elasticsearch-curator<3.5.0
+  - elasticsearch>=2.0.0,<3.0.0
+  - elasticsearch-curator==4.0.4
 
 # This sets the cluster name
 elasticsearch_cluster: openstack


andymcc:
This won't work until: #1282

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Liberty-PR: Fail

andymcc:
recheck_all

evrardjp:
👍

rpcd/playbooks/roles/elasticsearch/defaults/main.yml
```diff
@@ -26,8 +26,8 @@ elasticsearch_apt_packages:
   - openjdk-7-jre-headless
 
 elasticsearch_pip_packages:
-  - elasticsearch<2.1.0
-  - elasticsearch-curator<3.5.0
+  - elasticsearch>=2.0.0,<3.0.0
+  - elasticsearch-curator==4.0.4
```
git-harry: Why does this need to be a specific version?

rpcd/etc/openstack_deploy/user_rpco_variables_defaults.yml
```diff
@@ -136,3 +136,7 @@ secure_cluster: true
 # List of secure flags to set on for a pool (options for the list are nodelete, nopgchange, nosizechange - prevents deletion, pg from changing and size from changing respectively).
 secure_cluster_flags:
   - nodelete
+
+# We need to override the OpenStack upper constraint for Mitaka which is version 1.9.0, We should be able to remove this in Newton as the upper constraint moves to 2.3.0
+repo_build_upper_constraints_overrides:
+  - "elasticsearch>=2.0.0,<3.0.0"
```
sigmavirus24: I know we're focused on 13.0 right now, but thinking a little forward to upgrades, if we're installing say 2.0.x somewhere, when we upgrade, won't that stick around because that satisfies this requirement? In other words, do we actually want `>=2.1.0`?

rpcd/playbooks/roles/elasticsearch/defaults/main.yml
sigmavirus24: Hm. We provide a range above and a pin here. Why are we pinning instead of providing a range with a bunch of versions to exclude?

evrardjp:
I think it's best if we'd pin.

rpcd/playbooks/roles/elasticsearch/defaults/main.yml
andymcc: elasticsearch-curator didn't provide a sensible "version range" in the same way as elasticsearch did, so I went with pinning it to a specific version - happy to change.

rpcd/etc/openstack_deploy/user_rpco_variables_defaults.yml
andymcc: I based that off the recommendation from the elasticsearch maintainers:
https://pypi.python.org/pypi/elasticsearch/2.3.0

Happy to change to be a specific version.

sigmavirus24:
> I'd feel more confortable if we pinned es pip pkg too.

Pinning is fine for me, I'm just thinking that since constraints are typically *pins* the constraints overrides should be a pin. The requirement in our `pip_packages` list can simply be a range.

rpcd/etc/openstack_deploy/user_rpco_variables_defaults.yml
sigmavirus24: I think in this var only we should use a pin (as upper constraints are typically pins). The range should be fine in `elasticsearch_pip_packages` though

sigmavirus24:
I haven't tested this but this looks good to me now. 👍

evrardjp:
Looks good. 👍

PR #1287, opened 2016-08-03
Update Contributing guidelines

git-harry:
Swim lanes allow the board to be split into categories of work that
represent the different work streams the team needs to manage. This
commit updates the contributing guidelines to define the available swim
lanes and how to use them to manage WIP.

This change updates the contributing guidelines to further clarify what
constitutes a single item of WIP.

The wording around blocking WIP has been adjusted to highlight the
importance of removing the blockage quickly.

The details regarding where to find the current WIP limits have been
emended and the use of a minimum WIP limit has been explained.

The role of support in identifying new WIP has been made explicit.

Document the use of the label status-dont-merge. Currently the
documentation only provides use of the label status-blocked which is not
appropriate when the pull request is simply being revised.

In general an upstream fix is not immediately consumable by RPCO, OSA
has processes to periodically update its SHAs to pull in newer code from
associated projects. This lag between a fix appearing in a code base and
being available to RPCO has the potential to block new work. This commit
adds documentation for a new column, Pending-SHA-update, that can be
used as a buffer so that this work is appropriately tracked but does not
block new WIP.

Connected https://github.com/rcbops/u-suk-dev/issues/268


git-harry:
@b3rnard0 @mancdaz @claco this is to document some process updates that we have either already implemented or have been identified as being necessary. I would like to get this merged today.

b3rnard0:
Reviewed this earlier so :+1: from me.

mancdaz:
My only question is around where the pending-sha-update column sits in the workflow. With the way it is proposed, you have the potential to have a lot of work completed in a single shot every two weeks, skewing the CFD. A task could be technically completed in a day, but not actually moved to done for 2 weeks, meaning it appears to spend 2 weeks in the system.

Technically if we're doing this for osa patches we should do the same for all rpco patches, placing them into a pending status until they are released in a tag. But then you have the same problem, in that you get a ton of work 'completed' every 2 weeks.

git-harry:
@mancdaz yes it will create steps in the CFD and may end up giving the appearance of flow debt in our data. When we discussed this yesterday there was the alternative approach of breaking the work out into stages but there seemed to be a desire to use in issue to track a fix all the way through. That column is not tracking an item till it appears in a tag but instead tracking an item till it appears in a branch. I believe tracking tags would need to be a separate step and the benefit of that if for communicating with support so is not necessarily something we need.

b3rnard0:
@mancdaz 

I would think it sits in-between needs review and done. It's a buffer queue that does not have a WIP.

So the other option that @git-harry alluded to was to move these into a separate swimlane. 

How else would we be able to reduce the batch size for these types of items without fundamentally changing how we complete them? In order for us to be as transparent about how long these will take to complete, we cannot hide the 2+ week delay that happen. Support would want an SLA that's based on the actual completion time, which includes the external dependency of how long things complete in upstream. Perhaps a future improvement is for us to think about how we reduce our long cycle time? Excluding the the idle time here is not helpful to Support and customers.

git-harry:
Picking-up/corrupting @b3rnard0's point about fundamental change. We could choose to change our process, for example as in the days of Chef we could put something in place to update the IRR and service project SHA's more regularly or allow that to be as part of these upstream issues, however I think before we start considering how to improve things we should put more emphasis on gathering metrics. I think we still need more work to get to the point where we are properly tracking our work. Once things are on a firmer footing we can look to use the data we've generated to identify areas to improve.

claco:
Really, my only concerns here is 'add it to another swimlane'.  I worry that it's going to become difficult to just look at a topic and see what is happening if I have to know when to switch between swimlanes, or sources, especially when a bug evolves to span multiple lanes, or splits into two cards, the original, and a blocker that are in two different sources or two different swimlanes.

git-harry:
@claco the current patch does not propose adding another swim lane. I think there are merits to that approach but it all depends on how we want to track the work and what information we are trying to communicate with the different completion statuses. The current suggestion closely matches current working practices without unnecessarily affecting throughput and so gives us a defined based from which to evolve.

claco:
Yeah, I was more refering to @b3rnard0  last comment "So the other option that @git-harry alluded to was to move these into a separate swimlane."

mancdaz:
@git-harry @b3rnard0 @claco so what's the real benefit of tracking only to the point the code lands in the rpco branch (as we do now for rpco bugs, and as we would for osa bugs with this proposed process change)? None of it is of real benefit until it is consumable in a release/tag. As in, it's not real throughput until it's saleable, to use the workplant analogy.

I'd say either we go the whole hog and track it all to tag, or we accept that what we have now tracks the actual work that we are doing to land the patches in branch (in the respective project, not into rpco) and consider that done. Then separately, accepting that there is 'work' in doing a release that you might not be tracking per se (doing the tag, sending an email, new sha bump), the releases happen.

I still think the extra column helps you keep tabs on what will come in as a result of the sha bump, that is important to you, but that it should not form part of the end-to-end time in the system...unless we go the whole hog as I described above.

git-harry:
@mancdaz I think this is missing the point. The purpose when starting Kanban is to model the process as is and then look to evolve it over time. Now admittedly what has been defined does not completely map to how things were done, partly because of WIP limits and partly because it's difficult to describe a process when that process was so loose. We have never waited to close an issue until the next release has been made. Upstream work has always been handled inconsistently, although the majority of upstream issues are not closed until the SHA update is done, this change is designed to make sure we are all pulling in the same direction, to prevent upstream work blocking other work, and to allow us to begin gathering useful data.

We can define throughput to be whatever we want in the sense that there is not requirement for the departure point to be one thing or another. I believe that we should agree with our customer as to the appropriate definition of done. While I agree that is worthwhile, that also creates problems with the tooling and I believe is feature creep when we are still trying to properly define what we are currently doing.

We can look to break this work up so that an upstream issue is simply an upstream issue but then there are a number of ways we could look to change things. There is no perfect answer at this point and I think it would be better to resist evolving things prematurely.

b3rnard0:
@mancdaz @claco I agree with @git-harry in that this issue resolves specifically our upstream tracking challenges. Tracking all of our work in this fashion is outside the scope of this change but it is a needed improvement that we are tracking here: https://github.com/rcbops/u-suk-dev/issues/240

I completely agree that we need to start thinking about our dev-done to release work stream for general rpc-o bugs, as that is a much more complete picture of our cycle time/SLA. However, we need to solve things incrementally so I would say let's fix this one issue (upstream tracking) and then add the larger release picture as separate issue.

mancdaz:
@git-harry @b3rnard0 you've melded into a single b3rnardarry entity and I can't distinguish either of you now. You are one.

Fair enough, let's try this and see how it goes.

b3rnard0:
@mancdaz What did you call me?!

mancdaz:
So based on the 👍 from b3nardarry and myself, this merges

PR #1288, opened 2016-08-03
Remove tag added to log events by logstash plugin

sigmavirus24:
The logstash-input-beats plugin is necessary to use filebeat to ship
logs to logstash. However, it defaults to always adding a tag to each
log event that indicates what codec it used to process the beats input.

Until https://github.com/logstash-plugins/logstash-input-beats/pull/114
is merged (in one shape or another) we need to just remove the tag from
all of our events.

Connects rcbops/u-suk-dev#261

### DIFF
--- a/rpcd/playbooks/roles/logstash/templates/02-general.conf
+++ b/rpcd/playbooks/roles/logstash/templates/02-general.conf
@@ -8,4 +8,10 @@ filter {
       add_field => { "received_at" => "%{@timestamp}" }
     }
   }
+
+  mutate {
+    # TODO(sigmavirus24): Replace this mutate with configuration once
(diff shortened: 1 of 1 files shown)


sigmavirus24:
Testing this out right now on my AIO

sigmavirus24:
👎 this is broken

sigmavirus24:
It was a simple typo `remove_tags` instead of `remove_tag`. This is working on my AIO

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

d34dh0r53:
:+1: looks good to me, this tag is just noise in our environment.

evrardjp:
👍

Issue #941, opened 2016-03-30
MAAS playbooks use different type of hostnames


BjoernT:
##### What version (commit) do you encounter this issue on?

r11.0.0

##### What type of issue is this (bug, feature, documentation, regression)?

Bug

##### What is the expected behavior?

All maas related playbooks should use one type of hostname, either inventory_hostname or ansible_hostname to prevent situations that the host has a FQDN but is only known with a short name in the inventory. That leads to issues that the the verify-maas.yml playbook looks up checks with a different hostname than the checks got created with

For example :
```
root@bjoern-aio3:/opt/rpc-openstack/rpcd/playbooks# grep hostname roles/rpc_maas/templates/nova_api_local_check.yaml.j2
label: "nova_api_local_check--{{ ansible_hostname }}"
        label                   : nova_api_local_status--{{ ansible_hostname }}
		
		
root@bjoern-aio3:/opt/rpc-openstack/rpcd/playbooks# grep hostname verify-maas.yml -B1
        {{ maas_rpc_scripts_dir }}/rpc-maas-tool.py verify-created
        --entity {{ inventory_hostname }}
--
        {{ maas_rpc_scripts_dir }}/rpc-maas-tool.py verify-status
        --entity {{ inventory_hostname }}
```

##### What is actually happening?

See above 

##### How can we reproduce this issue?

Use short hostnames in the inventory and long hostnames on host level
and run through the MAAS life cycle

##### How can we verify your issue is addressed?

All playbooks run successfully when installing them and running the verify-maas.yml playbook

##### If this is a large issue, what can be done immediately to alleviate the issue while working on a longer-term solution?

Normal prio

##### What is your evaluation on the importance of this issue, and why?

None

----------------------------------------------------------------------------------------------------------------------------
###### For each pull request, update the branch line to reference it.
- [x] master
- [x] liberty-12.2
- [x] kilo

##### Upgrade impact. Consider the impact on moving between major, minor and patch releases.
- [x] This change does not impact upgrades
- [ ] This change addresses a potential upgrade impact
- [ ] The upgrade impact is being tracked by another issue.

The issue tracking the upgrade impact is:

##### Testing impact
- [x] An issue/pull-request has been create or exists to track the testing gap, and is referenced below
- [ ] This change includes additional testing
- [ ] This change is tested by the existing setup

Additional testing links: https://github.com/rcbops/u-suk-dev/issues/283

##### Release note
- [ ] This change includes a release note.
[...]

PR #1289, opened 2016-08-03
Use ansible_hostname in verify_maas.yml

nategraf:
All of the maas check templates use ansible_hostname to generate
check names, and prior to this commit verify_maas.yml used
inventory_hostname to generate the check names to verify.
This led to an error where verify_maas.yml would fail when
ansible_hostname and inventory_hostname were different

This commit fixes this by using ansible_hostname within
verify_maas.yml to be consistent with the check templates.

Connects #941

### DIFF
--- a/rpcd/playbooks/verify-maas.yml
+++ b/rpcd/playbooks/verify-maas.yml
@@ -23,7 +23,7 @@
         . {{ maas_venv_bin }}/activate
         {% endif %}
         {{ maas_rpc_scripts_dir }}/rpc-maas-tool.py verify-created \
-        --entity {{ inventory_hostname }} \
+        --entity {{ inventory_hostname }}{{ maas_fqdn_extension }} \
         {% for ec in maas_excluded_checks %} --excludedcheck {{ec}}
         {% endfor %}
       register: verify_maas
@@ -43,7 +43,7 @@
         . {{ maas_venv_bin }}/activate
         {% endif %}
         {{ maas_rpc_scripts_dir }}/rpc-maas-tool.py verify-status \
-        --entity {{ inventory_hostname }}
+        --entity {{ inventory_hostname }}{{ maas_fqdn_extension }}
       register: verify_status
       failed_when: verify_status.rc != 0
       changed_when: False


alextricity25:
:+1: makes sense to me.

git-harry:
👎 I think the assessment of what is going on here is incorrect.

inventory_hostname is used for some check names
```
$ git grep inventory_hostname rpcd/playbooks/roles/rpc_maas/templates/
rpcd/playbooks/roles/rpc_maas/templates/cpu_check.yaml.j2:label             : cpu_check--{{ inventory_hostname|quote }}
rpcd/playbooks/roles/rpc_maas/templates/cpu_check.yaml.j2:        label                   : idle_percent_average--{{ inventory_hostname|quote }}
rpcd/playbooks/roles/rpc_maas/templates/hp-memory.yaml.j2:        label                   : hp-memory--{{ inventory_hostname|quote }}
rpcd/playbooks/roles/rpc_maas/templates/hp-processors.yaml.j2:        label                   : hp-processors--{{ inventory_hostname|quote }}
rpcd/playbooks/roles/rpc_maas/templates/hp-vdisk.yaml.j2:        label                   : hp-disk--{{ inventory_hostname|quote }}
rpcd/playbooks/roles/rpc_maas/templates/memory_check.yaml.j2:label             : memory_check--{{ inventory_hostname|quote }}
rpcd/playbooks/roles/rpc_maas/templates/memory_check.yaml.j2:        label                   : memory_check--{{ inventory_hostname|quote }}
rpcd/playbooks/roles/rpc_maas/templates/network_throughput.yaml.j2:label: network_throughput_{{ item.0.name }}--{{ inventory_hostname|quote }}
rpcd/playbooks/roles/rpc_maas/templates/openmanage-memory.yaml.j2:        label                   : openmanage-memory--{{ inventory_hostname|quote }}
rpcd/playbooks/roles/rpc_maas/templates/openmanage-processors.yaml.j2:        label                   : openmanage-processors--{{ inventory_hostname|quote }}
rpcd/playbooks/roles/rpc_maas/templates/openmanage-vdisk.yaml.j2:        label                   : openmanage-vdisk--{{ inventory_hostname|quote }}
```

Check names and what is used in the template should not matter for the purpose of verifying the checks. Checks are associated with an entity via IDs and not labels. I believe the important thing here is that verify-maas.yml is using the same entity name as is used in [rpc_maas](https://github.com/rcbops/rpc-openstack/blob/master/rpcd/playbooks/roles/rpc_maas/tasks/host_setup.yml#L24)

Given that
```
entity_name: "{{ inventory_hostname }}{{ maas_fqdn_extension }}"
```
is what is used to configure the agent that is what should be used to test it. Also, I think ansible_nodename would probably be the most appropriate fact to use if this was to be tidied up given that ansible_hostname = ansible_nodename.split('.')[0].

alextricity25:
I didn't realize **some** of the check config templates use inventory_hostname.
i'm going to :-1: this until that's addressed

git-harry:
@alextricity25 my point is that the use of inventory_hostname in the checks is irrelevant. It is about the name used to create the entity when the server is provisioned and making sure that is how verify-maas references the entity.

cloader89:
@BjoernT @git-harry please review the changes and give us feedback.

nategraf:
Thank you @alextricity25  and @git-harry for your insight. I now have a better understanding of what is going on. Submitted a new commit below is the commit message.
------------------------------
 Use proper entity name in verify_maas.yml

verify_maas.yml uses the entity name in a call to the maas
tool to verify the configuration of maas checks. Prior
to this commit verify_mass.yml used the inventory_hostname
fact to set the entity name it would check. However, the
host_setup.yml rpc_mass task would use a concatination of
inventory_hostname and the maas_fqdn_extension facts to
set the entity name on creatiion. This means that if
mass_fqdn_extension is non-empty the verify maas playbook
would fail.

This commit fixes this issue by changing the entity name
variable in verify_maas.yml to match how it is set up in
the host_setup.yml task.

Connects #941

rpcd/playbooks/verify-maas.yml
```diff
@@ -23,7 +23,7 @@
         . {{ maas_venv_bin }}/activate
         {% endif %}
         {{ maas_rpc_scripts_dir }}/rpc-maas-tool.py verify-created \
-        --entity {{ inventory_hostname }} \
+        --entity {{ inventory_hostname }}{{ maas_fqdn_extension }} \
```
Frank-ZhangXin: verify-mass is still using inventory_hostname, which doesn't give solution that mix using hostname (ansible_hostname and inventory_name) in the same role (maas).

Frank-ZhangXin: {{ maas_fqdn_extension }} is set in rpcd/etc/openstack_deploy/user_rpco_variables_defaults.yml , but by default it's empty. So I believe it makes no difference only if user or tester gives customized FQDN. That means we need to define FQDN then see what happens/verify this.

nategraf: The issue we face is that we cannot change the templates used by setup maas without upgrade impact. Although that inconsistency needs to be cleaned up, it does not cause and failures as far as I am aware. The "--entity" option here needs needs to have the same value as when the entity is setup in the host_setup.yml task, or verify_mass.yml will fail. As you said, the maas_fqdn_extension fact is blank by default, which is why this script does not fail under default builds. However, there were examples of error (as can be seem from Thomas's comment on issue 190) caused by this in environments where the maas_fqdn_extension is set.

Frank-ZhangXin: @nategraf yeah, agree with you but I said maybe we need test a case like Thomas met, which has a defined FQDN.

git-harry: The code in this repo is not used to create the MaaS entities. The entities are created as part of the server provisioning process prior to RPCO being deployed.

The templates do not know about the entity directly. The use of ansible_hostname or inventory_hostname in forming the label by the templates is simply to make it easier to tell which host a check belongs to when reading the API response data.

Checks/alarms get associated with an entity by the agent that uses the templates to create them.

[This](https://github.com/rcbops/rpc-openstack/blob/master/rpcd/playbooks/roles/rpc_maas/tasks/raxmon_agent_install.yml#L59-L134) section of code associates the entity with the agent and so with the host. The form of entity name used is "{{ inventory_hostname }}{{ maas_fqdn_extension }}". If this was not correct, the checks/alarms would not be created or would be associated with the incorrect entity.

My understanding is that it is only verification that fails, in which case using the same definition of entity name, as was used to create the checks/alarms, should be sufficient to solve this issue.

andymcc: Another important thing to note. The template files that are linked in the bug report have nothing to do with the entity name - so I believe that is adding confusion. Those are the label names for the alarms/checks - these could be anything, e.g. "mymonitor" and the entity name would still remain the same.

The crux of the issue is that the entity is not being found correctly, Harry's final 2 paragraphs sum up what needs to be changed, as it is only the verify that is failing. If the entity was not found at creation time the maas play would fail earlier and monitors would not be created.

andymcc:
👍  This PR looks correct to me.

ansible_hostname vs inventory_hostname is a red herring, we use ansible_hostname for containers (which are not entities - the physical host would be) so you can have a "local" check for a container, and it will identify the container it is monitoring so you would know when troubleshooting an alert. Imagine a situation where you have 2 glance_api containers on a host, you would want to monitor both but identify them. These are labels, they are not representative of the entity and could literally be any string - they will show up in MaaS with that label, for the entity that is configured.

The issue is that the check doesn't happen against the entity that is used when creating the alerts - e.g. "{{ inventory_hostname }}{{ maas_fqdn_extension }}" so the verification is failed.

The entity_name is referenced in tasks/host_setup.yml for the rpc_maas role, and is what Nate has put in this PR. I believe this is the best and only real solution to this problem so I am happy for this merge.

alextricity25:
Thanks @git-harry @andymcc for the explanation. I have a much better understanding on entity names - they are not the same thing as check labels! Having said that, the verify-maas playbook should use the same entity name that was registered with the agent. Which is what is being address by this PR. 
:+1:

BjoernT:
👍

Issue #1301, opened 2016-08-09
virtualenv is not present on multinode installs: setup-maas.yml fails


mancdaz:
##### On which version(s) do you encounter this issue?

heads/master

##### What type of issue is this? Please tick the appropriate box.
- [x] bug - existing functionality that does not work for you as documented
- [ ] feature - new functionality that you would like to see added

##### What is the expected behavior?
The setup-maas playbook completes successfully

##### What is actually happening?
The setup-maas playbook is terminating prematurely because we are attempting to do things inside a virtualenv (the first task we hit is to update pip), but nowhere does the setup-maas playbook explicitly install virtualenv. Example output:

```
TASK: [rpc_maas | Update pip] *************************************************
failed: [533810-node14] => {"failed": true}
msg: Failed to find required executable virtualenv
failed: [533800-node11] => {"failed": true}
msg: Failed to find required executable virtualenv
failed: [533808-node12] => {"failed": true}
msg: Failed to find required executable virtualenv
failed: [533809-node13] => {"failed": true}
msg: Failed to find required executable virtualenv
ok: [533811-node15]
```

##### What were you attempting to do when you encountered this error (give as much detail as possible)? 

Testing the head of master on a multinode deployment prior to cutting a first r13.0.0rc1 tag

##### How can we reproduce this issue?

Install master on a multinode cluster, using the deploy script, and export DEPLOY_MAAS=yes

##### How can we verify your issue is addressed?

On a multinode deployment, run setup-maas and then verify-maas - all should work as expected

##### If you are using or know of a workaround please describe the necessary steps?

Manually install virtualenv on each host (```pip install virtualenv```)

##### What is your evaluation on the importance of this issue, and why?

Quite important since a full deployment won't complete for QE

----------------------------------------------------------------------------------------------------------------------------
###### For each pull request, update the branch line to reference it.
- [x] master (https://github.com/rcbops/rpc-openstack/pull/1302)

##### Upgrade impact. Consider the impact on moving between major, minor and patch releases.
- [x] This change does not impact upgrades
- [ ] This change addresses a potential upgrade impact
- [ ] The upgrade impact is being tracked by another issue.

The issue tracking the upgrade impact is:

##### Testing impact
- [ ] An issue/pull-request has been create or exists to track the testing gap, and is referenced below
- [ ] This change includes additional testing
- [x] This change is tested by the existing setup

Additional testing links:

##### Release note
- [ ] This change includes a release note.
[...]

PR #1302, opened 2016-08-09
Install virtualenv pip package for rpc-maas role

mancdaz:
In order to install the maas environment inside a virtualenv, we need to
ensure we are actually installing the virtualenv pip package. In an AIO
this is done by the nova role on the compute 'container', meaning it
actually gets installed on the host, but in a multinode we need to do
this explicitly.

Connected https://github.com/rcbops/rpc-openstack/issues/1301

### DIFF
--- a/rpcd/playbooks/roles/rpc_maas/defaults/main.yml
+++ b/rpcd/playbooks/roles/rpc_maas/defaults/main.yml
@@ -442,6 +442,12 @@ maas_apt_packages:
   - mariadb-client
 
 #
+# maas_requires_pip_packages: Pip packages needed outside the virtualenv
+#
+maas_requires_pip_packages:
+  - virtualenv
+
+#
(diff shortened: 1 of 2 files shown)


Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

alextricity25:
:+1: once gate checks pass

sigmavirus24:
👍 once the gate passes

mattt416:
Got enough of a multi-node going to replicate the error, and the attached fix did get me past that error.

evrardjp:
Makes sense, not tested. I'd give my 👍 too if it passes gates.

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Liberty-PR: Fail

sigmavirus24:
recheck_all

git-harry:
👍

Issue #1303, opened 2016-08-09
maas templates point to wrong location for maas plugins


mancdaz:
##### On which version(s) do you encounter this issue?

heads/master

##### What type of issue is this? Please tick the appropriate box.
- [x] bug - existing functionality that does not work for you as documented
- [ ] feature - new functionality that you would like to see added

##### What is the expected behavior?

All maas checks and alarms get configured correctly on all hosts in a multinode deployment

##### What is actually happening?

maas checks and alarms get configured correctly on the deployment host, but no other nodes in a multinode deployment.  
This is due to the switch to running maas plugins from a venv (commit 3cdd0727d98537b91605d271d4d0d760adc59681) where we started using the ```maas_plugin_dir``` variable as the location of the installed plugins. The current value of that var is ```/opt/rpc-openstack-maas/plugins``` - on the deployment host this directory exists, but does not on any other node so all checks on those nodes fail.
We need to switch to using the correct location of the installed plugins, ```/usr/lib/rackspace-monitoring-agent/plugins/```

This is what an incorrect template looks like:

```
# cat /etc/rackspace-monitoring-agent.conf.d/cinder_api_local_check--533808-node12_cinder_api_container-6d2d613c.yaml
type: agent.plugin
label: "cinder_api_local_check--533808-node12-cinder-api-container-6d2d613c"
disabled    : false
period      : "60"
timeout     : "30"
details     :
    file    : run_plugin_in_venv.sh
    args    : ["/opt/rpc-openstack/maas/plugins/cinder_api_local_check.py", "10.141.0.121"]
alarms      :
    cinder_api_local_status :
        label                   : cinder_api_local_status--533808-node12-cinder-api-container-6d2d613c
        notification_plan_id    : "npTechnicalContactsEmail"
        criteria                : |
            :set consecutiveCount=3
            if (metric["cinder_api_local_status"] != 1) {
                return new AlarmStatus(CRITICAL, "API unavailable");
            }
```

##### What were you attempting to do when you encountered this error (give as much detail as possible)? 

Do a multinode deployment, and run setup-maas.yml

##### How can we reproduce this issue?

Run a multinode deployment using master, and run setup-maas.yml, then verify-maas.yml and see the failures. Then log in to intelligence.rackspace.com and check - all nodes apart from the deployment host will have failing checks.

##### How can we verify your issue is addressed?

Run a multinode deployment, setup-maas.yml and then verify-maas.yml - it shoudl complete successfully

[...]

PR #1304, opened 2016-08-09
use correct location for maas plugins

mancdaz:
With the switch to using maas from a venv, we moved to using the
maas_plugin_dir var as a reference for the location of the installed
plugins.  The value of this var, however, was the location of plugins
inside the git clone of rpc-openstack, rather than the location of
where the plugins get installed. This meant that only a deployment host
had the plugins in the correct location.

This commit switches the value of this var to point to the correct
installed location, whilst creating a new var to be used by the only
other task that was using the original value. It is done in this way to
minimise the amount of change needed to fix this issue.

Additionally, some areas where we were hardcoding the path to the
plugin location are now set to use the var value.

Connected https://github.com/rcbops/rpc-openstack/issues/1303

### DIFF
--- a/rpcd/playbooks/roles/rpc_maas/tasks/raxmon_agent_install.yml
+++ b/rpcd/playbooks/roles/rpc_maas/tasks/raxmon_agent_install.yml
@@ -14,23 +14,30 @@
 # limitations under the License.
 
 - name: Install plugin dir
-  file: path=/usr/lib/rackspace-monitoring-agent/plugins/ state=directory owner=root group=root mode=0755
+  file:
+    path: "{{ maas_plugin_dir }}"
+    state: directory
+    owner: "root"
+    group: "root"
+    mode: "0755"
 
 - name: Install MAAS Plugins
   synchronize:
-    src={{ maas_plugin_dir }} dest=/usr/lib/rackspace-monitoring-agent/plugins/ delete=yes
+    src: "{{ maas_source_plugin_dir }}"
+    dest: "{{ maas_plugin_dir }}"
+    delete: yes
 
 - name: Drop in wrapper script to run maas plugins in venv
   template:
     src: "run_plugin_in_venv.sh.j2"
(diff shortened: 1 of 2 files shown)


rpcd/playbooks/roles/rpc_maas/tasks/raxmon_agent_install.yml
```diff
@@ -14,23 +14,23 @@
 # limitations under the License.
 
 - name: Install plugin dir
-  file: path=/usr/lib/rackspace-monitoring-agent/plugins/ state=directory owner=root group=root mode=0755
+  file: path={{ maas_plugin_dir }} state=directory owner=root group=root mode=0755
```
evrardjp: could we use the more traditional yaml dict notation instead?

evrardjp:
I didn't test it. It seems generally good (doing what it claims).

It could look better if we use our standard osa's upstream notation.

mattt416:
Patched my  deploy and it fixed up my non-deploy node checks w/ the correct path to the plugins, so LGTM.

andymcc:
LGTM functionally/change wise. I confirmed there are no other sections that should be changed, and that the changes are correct path wise.
I agree with JP's one comment though - if we could change the task structure (of the first 2 tasks) to match the other tasks it would be great.
I realise that's technically an unrelated change (functionality wise), but it would be nice to get that fixed and I'm skeptical that anybody else would submit a separate patch for that.

On that note I'm still 👍  without the change, but would be a nice fix.

andymcc:
I pushed the change including the reformat - (amended and made the PR), for the sake of efficiency and getting this PR through.

git-harry:
@evrardjp @andymcc I don't think we should be changing another part of the code. While minor it should be handled separately and approved like any other piece of work, all work comes at a cost and with the potential to generate more work. On the specifics we have no documented policy regarding tasks and we have no automation enforcing this style. More generally mixing two sets of changes into one commit makes it harder to review.

git-harry:
👍 there are a couple of nits but otherwise looks fine.

Issue #1306, opened 2016-08-10
verify-maas only works on an AIO


mancdaz:
###### Please provide the output to 'git describe --all' for  all deployments where you have confirmed it occurs.

heads/master

##### What type of issue is this? Please tick the appropriate box.
- [x] bug - existing functionality that does not work for you as documented
- [ ] feature - new functionality that you would like to see added

##### What is the expected behavior?
The verify-maas.yml playbook completes successfully, or rather is at least able to run (even if some alarms are failing, which could be legitimate)

##### What is actually happening?
On all nodes except the deployment node, ansible fails because the task is trying to use the ```rpc-maas-tool.py``` script from ```/opt/rpc-openstack/scripts```, and that directory only exists on the deployment host.

```
TASK: [Verify Checks & Alarms are registered] *********************************
ok: [533800-node11]
failed: [533811-node15] => {"attempts": 8, "changed": false, "cmd": ". /openstack/venvs/maas-UNKNOWN_RELEASE/bin/activate\n /opt/rpc-openstack/scripts/rpc-maas-tool.py verify-created --entity 533811-node15.ukospoc.com", "delta": "0:00:00.005510", "end": "2016-08-09 16:54:33.766599", "failed": true, "failed_when_result": true, "rc": 127, "start": "2016-08-09 16:54:33.761089", "stdout_lines": [], "warnings": []}
stderr: /bin/sh: 2: /opt/rpc-openstack/scripts/rpc-maas-tool.py: not found
msg: Task failed as maximum retries was encountered
failed: [533810-node14] => {"attempts": 8, "changed": false, "cmd": ". /openstack/venvs/maas-UNKNOWN_RELEASE/bin/activate\n /opt/rpc-openstack/scripts/rpc-maas-tool.py verify-created --entity 533810-node14.ukospoc.com", "delta": "0:00:00.004363", "end": "2016-08-09 16:54:33.895546", "failed": true, "failed_when_result": true, "rc": 127, "start": "2016-08-09 16:54:33.891183", "stdout_lines": [], "warnings": []}
stderr: /bin/sh: 2: /opt/rpc-openstack/scripts/rpc-maas-tool.py: not found
msg: Task failed as maximum retries was encountered
[...]

PR #1309, opened 2016-08-10
Copy rpc-maas-tool to all nodes

mancdaz:
In order to run properly, rpc-maas-tool needs to be on the node on
which it is doing the checks, because it needs to be able to
interrogate the maas conf files that have been laid down. By default,
we assume that it's present in /opt/rpc-openstack/scripts - which is
fine for an AIO but not for a multinode deployment.

This commit copies the file to the same location on all hosts in a
deployment so that the verify-maas playbook (which calls this script)
can complete successfully.

Note, I did investigate an alternative approach using the script
module, but I was unable to get it to work properly due to the
complexities of trying to run the script inside a virtual environment
whilst also using jinja to prepare args. The fix in this commit seemed
like the least complex fix to a pretty simple problem.

Connected https://github.com/rcbops/rpc-openstack/issues/1306

### DIFF
--- a/rpcd/playbooks/verify-maas.yml
+++ b/rpcd/playbooks/verify-maas.yml
@@ -17,6 +17,15 @@
   hosts: hosts
   user: root
   tasks:
+    - name: ensure remote directory
+      file:
+        path: "{{ maas_rpc_scripts_dir }}"
+        state: directory
+    - name: copy script to all hosts
+      copy:
+        src: "{{ maas_rpc_scripts_dir }}/rpc-maas-tool.py"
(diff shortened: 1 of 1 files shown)


mattt416:
Speaking of rpc-maas-tool.py, did you run into any issues w/ concurrent.futures not being present on your nodes?  I couldn't run this until I installed python-concurrent.futures.

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Liberty-PR: Fail

git-harry:
recheck_all

git-harry:
👍

Issue #1308, opened 2016-08-10
cinder-backup service off by default


mancdaz:
##### On which version(s) do you encounter this issue?

heads/master

##### What type of issue is this? Please tick the appropriate box.
- [x] bug - existing functionality that does not work for you as documented
- [ ] feature - new functionality that you would like to see added

##### What is the expected behavior?

cinder-backup service is running

##### What is actually happening?

cinder-backup service is not running, but we are now monitoring for it so this shows up as a monitoring alert.

This is due to the fact that by default, cinder-backup is disabled:

https://github.com/openstack/openstack-ansible-os_cinder/blob/stable/mitaka/defaults/main.yml#L117

In an AIO, the variable ```cinder_service_backup_program_enabled``` gets set to ```true```, so our AIO gating misses this issue. In rpc-o, we are not overriding the default value of ```false``` anywhere, so in a multinode we don't actually deploy cinder-backup. But we do monitor for it, as we should.

##### What were you attempting to do when you encountered this error (give as much detail as possible)? 

Deploy on a multinode, and run setup-maas. I noticed that the cinder backup check was failing, and on further investigation it was a 'correct fail', in that cinder-backup was indeed not running.

##### How can we reproduce this issue?

Do a non-aio deployment and then run setup-maas. notice that on your cinder node, cinder-backup is not running

##### How can we verify your issue is addressed?

do a deployment and note that cinder-backup is running and after running setup-maas, the cinder_backup alarm is not in alert status

##### If you are using or know of a workaround please describe the necessary steps?

override the default in one of the var files in /etc/openstack_deploy/ and re-run the os-cinder.yml playbook

##### What is your evaluation on the importance of this issue, and why?

It will cause a starightforward default deployment using deploy.sh to fail, because verify-maas will fail. Important for r13.0.0rc1

----------------------------------------------------------------------------------------------------------------------------
###### For each pull request, update the branch line to reference it.
- [x] master https://github.com/rcbops/rpc-openstack/pull/1314
- [x] liberty-12.2 https://github.com/rcbops/rpc-openstack/pull/1338
- [x] kilo (https://github.com/rcbops/rpc-openstack/pull/1324)

##### Upgrade impact. Consider the impact on moving between major, minor and patch releases.
- [x] This change does not impact upgrades
- [ ] This change addresses a potential upgrade impact
- [ ] The upgrade impact is being tracked by another issue.

The issue tracking the upgrade impact is:

##### Testing impact
[...]

PR #1312, opened 2016-08-11
Enable cinder-backup service

mancdaz:
By default, cinder-backup is disabled in OSA. Recently, we fixed the
monitoring that we had in place for cinder-backups and discovered that
the cinder-backup service is not running.

This commit sets the necessary variable to ensure that the
cinder-backup service gets configured/deployed.

Note: by default, cinder will be configured to use swift as a backup
target. Further configuration would need to be done by the customer to
enable different backup targets.

Connected https://github.com/rcbops/rpc-openstack/issues/1308

### DIFF
--- a/rpcd/etc/openstack_deploy/user_osa_variables_defaults.yml
+++ b/rpcd/etc/openstack_deploy/user_osa_variables_defaults.yml
@@ -73,6 +73,9 @@ keystone_token_driver: "sql"
 # Galera overrides
 galera_cluster_name: rpc_galera_cluster
 
+# Cinder overrides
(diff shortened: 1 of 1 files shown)


sigmavirus24:
Haven't tested but by looking at OSA this should just work. 👍

mattt416:
This is probably a bad user experience to auto-enable, as it'd default to a swift backend which most deployments won't have.  I think the monitoring should be conditional based on whether or not the service is running.

mancdaz:
Thanks @mattt416 I will look to conditionally enable monitoring instead

PR #1314, opened 2016-08-11
Conditionally monitor cinder-backup service

mattt416:
The cinder-backup service behaves a little differently as the group
will always exist but the service is only started when the
cinder_service_backup_program_enabled variable is set to True.  As a
result, we remove the cinder-backup check from
openstack_service_local_checks_list and add it to a new variable called
cinder_backup_checks_list.  We then add a new task that adds the checks
in cinder_backup_checks_list conditional on a variable called
maas_monitor_cinder_backup.  The variable maas_monitor_cinder_backup
is assigned the value of cinder_service_backup_program_enabled.
Lastly, we update the deploy script to set
cinder_service_backup_program_enabled to true for testing purposes.

NOTE: the OSA AIO already sets cinder_service_backup_program_enabled to
true, however that user variable file is read first and has a lower
precedence than the other variable files.

Connected #1308

### DIFF
--- a/rpcd/etc/openstack_deploy/user_osa_variables_defaults.yml
+++ b/rpcd/etc/openstack_deploy/user_osa_variables_defaults.yml
@@ -137,3 +137,5 @@ neutron_neutron_conf_overrides:
 # Container repos
 lxc_container_template_main_apt_repo: "https://mirror.rackspace.com/ubuntu"
 lxc_container_template_security_apt_repo: "https://mirror.rackspace.com/ubuntu"
+
+cinder_service_backup_program_enabled: false
--- a/rpcd/etc/openstack_deploy/user_rpco_variables_defaults.yml
+++ b/rpcd/etc/openstack_deploy/user_rpco_variables_defaults.yml
@@ -84,6 +84,8 @@ maas_filesystem_critical_threshold: 90.0
 #    warning_threshold: 80.0
 #    critical_threshold: 90.0
 
+maas_monitor_cinder_backup: "{{ cinder_service_backup_program_enabled }}"
+
 # overrides for the nova_cloud_stats  maas plugin
 cloud_resource_cpu_allocation_ratio: "{{ nova_cpu_allocation_ratio }}"
 cloud_resource_mem_allocation_ratio: "{{ nova_ram_allocation_ratio }}"
--- a/rpcd/playbooks/roles/rpc_maas/defaults/main.yml
+++ b/rpcd/playbooks/roles/rpc_maas/defaults/main.yml
@@ -366,7 +366,6 @@ kernel_checks_list:
 openstack_service_local_checks_list:
   - { name: "cinder_api_local_check", group: "cinder_api" }
   - { name: "cinder_scheduler_check", group: "cinder_scheduler" }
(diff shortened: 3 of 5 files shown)


sigmavirus24:
This looks great to me! 👍

mancdaz:
Will test this.

mancdaz:
I get an error if I run this and do not enable the backup program:

```
fatal: [533811-node15] => error while evaluating conditional: cinder_service_backup_program_enabled | bool
```

I think because the setup-maas playbook doesn't know about that var as it only exists in the os-cinder role unless we override it. I guess we can default it to false in the rpc-maas role defaults, or in the playbook where we evaluate it. I tested the latter and it seemed to work:

```
cinder_service_backup_program_enabled | default('False') | bool
```

I can confirm once I set a default, the logic worked as expected to either not lay the config file down if the var was false, or to lay it down if the var was true.

mattt416:
Derp, good catch @mancdaz.  Looking at this now.

mattt416:
I need to do a touch more testing here before this one merges.

mattt416:
Hi @mancdaz, this should resolve the issue you ran into.  Please give this a go and let me know if anything else needs updating.

Apsu:
RPC-AIO-Liberty-PR: Fail

mancdaz:
recheck_swift

mancdaz:
Hi @mattt416 

The addition of the extra var feels like it's overcomplicating things a bit to me. Wouldn't it just be easier to have:

```
- include: ensure_local_checks.yml
  vars:
    checks: "{{ cinder_backup_checks_list }}"
  when:
    - inventory_hostname in groups["cinder_backup"]
    - cinder_service_backup_program_enabled | default(false) | bool
```

Then the var is always false, unless it's overridden (and therefore visible to the playbook). Is there a reason we need the extra var?

rpcd/playbooks/group_vars/all.yml
```diff
@@ -28,3 +28,7 @@ migrations_dir: /etc/openstack_deploy/migrations
 
 # Ceph Conf Directory Mode
 conf_directory_mode: 755
+
+# This should default to the same default for the
+# cinder_service_backup_program_enabled variable in the os_cinder role.
+maas_monitor_cinder_backup: "{{ cinder_service_backup_program_enabled | default(false) }}"
```
git-harry: My preference is to explicitly set cinder_service_backup_program_enabled, the way this has been managed previously is by setting the OSA var and the rpc_maas var in the overrides files.

mattt416:
@mancdaz @git-harry I've pushed another change, which effectively removes the group_vars entry.  Adding defaults to the rpc user variables files is going to break the AIO so I've just left it like this for now.

Apsu:
RPC-AIO-Liberty-PR: Fail

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

git-harry:
👍

mancdaz:
👍  pending checks. I have already tested the general logic being applied here.

Apsu:
RPC-AIO-Liberty-PR: Fail

PR #1325, opened 2016-08-15
Add upgrade playbook for r13.1

sigmavirus24:
All upgrade related playbooks and tasks will be launched from this one
over-arching playbook to make running and testing upgrades simpler.

Connects rcbops/u-suk-dev#292

### DIFF
--- a/rpcd/playbooks/run-upgrade.yml
+++ b/rpcd/playbooks/run-upgrade.yml
@@ -1,0 +1,19 @@
+---
+# Copyright 2016, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+- name: Run upgrade from Liberty to Mitaka
+  hosts: all
+  user: root
+  tasks: []


sigmavirus24:
cc @rcbops/support-feedback

d34dh0r53:
Lint failing with `ERROR: expecting dict; got: osa-after-upgrade.yml, error in /home/travis/build/rcbops/rpc-openstack/rpcd/playbooks/roles/post-upgrade/tasks/main.yml`

d34dh0r53:
@sigmavirus24 I think you may need to add `---` and a comment or a dummy var to make these valid YAML files.

Apsu:
RPC-AIO-Liberty-PR: Fail

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

hughsaunders:
recheck_all

sigmavirus24:
@mattt416 was curious what this was for. To answer his question:

We've had several discussions with our support team and they strongly dislike and distrust upgrade bash scripts. They would be more confident if they had upgrade playbooks to do the work that the bash scripts tend to do. This is the start of the work towards that end.

mattt416:
Sorry @sigmavirus24, I understand the reasoning behind converting the bash script to ansible, but I was more curious if we need actual roles for this versus just having individual playbooks.

sigmavirus24:
Ah, so we could. I think we've just tended towards roles for organizational purposes.

sigmavirus24:
Gah, forgot to update the commit message.

sigmavirus24:
Based on this morning's mitaka stand-up. We're not going to use roles but instead we're going to use playbooks and factor those into roles as necessary

Apsu:
RPC-AIO-Liberty-PR: Fail

Apsu:
RPC-AIO-Liberty-PR: Fail

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

sigmavirus24:
recheck_all

git-harry:
👍 @evrardjp, @mattt416 would you please review this.

mattt416:
LGTM, but I think one of the first tasks to go into this is to prompt to proceed, or something of that nature.

PR #1327, opened 2016-08-16
Add a migration helper script

prometheanfire:
Adds a helper script to help migrate overrides from the old to
new locations.

Connects https://github.com/rcbops/u-suk-dev/issues/186

### DIFF
--- a/scripts/migrate-yaml.py
+++ b/scripts/migrate-yaml.py
@@ -1,0 +1,192 @@
+#!/usr/bin/env python
+# Copyright 2016, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# (c) 2016, Matthew Thode <matt.thode@rackspace.com>
+
+"""
+Print out values that have changed, by default just additions and changes.
+
+Additions can be taken as is and placed into an overrides file, but should
+you should verify you still want to keep it.
+
+Removals are not printed by default because not having an override is the
+same as taking a default.
+
+Changes are printed, showing what you overrode and what is getting overridden.
+
+CAUTION: danger-mode is only if you KNOW what you are overriding doesn't
+    need to change, warrantee is void if used.
+"""
+
+from __future__ import print_function
+import argparse
+from collections import namedtuple
+from deepdiff import DeepDiff
+import yaml
+
+
+def parse_args():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--overrides', dest='overrides', required=True,
+                        help='The path to the file with overrides.')
+    parser.add_argument('--info', dest='info', type=bool, default=False,
+                        help='Shows what is NOT overridden.')
+    parser.add_argument('--danger-mode', dest='danger',
+                        type=bool, default=False,
+                        help='Print output usable as an overrides file.')
+    return parser.parse_args()
+
+
+def root_cleanup(var):
+    """Cleans up what deepdiff returns so we can have some cleaner output.
+
+    :param var: the key you wish to clean up, just a string
+    :return: the cleaned up string
+    """
+    prefix = "root['"
+    suffix = "']"
+    if var.startswith(prefix):
+        var = var[len(prefix):]
+    if var.endswith(suffix):
+        var = var[:-len(suffix)]
+    var = var.replace("']['", '.')
+    return var
+
+
+def parsed_yaml_from(path):
+    """Load yaml data and return a python object
+
+    :raises: IOError if the target path is invalid
+    :raises: ParserError if the yaml is invalid
+    :param path: path of yaml data to load
+    :return: dictionary loaded from yaml data
+    """
+    try:
+        with open(path) as f:
+            data = f.read()
+    except IOError:
+        raise SystemExit('Error loading file: %s\n  Did you mistype?' % path)
+    else:
+        try:
+            loaded_data = yaml.safe_load(data)
+        except yaml.ParserError:
+            raise SystemExit('The yaml in %s is non-compliant and failed'
+                             'to load' % path)
+
+    if loaded_data is None:
+        return {}
+    else:
+        return loaded_data
+
+
+def do_the_diff(defaults, overrides):
+    """Diff two yaml loaded objects, returning diffs in a namedtuple
+
+    :param defaults: File you wish to compare against, the base.
+    :param overrides: File you wish to compare, the mask.
+    :return: what is added, taken away, changed(added) and changed(removed).
+    :rtype: namedtuple
+    """
+    added = {}
+    removed = {}
+    changed = {}
+    changed_types = {}
+    parsed_diff = namedtuple('parsed_yaml_diff',
+                             ['added', 'removed',
+                              'changed_added', 'changed_removed'])
+
+    diff = DeepDiff(defaults, overrides, verbose_level=2)
+
+    for old_key, value in diff.get('dictionary_item_added', {}).items():
+        added[root_cleanup(old_key)] = value
+    for old_key, value in diff.get('dictionary_item_removed', {}).items():
+        removed[root_cleanup(old_key)] = value
+    for old_key, value in diff.get('values_changed', {}).items():
+        changed[root_cleanup(old_key)] = value
+    for old_key, value in diff.get('type_changes', {}).items():
+        changed_types[root_cleanup(old_key)] = value
+
+    # Have to recast the values as dictionaries and combine type/value changes
+    new_changed = {}
+    new_changed_types = {}
+    for key in changed:
+        new_changed[key] = dict(changed[key])
+    for key in changed_types:
+        new_changed_types[key] = dict(changed_types[key])
+    new_changed.update(new_changed_types)
+
+    # we make two different yaml dump-able dictionaries, one for new values,
+    # one for old (what is being overridden)
+    new_changed_dict = {}
+    old_changed_dict = {}
+    for key in new_changed:
+        new_changed_dict[key] = new_changed[key]['new_value']
+        old_changed_dict[key] = new_changed[key]['old_value']
+
+    return parsed_diff(added=added, removed=removed,
+                       changed_added=new_changed_dict,
+                       changed_removed=old_changed_dict)
+
+
+def main(default_files):
+    """Manage the work of loading the files and printing the output as needed
+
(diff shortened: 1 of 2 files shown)


Apsu:
RPC-AIO-Liberty-PR: Fail

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

prometheanfire:
also, we now pass flak8 everywhere, but I increased the line length to 120, because that's sane in my view (and now mater how much we wish it, we don't live in the 80s).

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

Apsu:
RPC-AIO-Liberty-PR: Fail

scripts/f5-config.py
```diff
@@ -550,12 +549,12 @@ def main():
     pubvirts = []
 
     if IS_LEM is True:
-        SEC_RULE=[
```
git-harry: This file looks to have been added by mistake.

tox.ini
```diff
@@ -57,4 +57,5 @@ commands =
 # Excluding our upstream submodule, and our vendored f5 configuration script.
 
 ignore=F403,H303,R
-exclude=openstack-ansible/*,f5-config.py
```
git-harry: Was this file added by mistake?

prometheanfire: no, but maybe should be a different commit.  This and the commit to f5-config get us fully compliant.

F403 prevents `from foo import *` and we don't do that anywhere I can see.
H303 Do not use wildcard * import (*), don't think we do that either...

git-harry: Yes, I don't think flake8 failures for f5-config.py are relevant to the purpose of this commit. I would suggest adding an issue to the backlog and it can be prioritised appropriately.

scripts/migrate-yaml.py
```diff
@@ -0,0 +1,164 @@
+#!/usr/bin/env python
+# Copyright 2014-2016, Rackspace US, Inc.
```
sigmavirus24: This script is brand new. The copyright need not be a range.

scripts/migrate-yaml.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+def root_cleanup(var):
+    """Cleans up what deepdiff returns so we can have some cleaner output.
+
+    :param var: the key you wish to clean up, just a string
+    :return: the cleaned up string
+    """
+    prefix = 'root[\''
+    suffix = '\']'
```
sigmavirus24: More simply these could be:

```py
prefix = "root['"
suffix = "']"
```

So you don't have to escape the quoting

scripts/migrate-yaml.py
```diff
@@ -0,0 +1,164 @@
...
+    var = var[var.startswith(prefix) and len(prefix):]
+    var = var[:var.endswith(suffix) and -len(suffix)]
+    var = var.replace('\'][\'', '.')
```
sigmavirus24: Same here `"']['"`

scripts/migrate-yaml.py
```diff
@@ -0,0 +1,164 @@
...
+    return var
+
+
+def get_config(path):
+    """Load yaml data if you can
+
+    :param path: path of yaml data to load
+    :return: dictionary
```
sigmavirus24: It's a dictionary parsed from the YAML data, right? If you were just documenting the return type (`:rtype:`) then `dictionary` would be suitable, but you should describe what the dictionary contains here

scripts/migrate-yaml.py
```diff
@@ -0,0 +1,164 @@
...
+    """
+    prefix = 'root[\''
+    suffix = '\']'
+    var = var[var.startswith(prefix) and len(prefix):]
+    var = var[:var.endswith(suffix) and -len(suffix)]
+    var = var.replace('\'][\'', '.')
+    return var
+
+
+def get_config(path):
```
sigmavirus24: Which config are we "getting"? Also would there be a better name or can the docstring be improved to explain this? As it is, "get_config" is vague and "Load yaml data if you can" is about as vague.

scripts/migrate-yaml.py
sigmavirus24: Might also document that this raises an IOError in ... cases where the file doesn't exist and it raises an IOError

scripts/migrate-yaml.py
```diff
@@ -0,0 +1,164 @@
...
+    try:
+        with open(path) as f:
+            data = f.read()
+    except IOError as e:
+        if e.errno == errno.ENOENT:
+            data = None
+        else:
+            raise e
+    else:
+        data = yaml.safe_load(data)
```
sigmavirus24: What happens if `safe_load` raises an exception?

Also, can we *please* not reuse `data` here? Having a new variable binding won't hurt the performance of the script, but not having a new variable will make it less maintainable going forward.

scripts/migrate-yaml.py
```diff
@@ -0,0 +1,164 @@
...
+    else:
+        return data
+
+
+def do_the_diff(defaults, overrides):
+    """Diff two yaml files, returning four dictionaries containing cleaned up key-value sets.
+
+    :param defaults: File you wish to compare against, the base.
+    :param overrides: File you wish to compare, the mask
+    :return: four dictionaries, what is added, what was taken away and what was changed.
```
sigmavirus24: Four dictionaries, only three of which are described? Which is it four, 3 or an off-by-one error?

sigmavirus24: Also if you're returning 4 items like this that are *very* similar, please create and use a namedtuple so it's easier (when debugging) to understand what is actually happening here.

scripts/migrate-yaml.py
```diff
@@ -0,0 +1,164 @@
...
+    olddict = {}
+    for key in new_changed:
+        newdict[key] = new_changed[key]['new_value']
+        olddict[key] = new_changed[key]['old_value']
+
+    return dict(added), dict(removed), dict(newdict), dict(olddict)
+
+if __name__ == '__main__':
+    # boilerplate goes here
+    args = parse_args()
```
sigmavirus24: This section of code is *untestable* unless it's in one (or preferably more) functions.

scripts/migrate-yaml.py
prometheanfire: This was just what it was called in the other, bad, one.  Anyway, fixed.

scripts/migrate-yaml.py
prometheanfire: done

scripts/migrate-yaml.py
prometheanfire: fixed

scripts/migrate-yaml.py
prometheanfire: fixed

scripts/migrate-yaml.py
prometheanfire: fixed

scripts/migrate-yaml.py
prometheanfire: done

scripts/migrate-yaml.py
prometheanfire: fixed both

scripts/migrate-yaml.py
prometheanfire: fixed

scripts/migrate-yaml.py
```diff
@@ -0,0 +1,164 @@
...
+    :return: four dictionaries, what is added, what was taken away and what was changed.
+    """
+    added = {}
+    removed = {}
+    changed = {}
+    changed_types = {}
+
+    diff = DeepDiff(defaults, overrides, verbose_level=2)
+
+    if 'dictionary_item_added' in diff.keys():
```
sigmavirus24:
```py
items_added = diff.get('dictionary_item_added', {})
for old_key, value in items_added.items():
   added[root_cleanup(old_key)] = value
```

scripts/migrate-yaml.py
```diff
@@ -0,0 +1,164 @@
...
+        for old_key in diff['dictionary_item_added'].keys():
+            added[root_cleanup(old_key)] = diff['dictionary_item_added'].pop(old_key)
+    if 'dictionary_item_removed' in diff.keys():
+        for old_key in diff['dictionary_item_removed'].keys():
+            removed[root_cleanup(old_key)] = diff['dictionary_item_removed'].pop(old_key)
```
prometheanfire: This whole section has lines that are too long and I don't have a good way of shortening it.  If someone has suggestions...

prometheanfire:
The only issue I have with this is how nesting is shown for changed sets.

If I add the following it shows up correctly

```
fooasd:
  alsdk:
    needs:
      more: nesting
  this: is_a_test
```

But if I change an existing var it'll be output as the following

`horizon_custom_uploads.logo.dest: foo`

scripts/migrate-yaml.py
```diff
@@ -0,0 +1,193 @@
...
+    except IOError as e:
+        if e.errno == errno.ENOENT:
+            loaded_data = None
+        else:
+            raise e
+    else:
+        try:
+            loaded_data = yaml.safe_load(data)
+        except yaml.ParserError as e:
+            raise e
```
sigmavirus24: You're still re-raising this which means you're going to get a stack trace when you run the script against bad YAML... Can you catch this in main() and print a helpful error message maybe?

scripts/migrate-yaml.py
```diff
@@ -0,0 +1,193 @@
...
+
+def root_cleanup(var):
+    """Cleans up what deepdiff returns so we can have some cleaner output.
+
+    :param var: the key you wish to clean up, just a string
+    :return: the cleaned up string
+    """
+    prefix = "root['"
+    suffix = "']"
+    var = var[var.startswith(prefix) and len(prefix):]
```
sigmavirus24: I just noticed this and I'm not thrilled about it. I know it works, but it's too clever and you're hiding things in the slice. It'd be more explicitly written using if statements. I don't know if other people agree.

Apsu:
RPC-AIO-Liberty-PR: Fail

scripts/migrate-yaml.py
prometheanfire: fixed both in place by calling SystemExit

scripts/migrate-yaml.py
prometheanfire: changed

Apsu:
RPC-AIO-Ceph-Liberty-PR: Fail

scripts/migrate-yaml.py
```diff
@@ -0,0 +1,192 @@
...
+
+def root_cleanup(var):
+    """Cleans up what deepdiff returns so we can have some cleaner output.
+
+    :param var: the key you wish to clean up, just a string
+    :return: the cleaned up string
+    """
+    prefix = "root['"
+    suffix = "']"
+    if var.startswith(prefix):
```
sigmavirus24: Thanks for updating this. 👍

Apsu:
RPC-AIO-Liberty-PR: Fail

sigmavirus24:
Pending gates and some documentation for support, 👍

d34dh0r53:
👍 Looks good

Issue #190, opened 2015-06-24
f5-config.py creates priority groups for repo servers


Mudpuppy12:
The current configuration for f5-configy.py scripts creates a pool for the repo servers with a priority group of 100, 95, 90 with a balancing scheme of fastest-node. This forces all the traffic to go through a single node that we discovered. They should be set to all the same priority with least connection node for balancing.

current output

create ltm pool /QE-IAD3-1/QE-1_POOL_REPO { load-balancing-mode fastest-node members replace-all-with { QE-1_NODE_569038-infra03_repo_container-45441f9f:8181 { priority-group 100 } QE-1_NODE_569037-infra02_repo_container-20bae48b:8181 { priority-group 95 } QE-1_NODE_568575-infra01_repo_container-9c514b02:8181 { priority-group 90 } } min-active-members 1 monitor /QE-IAD3-1/QE-1_MON_HTTP_REPO }

something like

create ltm pool /QE-IAD3-1/QE-1_POOL_REPO { load-balancing-mode least-connections-node members replace-all-with { QE-1_NODE_569038-infra03_repo_container-45441f9f:8181 { priority-group 100 } QE-1_NODE_569037-infra02_repo_container-20bae48b:8181 { priority-group 100 } QE-1_NODE_568575-infra01_repo_container-9c514b02:8181 { priority-group 100 } } min-active-members 1 monitor /QE-IAD3-1/QE-1_MON_HTTP_REPO }

----------------------------------------------------------------------------------------------------------------------------
###### For each pull request, update the branch line to reference it.
- [x] master (https://github.com/rcbops/rpc-openstack/pull/1290)
- [x] mitaka-13.0 (#1322)
- [x] liberty-12.2 (https://github.com/rcbops/rpc-openstack/pull/1337)
- [x] kilo (https://github.com/rcbops/rpc-openstack/pull/1336)

##### Upgrade impact. Consider the impact on moving between major, minor and patch releases.
- [x] This change does not impact upgrades
- [ ] This change addresses a potential upgrade impact
- [ ] The upgrade impact is being tracked by another issue.

The issue tracking the upgrade impact is:

##### Testing impact
- [ ] An issue/pull-request has been create or exists to track the testing gap, and is referenced below
- [x] This change includes additional testing
- [ ] This change is tested by the existing setup

Additional testing links:

##### Release note
- [ ] This change includes a release note.
- [x] This change does not require a release note.

##### Does your change require documentation?
[...]

PR #1337, opened 2016-08-18
Enable active-active balancing of repo servers

mancdaz:
Prior to this commit repo servers were configured in an F5 pool
which assigns a unique priority to each server. This means
that all requests are forwarded to the first server in the pool
until that server become unavailible. This is active-passive
load balancing.

This commit disables prioritization of of repo servers, which
enables active-active load balancing and forwards connections
under the fastest-connection policy. All servers in the pool
are now treated equally. This change is benificial in that the
load is now spread throughout the repo servers and the
deployment can benifit from the combined power.

Connects #190

(cherry picked from commit 5340c58d8574873901638c3968c2e86d2b12d34e)

### DIFF
--- a/scripts/f5-config.py
+++ b/scripts/f5-config.py
@@ -326,7 +326,6 @@ POOL_PARTS = {
         'backend_port': 8181,
         'mon_type': '/' + PART + '/' + PREFIX_NAME + '_MON_HTTP_REPO',
         'group': 'pkg_repo',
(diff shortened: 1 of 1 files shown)


hughsaunders:
👍 I don't know about F5s or the script, but I approve of spreading the load, and matching HAProxy behaviour.

sigmavirus24:
👍

PR #1340, opened 2016-08-19
Add the RPC plays for upgrading

andymcc:
Take the playbooks that were called as part of the old rpc-upgrade.sh
script and add them to the run-upgrade.yml playbook as includes.

Additionally, there is a final task that was run as part of the script
that should now run against "hosts" after the logging is setup.

Connects https://github.com/rcbops/u-suk-dev/issues/298

### DIFF
--- a/rpcd/playbooks/run-upgrade.yml
+++ b/rpcd/playbooks/run-upgrade.yml
@@ -17,3 +17,21 @@
   hosts: all
   user: root
   tasks: []
+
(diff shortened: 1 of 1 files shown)


git-harry:
This seems to be breaking out deploy.sh, while the card does imply that I'm not sure this was every explicitly discussed. Deconstructing deploy.sh is a lot of extra work, this is probably worth some discussion in our meeting later.

hughsaunders:
recheck_swift

Issue #1345, opened 2016-08-22
The horizon-extensions upgrade fails on repo-build


d34dh0r53:
##### What type of issue is this? Please tick the appropriate box.
- [x] bug - existing functionality that does not work for you as documented
- [ ] feature - new functionality that you would like to see added

##### What is the expected behavior?
The horizon-extensions pip should build correctly

##### What is actually happening?
```
Collecting horizon_extensions from git+file:///var/www/repo/openstackgit/horizon-extensions@master#egg=horizon_extensions&gitname=horizon-extensions&projectgroup=all&subdirectory=horizon-extensions/ (from -c /var/www/repo/os-releases/13.3.1/requirements_constraints.txt (line 11))
  Cloning file:///var/www/repo/openstackgit/horizon-extensions (to master) to /tmp/openstack-builder/horizon-extensions
  Running command git clone -q file:///var/www/repo/openstackgit/horizon-extensions /tmp/openstack-builder/horizon-extensions
  Running command git show-ref
  fb690bc528499bbf9aebba3ab0cce0b4dffd9e35 refs/heads/master
  fb690bc528499bbf9aebba3ab0cce0b4dffd9e35 refs/remotes/origin/HEAD
  fb690bc528499bbf9aebba3ab0cce0b4dffd9e35 refs/remotes/origin/master
  b9b7a9acd9947ae5a0b12f195a136515b9d5e737 refs/tags/10.0.0rc1
  fe79499dc554a98c0c32d587541335e01309443d refs/tags/10.0.0rc4
  ca1bff63690919ec3ae76e5cffd502379343dceb refs/tags/10.0.0rc5
  1a41afa384800f7b13433560e664c536236a1c36 refs/tags/11.0.0rc1
  d2ddea42ca66f251555dde7d7dcc02e34ecadb76 refs/tags/11.0.0rc2
  25603468347d2adca40f99b1e1d1601feb4e2b2b refs/tags/11.0.0rc3
  d12619e468c6c0ae33245ce8a6d70fdd51358cd6 refs/tags/11.0.0rc4
  d0a8fa7831b5560903d8a7def88d0763511849ae refs/tags/11.0.0rc5
  3a3f0e9a71558a5877250e608407ab3bc538493e refs/tags/9.0.1
  a3378af1c5a6ce66f7a2f76f23facb149090e931 refs/tags/r11.0.0
  7d1c13964e5068805910c9def7ea53e6962f7060 refs/tags/r11.0.0rc10
[...]

Issue #357, opened 2015-08-20
MaaS playbooks fail on duplicate entities


sigmavirus24:
When MaaS has duplicate entries for an entity, the parsing in the playbooks can fail which causes the playbooks to fail and be unusable. I'm not sure how this should look exactly, but this is something we need to fix.

@BjoernT can you fill in more detail about this (since you seem to know MaaS better than the rest of us)?

git-harry:
The playbooks should fail on duplicate entities. There should never be duplicate entities, that is to say entities with the same label, if there are that needs investigating and cleaning up before running the maas playbook.

nrb:
Agreed with @git-harry, we did however see duplicate checks (not entities, but a specific disk space check).

claco:
Halting with an error is the desired outcome for this problem as it should be rare and requires MaaS intervention upstream.

PR #1348, opened 2016-08-23
[WIP] Create upgrade script for gating purposes

mattt416:
This hasn't been tested; it's being updated so we can start to hash it
out in earnest.

TBD: 1. What should this script be called?  Will it be used by jenkins
        only?
     2. How do we handle the upgrade of rcp-openstack bits?  In liberty
        version this was done by running scripts/deploy.sh.
     3. Do we need to run pip-lockdown.yml?  We will need to test this
        without running that.
     4. Is there any reason why we'd need to re-run the ceph playbooks
        when we're not changing ceph roles or the deployed version of
        ceph?
     5. Should this be a script or playbook?  This can easily be
        converted to a playbook but this is largely just documenting
        the manual steps required to perform the upgrade.
     6. The variable munging stuff can potentially be moved out into a
        separate playbook/script which can then be included here and by
        support, but we will need to wait to determine how exactly this
        is going to be run (do we use --danger DANGER or not?) before
        we do that work.

BLOCKERS: 1. migrate-yaml.py is stil being worked on and at present
             will not create 100% usable variables files.

TODO: 1. This script should get confirmation from the user before
         starting the upgrade.  This is dangerous command and not
         getting validation from the user before running could be
         catastrophic.
      2. Since openstack-ansible's run-upgrade.sh also runs
         bootstrap-ansible.sh, we need to find a way to skip that
         when we run it here.

Connected https://github.com/rcbops/u-suk-dev/issues/357
Connected https://github.com/rcbops/rpc-openstack/issues/1345
Connected https://github.com/rcbops/u-suk-dev/issues/347


sigmavirus24:
Should this move #357 into needs review?

alextricity25:
> 2. How do we handle the upgrade of rcp-openstack bits? In liberty
   version this was done by running scripts/deploy.sh.

I'm not sure why we would call deploy.sh. Essentially, if we re-run that with DEPLOY_OA='false', the difference between that and site.yml are: 
* Run pw-token-gen.py on user_osa_secrets.yml and user_rpco_secrets.yml (not sure if it would overwrite what's already there)
* Run verify-mass.yml

I'd say there is nothing wrong with site.yml. It's a little more clear IMO.

sigmavirus24:
@alextricity25 I believe there was an issue to get rid of `site.yml`.

mattt416:
@alextricity25 @sigmavirus24 One issue I know of is that a deploy may not have MaaS deployed.  What confuses me though is that in https://github.com/rcbops/rpc-openstack/blob/liberty-12.2/scripts/upgrade.sh we did not set DEPLOY_MAAS so that script would never upgrade the MaaS bits.  Anyone know how we handled this?

alextricity25:
@mattt416 That's a good catch. I'm not sure about that one TBH, but what you said makes perfect sense. It doesn't look like the MaaS bits were ever upgraded.

alextricity25:
@mattt416 probably explains why support included that as part of their maintenance plan:
https://github.com/rsoprivatecloud/Maintenance-Plans/blob/master/RPC%2011.x:%20Upgrade%20to%20RPC%2012.x#L268

sigmavirus24:
> Anyone know how we handled this?

Upgrades for 12.2 were tricky w/r/t MaaS. We changed how MaaS worked and thus needed separate playbooks for upgrading MaaS

alextricity25:
This worked on a basic L->M upgrade for me(no MaaS, no ceph, no openstack resources provisioned). @mattt416, i've given my answers to your questions below:

> TBD: 1. What should this script be called? Will it be used by jenkins
    only?

If this script is only meant to be run by devs or the gate, then test-upgrade.sh is fine

> How do we handle the upgrade of rcp-openstack bits? In liberty
    version this was done by running scripts/deploy.sh.

In my opinion we shouldn't be using deploy.sh for upgrading he RPCO bits. Things could go wrong if people have some left over environment variables set or forget to set the right ones(i.e DEPLOY_OA). I don't see anything wrong with just invoking the playbooks included in site.yml, and wrapping setup-maas.yml around a conditional checking if DEPLOY_MAAS is set.

> Do we need to run pip-lockdown.yml? We will need to test this
    without running that.

I'll test that now. I'll edit this answer once it's finished.

> Is there any reason why we'd need to re-run the ceph playbooks
when we're not changing ceph roles or the deployed version of
ceph?

I think we should worry about that once we get https://github.com/rcbops/u-suk-dev/issues/209 in the queue :)

> Should this be a script or playbook? This can easily be
converted to a playbook but this is largely just documenting
the manual steps required to perform the upgrade.

Keep a script for now, and see how important is to convert it to a playbook later. 

> The variable munging stuff can potentially be moved out into a
separate playbook/script which can then be included here and by
support, but we will need to wait to determine how exactly this
is going to be run (do we use --danger DANGER or not?) before
we do that work.

I honestly don't think this should be automated at all in practice. For the sake of testing though, using --danger for now is okay.

git-harry:
@mattt416 @sigmavirus24 @alextricity25 if you look at site.yml, it does not match what is in deploy.sh. In addition, deploy.sh uses environment variables which site.yml cannot understand, the use of site.yml and deploy.sh are not equivalent. We have a slightly muddled interface for enabling different parts of the product, we have to set env and conf bits for Ansible but then also have additional variables used by the scripts. I think that deploy.sh and the upgrade process should use the same method to run the RPCO playbooks.

@mattt416 if you are going to solving everything in the one commit, please make sure to add additional cards to the backlogs for the things that need to be revisited to make them more support friendly.

mattt416:
Closing this as the expectation is that this work is done in smaller chunks.

Issue #1346, opened 2016-08-23
migrate-yaml.py not merging lists correctly


mattt416:
###### Please provide the output to 'git describe --all' for  all deployments where you have confirmed it occurs.

master (mitaka)

##### What type of issue is this? Please tick the appropriate box.
- [x] bug - existing functionality that does not work for you as documented
- [ ] feature - new functionality that you would like to see added

##### What is the expected behavior?

The migrate-yaml.py script should merge variable files correctly.

##### What is actually happening?

If the DeepDiff library sees a changed element in a list, it returns just that changed element.  That means migrate-yaml.py needs to go back and find out what the full list is, and then return the list w/ the changed element.  A quick example:

```
root@liberty-aio:/opt/rpc-openstack/scripts# cat 1 2 test.py
repo_build_upper_constraints_overrides:
- elasticsearch<2.1.0
- foo
repo_build_upper_constraints_overrides:
- elasticsearch>=2.1.0
- foo
from deepdiff import DeepDiff
import yaml

f1 = yaml.load(open('1', 'r'))
f2 = yaml.load(open('2', 'r'))

d = DeepDiff(f1, f2)
print d['values_changed']
root@liberty-aio:/opt/rpc-openstack/scripts# python test.py
{"root['repo_build_upper_constraints_overrides'][0]": {'new_value': 'elasticsearch>=2.1.0', 'old_value': 'elasticsearch<2.1.0'}}
root@liberty-aio:/opt/rpc-openstack/scripts#
```

##### What were you attempting to do when you encountered this error (give as much detail as possible)? 

Merge old user_variables.yml file on a liberty deploy w/ the new variable files in rpco's master branch.

##### How can we reproduce this issue?

Deploy a liberty AIO, then run:

```
# /opt/rpc-openstack/scripts/migrate-yaml.py --overrides /etc/openstack_deploy/user_variables.yml --danger-mode DANGER
```

You'll see the following returned:

```
repo_build_upper_constraints_overrides'][0]: elasticsearch<2.1.0
```

##### How can we verify your issue is addressed?

Ensure that changed lists have the entire list rewritten to output, not just the changed element.

##### What is your evaluation on the importance of this issue, and why?

This blocks release of r13.1.0rc1.

----------------------------------------------------------------------------------------------------------------------------
###### For each pull request, update the branch line to reference it.
- [ ] master
- [ ] mitaka-13.0
- [ ] liberty-12.2
- [ ] kilo

[...]

PR #1349, opened 2016-08-23
Allow processing of lists in migrate-yaml

prometheanfire:
Fixes: #1346 

Connects #1346

### DIFF
--- a/scripts/migrate-yaml.py
+++ b/scripts/migrate-yaml.py
@@ -16,24 +16,13 @@
 # (c) 2016, Matthew Thode <matt.thode@rackspace.com>
 
 """
-Print out values that have changed, by default just additions and changes.
-
-Additions can be taken as is and placed into an overrides file, but should
-you should verify you still want to keep it.
-
-Removals are not printed by default because not having an override is the
-same as taking a default.
-
-Changes are printed, showing what you overrode and what is getting overridden.
-
-CAUTION: danger-mode is only if you KNOW what you are overriding doesn't
-    need to change, warrantee is void if used.
+Write a file containing new overrides with the new defaults stripped from the
+old overrides and conflicts listed in their own keys.
 """
 
 from __future__ import print_function
 import argparse
-from collections import namedtuple
-from deepdiff import DeepDiff
+import os
 import yaml
 
 
@@ -41,28 +30,11 @@ def parse_args():
     parser = argparse.ArgumentParser()
     parser.add_argument('--overrides', dest='overrides', required=True,
                         help='The path to the file with overrides.')
-    parser.add_argument('--info', dest='info', type=bool, default=False,
-                        help='Shows what is NOT overridden.')
-    parser.add_argument('--danger-mode', dest='danger',
-                        type=bool, default=False,
-                        help='Print output usable as an overrides file.')
+    parser.add_argument('--defaults', dest='defaults', required=True,
+                        help='The path to the file with defaults.')
+    parser.add_argument('--output-file', dest='output_file', required=True,
+                        help='The path to the new overrides file location')
     return parser.parse_args()
-
-
-def root_cleanup(var):
-    """Cleans up what deepdiff returns so we can have some cleaner output.
-
-    :param var: the key you wish to clean up, just a string
-    :return: the cleaned up string
-    """
-    prefix = "root['"
-    suffix = "']"
-    if var.startswith(prefix):
-        var = var[len(prefix):]
-    if var.endswith(suffix):
-        var = var[:-len(suffix)]
-    var = var.replace("']['", '.')
-    return var
 
 
 def parsed_yaml_from(path):
@@ -81,8 +53,8 @@ def parsed_yaml_from(path):
     else:
         try:
             loaded_data = yaml.safe_load(data)
-        except yaml.ParserError:
-            raise SystemExit('The yaml in %s is non-compliant and failed'
+        except yaml.parser.ParserError:
+            raise SystemExit('The yaml in %s is non-compliant and failed '
                              'to load' % path)
 
     if loaded_data is None:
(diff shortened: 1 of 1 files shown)


prometheanfire:
so.. this script assumes the following

* support update overrides based on release notes

This is needed if the old default is `foo=foo-alpha`, the old override is `foo=foo-alpha`, the new default (mentioned in a release note) is `foo=foo-beta`.  If support does not update their override (or preferably remove it), the new override will be `foo=foo-alpha` and could break things.

As long as this is done we'll be fine (we are generating release notes with this info in mitaka), the hard part is that we haven't been doing release notes in liberty.  Once this migration has been done support is already tasked with reading the release (upgrade) notes and updating their overrides as needed.

rpcd/etc/openstack_deploy/user_rpco_variables_defaults.yml
```diff
@@ -141,4 +141,4 @@ secure_cluster_flags:
 
 # We need to override the OpenStack upper constraint for Mitaka which is version 1.9.0, We should be able to remove this in Newton as the upper constraint moves to 2.3.0
 repo_build_upper_constraints_overrides:
-  - "elasticsearch==2.3.0"
+  - "elasticsearch===2.3.0"
```
mattt416: Hi @prometheanfire, why the need for this change?

prometheanfire: well, it now matches the syntax requirements lays down, not sure if strictly needed, but more of a correctness fix

scripts/migrate-yaml.py
```diff
@@ -91,6 +97,36 @@ def parsed_yaml_from(path):
...
+        for list_name in list_of_lists:
+            if key.startswith(list_name):
+                if list_name in obj.keys():
+                    obj[list_name].append(obj[key])
+                else:
+                    obj[list_name] = [obj[key]]
+    # remove the old bad 'list keys'
+    for key in obj.keys():
+        for list_name in list_of_lists:
+            if key.startswith(list_name):
```
prometheanfire: can probably combine the ifs

mattt416:
@prometheanfire spotted one issue w/ lists.  If my user_variables.yml has this:

```
repo_build_upper_constraints_overrides:
- elasticsearch<2.1.0
- test
```

... and user_rpco_variables_defaults.yml has this:

```
repo_build_upper_constraints_overrides:
  - "elasticsearch==2.3.0"
  - "test"
```

... I'll get this spit out using migrate-yaml.py:

```
==== added the following ====
repo_build_upper_constraints_overrides:
- elasticsearch<2.1.0


==== removed the following list items ====
repo_build_upper_constraints_overrides:
- elasticsearch==2.3.0
```

That is using the --combine-old-new-lists flag, and if I add --danger DANGER, I see:

```
repo_build_upper_constraints_overrides:
- elasticsearch<2.1.0
- elasticsearch==2.3.0
```

Since both have the `- test` list item, I'd expect that to remain.  Even tho the `- test` item is in the defaults file, ansible does not merge lists specified in different overrides files (at least, to my knowledge!)  It does have a `hash_behavior` option which can merge hashes, but we're not using this anywhere.

mattt416:
@prometheanfire, noticed the following behaviour w/ dicts:

My user_variables.yml has this:

```
horizon_custom_uploads:
  favicon:
    dest: img/favicon.ico
    src: '{{ rackspace_static_files_folder }}/favicon.ico'
  logo:
    dest: img/logo.png
    src: '{{ rackspace_static_files_folder }}/logo.png'
  logo-splash:
    dest: img/logo-splash.png
    src: '{{ rackspace_static_files_folder }}/logo-splash.png'
  test:
    dest: this
    src: that
```

... and my user_osa_variables_defaults.yml has this:

```
horizon_custom_uploads:
  logo-splash:
    src: "{{ rackspace_static_files_folder }}/logo-splash.png"
    dest: img/logo-splash.png
  logo:
    src: "{{ rackspace_static_files_folder }}/logo.png"
    dest: img/logo.png
  favicon:
    dest: img/favicon.ico
```

When I run migrate-yaml.py, I see this:

```
horizon_custom_uploads.test:
  dest: this
  src: that
```

Is that expected?

prometheanfire:
yes

alextricity25:
I noticed the same @mattt416 did with the dictionary when using the following user_variables.yml entry:
```
neutron_neutron_conf_overrides:
  DEFAULT:
    l3_ha: False
  nova:
    endpoint_type: internal
```
This entry isn't defined in /opt/rpc-openstack/rpcd/etc/openstack_deploy/user_osa_variables_defaults.yml, but the migrate-yaml.py script still strips out the "l3_ha" part:
```
neutron_neutron_conf_overrides.nova:
  endpoint_type: internal
```

prometheanfire:
@alextricity25 --info should get you all removed items iirc.

it was brought up that it might be better to not combine the two defaults files and compare them to one file but to compare files based on their use.  Just compare osa defaults to osa overrides and likewise with rpco.  This would change the script to take in two files, the default and overrides but sounds like a good idea.

The mapping we would put in the upgrade script would be like this iirc.

user_osa_defaults == user_variables
user_rpco_defaults == user_extra_variables

Let me know if this change is wanted and I can make it fairly easily.

rpcd/etc/openstack_deploy/user_rpco_variables_defaults.yml
odyssey4me: Please don't do that. We change what requirements lays down to an upper cap and that change of format is not applied to the contents of this var list, so if pip doesn't understand it then it will break the repo build.

mattt416:
Hi @alextricity25, that weird dict behaviour is because it was assumed that the dicts in different files would merge once ansible runs but as that is not the case we need to just print out the entire dict from the original overrides file.  The same should happen with lists -- if a list is specified in the  overrides file and it either doesn't exist in the defaults file or is different than the default value then it should just spit out the original value.  It'd be great if it could add a comment before (`--danger DANGER` mode here) showing the updated default and instructing the deployer to examine the difference.

On a side note @prometheanfire, now that we're not doing any weird list/dict merging do you actually need to use the deepmerge library?  Not having this dependency would actually be good.

mattt416:
Since you can't insert comments w/ PyYAML, how about doing something like this where conflicting defaults are added to the YAML itself:

```
MITAKA_UPGRADE_CONFLICTING_DEFAULTS:
  cinder_service_backup_program_enabled: false
  keystone_database_max_pool_size: '{{ db_max_pool_size }}'
  keystone_database_pool_timeout: '{{ db_pool_timeout }}'
  keystone_token_driver: sql
  keystone_token_provider: uuid
  neutron_db_max_overflow: '{{ db_max_overflow }}'
  neutron_db_max_pool_size: '{{ db_max_pool_size }}'
  neutron_db_pool_timeout: '{{ db_pool_timeout }}'
  neutron_neutron_conf_overrides:
    DEFAULT:
      l3_ha: false
  nova_db_max_overflow: '{{ db_max_overflow }}'
  nova_db_max_pool_size: '{{ db_max_pool_size }}'
  nova_db_pool_timeout: '{{ db_pool_timeout }}'
aodh_db_ip: 172.29.236.100
aodh_db_port: 27017
aodh_db_type: mongodb
apply_security_hardening: false
ceilometer_api_workers: 2
ceilometer_collector_workers: 2
ceilometer_db_ip: 172.29.236.100
ceilometer_db_port: 27017
ceilometer_db_type: mongodb
ceilometer_notification_workers: 2
cinder_osapi_volume_workers: 2
cinder_service_backup_program_enabled: true
debug: true
galera_innodb_buffer_pool_size: 512M
galera_innodb_log_buffer_size: 32M
galera_wsrep_provider_options:
- option: gcache.size
  value: 32M
glance_api_workers: 2
glance_default_store: rbd
glance_registry_workers: 2
heat_api_workers: 2
heat_engine_workers: 2
horizon_wsgi_processes: 2
horizon_wsgi_threads: 2
keystone_database_max_pool_size: 120
keystone_database_pool_timeout: 60
keystone_token_driver: keystone.token.persistence.backends.sql.Token
keystone_token_provider: keystone.token.providers.uuid.Provider
keystone_wsgi_processes: 2
lxc_cache_resolvers:
- nameserver 172.16.38.2
lxc_container_template_apt_components:
- main
- universe
lxc_net_address: 10.255.255.1
lxc_net_dhcp_range: 10.255.255.2,10.255.255.253
lxc_net_netmask: 255.255.255.0
neutron_db_max_overflow: 60
neutron_db_max_pool_size: 120
neutron_db_pool_timeout: 60
neutron_l3_agent_ini_overrides:
  DEFAULT:
    handle_internal_only_routers: true
neutron_metadata_workers: 1
neutron_ml2_conf_ini_overrides:
  ml2:
    extension_drivers: ''
neutron_neutron_conf_overrides:
  nova:
    endpoint_type: internal
neutron_rpc_workers: 1
nova_conductor_workers: 2
nova_db_max_overflow: 60
nova_db_max_pool_size: 120
nova_db_pool_timeout: 60
nova_libvirt_images_rbd_pool: vms
nova_metadata_workers: 2
nova_osapi_compute_workers: 2
repo_build_upper_constraints_overrides:
- elasticsearch<2.1.0
swift_account_server_workers: 2
swift_container_server_workers: 2
swift_object_server_workers: 2
swift_proxy_server_workers: 2
tempest_public_subnet_cidr: 172.29.248.0/22
tempest_volume_backup_enabled: true
```

The operator can then go through and figure out what needs fixing and then remove the `MITAKA_UPGRADE_CONFLICTING_DEFAULTS` entry.

prometheanfire:
Ya, once we figure out the final behavior we can prune the code.

The first merge is going to be by hand and then I'll make the program output that same data (comments notwithstanding).

I get demo files we can use 'soon'.

scripts/migrate-yaml.py
```diff
@@ -96,97 +67,55 @@ def do_the_diff(defaults, overrides):
...
+                                explicit_start=True))
+        fh.flush()
+
+    # Add the warrantee sticker
+    print('The output assumes you have been pruning your old overrides '
+          'files as needed and as suggested by release notes.')
+    if 'NEW_DEFAULTS' or 'OLD_OVERRIDES' in returned_combined.keys():
+        print('\nYour output has conflicts, examine it to make sure '
+              'values are as you wish them to be, check "new_defaults" and '
+              '"old_overrides" specifically as you may or may not need '
```
mattt416: this should be `NEW_DEFAULTS` and `OLD_DEFAULTS`

prometheanfire: I don't think so, at least personally I'm considering these to be new defaults and old overrides.  But that's just how I see it, man.

http://s2.quickmeme.com/img/1e/1e48392d8df59bcfbc11f7decbcac7a1155839151e5fece526e011d44ccd468e.jpg

scripts/migrate-yaml.py
```diff
@@ -96,97 +67,55 @@ def do_the_diff(defaults, overrides):
...
-        print(yaml.safe_dump(returned_diff.changed_added,
-                             default_flow_style=False))
-    if returned_diff.changed_removed and args.danger is False:
-        print('\n==== changed the following (removed) ====')
-        print(yaml.safe_dump(returned_diff.changed_removed,
-                             default_flow_style=False))
+    returned_combined = do_the_diff(defaults=parsed_yaml_from(args.defaults),
+                                    overrides=parsed_yaml_from(args.overrides))
+
+    with open(args.output_file, 'wb') as fh:
```
mattt416: Any value in checking if `output_file` exists, before overwriting it?  May be worth just throwing an error if it already exists so we don't allow a deployer to inadvertently blow away an existing file.

mattt416:
Works as expected, but a few suggestions before this goes in.

scripts/migrate-yaml.py
prometheanfire: Fair point, I'll wait til tomorrow before I add this type of thing so the UK can comment if they want.

scripts/migrate-yaml.py
mattt416: Referring to the lower case reference in the error message.

prometheanfire: oh, just the case, ya, next round then

scripts/migrate-yaml.py
```diff
@@ -16,55 +16,26 @@
...
     parser.add_argument('--overrides', dest='overrides', required=True,
                         help='The path to the file with overrides.')
-    parser.add_argument('--info', dest='info', type=bool, default=False,
-                        help='Shows what is NOT overridden.')
-    parser.add_argument('--danger-mode', dest='danger',
-                        type=bool, default=False,
-                        help='Print output usable as an overrides file.')
+    parser.add_argument('--defaults', dest='defaults', required=True,
+                        help='The path to the file with defaults.')
+    parser.add_argument('--output-file', dest='output_file', required=True,
```
sigmavirus24: argparse allows you to define this type as a File so you get an open file handler

alextricity25:
I like this new behavior and would love to see some more documentation around usage.
Maybe the upgrade docs are a good place for that, but last I checked L->M upgrade docs weren't up yet.
Giving this my :+1: now to get this rolling

scripts/migrate-yaml.py
prometheanfire: done

scripts/migrate-yaml.py
prometheanfire: done

sigmavirus24:
👎 until Matt's and my feedback is addressed.

Regarding whether or not we want to override `--output-file` we could always just open it append mode that way we never lose the previous run's data.

scripts/migrate-yaml.py
prometheanfire: I'd rather not change it at this point, but will if others really want to.

prometheanfire:
As far as docs go, I'm not sure if we should do that now or when the upgrade story is closer to completion.

scripts/migrate-yaml.py
```diff
@@ -96,97 +68,58 @@ def do_the_diff(defaults, overrides):
...
+                                    overrides=parsed_yaml_from(args.overrides))
+
+    if os.path.isfile(args.output_file):
+        raise SystemExit('The file you are creating already exists, '
+                         'please move it, we will not overwrite the file.')
+    with open(args.output_file, 'wb') as fh:
+        fh.write(yaml.safe_dump(returned_combined,
+                                default_flow_style=False,
+                                explicit_start=True))
+        fh.flush()
```
sigmavirus24: You shouldn't need to flush anything. When the file is closed it will be flushed. Did you see something that indicated otherwise?

Also, you're opening the file as bytes. Does `yaml.safe_dump` always return bytes or does it return text?

prometheanfire: ya, both are not needed, removed

sigmavirus24:
👀 Looks okay to me. 👍

mattt416:
👍

PR #1355, opened 2016-08-24
Remove horizon-extensions cache from repo_all

mattt416:

1. Removes the horizon-extensions clone from the repo servers as the
   repo has changed and cannot be updated using the method implemented
   in repo-install.yml.
2. Renames upgrade.sh to test-upgrade.sh.  This script will be used for
   testing and gating purposes only, and should not be used for
   production upgrades from Liberty->Mitaka.
3. Adds a new release note outlining step required to clean up repo
   server.

Connected https://github.com/rcbops/rpc-openstack/issues/1345

### DIFF
--- a/releasenotes/notes/rm-horizon-extensions-from-repo-server-3d441a4d0d1c6ad0.yaml
+++ b/releasenotes/notes/rm-horizon-extensions-from-repo-server-3d441a4d0d1c6ad0.yaml
@@ -1,0 +1,6 @@
+---
+upgrade:
+  - As the horizon-extensions repo changed in v13.0.0, operators
+    should run ``ansible -m file -a
+    "path=/var/www/repo/openstackgit/horizon-extensions state=absent"
+    repo_all`` prior to upgrading to this release.
--- a/scripts/test-upgrade.sh
+++ b/scripts/test-upgrade.sh
@@ -14,5 +14,18 @@
 # limitations under the License.
 #
 # (c) 2015, Nolan Brubaker <nolan.brubaker@rackspace.com>
-echo 'This release does not support upgrading from Liberty.'
-exit 1
+
+set -eux -o pipefail
+
+BASE_DIR=$( cd "$( dirname ${0} )" && cd ../ && pwd )
+OA_DIR="$BASE_DIR/openstack-ansible"
+RPCD_DIR="$BASE_DIR/rpcd"
+
+# TASK #1
+# Bug: https://github.com/rcbops/rpc-openstack/issues/1345
+# Issue: horizon-extensions now gets pulled from a different location; the
+#        git cache on the repo_all containers needs to be deleted so we can
+#        grab the new repo.  Failing to do this will cause repo-install.yml
+#        to fail to run.
+cd ${RPCD_DIR}/playbooks
+ansible -m file -a "path=/var/www/repo/openstackgit/horizon-extensions state=absent" repo_all


mattt416:
I ended up updating an existing relnote, should we backport that to 13.0.0rc1 for consistency?

alextricity25:
:+1: Thanks for getting the test-upgrade.sh in there :)

releasenotes/notes/horizon-extensions-in-tree-91f2ed817fde11aa.yaml
```diff
@@ -2,3 +2,8 @@
 other:
   - The horizon extensions are now provided in the
     rpc-openstack repository.
+upgrade:
```
sigmavirus24: @git-harry can confirm or deny, but I don't think we're supposed to amend release notes after the fact. Rather we should be generating new ones.

sigmavirus24:
Pending clarification on the use of release notes by @git-harry I'm 👍 to the rest of this.

releasenotes/notes/horizon-extensions-in-tree-91f2ed817fde11aa.yaml
git-harry: @sigmavirus24 makes a good point. As I understand it, the release note is listed under the first tag against which it appears. I think the original release note appears in r13.0.0rc1 and so I believe that means we should have a separate note. If the change was to correct or clarify the existing comment then it would makes sense but in this case I believe it should be separate.

mattt416:
Hi All,

As discussed, an alternative here would be to tear down the repo containers and simply rebuild.  This will ensure that we don't hit any other lurking issues w/ the repo containers.  The downside being that it may take a while longer to run repo-install.yml since we now have no cached pool.

If no objects, I'll get moving on this tomorrow.

--Matt

mattt416:
We discussed an alternate approach involving nuking the repo containers in our call today.  Let's hold off on merging this for the time being until we know which way we're going to go.

sigmavirus24:
> As discussed, an alternative here would be to tear down the repo containers and simply rebuild.

It seems I missed this discussion.

> The downside being that it may take a while longer to run repo-install.yml since we now have no cached pool.

Frankly I see that as a something that precludes this approach. We've had several meetings with support in which speed is of the utmost importance for them when performing an upgrade. As it is, Kilo to Liberty upgrades take all too long for their sensibilities. I'd prefer not destroying those containers.

git-harry:
@mattt416 @sigmavirus24 after further investigation I don't believe rebuilding the repo servers is of any use with respect to the purpose of this pull request

```
stable/mitaka:playbooks/repo-server.yml:          - "lxc.mount.entry=/openstack/{{ container_name }} var/www none bind 0 0"
```

/var/www is on the host and so rebuilding the containers will not have any impact on the existence of /var/www/repo/openstackgit/horizon-extensions. I think there may still be a case for rebuilding the containers, but this pull request should not be delayed by that discussion.

I would still like to get a little more information regarding the best way to add the release note. @odyssey4me would you mind reviewing the comments relating to the placement of the release note for this commit and giving your thoughts.

odyssey4me:
If the lxc-containers-destroy.yml playbook is used to destroy the containers, then the data on the bind mount goes with it thanks to https://github.com/openstack/openstack-ansible/blob/stable/mitaka/playbooks/lxc-containers-destroy.yml#L34.

Before throwing away the option of rebuilding the containers, it's worth verifying how long it takes. If the rpc-repo is up to date, the rebuild will simply be using pip to download the pre-built wheels from that as a source rather than downloading the source and completing the build. Regardless IIRC the repo build in Mitaka is *way* shorter than in Liberty/Kilo because we cut out the unnecessary global-requirements wheel build and parallelised as many actions as we could. IIRC it takes less than 15 mins for the complete container and repo rebuild from scratch, especially if you have pre-build wheels in one of the indexes used.

Of course another approach here is to simply remove the horizonextensions folder in repo/openstackgit  in the master repo container.

@git-harry IIRC if a release note is changed then it is detected as being new for this tag so when the release notes build again it will disappear from where it previously showed and only show in the new tag. This is only of importance if you've published the release note before. Within OSA the rule of thumb is that if the note has been published in a tag then it must be left alone. Be more careful with ensuring that they're correct when reviewing the patch that's putting it there.
If a release note was back-ported to a stable branch, leave the stable one alone. If the development head patch has a release note which needs corrections then go ahead and correct - just don't touch the stable one.

evrardjp:
👍  seems ok from the code and release note but not tested.

git-harry:
👍 tested the script, all looks good.

sigmavirus24:
recheck_all

PR #1367, opened 2016-08-30
Enable LBaaSv2 Horizon panels if LBaaSv2 in use

major:
This patch enables the LBaaSv2 panels in Horizon after a Mitaka
upgrade if LBaaSv2 is already in use. It can be run multiple times
if needed.

Connects rcbops/u-suk-dev#215

### DIFF
--- a/scripts/test-upgrade.sh
+++ b/scripts/test-upgrade.sh
@@ -20,6 +20,7 @@ set -eux -o pipefail
 BASE_DIR=$( cd "$( dirname ${0} )" && cd ../ && pwd )
 OA_DIR="$BASE_DIR/openstack-ansible"
 RPCD_DIR="$BASE_DIR/rpcd"
+OSA_VARS="/etc/openstack_deploy/user_osa_variables_defaults.yml"
 
 # TASK #1
 # Bug: https://github.com/rcbops/rpc-openstack/issues/1345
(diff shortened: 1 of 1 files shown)


major:
This isn't terribly elegant -- I know -- but I'm definitely open to suggestions.

mattt416:
For my own understanding -- test-upgrade.sh will be used for gate testing only.  Is this feature something we'll be testing in the gate?

major:
It should probably be tested in the gate since some customers might be using it as a tech preview in 12.2 or 13.0.  However, where should it live long-term so that it's in the normal upgrade path?

scripts/test-upgrade.sh
```diff
@@ -29,3 +30,15 @@ RPCD_DIR="$BASE_DIR/rpcd"
...
+
+# TASK #2
+# Bug: https://github.com/rcbops/u-suk-dev/issues/215
+# issue: Enable the Horizon panels for LBaaSv2 if the environment has LBaaSv2
+#        enabled already. Check for anything matching "LoadBalancerPluginv2"
+#        in the OSA config and enable the Horizon panels via the RPC defaults
+#        config.
+if grep -q "LoadBalancerPluginv2" /etc/openstack_deploy/user_variables.yml ; then
+    if ! grep -q "^horizon_enable_neutron_lbaas" "${OSA_VARS}" ; then
+        echo "horizon_enable_neutron_lbaas: True" | tee -a "${OSA_VARS}"
```
odyssey4me: Don't we already have this auto-enablement via group_vars in OSA? If not, then why not add it there?

git-harry: I would have thought it was a bit late in the day to be changing behaviour in stable/mitaka however if upstream with accept the change this sounds like the ideal solution to me. 

Alternatively, I think this is a documentation task, we just need to document that when setting the neutron var to enable LBaaSv2 it is recommended to also set "horizon_enable_neutron_lbaas: True". These vars should be set in /etc/openstack_deploy/user_osa_variables_overrides.yml.

As to testing, I don't believe we currently test this in the AIO, if it's still in tech preview and so not the default, I would be inclined to leave it out and rely on QE. We could also look to add additional AIO jobs. What is the likelihood that enabling it could hide issues that present when it's not enabled?

major:
Let's forget this PR. I'll try to get this done by default in OSA today.

PR #1372, opened 2016-08-31
Destroy repo_server containers in test-upgrade.sh

alextricity25:
This commit adds bits to the test-upgrade.sh so the repo_server
containers are destroyed per https://github.com/rcbops/u-suk-dev/issues/199

Connects https://github.com/rcbops/u-suk-dev/issues/199

### DIFF
--- a/scripts/test-upgrade.sh
+++ b/scripts/test-upgrade.sh
@@ -22,10 +22,11 @@ OA_DIR="$BASE_DIR/openstack-ansible"
 RPCD_DIR="$BASE_DIR/rpcd"
 
 # TASK #1
-# Bug: https://github.com/rcbops/rpc-openstack/issues/1345
-# Issue: horizon-extensions now gets pulled from a different location; the
-#        git cache on the repo_all containers needs to be deleted so we can
-#        grab the new repo.  Failing to do this will cause repo-install.yml
-#        to fail to run.
-cd ${RPCD_DIR}/playbooks
-ansible -m file -a "path=/var/www/repo/openstackgit/horizon-extensions state=absent" repo_all
+# Bug: https://github.com/rcbops/u-suk-dev/issues/199
+# Issue: To avoid any dependency issues that have occured in the past, the
+#        repo_server containers are re-built. This was done as part of the
+#        OSA upgrade script in liberty, but is not done in the OSA mitaka
+#        upgrade script. Thus, we destroy the containers here.
+cd ${OA_DIR}/playbooks
+openstack-ansible lxc-containers-destroy.yml --limit repo_all
+openstack-ansible setup-hosts.yml --limit repo_all


evrardjp:
Are we sure we want to use lxc-container-destroy?

This is gonna destroy logs, etc.

Maybe we should have a --skip-tags container-directories ?

sigmavirus24:
@evrardjp if there's a chance leaving logs around, etc. will cause instability/reproducibility problems for support, then it's likely we'll end up destroying them in the end anyway

scripts/test-upgrade.sh
```diff
@@ -29,3 +29,12 @@ RPCD_DIR="$BASE_DIR/rpcd"
 #        to fail to run.
 cd ${RPCD_DIR}/playbooks
 ansible -m file -a "path=/var/www/repo/openstackgit/horizon-extensions state=absent" repo_all
```
sigmavirus24: We don't need this if we're going to destroy the repo containers, right? Let's remove this now.

sigmavirus24:
This needs a bit more work. 👎

alextricity25:
@evrardjp: this is the same pattern OSA uses in their K->L upgrade script. Nothing new here.
@sigmavirus24 Good point! I'll remove that

rackernexus:
If you want to double-check your formatting, check [the preview](https://staging.developer.rackspace.com/staging.horse/build-2351c57221/docs/private-cloud/rpc/release-notes/).

hughsaunders:
👍  pending jenkins.

evrardjp:
fine for me if everybody agrees on what it does.

sigmavirus24:
👍 once the checks pass

prometheanfire:
:+1: once passing

alextricity25:
recheck_ceph

alextricity25:
:-1: In the OSA mitaka upgrade script, there is a [playbook](https://github.com/openstack/openstack-ansible/blob/stable/mitaka/scripts/run-upgrade.sh#L153) that targets the repo_server containers, and it runs before the setup-hosts.yml playbook.
I'm going to add ``setup-hosts.yml --limit repo_all`` here to rebuild the containers before said playbook is run.

rackernexus:
:mag_right: [Preview](https://staging.developer.rackspace.com/staging.horse/build-859ff14f7f/docs/private-cloud/rpc/release-notes/)

alextricity25:
Alright, I made the change. Please review the latest commit :)

git-harry:
👍  this looks good to me. OSA's run-upgrade.sh script does create some potential issues because it performs certain tasks related to the vars and env files however in practise I don't believe they will cause any problems.

mattt416:
LGTM 👍

PR #1375, opened 2016-09-01

git-harry:
the default value.

Connects #973

### DIFF
--- a/releasenotes/notes/override-l2pop-default-d7426c07d8b72f2e.yaml
+++ b/releasenotes/notes/override-l2pop-default-d7426c07d8b72f2e.yaml
@@ -1,0 +1,10 @@
+---
+issues:
+  - The default value for `neutron_l2_population` has been
+    overridden, which will require deployers to recreate
+    any VXLAN interfaces created with `neutron_l2_population`
+    set `False`. Not recreating these interfaces will lead to
+    connectivity issues for virtual machines.
+upgrade:
+  - Based on feedback from Support, the default value for
+    `neutron_l2_population` has been overridden to `True`.
--- a/rpcd/etc/openstack_deploy/user_osa_variables_defaults.yml
+++ b/rpcd/etc/openstack_deploy/user_osa_variables_defaults.yml
@@ -128,6 +128,11 @@ repo_build_pip_extra_indexes:
 # Docs: http://docs.openstack.org/developer/openstack-ansible-security/
 # apply_security_hardening: true
 
+## Enable Neutron l2_population
+# We are overriding the default value for neutron_l2_population. Please see
+# https://github.com/rcbops/rpc-openstack/issues/973 for further details.
+neutron_l2_population: True
+
 # Based on https://github.com/rcbops/rpc-openstack/issues/1149
 # L3HA has to be disabled until all major issues are fixed.
 neutron_neutron_conf_overrides:


rackernexus:
:eyes: [Preview](https://staging.developer.rackspace.com/staging.horse/build-ef28770f60/docs/private-cloud/rpc/release-notes/)

evrardjp:
👎  I don't agree with the decision taken for reasons I will send by email. 

Even if this is the decision is set in stone, I think we should properly document this, and properly give a documentation on how to switch back and forth to/away from l2pop.

git-harry:
recheck_all

sigmavirus24:
@evrardjp did you ever send that email? I have not seen it.

d34dh0r53:
I have not seen the email from @evrardjp either, but this does what it says it does so 👍 from me.

sigmavirus24:
I suspect the meat of @evrardjp's complaint to be that Neutron cores told OSA to turn off l2pop because it's inefficient and the more we diverge from OSA, the more @evrardjp becomes uncomfortable.

I think the thing to realize is that this isn't a matter of migrating off of l2pop, it's a matter of "not using l2pop has significantly degraded the customer experience and enabling l3ha degrades the customer experience." As much as it would be nice to just accept all of OSA's defaults, those defaults no longer seem to be driven by actual experience but instead by core reviewers on each project who have an ideal they're striving for. Those ideals, some times, are not practical or production ready when they think they are. We can't be restrained by those ideals otherwise we won't have customers or jobs.

prometheanfire:
Part of the problem is that upstream isn't testing with our settings anymore either.  I'm fine with setting this on upgrade where it is already set, as an override.  But I don't think we should set this to true as default. :-1:

prometheanfire:
as far as performance goes, are there open bugs related to this upstream?

prometheanfire:
also, the performance bugs got fixed in mitaka iirc....

sigmavirus24:
@prometheanfire upstream is not maintaining l2pop anymore either. I think @alextricity25 knows more details about our problems with l3ha and why it's not suitable.

prometheanfire:
I almost feel that we should define our desired state and then just work toward getting there, we seem to be going in circles.

sigmavirus24:
We can definitely put together Use Cases and Acceptance Criteria, plan out a spike, and work towards getting there. Let's start working with @b3rnard0 on planning that as part of r13.3 or some other version so we have time to gather the UC/AC from the right people, figure out who should work on the spike, etc.

cloader89:
Why not set it as a user configurable variable? Just document where this can be changed, add the variable inside the upgrade playbook.

prometheanfire:
I'm fine with working on it, I worry that if we push it to 13.3 that we'd be flip flopping on this, though keeping it as True isn't really a change, so it's just more pushing it down the road...

prometheanfire:
Users can already override it :D  as for a new variable for the upgrade, that's a possibility, but I'd rather avoid choosing a solution before we have the problem laid out.

cloader89:
I'm saying, have the variable be set to the previous value and if necessary, it'd update as its being upgraded =)

prometheanfire:
I think we should default to the preferred setting (meaning keeping it off).  In any case I don't think we should decide on a solution when we don't know the full problem domain.

git-harry:
Apologies for not updating this pull request. There was a meeting earlier today between engineering and support, the outcome was an agreement that L2 population should be enabled. Support are not currently able to manage use of the upstream default except in exceptional circumstances. A card, https://github.com/rcbops/u-suk-dev/issues/386, has been created to track the work required to support disabling the L2 population setting by default in a future release.

PR #1377, opened 2016-09-02

git-harry:
the default value.

Connects #973

(cherry picked from commit d2ff1b40cec8208ac6e116aff7405c67f931d4b4)

### DIFF
--- a/releasenotes/notes/override-l2pop-default-d7426c07d8b72f2e.yaml
+++ b/releasenotes/notes/override-l2pop-default-d7426c07d8b72f2e.yaml
@@ -1,0 +1,10 @@
+---
+issues:
+  - The default value for `neutron_l2_population` has been
+    overridden, which will require deployers to recreate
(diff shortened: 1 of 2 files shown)


sigmavirus24:
This is a straight-forward backport. Looks good. I have one question about how we backport release notes though: Do we need to regenerate the name of the file or is reno/our rackernexus stuff smart enough to know which release they belong to?

git-harry:
@sigmavirus24 as I understand it they will belong to both releases. We have two separate branches and so reno treats them as such. In cases such as this maybe we should consider removing the release note from master?

sigmavirus24:
> In cases such as this maybe we should consider removing the release note from master?

I wouldn't think so but I'm also not sure. I'm fine with leaving this as is as an experiment to see what happens.

👍 from me once the tests pass if it wasn't obvious from my last comment.

git-harry:
recheck_ceph

alextricity25:
:+1: Pending checks.

PR #1379, opened 2016-09-02

git-harry:
the default value.

Connects #973


d34dh0r53:
:+1:

sigmavirus24:
Looks to be a straight-forward backport. 👍

PR #1388, opened 2016-09-07
Allow unauthenticated APT Mirror

mancdaz:
Openstack infra provide APT mirrors, but they are not signed. In order
to make use of these for gating it is necessary to disable security
hardening rule V-38462.

Connects rcbops/u-suk-dev#300

(cherry picked from commit 72a51b435cefeacb1e851209d9ed4b66dd931475)


evrardjp:
recheck_swift

hughsaunders:
recheck_swift got serverfault which I suspect is the new incarnation of portnotfound

major:
I'm :+1: on this. We do a similar thing in OSA already.

mancdaz:
👍

PR #1390, opened 2016-09-07
Allow unauthenticated APT Mirror

mancdaz:
Openstack infra provide APT mirrors, but they are not signed. In order
to make use of these for gating it is necessary to disable security
hardening rule V-38462.

Connects rcbops/u-suk-dev#300

(cherry picked from commit 72a51b435cefeacb1e851209d9ed4b66dd931475)


hughsaunders:
recheck_ceph

evrardjp:
recheck_ceph

major:
:-1: We don't have security hardening in kilo? Is this necessary?

hughsaunders:
Excellent point @major.

PR #1395, opened 2016-09-07
Add OSA upgrade script to test-upgrade.sh

git-harry:
One of the steps when performing an upgrade from Liberty to Mitaka will
be to run OSA's upgrade script, run-upgrade.sh. To enable us to perform
a tested upgrade this commit adds run-upgrade.sh to the RPCO test script
test-upgrade.sh.

A release note is also added to ensure that security hardening is
disabled where necessary. Currently deploy.sh manages that with an
enviroment variable but that is not feasible in this situation.

Connected https://github.com/rcbops/u-suk-dev/issues/392

PR #1396, opened 2016-09-07
Added RPCO bits to test-upgrade and..

alextricity25:
Also moved the rpc playbook bits from deploy.sh into
it's own script, deploy-rpc-playbooks.sh. This is so
those bits can be easily included in test-upgrades.sh and in
deploy.sh.
I moved the run_ansible function previously defined in deploy.sh to a new file, ``functions.sh``, that is
then sourced by both ``deploy.sh`` and ``test-upgrades.sh``.
The bash environment variables is another matter. For now, I just exported what seemed necessary into the environment so the variables can be carried over to the next script. However, we should look into creating a more streamlined way to manage these variables later. Perhaps move the variables into their own file, that other scripts can then source.

Connects https://github.com/rcbops/u-suk-dev/issues/393

### DIFF
--- a/scripts/deploy-rpc-playbooks.sh
+++ b/scripts/deploy-rpc-playbooks.sh
@@ -1,0 +1,31 @@
+#!/usr/bin/env bash
+
+set -e -u -x
+set -o pipefail
+
+source ${BASE_DIR}/scripts/functions.sh
+
+# begin the RPC installation
+cd ${RPCD_DIR}/playbooks/
+
+# configure everything for RPC support access
+run_ansible rpc-support.yml
+
+# configure the horizon extensions
+run_ansible horizon_extensions.yml
+
+# deploy and configure RAX MaaS
+if [[ "${DEPLOY_MAAS}" == "yes" ]]; then
+  run_ansible setup-maas.yml
+  run_ansible verify-maas.yml
+fi
+
+# deploy and configure the ELK stack
+if [[ "${DEPLOY_ELK}" == "yes" ]]; then
+  run_ansible setup-logging.yml
+
+  # deploy the LB required for the ELK stack
+  if [[ "${DEPLOY_HAPROXY}" == "yes" ]]; then
+    run_ansible haproxy.yml
+  fi
+fi
--- a/scripts/test-upgrade.sh
+++ b/scripts/test-upgrade.sh
@@ -18,34 +18,68 @@
 
 set -eux -o pipefail
 
-BASE_DIR=$( cd "$( dirname ${0} )" && cd ../ && pwd )
-OA_DIR="$BASE_DIR/openstack-ansible"
-RPCD_DIR="$BASE_DIR/rpcd"
+export BASE_DIR=$( cd "$( dirname ${0} )" && cd ../ && pwd )
+export OA_DIR="$BASE_DIR/openstack-ansible"
+export RPCD_DIR="$BASE_DIR/rpcd"
+
+source ${BASE_DIR}/scripts/functions.sh
 
 # TASK #0
 # Bug: https://github.com/rcbops/u-suk-dev/issues/347
 # Issue: Variables files have been reorganized, run a migration script to move
 #        them to the new location.
-"${BASE_DIR}"/scripts/migrate-yaml.py --for-testing-take-new-vars-only \
-    --defaults "${RPCD_DIR}"/etc/openstack_deploy/user_rpco_variables_defaults.yml \
-    --overrides /etc/openstack_deploy/user_extras_variables.yml \
-    --output-file /etc/openstack_deploy/user_rpco_variables_overrides.yml
-"${BASE_DIR}"/scripts/migrate-yaml.py --for-testing-take-new-vars-only \
-  --defaults "${RPCD_DIR}"/etc/openstack_deploy/user_osa_variables_defaults.yml \
-  --overrides /etc/openstack_deploy/user_variables.yml \
-  --output-file /etc/openstack_deploy/user_osa_variables_overrides.yml
+# WARNING: In no way should the migrate-yaml.py script be run with
+#          --for-testing-take-new-vars-only in production environments.
+if [[ ! -f "/etc/openstack_deploy/user_rpco_variables_overrides.yml" ]]; then
+  "${BASE_DIR}"/scripts/migrate-yaml.py --for-testing-take-new-vars-only \
+      --defaults "${RPCD_DIR}"/etc/openstack_deploy/user_rpco_variables_defaults.yml \
+      --overrides /etc/openstack_deploy/user_extras_variables.yml \
+      --output-file /etc/openstack_deploy/user_rpco_variables_overrides.yml
+  rm -f /etc/openstack_deploy/user_extras_variables.yml
+fi
+if [[ ! -f "/etc/openstack_deploy/user_osa_variables_overrides.yml" ]]; then
+  "${BASE_DIR}"/scripts/migrate-yaml.py --for-testing-take-new-vars-only \
+    --defaults "${RPCD_DIR}"/etc/openstack_deploy/user_osa_variables_defaults.yml \
+    --overrides /etc/openstack_deploy/user_variables.yml \
+    --output-file /etc/openstack_deploy/user_osa_variables_overrides.yml
+  rm -f /etc/openstack_deploy/user_variables.yml
+fi
+# TODO: We need to eventually find a proper way to migrate these secret variables.
+#       For now, we just copy existing secret variables defined previously on the
+#       liberty install. This won't cause any problems since there aren't any new
+#       secret variables in mitaka that are not defined in liberty.
+if [[ -f "/etc/openstack_deploy/user_extras_secrets.yml" ]]; then
+  mv /etc/openstack_deploy/user_extras_secrets.yml \
+     /etc/openstack_deploy/user_rpco_secrets.yml
+fi
+if [[ -f "/etc/openstack_deploy/user_secrets.yml" ]]; then
+  mv /etc/openstack_deploy/user_secrets.yml \
+     /etc/openstack_deploy/user_osa_secrets.yml
+fi
 cp -a ${RPCD_DIR}/etc/openstack_deploy/*defaults* /etc/openstack_deploy
-rm /etc/openstack_deploy/user_extras_variables.yml /etc/openstack_deploy/user_variables.yml
 
 # TASK #1
-# Bug: https://github.com/rcbops/u-suk-dev/issues/293
+# Update Ansible
+if [[ -f "/root/.pip/pip.conf" ]]; then
+  rm /root/.pip/pip.conf
+fi
+cd ${OA_DIR}
+bash scripts/bootstrap-ansible.sh
+
+# TASK #2
+# Update the rpc-openstack galaxy modules
+ansible-galaxy install --role-file=${BASE_DIR}/ansible-role-requirements.yml --force \
+--roles-path=${RPCD_DIR}/playbooks/roles
+
+# TASK #3
+# Bug:   https://github.com/rcbops/u-suk-dev/issues/293
 # Issue: We need to analyze support's maintenance plan to determine which
 # pieces we can orchestrate with ansible. This should be added to a
 # pre-upgrade task list.
 cd ${RPCD_DIR}/playbooks
 openstack-ansible rpc-pre-upgrades.yml
 
-# TASK #2
+# TASK #4
 # Bug: https://github.com/rcbops/u-suk-dev/issues/199
 # Issue: To avoid any dependency issues that have occured in the past, the
 #        repo_server containers are re-built. This was done as part of the
@@ -55,8 +89,8 @@ cd ${OA_DIR}/playbooks
 openstack-ansible lxc-containers-destroy.yml --limit repo_all
 openstack-ansible setup-hosts.yml --limit repo_all
 
-# TASK #3
-# Bug: https://github.com/rcbops/u-suk-dev/issues/374
+# TASK #5
+# Bug:   https://github.com/rcbops/u-suk-dev/issues/374
 # Issue: Neutron has no migration to correctly set the MTU on existing networks
 #        created in liberty or below.  This work-around sets the MTU on these
 #        networks before the upgrade starts so that instances booted on these
@@ -71,7 +105,7 @@ cd ${OA_DIR}/playbooks
 ansible galera_all[0] -m shell -a "mysql --verbose -e \"${VLAN_FLAT}\" neutron"
 ansible galera_all[0] -m shell -a "mysql --verbose -e \"${VXLAN}\" neutron"
 
-# TASK #4
+# TASK #6
 # https://github.com/rcbops/u-suk-dev/issues/392
 # Upgrade openstack-ansible
 pushd ${OA_DIR}
@@ -79,8 +113,17 @@ export I_REALLY_KNOW_WHAT_I_AM_DOING=true
 echo "YES" | ${OA_DIR}/scripts/run-upgrade.sh
 popd
 
-# TASK #5 TODO: This task should be moved somewhere towards the bottom once
-#               we get a more robust script going.
+# TASK #7
+# https://github.com/rcbops/u-suk-dev/issues/393
+# Run deploy-rpc-playbooks.sh
+# Ultimitely, this will run the RPCO playbooks.
+cd ${BASE_DIR}
+export DEPLOY_HAPROXY=${DEPLOY_HAPROXY:-"no"}
+export DEPLOY_ELK=${DEPLOY_ELK:-"yes"}
+export DEPLOY_MAAS=${DEPLOY_MAAS:-"yes"}
+bash scripts/deploy-rpc-playbooks.sh
+
+# TASK #8
 # Bug: https://github.com/rcbops/u-suk-dev/issues/366
 # Description: Run post-upgrade tasks.
 #              For a detailed description, please see the README in
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -14,26 +14,23 @@ export DEPLOY_CEILOMETER="no"
 export DEPLOY_CEPH=${DEPLOY_CEPH:-"no"}
 export DEPLOY_SWIFT=${DEPLOY_SWIFT:-"yes"}
 export DEPLOY_HARDENING=${DEPLOY_HARDENING:-"yes"}
-export FORKS=${FORKS:-$(grep -c ^processor /proc/cpuinfo)}
-export ANSIBLE_PARAMETERS=${ANSIBLE_PARAMETERS:-""}
 export ANSIBLE_FORCE_COLOR=${ANSIBLE_FORCE_COLOR:-"true"}
 export BOOTSTRAP_OPTS=${BOOTSTRAP_OPTS:-""}
 export UNAUTHENTICATED_APT=${UNAUTHENTICATED_APT:-no}
 
-OA_DIR='/opt/rpc-openstack/openstack-ansible'
-RPCD_DIR='/opt/rpc-openstack/rpcd'
-RPCD_VARS='/etc/openstack_deploy/user_rpco_variables_defaults.yml'
-RPCD_SECRETS='/etc/openstack_deploy/user_rpco_secrets.yml'
+export BASE_DIR='/opt/rpc-openstack'
+export OA_DIR='/opt/rpc-openstack/openstack-ansible'
+export RPCD_DIR='/opt/rpc-openstack/rpcd'
+export RPCD_VARS='/etc/openstack_deploy/user_rpco_variables_defaults.yml'
+export RPCD_SECRETS='/etc/openstack_deploy/user_rpco_secrets.yml'
 
-function run_ansible {
-  openstack-ansible ${ANSIBLE_PARAMETERS} --forks ${FORKS} $@
-}
+source ${BASE_DIR}/scripts/functions.sh
 
 # begin the bootstrap process
 cd ${OA_DIR}
 
-# bootstrap ansible and install galaxy roles (needed whether AIO or multinode)
 ./scripts/bootstrap-ansible.sh
+
 # This removes Ceph roles downloaded using their pre-Ansible-Galaxy names
 ansible-galaxy remove --roles-path /opt/rpc-openstack/rpcd/playbooks/roles/ ceph-common ceph-mon ceph-osd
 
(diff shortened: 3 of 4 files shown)


rackernexus:
HULK [PREVIEW](https://staging.developer.rackspace.com/staging.horse/build-f3dd8ff5bd/docs/private-cloud/release-notes/)

rackernexus:
If you want to double-check your formatting, check [the preview](https://staging.developer.rackspace.com/staging.horse/build-68e18c81c7/docs/private-cloud/release-notes/).

rackernexus:
:eyes: [Preview](https://staging.developer.rackspace.com/staging.horse/build-02b1aa6a9e/docs/private-cloud/release-notes/)

alextricity25:
This code is ready for review, but I want to wait for https://github.com/rcbops/rpc-openstack/pull/1395 to merge so I can deal with the merge conflicts before they come back to haunt me.

rackernexus:
HULK [PREVIEW](https://staging.developer.rackspace.com/staging.horse/build-6a5daa3da0/docs/private-cloud/release-notes/)

rackernexus:
HULK [PREVIEW](https://staging.developer.rackspace.com/staging.horse/build-e2e8d99c7b/docs/private-cloud/release-notes/)

scripts/test-upgrade.sh
```diff
@@ -22,30 +22,54 @@ BASE_DIR=$( cd "$( dirname ${0} )" && cd ../ && pwd )
...
+    --defaults "${RPCD_DIR}"/etc/openstack_deploy/user_osa_variables_defaults.yml \
+    --overrides /etc/openstack_deploy/user_variables.yml \
+    --output-file /etc/openstack_deploy/user_osa_variables_overrides.yml
+fi
+# TODO: We need to eventually find a proper way to migrate these secret variables.
+#       For now, we just copy existing secret variables defined previously on the
+#       liberty install. This won't cause any problems since there aren't any new
+#       secret variables in mitaka that are not defined in liberty.
+cp /etc/openstack_deploy/user_extras_secrets.yml \
+   /etc/openstack_deploy/user_rpco_secrets.yml
```
mattt416: Do you skip moving user_secrets.yml to user_osa_secrets.yml since that's handled by deploy.sh?

scripts/test-upgrade.sh
```diff
@@ -55,16 +79,25 @@ cd ${OA_DIR}/playbooks
...
-# TASK #4 TODO: This task should be moved somewhere towards the bottom once
-#               we get a more robust script going.
+# TASK #5
+# https://github.com/rcbops/u-suk-dev/issues/393
+# Run deploy.sh with MaaS and don't run the OSA playbooks.
+# Ultimitely, this will run the RPCO playbooks.
+cd ${BASE_DIR}
+export DEPLOY_MAAS=${DEPLOY_MAAS:-'yes'}
+export DEPLOY_OA='no'
+export BOOTSTRAP_ANSIBLE='no'
```
mattt416: deploy.sh doesn't use BOOTSTRAP_ANSIBLE at the moment, is that a work in progress?

alextricity25:
This is ready for review.
For those who are testing: Keep in mind that unless you pull in[ this patch](https://review.openstack.org/#/c/360539/), or wait until the next SHA bump, the post-upgrade playbooks is expected to fail.

scripts/test-upgrade.sh
alextricity25: I added it in as part of this PR. See changes in deploy.sh

scripts/test-upgrade.sh
alextricity25: That's correct!

scripts/test-upgrade.sh
mattt416: *goes to get more coffee*

mattt416:
Testing this now on a liberty deploy.

mattt416:
My upgrade has progressed to the rpc-openstack bits, so this LGTM 👍

rackernexus:
:mag_right: [Preview](https://staging.developer.rackspace.com/staging.horse/build-f8e5d924df/docs/private-cloud/release-notes/)

scripts/test-upgrade.sh
```diff
@@ -22,30 +22,54 @@ BASE_DIR=$( cd "$( dirname ${0} )" && cd ../ && pwd )
 OA_DIR="$BASE_DIR/openstack-ansible"
 RPCD_DIR="$BASE_DIR/rpcd"
 
+# Create a backup of /etc/openstack_deploy, if one doesn't already exist
+# NOTE: This gets done by openstack-ansible's scripts/run-upgrade.sh script,
+#       but if we wait until then it'll be too late.
+if [[ ! -d "/etc/openstack_deploy.MITAKA" ]]; then
+  cp -a /etc/openstack_deploy /etc/openstack_deploy.MITAKA
```
sigmavirus24: if this is the backup, should it be called "MITAKA"? It's a back up of Liberty's deploy directory, right?

evrardjp: That's how we do in OSA. It's fine for me if we say it's because it's happening during mitaka upgrade that we do that :)

scripts/test-upgrade.sh
```diff
@@ -22,30 +22,54 @@ BASE_DIR=$( cd "$( dirname ${0} )" && cd ../ && pwd )
...
-rm /etc/openstack_deploy/user_extras_variables.yml /etc/openstack_deploy/user_variables.yml
+rm /etc/openstack_deploy/user_extras_variables.yml \
+   /etc/openstack_deploy/user_variables.yml        \
+   /etc/openstack_deploy/user_extras_secrets.yml
 
 # TASK #1
-# Bug: https://github.com/rcbops/u-suk-dev/issues/293
+# Update Ansible
+cd ${OA_DIR}
+bash scripts/bootstrap-ansible.sh
```
evrardjp: Why do we need to bootstrap-ansible again? Would it be possible to use the first bootstrap ansible to install ansible and the roles, and here just refresh the roles (only if needed)?

scripts/test-upgrade.sh
evrardjp: good to remove that part.

scripts/test-upgrade.sh
alextricity25: I honestly don't remember why I needed bootstrap ansible here.
What did you have in mind? ``ansible-galaxy --install`` to update the roles?

andymcc: You need the openstack_hosts role in order to run the setup_hosts role in Task #3
The other option is looking at whether we need to rebuild the containers here... OSA upgrade will do a container-create (except for galera_all) so it should recreate them, the only issue would be if the repo servers are needed before that happens.

alextricity25: @andymcc Good catch on ``pip.conf``. I'll add that task in.
RE: repo-containers, The reason we needed them up before the OSA upgrade script is because that script runs [a playbook](https://github.com/openstack/openstack-ansible/blob/stable/mitaka/scripts/run-upgrade.sh#L152) that targets repo-containers before running ``setup-hosts.yml``. And if they are not there before it's run, it will fail :(

andymcc: Ahh ok cool - I was going to test that, makes sense - would've been nicer the other way but I did the same thing to get it going earlier, so lets roll with that.

scripts/test-upgrade.sh
sigmavirus24: > That's how we do in OSA.

I'm fairly certain that RPCO has done it differently in the past though. It also makes more sense to me to label the directory with the release for which it works.

@d34dh0r53 and @Apsu you two seemed to be confused by this too on Thursday and Friday. Thoughts?

scripts/test-upgrade.sh
andymcc: Can we add:
if [[ -f "/etc/openstack_deploy/user_extra_secrets.yml" ]]; then
  cp ...
  cp ...
  rm ...
fi

rackernexus:
HULK [PREVIEW](https://staging.developer.rackspace.com/staging.horse/build-0926a96e24/docs/private-cloud/release-notes/)

alextricity25:
Ready for testing

rackernexus:
HULK [PREVIEW](https://staging.developer.rackspace.com/staging.horse/build-8598d8dd25/docs/private-cloud/release-notes/)

scripts/test-upgrade.sh
andymcc: I agree with @sigmavirus24 it makes more sense (to me at least) for it to be openstack_deploy.LIBERTY - since that would be the version for which the variables are setup, its a simple change too.

scripts/test-upgrade.sh
```diff
@@ -22,30 +22,57 @@ BASE_DIR=$( cd "$( dirname ${0} )" && cd ../ && pwd )
...
+  cp -a ${RPCD_DIR}/etc/openstack_deploy/*defaults* /etc/openstack_deploy
+  rm /etc/openstack_deploy/user_extras_variables.yml \
+     /etc/openstack_deploy/user_variables.yml        \
+     /etc/openstack_deploy/user_extras_secrets.yml
+fi
 
 # TASK #1
-# Bug: https://github.com/rcbops/u-suk-dev/issues/293
+# Update Ansible
+rm /root/.pip/pip.conf
```
andymcc: Should we do an if on this, or a || true so that the rm will succeed even if the file doesn't exist.

alextricity25: Good point. I'll add that now

rackernexus:
:eyes: [Preview](https://staging.developer.rackspace.com/staging.horse/build-eb0358d94b/docs/private-cloud/release-notes/)

rackernexus:
If you want to double-check your formatting, check [the preview](https://staging.developer.rackspace.com/staging.horse/build-8760b84c90/docs/private-cloud/release-notes/).

git-harry:
👎 this looks to be adding a lot of complexity. I think we should consider breaking out bits of deploy.sh so that both deploy.sh and test-upgrade.sh are using the same scripts instead of trying to make deploy.sh have more configurability. So for example below is the deploy.sh code to run the RPC playbooks, why not break that out into a separate script and have both deploy.sh and test-upgrade.sh use it. I think that would make it clearer what is shared between the two scripts, the current deploy.sh has already proved problematic for support to understand, this patch seems to be increasing the complexity.

```
# begin the RPC installation
cd ${RPCD_DIR}/playbooks/

# configure everything for RPC support access
run_ansible rpc-support.yml

# configure the horizon extensions
run_ansible horizon_extensions.yml

# deploy and configure RAX MaaS
if [[ "${DEPLOY_MAAS}" == "yes" ]]; then
  run_ansible setup-maas.yml
  run_ansible verify-maas.yml
fi

# deploy and configure the ELK stack
if [[ "${DEPLOY_ELK}" == "yes" ]]; then
  run_ansible setup-logging.yml

  # deploy the LB required for the ELK stack
  if [[ "${DEPLOY_HAPROXY}" == "yes" ]]; then
    run_ansible haproxy.yml
  fi  
fi
```

scripts/test-upgrade.sh
```diff
@@ -22,30 +22,59 @@ BASE_DIR=$( cd "$( dirname ${0} )" && cd ../ && pwd )
...
+fi
 
 # TASK #1
-# Bug: https://github.com/rcbops/u-suk-dev/issues/293
+# Update Ansible
+if [[ -f "/root/.pip/pip.conf" ]]; then
+  rm /root/.pip/pip.conf
+fi
+cd ${OA_DIR}
+bash scripts/bootstrap-ansible.sh
```
evrardjp: why are we rebootstrapping ansible?

Shouldn't we just do one bootstrap once (install ansible and roles), and skip the multiple bootstrap?

evrardjp:
Agreed with @git-harry 
let's split in small bits, and re-use just what we need if possible.

scripts/test-upgrade.sh
alextricity25: We need the roles to be in place before we run ``setup-hosts.yml`` in TASK#3.
I'm not sure what you mean by "rebootstrapping ansible". This is the first invocation of ``bootstrap-ansible.sh`` in the script, no?

Apsu: Line 38 does as well, if the variable is set.

alextricity25:
I've incorporated the feedback above. Once I finish testing it out I'll submit the latest commit.

rackernexus:
Your content preview [is now ready](https://staging.developer.rackspace.com/staging.horse/build-89a75b2153/docs/private-cloud/release-notes/). :bow:

rackernexus:
HULK [PREVIEW](https://staging.developer.rackspace.com/staging.horse/build-633d023962/docs/private-cloud/release-notes/)

alextricity25:
recheck_swift

rackernexus:
:mag_right: [Preview](https://staging.developer.rackspace.com/staging.horse/build-18c3592dc6/docs/private-cloud/release-notes/)

rackernexus:
HULK [PREVIEW](https://staging.developer.rackspace.com/staging.horse/build-15ab25748b/docs/private-cloud/release-notes/)

scripts/deploy-rpc-playbooks.sh
```diff
@@ -0,0 +1,37 @@
...
+
+# deploy and configure the ELK stack
+if [[ "${DEPLOY_ELK}" == "yes" ]]; then
+  run_ansible setup-logging.yml
+
+  # deploy the LB required for the ELK stack
+  if [[ "${DEPLOY_HAPROXY}" == "yes" ]]; then
+    run_ansible haproxy.yml
+  fi
+fi
```
mattt416: Should this script be able to be run independently?  I see benefit in that, but then it means having another place where default variables need to be maintained (currently 3 places).

alextricity25:
@all Please review the latest commit. Thank you!

rackernexus:
HULK [PREVIEW](https://staging.developer.rackspace.com/staging.horse/build-2c76369a0a/docs/private-cloud/release-notes/)

scripts/test-upgrade.sh
```diff
@@ -18,34 +18,65 @@
...
+export BASE_DIR=$( cd "$( dirname ${0} )" && cd ../ && pwd )
+export OA_DIR="$BASE_DIR/openstack-ansible"
+export RPCD_DIR="$BASE_DIR/rpcd"
+
+source ${BASE_DIR}/scripts/functions.sh
+
+# Create a backup of /etc/openstack_deploy, if one doesn't already exist
+# NOTE: This gets done by openstack-ansible's scripts/run-upgrade.sh script,
+#       but if we wait until then it'll be too late.
+if [[ ! -d "/etc/openstack_deploy.MITAKA" ]]; then
```
mattt416: Looks like this is still referencing the old `MITAKA`.  Side note, I created https://review.openstack.org/#/c/369650/ so that upstream OSA upgrade backs stuff up to /etc/openstack_deploy.LIBERTY, rather than /etc/openstack_deploy.MITAKA.  In liberty of OSA it backed up to /etc/openstack_deploy.KILO.

With that said, I almost feel like this script should back up somewhere else, ie. /etc/openstack_deploy.LIBERTY_RPCO or whatever.  The reason being that when OSA upgrade commences and backs up these files it's going to overwrite the original backups we had in /etc/openstack_deploy.LIBERTY.  It would be good to have a pristine backup of /etc/openstack_deploy which we know OSA won't touch that support can resort to if needed.  Thoughts?

scripts/test-upgrade.sh
```diff
@@ -18,34 +18,65 @@
...
-    --overrides /etc/openstack_deploy/user_extras_variables.yml \
-    --output-file /etc/openstack_deploy/user_rpco_variables_overrides.yml
-"${BASE_DIR}"/scripts/migrate-yaml.py --for-testing-take-new-vars-only \
-  --defaults "${RPCD_DIR}"/etc/openstack_deploy/user_osa_variables_defaults.yml \
-  --overrides /etc/openstack_deploy/user_variables.yml \
-  --output-file /etc/openstack_deploy/user_osa_variables_overrides.yml
-cp -a ${RPCD_DIR}/etc/openstack_deploy/*defaults* /etc/openstack_deploy
-rm /etc/openstack_deploy/user_extras_variables.yml /etc/openstack_deploy/user_variables.yml
+if [[ ! -f "/etc/openstack_deploy/user_rpco_variables_overrides.yml" ]]; then
+  "${BASE_DIR}"/scripts/migrate-yaml.py --for-testing-take-new-vars-only \
```
mattt416: It'd probably be worthwhile pointing out that `--for-testing-take-new-vars-only` should in no way be used in production.

scripts/test-upgrade.sh
alextricity25: Good thinking. This is actually [already done in the pre-upgrade playbooks](https://github.com/rcbops/rpc-openstack/blob/master/rpcd/playbooks/roles/rpc_pre_upgrade/tasks/backup.yml), so it can be removed.

rackernexus:
HULK [PREVIEW](https://staging.developer.rackspace.com/staging.horse/build-b416349db4/docs/private-cloud/release-notes/)

alextricity25:
I was tempted to move all of the bash environment variables into one file that we could source(that way we have one place for all of these variables instead of three), but I'm not so sure if that work should be part of this card. 
@mattt416 I've addressed your comments.
The latest commit is ready for review. Thanks.

rackernexus:
:mag_right: [Preview](https://staging.developer.rackspace.com/staging.horse/build-a368506ccb/docs/private-cloud/release-notes/)

rackernexus:
:mag_right: [Preview](https://staging.developer.rackspace.com/staging.horse/build-d09d31705f/docs/private-cloud/release-notes/)

rackernexus:
If you want to double-check your formatting, check [the preview](https://staging.developer.rackspace.com/staging.horse/build-ea4cdff1ab/docs/private-cloud/release-notes/).

scripts/deploy-rpc-playbooks.sh
git-harry: Yes, this script has to be able to be run separately. The whole idea is that each step in test-upgrade.sh is either something support needs to be able to run or test 'glue'.

scripts/test-upgrade.sh
```diff
@@ -18,34 +18,59 @@
...
+     /etc/openstack_deploy/user_rpco_secrets.yml
+fi
 cp -a ${RPCD_DIR}/etc/openstack_deploy/*defaults* /etc/openstack_deploy
-rm /etc/openstack_deploy/user_extras_variables.yml /etc/openstack_deploy/user_variables.yml
 
 # TASK #1
-# Bug: https://github.com/rcbops/u-suk-dev/issues/293
+# Update Ansible
+if [[ -f "/root/.pip/pip.conf" ]]; then
+  rm /root/.pip/pip.conf
```
git-harry: Why is this required? bootstrap-ansible.sh falls back on --isolated where necessary. Did you see an error?

andymcc:
We still need to setup the ceph playbooks, its possible the directory is moved and a new one cloned - so the roles won't exist - causing MaaS to fail straight away.
e.g. these lines: https://github.com/rcbops/rpc-openstack/blob/master/scripts/deploy.sh#L37-L41

git-harry:
@andymcc I noted that on https://github.com/rcbops/u-suk-dev/issues/388 but I didn't realise MaaS was dependent on them. Maybe we should have a bootstrap script that does bootstrap-ansible.sh and these RPCO bits together.

andymcc:
Ok with the ceph change - this worked for me - there is one bug in the rpc_post_upgrade role, which will require a backport in OSA (or a work around - which may be better for now in RPC) - will create a card for that.

I'm happy to merge this once the Ceph bit is fixed.

scripts/test-upgrade.sh
andymcc: It spits out errors that can be confusing - although it will work.

rackernexus:
Your content preview [is now ready](https://staging.developer.rackspace.com/staging.horse/build-e7b2a36bec/docs/private-cloud/release-notes/). :bow:

scripts/test-upgrade.sh
```diff
@@ -71,16 +101,25 @@ cd ${OA_DIR}/playbooks
...
+# https://github.com/rcbops/u-suk-dev/issues/393
+# Run deploy-rpc-playbooks.sh
+# Ultimitely, this will run the RPCO playbooks.
+cd ${BASE_DIR}
+export DEPLOY_HAPROXY=${DEPLOY_HAPROXY:-"no"}
+export DEPLOY_ELK=${DEPLOY_ELK:-"yes"}
+export DEPLOY_MAAS=${DEPLOY_MAAS:-"yes"}
+bash scripts/deploy-rpc-playbooks.sh
+
+# TASK #7
```
git-harry: 8

scripts/test-upgrade.sh
andymcc: I believe thats in the deploy.sh script which doesn't actually get called by test-upgrade.sh
scripts/deploy-rpc-playbooks.sh does get called, but that doesn't perform a bootstrap.

We do end up with 2 bootstraps unfortunately, because one happens as part of the openstack-ansible run-upgrade.sh, but that seems to be unavoidable if we want to remove and recreate the repo servers (Task 4). Or at least we do need to take some actions and performing a bootstrap just seems simpler than (for example) installing a subset of roles and hoping all dependencies etc are met.

git-harry:
👎 scripts/deploy-rpc-playbooks.sh doesn't look like something that support could easily run, they have to set all sorts of variables just to run it.

scripts/test-upgrade.sh
evrardjp: Agreed it's just on the deploy node anyway.

scripts/test-upgrade.sh
evrardjp: Well, now that we broke the test-upgrade in multiple script this occurence is indeed the first one.

Because we run the osa bootsrap and then install the rpc roles through galaxy I think we are fine.

Yes, we re-bootstrap during the upgrade, but it seems a pain to change in OSA, not RPC.

So, we're good.

alextricity25:
@git-harry I disagree that ``deploy-rpc-playbooks.sh`` should be something that can be independently run. It should only be invoked by either ``deploy.sh`` or ``test-upgrades.sh``. If someone wants to run the rpc-playbooks, they can just ``cd /opt/rpc-openstack/rpcd/playbooks`` and run them from there(which is what I hope support would be doing when upgrading an environment that's not for testing)
I don't agree with setting defaults for these variables in various places that we'll have to keep track of. i.e.
```
export ${DEPLOY_MAAS:-'yes'}
..
..
```

I think the right way to go would be to streamline the way all of these variables are set. We could munge them into one file, then have these deployment/upgrade scripts source that file. I don't believe that's part of the scope of this card, though.

andymcc:
👍  This worked for me in a full multi-node upgrade.

mattt416:
I'm going to test this out today, but generally this looks OK to me.  We'll still need to figure out where the ceph playbooks are re-run but we can add that as we start digging into the ceph bits.  Only final thing that I can think of that is needed (now that we do not include deploy.sh) is moving user_secrets.yml to user_osa_secrets.yml.

alextricity25:
recheck_ceph

mattt416:
Tested, had some unrelated failures along the way (ssh timeouts etc.) but this script otherwise looks 👍 to me.

alextricity25:
recheck_ceph

alextricity25:
recheck_ceph

git-harry:
recheck_ceph

hughsaunders:
recheck_ceph

git-harry:
recheck_ceph

git-harry:
recheck_ceph

evrardjp:
recheck_ceph -- maas issue.

mancdaz:
recheck_ceph

alextricity25:
recheck_ceph

alextricity25:
recheck_ceph

alextricity25:
recheck_ceph

mancdaz:
recheck_ceph

alextricity25:
recheck_all

Apsu:
recheck_ceph

mancdaz:
recheck_swift

mattt416:
recheck_ceph

alextricity25:
recheck_ceph !!!!!

mancdaz:
w00t!!

PR #1398, opened 2016-09-08
Set rpc_openstack_repo var using git_repo_info

mattt416:
This commit updates setup-maas.yml and verify-maas.yml to discover and
set rpc_openstack_repo variable (which is used to set rpc_release).
This is needed by both playbooks (and the rpc_maas role) to correctly
create and use the maas venv in /openstack/venvs.  Unless these
playbooks are run in conjunction with rpc-support.yml (which runs
git_repo_info), the value of rpc_release will default to UNKNOWN.
By running in git_repo_info in all the necessary playbooks and roles
means that rpc-support.yml, setup-maas.yml, and verify-maas.yml can
all be run independently and still have the same values for
rpc_openstack_repo and rpc_release.

Connects https://github.com/rcbops/u-suk-dev/issues/397

(cherry picked from commit 2e10437fe437e51f328e6fbc49ce0f139f149183)

### DIFF
--- a/rpcd/playbooks/setup-maas.yml
+++ b/rpcd/playbooks/setup-maas.yml
@@ -24,5 +24,16 @@
 - name: Install MaaS
   hosts: hosts:all_containers
   user: root
+  pre_tasks:
+    # This is needed to set the rpc_openstack_repo and rpc_release variables
+    # correctly
+    - name: Discover rpc-openstack version
+      git_repo_info:
+        path: "/opt/rpc-openstack"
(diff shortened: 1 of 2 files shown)


alextricity25:
:+1: Pending checks

andymcc:
LGTM too - pending release of 13.0.0

evrardjp:
LGTM too. 
Let me do a follow-up commit to remove the UNKNOWN bits, and make sure the playbook fails when the variable is undefined.

Issue #1404, opened 2016-09-09
L->M upgrades break MaaS due to pip_lockdown dependency


andymcc:
##### What type of issue is this? Please tick the appropriate box.
- [x] bug - existing functionality that does not work for you as documented
- [ ] feature - new functionality that you would like to see added

##### What is the expected behavior?
rpc_maas role should run pip_lockdown so that the correct versions of dependencies are installed.
##### What is actually happening?
During the OSA upgrade script the pip.conf files are removed, and since rpc_maas doesn't depend on pip_lockdown so newer versions of pip packages will be deployed. (For example, keystone version 3.5.0 is installed when it should install 2.3.1).
##### What were you attempting to do when you encountered this error (give as much detail as possible)? 
Upgrade testing for MaaS
##### How can we reproduce this issue?
Run a multi-node cluster in Liberty and follow the upgrade path from L->M (there has to be at least one node that isn't a physical service, like swift-storage, cinder-volumes(lvm) or nova-compute - since these will get pip_lockdown as part of OSA).  - in greenfields these files aren't removed and the pip-lockdown rpc play is run in the delpoy.sh script, so this only affects upgrades.
##### How can we verify your issue is addressed?
Test the maas plugins (or verify-maas)
##### If you are using or know of a workaround please describe the necessary steps?
You can run the pip-lockdown.yml playbook before updating MaaS, or you can rm -rf /openstack/venv/
##### What is your evaluation on the importance of this issue, and why?
High. We need this to work for upgrades to be successful.

----------------------------------------------------------------------------------------------------------------------------
###### For each pull request, update the branch line to reference it.
- [x] master
- [ ] mitaka-13.0
- [ ] liberty-12.2
- [ ] kilo

##### Upgrade impact. Consider the impact on moving between major, minor and patch releases.
- [ ] This change does not impact upgrades
- [x] This change addresses a potential upgrade impact
- [ ] The upgrade impact is being tracked by another issue.

The issue tracking the upgrade impact is:

##### Testing impact
- [x] An issue/pull-request has been create or exists to track the testing gap, and is referenced below
- [ ] This change includes additional testing
- [ ] This change is tested by the existing setup

Additional testing links:
Upgrade testing is being create:; https://waffle.io/rcbops/u-suk-dev/cards/57bf0de134ad3540017010d5

##### Release note
- [ ] This change includes a release note.
- [x] This change does not require a release note.

##### Does your change require documentation?
[...]

PR #1405, opened 2016-09-09
Resolve pip_lock_down and maas_common issues

andymcc:
The rpc_maas role needs to depend on the pip_lock_down role, otherwise
the openstack clients and other pip packages are not installed from the
repo server.

To prevent this issue occuring on other roles that perform pip installs,
and to keep the roles consistent, this PR adds the pip_lock_down
dependency to the rpc_support and horizon_extensions roles.

This will prevent us from having to run "pip-lockdown" against hosts
unnecessarily.

Additionally, in the next version of python-keystoneclient
apiclient.exceptions will be removed, as per:
https://review.openstack.org/#/c/257127/ - we can fix this in advance by
following the recommendation and use keystoneclient.exceptions. This
won't affect current releases but will future proof for upgrades to the
Newton cycle (and above).

Connects #1404

### DIFF
--- a/maas/plugins/maas_common.py
+++ b/maas/plugins/maas_common.py
@@ -195,7 +195,7 @@ else:
         return nova
 
 try:
-    from keystoneclient.openstack.common.apiclient import exceptions as k_exc
+    from keystoneclient import exceptions as k_exc
     from keystoneclient.v2_0 import client as k2_client
     from keystoneclient.v3 import client as k3_client
 except ImportError:
--- a/rpcd/playbooks/roles/horizon_extensions/meta/main.yml
+++ b/rpcd/playbooks/roles/horizon_extensions/meta/main.yml
@@ -15,3 +15,6 @@ galaxy_info:
     - heat
     - development
     - openstack
+
+dependencies:
+  - pip_lock_down
--- a/rpcd/playbooks/roles/rpc_maas/meta/main.yml
+++ b/rpcd/playbooks/roles/rpc_maas/meta/main.yml
@@ -29,3 +29,6 @@ galaxy_info:
     - MaaS
     - development
     - openstack
+
+dependencies:
+  - pip_lock_down
--- a/rpcd/playbooks/roles/rpc_support/meta/main.yml
+++ b/rpcd/playbooks/roles/rpc_support/meta/main.yml
@@ -27,3 +27,6 @@ galaxy_info:
     - support
     - rackspace
     - backups
+
+dependencies:
+  - pip_lock_down


andymcc:
recheck_all

andymcc:
failed to clone rsyslog_server - but manually that seems to work ok - so rechecking.

andymcc:
recheck_swift

evrardjp:
testing this right now.

mattt416:
This change looks good to me, but two minor questions:

1. Should we make pip_lock_down a dependency in all rpcd roles?  This fixes maas but rpc_support may be susceptible to similar breakage if a deployer does not run pip_lockdown.yml during the upgrade process.
2. The linked bug mentions `keystoneclient.apiclient.exceptions`, we're using `keystoneclient.openstack.common.apiclient.exceptions`.  They're both deprecated so I think the change is right, I'm just wondering if we're linking to the wrong bug?

--Matt

BjoernT:
Yepp @mattt416  see https://github.com/rcbops/rpc-openstack/pull/1409

andymcc:
@matt416 good catch on the PR - there were 2, I found the appropriate one.

evrardjp:
Do you think horizon_extensions and rpc-support should benefit from this approach too?
It's probably better to lock them down too, right?

evrardjp:
Well I just saw your comment on the other PR.
My analysis is the same, so we're good to go I guess.

The test-upgrade.sh still fails for me, but not linked to this PR.

rackernexus:
Your content preview [is now ready](https://staging.developer.rackspace.com/staging.horse/build-53ca97ade6/docs/private-cloud/release-notes/). :bow:

rackernexus:
:eyes: [Preview](https://staging.developer.rackspace.com/staging.horse/build-7f7e67e036/docs/private-cloud/release-notes/)

mattt416:
Still waiting on gate, but this looks 👍 to me.

BjoernT:
👎  We need to make sure that the pip_lock_down role is dependent on other roles too like rpc_support or the maas extensions

andymcc:
@BjoernT if you look at the PR it does address rpc_support, as well as horizon_extensions and rpc_maas. The kibana and elasticsearch roles are already dependent on it - all other roles don't actually install any pip packages and so don't need pip_lock_down.

BjoernT:
@andymcc  OOps my bad we are 👍  here then

sigmavirus24:
👍 looks good to me pending the gate

mancdaz:
👍  pending checks

andymcc:
recheck_swift

PR #1408, opened 2016-09-09
Handle cases where beaver/swift are not installed

mattt416:
This commit updates rpc_pre_upgrade to only run gather_swift_stats.yml
when swift is installed (groups['swift_proxy']|length > 0).
Additionally, we update log_rotation.yml to first check if beaver is
installed, and then only restart it if the init script exists.

Connects https://github.com/rcbops/u-suk-dev/issues/407

### DIFF
--- a/rpcd/playbooks/roles/rpc_pre_upgrade/tasks/main.yml
+++ b/rpcd/playbooks/roles/rpc_pre_upgrade/tasks/main.yml
@@ -25,6 +25,7 @@
 
 # Gather swift-recon md5 output
 - include: gather_swift_stats.yml
+  when: groups['swift_proxy']|default([])|length > 0
 
 # Perform log rotation and cleanup
 - include: log_rotation.yml
--- a/rpcd/playbooks/roles/rpc_pre_upgrade/tasks/log_rotation.yml
+++ b/rpcd/playbooks/roles/rpc_pre_upgrade/tasks/log_rotation.yml
@@ -52,9 +52,17 @@
   delegate_to: "{{ item }}"
   with_items: "{{ groups['rsyslog_container'] }}"
 
-- name: Restart beaver
+- name: Check if beaver is installed
+  stat:
+    path: /etc/init.d/beaver
+  register: beaver_init_script
+  with_items: "{{ groups['all'] }}"
+  delegate_to: "{{ item }}"
+
+- name: Restart beaver if installed
   service:
     name: beaver
     state: restarted
-  delegate_to: "{{ item }}"
-  with_items: "{{ groups['all'] }}"
+  with_items: beaver_init_script.results
+  when: item.stat.exists
+  delegate_to: "{{ item.item }}"


rackernexus:
:eyes: [Preview](https://staging.developer.rackspace.com/staging.horse/build-b5459d5468/docs/private-cloud/release-notes/)

rpcd/playbooks/roles/rpc_pre_upgrade/tasks/main.yml
```diff
@@ -25,6 +25,8 @@
 
 # Gather swift-recon md5 output
 - include: gather_swift_stats.yml
+  when: >
```
evrardjp: nit: we don't need the newline here.

rpcd/playbooks/roles/rpc_pre_upgrade/tasks/main.yml
```diff
@@ -25,6 +25,8 @@
...
+    groups['swift_proxy']|length > 0
```
evrardjp: groups['swift_proxy']|default([])| length > 0
instead?

evrardjp:
not sure about the groups, maybe we have to check the env.d

rpcd/playbooks/roles/rpc_pre_upgrade/tasks/main.yml
evrardjp: According to @andymcc , we define the swift_proxy group by default, so checking the length is enough.

rackernexus:
Your content preview [is now ready](https://staging.developer.rackspace.com/staging.horse/build-0ffdf9e2e1/docs/private-cloud/release-notes/). :bow:

mattt416:
recheck_lint

mattt416:
recheck_ceph

andymcc:
If you are concerned that the swift_proxy group won't exist in env.d you could do a "if swift_proxy is defined and swift_proxy | length > 0" but it is a default env file that should exist. So this lgtm.

There are similar ones, for example if logging doesnt exist this fails similarly - is it safe to assume we always have logging?

mattt416:
I was under the assumption that any RPCO deploy would have the logging bits deployed.  Can you confirm, @BjoernT?  If not, we'll need to update log_rotation.yml so that beaver is only restarted if it's actually installed.

BjoernT:
@mattt416 Today most deployments have logging nodes configured, 1-2 customers definitely removed this config since they are using their own logging infrastructure.  Hence, don't rely on logging node, container and/or ansible groups. the rotation would need to be dependent on if a logging host actually exist

evrardjp:
@mattt416 We have multiple scenario in the deploy.sh , some don't deploy logging, some don't deploy swift... I think we should verify the group existence and length in the cases explained above.

It's maybe worth another work/card to check that all the assumptions we do are correct in our roles.

git-harry:
recheck_ceph

rpcd/playbooks/roles/rpc_pre_upgrade/tasks/main.yml
```diff
@@ -25,6 +25,7 @@
 
 # Gather swift-recon md5 output
 - include: gather_swift_stats.yml
+  when: groups['swift_proxy']|default([])|length > 0
```
git-harry: The pre-upgrade Swift task have the condition
```
when: inventory_hostname in groups['swift_proxy']
```
In general there seems to be a lot of overlap between the Swift bits of the pre and post roles, not suggesting that this patch should address that but maybe we should be consistent between the two. Is one condition better than the other?

evrardjp: Well, pre-upgrades playbook runs on localhost (which is terrible IMO but that's another story), where post-upgrades run on galera_all:rabbitmq:utility:swift_proxy:elasticsearch_container (which is less terrible but let's go on).
In that sense, the post_upgrades "inventory_hostname in groups['whatever']" will never work on pre-upgrades.

Let's continue with what's needed. Swift, log rotation isn't forced to be deployed, and the deployer could even have removed the env.d, so it's good to ensure that the group exists and have nodes in it before doing anything.

Two methods are good for that.
```
when:
  - groups['swift_proxy'] is defined
  - groups['swift_proxy']|length > 0

```
or the shorter version:
`when: groups['swift_proxy']|default([])|length > 0`

evrardjp:
👎  We should also apply this for log_rotation

mattt416:
We can fix the log rotation bits in here.  However, in essence the rsyslog stuff will always be installed, right?  Which means the only task we're concerned with is the restart of beaver.  I was looking at this yesterday and hit a bit of a wall since everything runs on localhost and then delegates where necessary.  Ordinarily you'd just check if /etc/init.d/beaver exists on a host, then restart it conditionally ... but things are a bit more tricky with the way we've constructed the role.

rackernexus:
:mag_right: [Preview](https://staging.developer.rackspace.com/staging.horse/build-339f341a2e/docs/private-cloud/release-notes/)

rackernexus:
:mag_right: [Preview](https://staging.developer.rackspace.com/staging.horse/build-42b9830426/docs/private-cloud/release-notes/)

rackernexus:
:eyes: [Preview](https://staging.developer.rackspace.com/staging.horse/build-c6b299cb12/docs/private-cloud/release-notes/)

evrardjp:
recheck_swift -- I see the build as aborted. I recheck to see the logs.

git-harry:
👍

evrardjp:
👍

PR #904, opened 2016-03-17
Retry MaaS alarm status verification, closes #903

git-harry:
Add retrying to the task verifying the status of MaaS alarms.

MaaS alarms can take several minutes, after a plugin starts returning
the correct data, to return to an okay state. This is causing the
upgrade script to fail intermittently. This change will allow more time
before reporting a failure.

Closes-issue: https://github.com/rcbops/rpc-openstack/issues/903

PR #1416, opened 2016-09-13
Retry Maas Status Verification

hughsaunders:
MaaS operations sometimes fail due to transient issues with the API.
This should not cause a gate run to fail, so MaaS operations should be
retried.

This bug is only present in kilo, but this is still a problem as we do
kilo deploys and upgrades from kilo.

This was fixed in master in rcbops/rpc-openstack#904
Connects rcbops/u-suk-dev#413


major:
:+1:

PR #1418, opened 2016-09-14
set OS_INTERFACE=internal for openstack commands

andymcc:
The "openstack compute service list" command will default to use the
publicURL, this will be addressed in OSA - but as a workaround for now
we can export the value to ensure this works.

Connects https://github.com/rcbops/u-suk-dev/issues/417

### DIFF
--- a/rpcd/playbooks/roles/rpc_post_upgrade/tasks/post-upgrade-utility.yml
+++ b/rpcd/playbooks/roles/rpc_post_upgrade/tasks/post-upgrade-utility.yml
@@ -16,6 +16,7 @@
 # Verify services are up and checking in
 - name: Grab relevant output of openstack compute service list
   shell: |
+    export OS_INTERFACE=internal
     . ~/openrc
     openstack compute service list -c Binary -c Host -c State -f value
   register: openstack_output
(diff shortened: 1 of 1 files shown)


rackernexus:
Your content preview [is now ready](https://staging.developer.rackspace.com/staging.horse/build-3f081788f4/docs/private-cloud/release-notes/). :bow:

mancdaz:
@andymcc perhaps worth adding an inline comment pointing at the OSA review so that we know when we can remove this workaround?
Or perhaps a card to ensure we track this for rc2. Or both.

git-harry:
A card to track any follow up work sounds good. Putting this off to rc2 should be fine assuming QE don't rely on the client for their testing or they can use the public endpoint. @jacobwagner is this likely to impact your testing?

I notice one of the [pre-upgrade steps](https://github.com/rcbops/rpc-openstack/blob/master/rpcd/playbooks/roles/rpc_pre_upgrade/tasks/gather_stats.yml#L48) uses the openstack client, is that unaffected because this is specific to newer versions of the client?

andymcc:
@git-harry its only the openstack compute section that fails, I added neutron to ensure I hadn't already exported the var.

On second thoughts - after speaking to Hugh - this won't break gating because it's not using an F5 or other scheme (NAT addresses etc) that would cause this to fail.

In other words - it'll work in the interim so we can forget about doing a temporary fix and just wait for the SHA bump - I've also spoken to Jesse and we'll get a SHA bump moving forward for OSA

PR #1423, opened 2016-09-16
Move beaver restart to it's own play

mancdaz:
The way we currently do the beaver restarts means that it takes an
inordinate amount of time to restart beaver across all hosts.

This commit moves the beaver restart tasks to their own play so that
they can be targeted at the correct hosts in the way that ansible is
designed to be used.

An example of the difference this makes to the timings, from an AIO:

OLD WAY:

```
PLAY RECAP ********************************************************************
rpc_pre_upgrade | Restart beaver if installed ------------------------- 445.65s
rpc_pre_upgrade | Check if beaver is installed ------------------------- 45.07s

real       8m13.294s
user       0m10.392s
sys        0m1.471s
```

NEW WAY:

```
PLAY RECAP ********************************************************************
Restart beaver if installed -------------------------------------------- 41.22s
Check if beaver is installed -------------------------------------------- 1.42s

real       0m51.494s
user       0m4.022s
sys        0m1.427s
```

Connected #1419

PR #1427, opened 2016-09-19
Remove beaver from existing containers

prometheanfire:
We basically do the beaver install tasks in inverse.

Connected rcbops/u-suk-dev#180

### DIFF
--- a/rpcd/playbooks/roles/rpc_pre_upgrade/tasks/main.yml
+++ b/rpcd/playbooks/roles/rpc_pre_upgrade/tasks/main.yml
@@ -29,4 +29,3 @@
 
 # Perform log rotation and cleanup
 - include: log_rotation.yml
-
--- a/rpcd/playbooks/rpc-pre-upgrades.yml
+++ b/rpcd/playbooks/rpc-pre-upgrades.yml
@@ -19,18 +19,63 @@
   roles:
     - { role: "rpc_pre_upgrade", tags: [ "rpc-pre-upgrades" ] }
 
-- name: restart beaver
+- name: remove beaver
   hosts: all
   tags:
     - rpc-pre-upgrades
+    - remove-beaver
   tasks:
     - name: Check if beaver is installed
       stat:
         path: /etc/init.d/beaver
       register: beaver_init_script
+      changed_when: false
 
-    - name: Restart beaver if installed
+    - name: Make sure beaver is stopped
       service:
         name: beaver
-        state: restarted
+        enabled: no
+        state: stopped
       when: beaver_init_script.stat.exists
+
+    - name: Remove beaver etc files
+      file:
+        path: /etc/beaver
+        state: absent
+
+    - name: Remove beaver init files
+      file:
+        path: /etc/init.d/beaver
+        state: absent
+
+    - name: Remove beaver logrotate files
+      file:
+        path: /etc/logrotate.d/beaver
+        state: absent
+
+    - name: Remove beaver log files
+      file:
+        path: /var/log/beaver
+        state: absent
+
+    - name: Check if beaver is installed
+      shell: pip show -q Beaver
+      register: is_beaver_installed
+      failed_when: false
+      changed_when: false
+
+    - name: Uninstall beaver pip package
+      pip:
+        name: beaver
+        state: absent
+      when: is_beaver_installed.rc == 0
+
+    - name: Remove beaver user
+      user:
+        name: beaver
+        state: absent
+
+    - name: Remove beaver group
+      group:
+        name: beaver
+        state: absent


rackernexus:
:eyes: [Preview](https://staging.developer.rackspace.com/staging.horse/build-f129b05042/docs/private-cloud/release-notes/)

rpcd/playbooks/roles/rpc_pre_upgrade/tasks/kill_beaver.yml
```diff
@@ -0,0 +1,57 @@
...
+# limitations under the License.
+
+- name: Make sure beaver is stopped
+  service:
+    name: beaver
+    enabled: no
+    state: stopped
+  tags:
+    - kill-the-beaver
+
```
mancdaz: Let's please use something a bit more professional sounding for the tag.
I'd also say that the tag on a single task isn't much use - you probably want to tag all these tasks.

rpcd/playbooks/roles/rpc_pre_upgrade/tasks/main.yml
```diff
@@ -30,3 +30,6 @@
 # Perform log rotation and cleanup
 - include: log_rotation.yml
 
+# Remove beaver from all the things
+- include kill_beaver.yml
+
```
mancdaz: can we change the name here. Perhaps uninstall_beaver ?

mancdaz: you're also missing a colon after 'include', hence the linting failures.

rackernexus:
:eyes: [Preview](https://staging.developer.rackspace.com/staging.horse/build-52c1661c59/docs/private-cloud/release-notes/)

rpcd/playbooks/roles/rpc_pre_upgrade/tasks/remove_beaver.yml
```diff
@@ -0,0 +1,57 @@
...
+  file:
+    path: /etc/logrotate.d/beaver
+    state: absent
+
+- name: Remove beaver log files
+  file:
+    path: /var/log/beaver
+    state: absent
+
+- name: uninstall beaver
```
sigmavirus24: We should probably sentence case this, but I don't think it's a blocker.

mancdaz: @sigmavirus24 agree - it's worth doing now while we can.

claco: Is that our norm, and do we happen to document that in contributing?

prometheanfire: addressed

rackernexus:
HULK [PREVIEW](https://staging.developer.rackspace.com/staging.horse/build-406a0133dc/docs/private-cloud/release-notes/)

rpcd/playbooks/roles/rpc_pre_upgrade/tasks/remove_beaver.yml
sigmavirus24: > Is that our norm, and do we happen to document that in contributing/style guide?

It is our norm. I don't know if we document it there.

rpcd/playbooks/roles/rpc_pre_upgrade/tasks/main.yml
prometheanfire: all this has been done...

prometheanfire: so how do I get this to go away?

prometheanfire:
recheck_ceph

rpcd/playbooks/roles/rpc_pre_upgrade/tasks/main.yml
sigmavirus24: lol, no clue. GitHub is terrible.

alextricity25:
Testing this now.

rpcd/playbooks/roles/rpc_pre_upgrade/tasks/remove_beaver.yml
```diff
@@ -0,0 +1,57 @@
...
+    state: absent
+
+- name: Remove beaver log files
+  file:
+    path: /var/log/beaver
+    state: absent
+
+- name: Uninstall beaver pip package
+  pip:
+    name: beaver
```
alextricity25:
```
root@cantu-testup-rpc24_utility_container-de23582e:~# pip freeze
appdirs==1.4.0
Babel==2.1.1
Beaver==33.3.0
```
Should this be ``Beaver`` and not ``beaver``?

prometheanfire: Functionally it doesn't mater.  For correctness it should be beaver, but I'd rather not change it now.

(openstack-infra has this problem with some requirements things)

sigmavirus24: pip normalizes the names. Technically "pip install Django" is the correct way to install that package but "pip install django" will also work, just like "pip uninstall beaver" will work. ;)

rpcd/playbooks/roles/rpc_pre_upgrade/tasks/remove_beaver.yml
```diff
@@ -0,0 +1,57 @@
+---
```
alextricity25: This has already been mentioned, but these tasks need to target the containers and hosts beaver is installed on. Not just localhost.

prometheanfire: yep, it's becoming it's own playbook

rackernexus:
:mag_right: [Preview](https://staging.developer.rackspace.com/staging.horse/build-5831ef445e/docs/private-cloud/release-notes/)

alextricity25:
I like the previous approach of having these tasks in the ``rpc-pre-upgrade.sh`` playbook. After all, the reason the role is there in the first place is so we had one place to perform tasks that needed to be done before an upgrade.
Was there something wrong with keeping the tasks where they were and using ``delegate_to`` to target all of the hosts?

prometheanfire:
I tried looking up if delegate_to can work for targeting all hosts and was unable to determine if so.

alextricity25:
Hrmmmm....Shouldn't it just be ``delegate_to: all``?

alextricity25:
Or maybe:
```
delegate_to: {{ item }}
with_items: {{ groups['all'] }}
```

rackernexus:
HULK [PREVIEW](https://staging.developer.rackspace.com/staging.horse/build-6d457c1ecb/docs/private-cloud/release-notes/)

prometheanfire:
ok, pushed

prometheanfire:
guess the -2 follows, and why do we use this instead of gerrit?

rpcd/playbooks/roles/rpc_pre_upgrade/tasks/remove_beaver.yml
```diff
@@ -0,0 +1,90 @@
...
+  delegate_to: {{ item }}
+  with_items: {{ groups['all'] }}
+  tags:
+    - remove-beaver
+
+- name: Uninstall beaver pip package
+  pip:
+    name: beaver
+    state: absent
+  ignore_errors: true
```
mancdaz: As previously discussed, I'd rather see us check what hosts beaver is installed on and then target this task only at those hosts. I really don't like the idea of accepting that we're going to have errors/fails due to the fact that beaver doesn't exist on a host, and then just ignoring them - we may end up ignoring an actual error.

evrardjp:
I'll wait for this to be in its own playbook (and targetting the proper hosts) to increase the speed of execution.

👎  in the meantime.

rpcd/playbooks/roles/rpc_pre_upgrade/tasks/remove_beaver.yml
prometheanfire: And as previously discussed I'd really rather not

mancdaz: Actually you shouldn't even need the ignore_errors (and so don't need to do what I'm asking either) - if the package isn't installed then it will just pass by with no changes. What is the ignore_errors there for?

prometheanfire: I tested it without and got errors (running this same task list multiple times.

sigmavirus24: > I tested it without and got errors (running this same task list multiple times).

What were the errors exactly?

mancdaz:
@evrardjp @sigmavirus24 a separate playbook, or perhaps a separate play within the rpc-pre-upgrades playbook...?

alextricity25:
Sorry @prometheanfire, I might of mis-lead you yesterday when asking if this could be part of the rpc-pre-upgrade role.
It would be better if this was it's own play within the rpc-pre-upgrade.yml playbook. Especially since we are targeting all of the hosts here.

sigmavirus24:
> a separate playbook, or perhaps a separate play within the rpc-pre-upgrades playbook...?

I suspect either could work really.

prometheanfire:
Well, I initially developed this as it's own playbook (all in one file).  If you are fine with that I can change back to that...

sigmavirus24:
> Well, I initially developed this as it's own playbook (all in one file). If you are fine with that I can change back to that...

I think the individual tasks make the most sense in the pre-upgrade role like you already have. They just need to be targeted differently than the rest of the tasks.

prometheanfire:
ok, I guess I'll wait while people argue over if it should be part of the playbook (as @sigmavirus24 wants) or it's own playbook (as @evrardjp wants)

rpcd/playbooks/roles/rpc_pre_upgrade/tasks/remove_beaver.yml
d34dh0r53: Beaver *should* be installed everywhere, every host, every container.  I personally think that it is safe to assume that it is and an a posteriori check is not required and will simply consume resources and time.

d34dh0r53: One final thought is that errors are cause for concern about the general state of the environment as it is indicative of things being added without all of the plays being completed  successfully.  I do think that we should ignore said errors but a release note stating that errors in the remove beaver role should at least warrant a sanity check.

d34dh0r53:
I think this belongs in the `rpc_pre_upgrade` role in place of the Beaver restart that takes so long.  Having this in it's own play doesn't serve a purpose in my eyes as I can't see a time when you would run it on it's own.  To me it logically fits in the `rpc_pre_upgrade` role with tags that allow you to run it exclusively should you ever want to.

rpcd/playbooks/roles/rpc_pre_upgrade/tasks/remove_beaver.yml
alextricity25: I still think these tasks should be moved into the ``rpc-pre-upgrade.yml`` file as it's own play(not it's own playbook file). This is because of the same reason stated [here](https://github.com/rcbops/rpc-openstack/pull/1423). Right now were are iterating through all of the hosts one-by-one, which isn't very efficient.

alextricity25:
I agree with Dave. Just add the changes in place of what is being done here: https://github.com/rcbops/rpc-openstack/pull/1423

prometheanfire:
OK, that's similar to how I did this to begin with.  I'll bring it up at the Mitaka meeting to make sure no one is going to block it if I do that.

mancdaz:
I still would ask what errors are being ignored in the pip uninstall task? Why do we expect errors there? Why is it ok to ignore them?

And yes, this PR also needs to remove the restart stuff (the patch for which was always an interim measure for rc1 to stop it taking so long)

rackernexus:
:eyes: [Preview](https://staging.developer.rackspace.com/staging.horse/build-3cae648fbf/docs/private-cloud/release-notes/)

prometheanfire:
Now, if people could remove their -2's that'd be nice.

rackernexus:
Your content preview [is now ready](https://staging.developer.rackspace.com/staging.horse/build-1bf59be48c/docs/private-cloud/release-notes/). :bow:

sigmavirus24:
> Now, if people could remove their -2's that'd be nice.

Those people need to re-review and update their reviews. GitHub doesn't auto-remove them and they're not -2's, GitHub just persists them because they've never used a real review tool before.

rpcd/playbooks/rpc-pre-upgrades.yml
```diff
@@ -19,18 +19,60 @@
   roles:
     - { role: "rpc_pre_upgrade", tags: [ "rpc-pre-upgrades" ] }
 
-- name: restart beaver
+- name: remove beaver beaver
```
sigmavirus24: "remove beaver beaver"?

sigmavirus24:
Oh wow, GitHub is so busted. I just updated my review to be a "comment" and it's saying I requested changes still.

rackernexus:
If you want to double-check your formatting, check [the preview](https://staging.developer.rackspace.com/staging.horse/build-148234ccda/docs/private-cloud/release-notes/).

sigmavirus24:
recheck_swift

sigmavirus24:
Approval pending tests (of course)

alextricity25:
recheck_ceph

rpcd/playbooks/rpc-pre-upgrades.yml
```diff
@@ -19,18 +19,60 @@
...
+    - name: Check if beaver is installed
+      shell: pip show -q Beaver
+      register: is_beaver_installed
+
+    - name: Uninstall beaver pip package
+      pip:
+        name: beaver
+        state: absent
+      when: is_beaver_installed.rc == 0
+
```
mancdaz: 2 things:

- ```pip show -q Beaver``` will always return 0, whether the package is installed or not, so this won't work
- we actually don't need the test for whether it's installed (I know I was earlier arguing for it, but that was when we were ignoring apparent errors that were encountered). The pip module with state absent will do the right thing

Other than that, thanks for moving this to a play in the playbook - a much better approach all round. Ultimately I think all tasks in this role should be moved up here and the role removed, but that's for another day.

mancdaz: OK I just tested this properly and well, it seems good old pip in my mac (crappy old 1.5.6) behaves differently to the version we're using (8.1.2), so I'm wrong about the return code always being 0.

And looking again to see if just letting the pip module do the right thing works, I can see the errors you were running in to. It seems the pip module is somewhat busted in ansible 1.9.4 (we checked, and it's fixed in 1.9.5 but breaks some lxc container stuff), in that it will error with an rc of 1 if the package is not installed.

Chatting with @evrardjp and @odyssey4me the best way we can handle this cleanly is to do something like (pseudo code):

failed_when stderr does not contain the string 'not installed' and stdout does not contain the string 'Successfully uninstalled'

prometheanfire: it works

    pip show -q beaver
    echo $?
    1

The pip module with state absent does not do the right thing if it's already missing.  I tested it.

The workaround used in this changeset works now and in the future (when we won't need it).  I don't see this as a compelling reason to hold this up.

mancdaz: or just add 

failed_when: false
changed_when: false 

in your 'pip show -q Beaver' task

mancdaz: actually this latter way is the only way that will work. The pip module is so broken that you can't do it the way I suggested originally.

So the check task becomes:

```
    - name: Check if beaver is installed
      shell: pip show -q Beaver
      register: is_beaver_installed
      failed_when: false
      changed_when: false
```

I just tested the above, and it works.

Sorry for all the gibbering, and thankyou for your patience.  With that change, I think we'll be good to go.

rackernexus:
:mag_right: [Preview](https://staging.developer.rackspace.com/staging.horse/build-f6f6335f3e/docs/private-cloud/release-notes/)

mancdaz:
👍  Tested working when beaver is, and is not, installed.

Thanks @prometheanfire !

alextricity25:
surprise surprise...
```
15:53:00 TASK: [Verify Checks & Alarms are registered] ********************************* 
16:18:47 [0;31mfailed: [jrpcaioiad-869] => {"attempts": 5, "changed": false, "cmd": ". /openstack/venvs/maas-r13.1.0rc1/bin/activate\n /opt/rpc-openstack/scripts/rpc-maas-tool.py verify-created --entity jrpcaioiad-869", "delta": "0:04:00.570247", "end": "2016-09-23 16:18:47.953679", "failed": true, "failed_when_result": true, "rc": 1, "start": "2016-09-23 16:14:47.383432", "stdout_lines": [], "warnings": []}[0m
16:18:47 [0;31mstderr: Traceback (most recent call last):
16:18:47   File "/opt/rpc-openstack/scripts/rpc-maas-tool.py", line 618, in <module>
16:18:47     cli = RpcMassCli()
16:18:47   File "/opt/rpc-openstack/scripts/rpc-maas-tool.py", line 214, in __init__
16:18:47     self.args.raxrcpath)
16:18:47   File "/opt/rpc-openstack/scripts/rpc-maas-tool.py", line 54, in __init__
16:18:47     self._add_links()
16:18:47   File "/opt/rpc-openstack/scripts/rpc-maas-tool.py", line 96, in _add_links
16:18:47     for entry in self.overview:
16:18:47   File "/openstack/venvs/maas-r13.1.0rc1/local/lib/python2.7/site-packages/libcloud/common/types.py", line 115, in __iter__
16:18:47     self._load_all()
16:18:47   File "/openstack/venvs/maas-r13.1.0rc1/local/lib/python2.7/site-packages/libcloud/common/types.py", line 141, in _load_all
16:18:47     value_dict=self._value_dict)
16:18:47   File "/openstack/venvs/maas-r13.1.0rc1/local/lib/python2.7/site-packages/rackspace_monitoring/drivers/rackspace.py", line 233, in _get_more
16:18:47     response = self.connection.request(value_dict['url'], params)
16:18:47   File "/openstack/venvs/maas-r13.1.0rc1/local/lib/python2.7/site-packages/rackspace_monitoring/drivers/rackspace.py", line 152, in request
16:18:47     raw=raw
16:18:47   File "/openstack/venvs/maas-r13.1.0rc1/local/lib/python2.7/site-packages/libcloud/common/openstack.py", line 227, in request
16:18:47     raw=raw)
16:18:47   File "/openstack/venvs/maas-r13.1.0rc1/local/lib/python2.7/site-packages/libcloud/common/base.py", line 862, in request
16:18:47     response = responseCls(**kwargs)
16:18:47   File "/openstack/venvs/maas-r13.1.0rc1/local/lib/python2.7/site-packages/libcloud/common/base.py", line 179, in __init__
16:18:47     headers=self.headers)
16:18:47 libcloud.common.exceptions.BaseHTTPError[0m
```

alextricity25:
recheck_swift

alextricity25:
recheck_swift

evrardjp:
Not tested but it looks good to me. 👍

PR #1430, opened 2016-09-20
Branding release note

evrardjp:
Our 12.2 branch is lacking release notes for its branding.

Connected #1088

### DIFF
--- a/releasenotes/notes/branding-de80e7d9c1392e69.yaml
+++ b/releasenotes/notes/branding-de80e7d9c1392e69.yaml
@@ -1,0 +1,9 @@
+---
+fixes:
+  - This release fixes a race condition where the
+    branding could be overriden depending on the playbook
+    execution order. Deployers have to make sure the
+    `horizon_custom_uploads` and the
+    `rackspace_static_files_folder` from
+    `rpcd/etc/user_variables.yml` are properly copied
+    into `/etc/openstack_deploy/user_variables.yml`.


evrardjp:
recheck_upgrade

claco:
recheck_upgrade

releasenotes/notes/branding-de80e7d9c1392e69.yaml
```diff
@@ -0,0 +1,9 @@
+---
+fixes:
+  - This release fixes a race condition where the
+    branding could be overriden depending on the playbook
+    execution order. Deployers have to make sure the
```
prometheanfire: I think if deployers have do do something it should probably go into another section (not sure which at this time though).

tomc603: @evrardjp The file name `user_variables.yml` is invalid in the master branch now.

tomc603: @prometheanfire I don't think these are instructions that don't already appear in the proper location.

sigmavirus24: Based on the conversation in #1088 this doesn't even sound like the right set of instructions.

Also, shouldn't this be described in a known-issue section? I feel like this is the description of a fix for a known bug that is not fixed.

git-harry:
👎 liberty-12.2 does not support the use of release notes (reno). Any documentation would need to be in https://github.com/rackerlabs/docs-rpc

evrardjp:
@git-harry My idea is to come to a consensus to the "what do we say" and "if we say it".
If we agree on the method of this backport (i.e. doc change with a proper message), then I'll work with docs to make it clean to the proper place.

@tomc603 These instructions will probably be moved to a proper place in the documentation section.

@sigmavirus24 If you know what the fix would be, feel free to comment, edit and/or submit another patchset. I'm pretty confident that if we are doing the branding with OSA (as we should be), there will be no problem of branding anymore. I'll be listening to your advise, as usual, anyway.

In any case, please don't merge this, but please continue to give feedback.

sigmavirus24:
@evrardjp I think what's missing is the context that the bug was fixed by https://github.com/rcbops/rpc-openstack/pull/1155 and that this is just noting that fact in our releasnotes/documentation. Is that correct?

If so, I think the explanation can be clearer, but that should address my concern about fixing this instead of just documenting a fix that didn't seem to match what was discussed on the original issue.

evrardjp:
@sigmavirus24 It's just the release note part you're right.
Indeed, the context may seem unclear. Therefore an explicit explanation will be given.

tomc603:
@evrardjp After confirming with Docs folks, is this issue OK to close?

git-harry:
Closing, this change is being made in the docs repo https://github.com/rackerlabs/docs-rpc/issues/619

Issue #1434, opened 2016-09-23
test-upgrade fails on "Get elastic search indicies"


evrardjp:
I've never deployed logging stash (or at least I didn't ask for it by setting deploy logging to no in my exports) and my upgrade fails on the post-upgrade role, when it tries to do actions on elastic search.

The task failing is: 
"Get elastic search indicies" (there is a typo on the task name btw), which returns msg: Socket error: [Errno 111] Connection refused to http://172.29.239.40:9200/_cat/indices?v

As you can see there is an elastic search url defined (weird). I didn't got that issue on my first AIO upgrades, so I have to check what would be the cause. 

Let's keep this card for the record.

mattt416:
Hi @evrardjp, I think the issue here is actually a bug in scripts/deploy.sh.  I've pushed a PR to address.

alextricity25:
I created a backport for this. Please review when you get the chance.

mattt416:
Closing this card since the backport has a separate issue for tracking.

git-harry:
@mattt416 this is now in master and mitaka-13.1, what other backports are required?

mattt416:
@git-harry What other backports are required to resolve this issue?  That should be it as far as I'm aware.

alextricity25:
I personally feel like no other backport is needed

git-harry:
@mattt416 Okay, not offering an opinion, just confirming that all our supported branches have been considered. I'll close this card.

PR #1439, opened 2016-09-23
Remove ELK config before inventory is generated

mattt416:
Currently, we rm env.d/{elasticsearch,logstash,kibana}.yml after we run
openstack-ansible, resulting in ELK-related items being in the
inventory.  This commit updates scripts/deploy.sh to rm these files
before, which means you will no longer get ELK containers if
DEPLOY_ELK != "yes".

Connects https://github.com/rcbops/rpc-openstack/issues/1434

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -142,6 +142,13 @@ fi
 ./scripts/pw-token-gen.py --file /etc/openstack_deploy/user_osa_secrets.yml
 ./scripts/pw-token-gen.py --file $RPCD_SECRETS
 
+# ensure that the ELK containers aren't created if they're not
+# going to be used
+# NOTE: this needs to happen before ansible/openstack-ansible is first run
+if [[ "${DEPLOY_ELK}" != "yes" ]]; then
+  rm -f /etc/openstack_deploy/env.d/{elasticsearch,logstash,kibana}.yml
+fi
+
 # Apply any patched files.
 cd ${RPCD_DIR}/playbooks
 openstack-ansible -i "localhost," patcher.yml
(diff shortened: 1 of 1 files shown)


rackernexus:
:mag_right: [Preview](https://staging.developer.rackspace.com/staging.horse/build-9cf4879e21/docs/private-cloud/release-notes/)

evrardjp:
Not tested but it looks good to me.

rackernexus:
If you want to double-check your formatting, check [the preview](https://staging.developer.rackspace.com/staging.horse/build-d521e3210f/docs/private-cloud/release-notes/).

rackernexus:
HULK [PREVIEW](https://staging.developer.rackspace.com/staging.horse/build-a4f9a3df08/docs/private-cloud/release-notes/)

git-harry:
👎 this looks to be moving the code into the AIO section of the script which I assume will cause problems for multinode deployments. I'm not entirely clear what is being solved by this, my understanding of the inventory was that it's dynamic in the sense that every time openstack-ansible is called the inventory gets generated. The section of code between where it was and where you've moved it to does not require the ELK containers to exist or try to create them.

mattt416:
Hi @git-harry,

I'm working off the original commit here -- https://github.com/rcbops/rpc-openstack/commit/4df376bfe312e3bacc9b66380db85b90e92e58b0.  When that was added, there was no instance of ansible or openstack-ansible being called before the env.d files were removed.

Anyway, I have tested two identical deploys where one had this patch and one didn't and the one that didn't ended up with the following containers, despite DEPLOY_ELK="no":

```
mitaka-elk-1_elasticsearch_container-2fe408d3
mitaka-elk-1_kibana_container-7bf43cfa
mitaka-elk-1_logstash_container-5e1e153f
```

I suppose once the inventory is first generated with groups and hosts defined, removing these env.d scripts does not change any of that (that'd be quite dangerous I'd imagine).

With that said, I did originally leave this removal outside of the DEPLOY_AIO block, but then moved it in for two reasons:

1. There are other removals of unwanted service configurations which only happens when DEPLOY_AIO="yes", so this keeps things consistent
2. We assume the deployer has correctly created /etc/openstack_deploy, so should we really be going in and removing files that they've put in?

I'm open to thoughts and suggestions, but will update my commit message to make this more clear.

mattt416:
@git-harry I've moved where this rm happens, but have left it outside of the DEPLOY_AIO block so we do not change behaviour here.

git-harry:
👍 in my testing this prevented ELK related containers from existing in the inventory.

Issue #1300, opened 2016-08-09
Reporting the remaining steps when upgrade.sh fails


git-harry:
###### Please provide the output to 'git describe --all' for  all deployments where you have confirmed it occurs.
all versions
##### What type of issue is this? Please tick the appropriate box.
- [x] bug - existing functionality that does not work for you as documented
- [ ] feature - new functionality that you would like to see added

##### What is the expected behavior?
If upgrade.sh fails it should return a list of the remaining steps required to complete the upgrade, similar to what is done by [openstack-ansible](https://github.com/openstack/openstack-ansible/blob/master/scripts/run-upgrade.sh#L66-L78).
##### What is actually happening?
If the upgrade fails while running OSA's run-upgrade.sh script, the remaining OSA steps will be reported but under no circumstances are RPCO specific steps listed.
##### What were you attempting to do when you encountered this error (give as much detail as possible)? 
This can be seen when there is a failure while running upgrade.sh
##### How can we reproduce this issue?
Run scripts/upgrade.sh and view the output if it fails.
##### How can we verify your issue is addressed?
If upgrade.sh fails it will list the remaining steps to be completed.
##### If you are using or know of a workaround please describe the necessary steps?
Reviewing the script will allow one to work out what is left to do.
##### What is your evaluation on the importance of this issue, and why?
This does not prevent upgrades from happening but does make failure recovery harder.

----------------------------------------------------------------------------------------------------------------------------
###### For each pull request, update the branch line to reference it.
- [ ] liberty-12.2

##### Upgrade impact. Consider the impact on moving between major, minor and patch releases.
- [ ] This change does not impact upgrades
- [ ] This change addresses a potential upgrade impact
- [ ] The upgrade impact is being tracked by another issue.

The issue tracking the upgrade impact is:

##### Testing impact
- [ ] An issue/pull-request has been create or exists to track the testing gap, and is referenced below
- [ ] This change includes additional testing
- [ ] This change is tested by the existing setup

Additional testing links:

##### Release note
- [ ] This change includes a release note.
- [ ] This change does not require a release note.

##### Does your change require documentation?
- [ ] Yes, and I submitted an issue to [Docs issues](https://github.com/rackerlabs/docs-rpc/issues "Docs issues")
- [ ] Yes, and I contributed a documentation patch to [docs-rpc](https://github.com/rackerlabs/docs-rpc "docs-rpc")
- [ ] No documentation required for this change

[...]

PR #1442, opened 2016-09-26
List the remaining steps in upgrade.sh when fails

cloader89:
This addresses the issue of not knowing the remaining steps in
upgrade.sh should it fail. It will print the commands for the
person doing the upgrade to essentially resume instead of
restarting.

Connects #1300


cloader89:
recheck_upgrade

Issue #1436, opened 2016-09-23
deploy.sh updates defaults files rather than add to overrides


mattt416:
###### Please provide the output to 'git describe --all' for  all deployments where you have confirmed it occurs.

Mitaka versions of scripts/deploy.sh should override variables by adding them to the respective overrides files, not overwriting them in the defaults files.

##### What type of issue is this? Please tick the appropriate box.
- [x] bug - existing functionality that does not work for you as documented
- [ ] feature - new functionality that you would like to see added

##### What is the expected behavior?

Variable overrides should be added to the overrides files, not updated in the defaults files.

##### What is actually happening?

Variables are being updated in the defaults files.

##### What were you attempting to do when you encountered this error (give as much detail as possible)? 

No error, just a bad practice.

##### How can we reproduce this issue?

Deploy an AIO using scripts/deploy.sh.

##### How can we verify your issue is addressed?

The overrides files are correctly populated with variable overrides.

##### What is your evaluation on the importance of this issue, and why?

----------------------------------------------------------------------------------------------------------------------------
###### For each pull request, update the branch line to reference it.
- [x] master https://github.com/rcbops/rpc-openstack/pull/1444
- [ ] mitaka-13.1
- [ ] mitaka-13.0
- [ ] liberty-12.2
- [ ] kilo

##### Upgrade impact. Consider the impact on moving between major, minor and patch releases.
- [ ] This change does not impact upgrades
- [ ] This change addresses a potential upgrade impact
- [ ] The upgrade impact is being tracked by another issue.

The issue tracking the upgrade impact is:

##### Testing impact
- [ ] An issue/pull-request has been create or exists to track the testing gap, and is referenced below
- [ ] This change includes additional testing
- [ ] This change is tested by the existing setup

Additional testing links:

##### Release note
- [ ] This change includes a release note.
- [ ] This change does not require a release note.

##### Does your change require documentation?
- [ ] Yes, and I submitted an issue to [Docs issues](https://github.com/rackerlabs/docs-rpc/issues "Docs issues")
- [ ] Yes, and I contributed a documentation patch to [docs-rpc](https://github.com/rackerlabs/docs-rpc "docs-rpc")
- [ ] No documentation required for this change

Additional documentation links (please add links if you have ticked yes):

git-harry:
[...]

PR #1444, opened 2016-09-27
[WIP] Update deploy.sh to write to overrides files

mattt416:
This commit makes four notable changes:

1. Replaces RPCD_VARS with RPCD_OVERRIDES since we only operate against
   the overrides file
2. Updates the DEPLOY_AIO block to write variables overrides to the
   respective overrides files, rather than updating/appending to the
   defaults files
3. Sets apply_security_hardening to True in
   user_osa_variables_defaults.yml, since this is the default value we
   wish to operate against.
4. Updates the DEPLOY_HARDENING block to exit if DEPLOY_HARDENING is
   not enabled on a non-AIO build.  This is done because our tooling
   (outside of an AIO deployment) should not be making modifications to
   the overrides files.  These files are specifically for
   support/deployer and if overrides are needed then these should be
   made manually.  We also move this block to the top of deploy.sh so
   we can exit early if this condition is met and also add a block to
   DEPLOY_AIO so that developers can still disable hardening when
   deploying on an AIO.

Connects https://github.com/rcbops/rpc-openstack/issues/1436

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -20,11 +20,22 @@ export UNAUTHENTICATED_APT=${UNAUTHENTICATED_APT:-no}
 
 export BASE_DIR='/opt/rpc-openstack'
 export OA_DIR='/opt/rpc-openstack/openstack-ansible'
+export OA_OVERRIDES='/etc/openstack_deploy/user_osa_variables_overrides.yml'
 export RPCD_DIR='/opt/rpc-openstack/rpcd'
-export RPCD_VARS='/etc/openstack_deploy/user_rpco_variables_defaults.yml'
+export RPCD_OVERRIDES='/etc/openstack_deploy/user_rpco_variables_overrides.yml'
 export RPCD_SECRETS='/etc/openstack_deploy/user_rpco_secrets.yml'
 
 source ${BASE_DIR}/scripts/functions.sh
+
+if [[ "$DEPLOY_AIO" != "yes" ]] && [[ "$DEPLOY_HARDENING" != "yes" ]]; then
+  echo "** DEPLOY_HARDENING should no longer be used **"
+  echo "To disable security hardening, please add the following line to"
+  echo "/etc/openstack_deploy/user_osa_variables_overrides.yml and then"
+  echo "re-run this script:"
+  echo ""
+  echo "apply_security_hardening: false"
+  exit 1
+fi
 
 # begin the bootstrap process
 cd ${OA_DIR}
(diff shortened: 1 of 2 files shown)


rackernexus:
HULK [PREVIEW](https://staging.developer.rackspace.com/staging.horse/build-516e0936d9/docs/private-cloud/release-notes/)

alextricity25:
The commit format needs updating, other than that the code LGTM. Testing it now on a greenfield mitaka build.

evrardjp:
It should fix our ceph issue and seems clear.

evrardjp:
This looks good for greenfield. Upgrades are not impacted, so it should be fine.

scripts/deploy.sh
```diff
@@ -66,17 +67,17 @@ if [[ "${DEPLOY_AIO}" == "yes" ]]; then
...
     sed -i "s/kibana_password:.*/kibana_password: ${ADMIN_PASSWORD}/" $RPCD_SECRETS
     # set the load balancer name to the host's name
-    sed -i "s/lb_name: .*/lb_name: '$(hostname)'/" $RPCD_VARS
+    echo "lb_name: '$(hostname)'" >> $RPCD_OVERRIDES
     # set the notification_plan to the default for Rackspace Cloud Servers
-    sed -i "s/maas_notification_plan: .*/maas_notification_plan: npTechnicalContactsEmail/" $RPCD_VARS
-    # the AIO needs this enabled to test the feature, but $RPCD_VARS defaults this to false
-    sed -i "s/cinder_service_backup_program_enabled: .*/cinder_service_backup_program_enabled: true/" /etc/openstack_deploy/user_osa_variables_defaults.yml
+    echo "maas_notification_plan: npTechnicalContactsEmail" >> $RPCD_OVERRIDES
+    # the AIO needs this enabled to test the feature, but user_rpco_variables_defaults defaults this to false
```
git-harry: user_rpco_variables_defaults.yml

scripts/deploy.sh
```diff
@@ -122,19 +123,28 @@ if [[ ! -f /etc/openstack_deploy/user_osa_secrets.yml ]] && [[ -f /etc/openstack
...
+if [[ ! -f "$OA_OVERRIDES" ]]; then
+  cp "${RPCD_DIR}"/etc/openstack_deploy/user_osa_variables_overrides.yml $OA_OVERRIDES
+fi
+if [[ ! -f "$RPCD_OVERRIDES" ]]; then
+  cp "${RPCD_DIR}"/etc/openstack_deploy/user_rpco_variables_overrides.yml $RPCD_OVERRIDES
+fi
+
 # Apply host security hardening with openstack-ansible-security
 # The is applied as part of setup-hosts.yml
 if [[ "$DEPLOY_HARDENING" == "yes" ]]; then
```
git-harry: I think this if block should be changed to something along the lines of:

```
if [[ "$DEPLOY_HARDENING" != "yes" ]]; then
  echo -e 'DEPLOY_HARDENING should no longer be used. Please add the following line to /etc/openstack_deploy/user_osa_variables_overrides.yml to disable security hardening.\n\napply_security_hardening: false\n'
  exit 1
fi
```

With the way that the variables files are split between defaults and overrides there is not an appropriate place for this script to write the var. The overrides files are supposed to come with a guarantee that we will not modify them and defaults files are supposed to contain defaults. In addition, given that the default is now to deploy security hardening, rerunning the script and forgetting to set DEPLOY_HARDENING to no on a deployment where it was disabled is likely to be problematic.

git-harry:
"1. Renames RPCD_VARS to RPCD_OVERRIDES to make it more clear that we're
   operating on the overrides files"

This line of the commit message feels a bit misleading to me. Not only is the name changed but also the value.

git-harry:
👎 forgot to add my vote with my previous comments. Mostly concerned about DEPLOY_HARDENING modifying the overrides file but in general I like the approach.

scripts/deploy.sh
alextricity25: I don't like how that involves manual intervention, but I do like how that catches the possibility of someone mistakenly deploying security hardening. :+1: from me.

scripts/deploy.sh
mattt416: Fixing

scripts/deploy.sh
mattt416: That sounds sensible.  At present `apply_security_hardening` is commented out in user_osa_variables_defaults.yml.  No reason for that not to be uncommented right?

rackernexus:
HULK [PREVIEW](https://staging.developer.rackspace.com/staging.horse/build-7adfc949b7/docs/private-cloud/release-notes/)

mattt416:
@git-harry @alextricity25 I think I've made all the requested changes.  Including @major so he can confirm these security hardening changes are satisfactory.

rackernexus:
:eyes: [Preview](https://staging.developer.rackspace.com/staging.horse/build-04cb5caad1/docs/private-cloud/release-notes/)

git-harry:
👍 this looks good to me

Issue #1437, opened 2016-09-23
os_neutron ns-metadata-proxy cleanup error attempting to kill PID


Apsu:
##### On which version(s) do you encounter this issue?
master upgrade from liberty-12.2

###### Please provide the output to 'git describe --all' for  all deployments where you have confirmed it occurs.
liberty-12.2 to master

##### What type of issue is this? Please tick the appropriate box.
- [x] bug - existing functionality that does not work for you as documented
- [ ] feature - new functionality that you would like to see added

Upstream bug: https://bugs.launchpad.net/openstack-ansible/+bug/1627185

##### What is the expected behavior?
Old Neutron ns-metadata-proxy processes are successfully killed prior to restarting into the upgraded versions

##### What is actually happening?
Sometimes the PID from the pgrep/grep is either no longer there or the filtering malforms it so this happens:

```
13:35:48 failed: [jrpcaioiad-3a8_neutron_agents_container-101ec306] => {"changed": true, "cmd": "for ns_pid in $(pgrep neutron-ns-meta); do\n echo $(readlink -f \"/proc/$ns_pid/exe\") | grep -qv \"13.3.4\"\n if [ $? -eq 0 ]; then\n (echo \"old metadata proxy pid found running clean up on $ns_pid\"; kill -9 \"$ns_pid\")\n fi\n done", "delta": "0:00:00.049977", "end": "2016-09-23 13:35:48.950575", "rc": 1, "start": "2016-09-23 13:35:48.900598", "warnings": []}
13:35:48 stderr: /bin/sh: 4: kill: No such process
13:35:48
13:35:48 /bin/sh: 4: kill: No such process
13:35:48 stdout: old metadata proxy pid found running clean up on 1337
13:35:48 old metadata proxy pid found running clean up on 2781
```

##### What were you attempting to do when you encountered this error (give as much detail as possible)? 
http://jenkins.propter.net/job/JJB-RPC-AIO_mitaka-upgrade this automatic gate job

##### How can we reproduce this issue?
Rerun the gate job a few times

##### How can we verify your issue is addressed?
Gate jobs don't fail from missing PIDs any more

##### If you are using or know of a workaround please describe the necessary steps?
Seems to be a race condition or process filtering oddity, not sure of a workaround without hand-modifying the Galaxy role

##### What is your evaluation on the importance of this issue, and why?
I was told this was high priority

----------------------------------------------------------------------------------------------------------------------------
###### For each pull request, update the branch line to reference it.
[...]

PR #1449, opened 2016-09-29
[Backport] Remove beaver from existing containers

alextricity25:
We basically do the beaver install tasks in inverse.
No need to restart beaver either.

Connected rcbops/u-suk-dev#180
Connected rcbops/u-suk-dev#467

(cherry picked from commit 04ae871fde8600b1533df7c5146134978d83a6c6)

### DIFF
--- a/rpcd/playbooks/rpc-pre-upgrades.yml
+++ b/rpcd/playbooks/rpc-pre-upgrades.yml
@@ -19,18 +19,63 @@
   roles:
     - { role: "rpc_pre_upgrade", tags: [ "rpc-pre-upgrades" ] }
 
-- name: restart beaver
+- name: remove beaver
   hosts: all
   tags:
     - rpc-pre-upgrades
+    - remove-beaver
   tasks:
     - name: Check if beaver is installed
       stat:
         path: /etc/init.d/beaver
       register: beaver_init_script
+      changed_when: false
 
-    - name: Restart beaver if installed
+    - name: Make sure beaver is stopped
       service:
         name: beaver
-        state: restarted
+        enabled: no
+        state: stopped
       when: beaver_init_script.stat.exists
+
+    - name: Remove beaver etc files
+      file:
+        path: /etc/beaver
+        state: absent
+
+    - name: Remove beaver init files
+      file:
+        path: /etc/init.d/beaver
+        state: absent
+
+    - name: Remove beaver logrotate files
+      file:
+        path: /etc/logrotate.d/beaver
+        state: absent
+
+    - name: Remove beaver log files
+      file:
+        path: /var/log/beaver
+        state: absent
+
+    - name: Check if beaver is installed
+      shell: pip show -q Beaver
+      register: is_beaver_installed
+      failed_when: false
+      changed_when: false
+
+    - name: Uninstall beaver pip package
+      pip:
+        name: beaver
+        state: absent
+      when: is_beaver_installed.rc == 0
+
+    - name: Remove beaver user
+      user:
+        name: beaver
+        state: absent
+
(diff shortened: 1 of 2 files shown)


alextricity25:
recheck_swift

alextricity25:
If my understanding of this is correctly, the upgrade job won't pass until https://github.com/rcbops/rpc-openstack/issues/1437 gets included in the next SHA bump. Marking as blocked for now.

git-harry:
recheck_upgrade

rpcd/playbooks/rpc-pre-upgrades.yml
```diff
@@ -19,18 +19,63 @@
...
   hosts: all
   tags:
     - rpc-pre-upgrades
+    - remove-beaver
   tasks:
     - name: Check if beaver is installed
       stat:
         path: /etc/init.d/beaver
       register: beaver_init_script
+      changed_when: false
```
Apsu: https://github.com/ansible/ansible-modules-core/blob/stable-2.1/files/stat.py#L409

This module never returns changed=True, so this guard is unnecesary.

odyssey4me: @Apsu Good call, although you may wish to reference https://github.com/ansible/ansible-modules-core/blob/stable-1.9/files/stat.py#L351 instead, seeing as that's the Ansible version being used. :)

alextricity25: That would need to be changed in the original commit first. I presume it will also need it's own card.
For the backport, I think ti's fine, right?

git-harry: Yeah, it's a backport and that's only a nit so I think it's fine.

git-harry:
👍  seems like a straight forward backport

rpcd/playbooks/rpc-pre-upgrades.yml
```diff
@@ -19,18 +19,63 @@
...
+        path: /etc/logrotate.d/beaver
+        state: absent
+
+    - name: Remove beaver log files
+      file:
+        path: /var/log/beaver
+        state: absent
+
+    - name: Check if beaver is installed
+      shell: pip show -q Beaver
```
Apsu: If you're not using shell specific features, don't use the `shell` module, use `command`. That said, while you could just add ` || true` to the end of the command with the `shell` module, there's a bigger issue. There's lot of types of errors that could occur aside from just beaver not being installed (pip not installed, pip failing, syntax issues in the args, path issues, environment/locale/encoding issues, etc).

I think this can be done way better than just hamfisting "never fails" and "never changes". For one thing, there's an `ignore_errors` key that can be set, which allows the check task to fail, but a slightly better way is to break this into smaller chunks. Example:

command: pip list
register: pip_packages

pip:
  name: beaver
  state: absent
when: pip_packages.stdout.find('Beaver') != -1

But! Since your goal is to uninstall the package anyway using the `pip` module in the next task, none of this even necessary!

Ansible modules are designed to be idempotent. And we can even confirm that the `pip` module works this with when uninstalling: https://github.com/ansible/ansible-modules-core/blob/stable-2.1/packaging/language/pip.py#L413

Meaning you can get rid of the shell command and checks entirely. All you need is this:

pip:
  name: beaver
  state: absent

Just as $deity intended :)

odyssey4me: If you try just using the absent state in Ansible 1.9.4 it fails, that's why this approach was taken @Apsu. This bug was fixed in Ansible 1.9.5, but that version has LXC module issues, as does 1.9.6.

PR #1451, opened 2016-09-29
Remove Holland from outside venv

mancdaz:
In mitaka, holland gets installed inside a venv, which means that we end
up with lingering holland pip packages outside the venv. To remove any
confusion, here we remove those packages as part of the post-upgrade.

Note, the pattern of '_check if holland is installed, then remove the pip
package if it is_' is used only because in ansible 1.9.4, the pip module
is broken and it won't just "do the right thing" if you declare the
package as absent (it will actually fail with an rc of 1 if the package
has already been uninstalled)

Connected https://github.com/rcbops/u-suk-dev/issues/266

### DIFF
--- a/rpcd/playbooks/roles/rpc_post_upgrade/tasks/post-upgrade-holland.yml
+++ b/rpcd/playbooks/roles/rpc_post_upgrade/tasks/post-upgrade-holland.yml
@@ -1,0 +1,30 @@
+---
+# Copyright 2016, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+- name: Check if holland is installed outside venv
+  shell: pip show -q holland
+  register: is_holland_installed
+  failed_when: false
+  changed_when: false
+
+- name: Uninstall holland pip package from outside venv
+  pip:
+    name: "{{ item }}"
+    state: absent
+  when: is_holland_installed.rc == 0
+  with_items:
+    - holland
+    - holland.lib.common
+    - holland.lib.mysql
--- a/rpcd/playbooks/roles/rpc_post_upgrade/tasks/main.yml
+++ b/rpcd/playbooks/roles/rpc_post_upgrade/tasks/main.yml
@@ -19,6 +19,11 @@
   tags:
     - galera
 
+- include: post-upgrade-holland.yml
+  when: inventory_hostname in groups['galera_all']
+  tags:
+    - galera
+
 # Run rabbitmq post upgrade tasks
 - include: post-upgrade-rabbitmq.yml
   when: inventory_hostname in groups['rabbitmq']
--- a/rpcd/playbooks/roles/rpc_post_upgrade/tasks/main.yml
+++ b/rpcd/playbooks/roles/rpc_post_upgrade/tasks/main.yml
@@ -19,7 +19,7 @@
   tags:
     - galera
 
-- include: post-upgrade-holland.yml
+- include: remove-holland.yml
   when: inventory_hostname in groups['galera_all']
   tags:
     - galera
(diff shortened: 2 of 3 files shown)


rackernexus:
HULK [PREVIEW](https://staging.developer.rackspace.com/staging.horse/build-7ee3f254ee/docs/private-cloud/release-notes/)

mancdaz:
recheck_ceph

```
20:45:55 msg: Zero entities with the label jrpcaioiad-a4d exist. Entities should be created as part of the hardware provisioning process, if missing, please consult the internal documentation for associating one with the device.
```

git-harry:
This commit seems to be changing the current way that rpc_post_upgrade behaves. While the role is a collection of tasks operating on a number of different subsets of hosts, as far as I understand it the role is only used to collect data on the current state of the deployment. This commit adds a post-upgrade cleanup task. I'm not sure if that makes a difference in practice but does suggest this role is ripe for significant rework for the next major release.

mancdaz:
recheck_ceph

Assuming this is a 'slow api' type error and not a real problem with the heat api check:

```
10:24:58 The following checks have config files but are not registered with the maas api: ['lb_api_check_heat_cloudwatch--jrpcaioiad-977', 'lb_api_check_keystone--jrpcaioiad-977'] 
10:24:58 The following alarms are present in config files but are not registered with the maas api: ['lb_api_alarm_heat_cloudwatch', 'lb_api_alarm_keystone']
```

mattt416:
@git-harry I think you raise a good point.  The beaver stuff was put into the pre-upgrade role, so I think if we're going to shove these sorts of tasks into these roles we should at least centralise them so that when we get to upgrading to newton we can just yank anything not applicable.  Doing that when it's all scattered will be harder to do and we'll probably end up running things that aren't necessary.

alextricity25:
Testing this now.
I realize now that these tasks files don't have the best names. ``remove-holland.yml`` makes more sense than ``post-upgrade-holland.yml``, IMO. It doesn't need to be changed right now. I think the changes look fine the way they are.

alextricity25:
Before running the playbook:
```
ansible galera_all -m shell -a "pip freeze | grep holland"
cantu-testup-ceph18_galera_container-65073c04 | success | rc=0 >>
holland==1.0.10
holland.backup.xtrabackup===1.0.10dev-r0
holland.lib.common===1.0.10dev-r0
holland.lib.mysql===1.0.10dev-r0Warning: cannot find svn location for holland.backup.xtrabackup===1.0.10dev-r0
Warning: cannot find svn location for holland.lib.mysql===1.0.10dev-r0
Warning: cannot find svn location for holland.lib.common===1.0.10dev-r0

cantu-testup-ceph18_galera_container-4e45d7ea | success | rc=0 >>
holland==1.0.10
holland.backup.xtrabackup===1.0.10dev-r0
holland.lib.common===1.0.10dev-r0
holland.lib.mysql===1.0.10dev-r0Warning: cannot find svn location for holland.backup.xtrabackup===1.0.10dev-r0
Warning: cannot find svn location for holland.lib.mysql===1.0.10dev-r0
Warning: cannot find svn location for holland.lib.common===1.0.10dev-r0

cantu-testup-ceph18_galera_container-99a4f544 | success | rc=0 >>
holland==1.0.10
holland.backup.xtrabackup===1.0.10dev-r0
holland.lib.common===1.0.10dev-r0
holland.lib.mysql===1.0.10dev-r0Warning: cannot find svn location for holland.backup.xtrabackup===1.0.10dev-r0
Warning: cannot find svn location for holland.lib.mysql===1.0.10dev-r0
Warning: cannot find svn location for holland.lib.common===1.0.10dev-r0
```
After running it:
```
root@cantu-testup-ceph18:/opt/rpc-openstack/rpcd/playbooks# ansible galera_all -m shell -a "pip freeze | grep holland"
cantu-testup-ceph18_galera_container-99a4f544 | success | rc=0 >>
holland.backup.xtrabackup===1.0.10dev-r0Warning: cannot find svn location for holland.backup.xtrabackup===1.0.10dev-r0

cantu-testup-ceph18_galera_container-4e45d7ea | success | rc=0 >>
holland.backup.xtrabackup===1.0.10dev-r0Warning: cannot find svn location for holland.backup.xtrabackup===1.0.10dev-r0

cantu-testup-ceph18_galera_container-65073c04 | success | rc=0 >>
holland.backup.xtrabackup===1.0.10dev-r0Warning: cannot find svn location for holland.backup.xtrabackup===1.0.10dev-r0
```
Maybe we should also delete ``holland.backup.xtrabackup``?

rpcd/playbooks/roles/rpc_post_upgrade/tasks/post-upgrade-holland.yml
```diff
@@ -0,0 +1,30 @@
...
+
+- name: Uninstall holland pip package from outside venv
+  pip:
+    name: "{{ item }}"
+    state: absent
+  when: is_holland_installed.rc == 0
+  with_items:
+    - holland
+    - holland.lib.common
+    - holland.lib.mysql
```
alextricity25: ``holland.backup.xtrabackup`` will also need to be deleted.

evrardjp:
Seems the concerns are resolved. I'm trying this too.

evrardjp:
👍

PR #1459, opened 2016-10-03
This is a test

git-harry:

PR #1460, opened 2016-10-03
Do not install binary pycparser packages

sigmavirus24:
pycparser's 2.14 wheel package on PyPI is broken and has been causing
breakage with cryptography and OpenStack in general. For rpc-openstack
in particular, ansible relies on cryptography and that dependency adds a
dependency on pycparser which has been breaking our gating.

Connects rcbops/u-suk-dev#474

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -5,6 +5,8 @@ pep8==1.5.7
 pyflakes==0.8.1
 mccabe==0.2.1 # capped for flake8
 bashate>=0.2 # Apache-2.0
+# Fix for https://waffle.io/rcbops/u-suk-dev/cards/57f258cd31f3e3f8009eb355
+--no-binary pycparser
 
 
 # this is required for the docs build jobs


rackernexus:
If you want to double-check your formatting, check [the preview](https://staging.developer.rackspace.com/staging.horse/build-e43ad5ef81/docs/private-cloud/release-notes/).

git-harry:
@sigmavirus24 only liberty-12.2 is affected, is the intention for this to be here temporarily? What is the effect of this flag?

sigmavirus24:
@git-harry in order:

- Have pull requests to branches other than liberty-12.2 passed gating? I didn't check, I expected from the chatter on openstack-dev that this wouldn't just affect liberty-12.2. 

- My intent was to have this here until pycparser releases a version without a broken wheel.

- The flag avoids installing the wheel that was uploaded for pycparser. The source distribution of pycparser is not broken, only the wheel.

git-harry:
@sigmavirus24 I tested [master](https://github.com/rcbops/rpc-openstack/pull/1459) and it was working. When I was looking earlier it appeared the failure was due to the [install of Ansible](https://github.com/rcbops/rpc-openstack/blob/liberty-12.2/.travis.yml#L5) which is no longer done in the Mitaka series.

If we only need it temporarily I would say it's only required in liberty-12.2 but if there is benefit to always doing it then we should put it in all branches.

sigmavirus24:
Sounds good. I'll just submit a similar change to the liberty branch then.

PR #1461, opened 2016-10-03
[Backport] Remove Holland from outside venv (#1451)

alextricity25:
* Remove Holland from outside venv

In mitaka, holland gets installed inside a venv, which means that we end
up with lingering holland pip packages outside the venv. To remove any
confusion, here we remove those packages as part of the post-upgrade.

Note, the pattern of 'check if holland is installed, then remove the pip
package if it is' is used only because in ansible 1.9.4, the pip module
is broken and it won't just "do the right thing" if you declare the
package as absent (it will actually fail with an rc of 1 if the package
has already been uninstalled)

Connected https://waffle.io/rcbops/u-suk-dev/cards/579f4f206d2c9e0a01448c6e

* Address comments in review

- change post-upgrade-holland.yml to remove-holland.yml
- add holland.backup.xtrabackup to list of pip packages to remove

(cherry picked from commit 612ff25882c24a4b17e0bcd75180593322914f68)

Connected rcbops/u-suk-dev#266

### DIFF
--- a/rpcd/playbooks/roles/rpc_post_upgrade/tasks/main.yml
+++ b/rpcd/playbooks/roles/rpc_post_upgrade/tasks/main.yml
@@ -19,6 +19,11 @@
   tags:
     - galera
 
+- include: remove-holland.yml
+  when: inventory_hostname in groups['galera_all']
+  tags:
+    - galera
+
 # Run rabbitmq post upgrade tasks
 - include: post-upgrade-rabbitmq.yml
   when: inventory_hostname in groups['rabbitmq']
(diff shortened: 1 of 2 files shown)


alextricity25:
rechek_upgrade

alextricity25:
recheck_upgrade

alextricity25:
recheck_upgrade

evrardjp:
recheck_upgrade

failed on neutron metadata service restart (this bug should be fixed by the next sha bump)

alextricity25:
Another failure due to MaaS alarms in the critical status after an upgrade.

alextricity25:
I'm not exactly sure what is causing the MaaS alarms to stay in CRITICAL status on the upgrade job, but I'm currently looking into it.

alextricity25:
recheck_upgrade

mattt416:
Straight-forward backport, LGTM

PR #1462, opened 2016-10-03
Do not install binary pycparser packages

sigmavirus24:
pycparser's 2.14 wheel package on PyPI is broken and has been causing
breakage with cryptography and OpenStack in general. For rpc-openstack
in particular, ansible relies on cryptography and that dependency adds a
dependency on pycparser which has been breaking our gating.

Connects rcbops/u-suk-dev#474

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,12 @@
 language: python
 
 before_script:
-- pip install tox
+- pip install -U tox pip
+# NOTE(sigmavirus24): pycparser 2.14's wheel on PyPI is broken. Using
+# --no-binary below ignores the wheel, downloads the source distribution and
+# avoids the issue when installing cryptography for ansible.
+# This is a temporary workaround and should be removed in due time.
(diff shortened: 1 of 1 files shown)


git-harry:
@sigmavirus24 this has failed with

```
[0K$ pip2 install --no-binary pycparser

Usage:   
  pip install [options] <requirement specifier> [package-index-options] ...
  pip install [options] -r <requirements file> [package-index-options] ...
  pip install [options] [-e] <vcs project url> ...
  pip install [options] [-e] <local project path> ...
  pip install [options] <archive url/path> ...

no such option: --no-binary

travis_time:end:06d86627:start=1475503526656143625,finish=1475503526857033123,duration=200889498
[0K
[31;1mThe command "pip2 install --no-binary pycparser" failed and exited with 2 during .[0m
```

Apsu:
Looks reasonable to me. 👍

git-harry:
The upstream issue has been resolved, unless there is another reason to make this change I think we can abandon this issue.

PR #1464, opened 2016-10-04
Test upstream review 381586

git-harry:
https://review.openstack.org/#/c/381586/

This is only for testing upgrades and should not be merged.

Connected https://github.com/rcbops/u-suk-dev/issues/476


git-harry:
Upgrade test passed.

PR #1471, opened 2016-10-07
Set correct user in rpc-pre-upgrades.yml

mattt416:
We're seeing some inconsistent $HOME references when the pip_lock_down
role runs, and suspect it may have something to do with this playbook
not using the correct user.

Connects rcbops/u-suk-dev#488

### DIFF
--- a/rpcd/playbooks/rpc-pre-upgrades.yml
+++ b/rpcd/playbooks/rpc-pre-upgrades.yml
@@ -21,6 +21,7 @@
 
 - name: remove beaver
   hosts: all
+  user: root
   tags:
     - rpc-pre-upgrades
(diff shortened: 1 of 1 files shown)


git-harry:
recheck_ceph tempest failure, Details: (ListImageFiltersTestJSON:setUpClass) Image ecbd54ad-2263-485f-9ec2-9be378616b12 failed to reach ACTIVE state(current state SAVING) within the required time (300 s)

alextricity25:
:+1:

git-harry:
👍 this may or may not fix the issue, we'll need more test runs to have confidence, but it's a reasonable change either way.

Issue #742, opened 2016-01-14
Poor default value for network_checks_list inside rpc_maas role


BjoernT:
Apparently we defined the interfaces eth0 to eth3 to be enabled for maas network checks  https://github.com/rcbops/rpc-openstack/blob/kilo/rpcd/playbooks/roles/rpc_maas/defaults/main.yml#L324-L328 but in reality bare metal machines are using now udev names for interfaces like em1. 
All new checks will practically fail due to missing network interfaces and we need to either empty this parameter or add a ansible facts discovery around the interfaces or setup limits for the bonding interfaces only
----------------------------------------------------------------------------------------------------------------------------
###### For each pull request, update the branch line to reference it.
- [x] master (#1401)
- [x] mitaka-13.1 (#1473)
- [x] mitaka-13.0 (#1474)
- [x] liberty-12.2 (#1453)
- [x] kilo (#1475)

##### Upgrade impact. Consider the impact on moving between major, minor and patch releases.
- [x] This change does not impact upgrades
- [ ] This change addresses a potential upgrade impact
- [ ] The upgrade impact is being tracked by another issue.

The issue tracking the upgrade impact is:

##### Testing impact
- [ ] An issue/pull-request has been create or exists to track the testing gap, and is referenced below
- [ ] This change includes additional testing
- [ ] This change is tested by the existing setup

Additional testing links:

##### Release note
- [ ] This change includes a release note.
- [x] This change does not require a release note.

##### Does your change require documentation?
- [ ] Yes, and I submitted an issue to [Docs issues](https://github.com/rackerlabs/docs-rpc/issues "Docs issues")
- [ ] Yes, and I contributed a documentation patch to [docs-rpc](https://github.com/rackerlabs/docs-rpc "docs-rpc")
- [x] No documentation required for this change

Additional documentation links (please add links if you have ticked yes):

BjoernT:
@byronmccollum

hughsaunders:
@BjoernT The default was changed to eth0 only in #719 this is in recognition that network_checks_list is not a suitable default for physical deploys or virtual tests. So network_checks_list will need to be overridden, probably in user_extras_variables for all physical deploys.

nrb:
So a completely empty value would be valid? As long as support personnel know to update that, I'm fine.

From what @hughsaunders is saying, support will need to define this, at least in the short term (since 11.1.1 is planned for tomorrow).

hughsaunders:
[...]

PR #1473, opened 2016-10-07
Set proper default NIC of network_check_list

sigmavirus24:
Default NIC of network_check_list is eth0 which is barely used and
always caused maas checks to fail. Now We only need to worry the
 state of bond interface. This change covers most of practical
  customer cases.

Connects #742
(cherry picked from commit e0f96706be580450af87bb816d2077bffc75d820)

### DIFF
--- a/rpcd/playbooks/roles/rpc_maas/defaults/main.yml
+++ b/rpcd/playbooks/roles/rpc_maas/defaults/main.yml
@@ -358,7 +358,7 @@ ceph_checks_list:
   - { name: "ceph_cluster_stats", group: "mons" }
 
 network_checks_list:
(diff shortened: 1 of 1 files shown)


tomc603:
Same backport as the others. 👍

alextricity25:
:+1: Pending checks

sigmavirus24:
Failed to restart Apache in the horizon extensions role

```
19:14:58 msg: (98)Address already in use: AH00072: make_sock: could not bind to address [::]:80
19:14:58 (98)Address already in use: AH00072: make_sock: could not bind to address 0.0.0.0:80
```

recheck

sigmavirus24:
recheck_upgrade

PR #1474, opened 2016-10-07
Set proper default NIC of network_check_list

sigmavirus24:
Default NIC of network_check_list is eth0 which is barely used and
always caused maas checks to fail. Now We only need to worry the
 state of bond interface. This change covers most of practical
  customer cases.

Connects #742
(cherry picked from commit e0f96706be580450af87bb816d2077bffc75d820)


tomc603:
Looks okay 👍

alextricity25:
:+1:

PR #1477, opened 2016-10-07
DO NOT MERGE TESTING JENKINS

alextricity25:
Jenkins does not set the HOME variable to /root, which causes pip configurations to be dropped in /jenkins on some containers and in turn, causing errors in gate. This PR is just a test to see what exactly will fix that.


alextricity25:
Closing in favor of https://github.com/rcbops/jenkins-rpc/pull/663

Issue #1478, opened 2016-10-10
Add a release note for ceph l>m upgrade changes


evrardjp:
As discussed in rcbops/rpc-openstack#1463, the ceph l>m upgrade needs a reno.

PR #1480, opened 2016-10-10
Introduce ceph upgrade release notes

evrardjp:
Ceph upgrades requires changes in the containers, and will change
the bind mounts for the ceph monitors. This should be the record
of it for the deployers.

Connected rcbops/rpc-openstack#1478


git-harry:
👍

Issue #1262, opened 2016-07-25
horizon_maas plugin "“horizon_local_status”"  throws errors when using keystone domains


brainerd123:
Hello,

When enabling Keystone Domains, the following MaaS Horizon check stops working.

"horizon_local_status"

:set consecutiveCount=3
if (metric["horizon_local_status"] != 1) {
    return new AlarmStatus(CRITICAL, "Horizon unavailable");
}

The check can no longer login to Horizon.

###### Please provide the output to 'git describe --all' for  all deployments where you have confirmed it occurs.git

root@726402-infra01:/opt/openstack-ansible# git describe --all
heads/stable/mitaka

##### What type of issue is this? Please tick the appropriate box.
- [x] bug - existing functionality that does not work for you as documented
- [ ] feature - new functionality that you would like to see added

##### What is the expected behavior?

The check should account for Domains in Horizon be able to login, which would clear the MaaS check.
##### What is actually happening?

When domains are enabled it can longer login to horizon so the check fails

##### What were you attempting to do when you encountered this error (give as much detail as possible)? 

Nothing, I just noticed the MaaS check failed when domains were enabled. 

##### How can we reproduce this issue?

Enable Domains in a Liberty or newer environment and test the MaaS check afterwards. 

##### How can we verify your issue is addressed?

Update this thread or provide me a new MaaS check to test in my environment. 

##### If you are using or know of a workaround please describe the necessary steps?

No other workarounds other than disabling the check

##### What is your evaluation on the importance of this issue, and why?

Medium importance.

----------------------------------------------------------------------------------------------------------------------------
###### For each pull request, update the branch line to reference it.
- [x] master (#1476)
- [x] mitaka-13.1 (#1488)
- [x] mitaka-13.0 (#1489)
- [x] liberty-12.2 (#1490)
- [ ] kilo

##### Upgrade impact. Consider the impact on moving between major, minor and patch releases.
- [x] This change does not impact upgrades
- [ ] This change addresses a potential upgrade impact
- [ ] The upgrade impact is being tracked by another issue.

The issue tracking the upgrade impact is:

##### Testing impact
- [ ] An issue/pull-request has been create or exists to track the testing gap, and is referenced below
- [x] This change includes additional testing
- [ ] This change is tested by the existing setup

Additional testing links:

screenshot of Horizon with Domains:

https://www.dropbox.com/s/nhsqtjwinj4tg3v/Domains.png?dl=0 

##### Release note
[...]

PR #1489, opened 2016-10-12
Search for and provide domain in horizon check

Frank-ZhangXin:
In multi-domain deployments of RPC OpenStack, Horizon will require the
user to enter a domain name. As such, we need to provide that
information when posting our login request to Horizon in our MaaS check
plugin.

To detect the multi-domain case, we search for an input field named
"domain" and only then do we provide it. To make the check slightly more
efficient we also refactor it to just parse the HTML once instead of 2
or 3 times.

Connects #1262
(cherry picked from commit b87c025c4befa6242daf70d79bf793f6fc76dcf7)


Frank-ZhangXin:
recheck_ceph
recheck_swift

tomc603:
Backport looks good 👍

PR #1496, opened 2016-10-13
Cap tox at 2.4.0 to avoid env substitution bug

git-harry:
Tox 2.4.0 was released this morning and broke the ability to use
environment substitution. To avoid this and keep our continuous
integration happy while we complete our work, we prevent travis from
installing any version of tox 2.4.0 or later.

Related-to tox-dev/tox#380
Connects rcbops/u-suk-dev#512

(cherry picked from commit 3e18ce8b06ab64cf2d509abefbef5e33f6a0e39f)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: python
 
 before_script:
-- pip install tox
+- pip install 'tox<2.4.0,>=2.0'
 
 script: tox
 


sigmavirus24:
👎 there's a new version of tox that fixes this appropriately. I'd like to see what happens with #1497 before we merge this so that we do not have to cap tox long term.

evrardjp:
👎

evrardjp:
#1499 out.

PR #1507, opened 2016-10-17
Remove pattern from Apache2 restart handler

mattt416:
We are seeing periodic failures where the Apache2 restart handler is
triggered but is unable to restart Apache2 successfully as it's already
running.

Reproducing this infrequent failure is difficult, but I suspect the
failure is due to the handler's service being defined with the pattern
parameter present, which does a process listing to determine the status
of the service.  The Apache2 service reports status correctly via the
init script itself, so using pattern is unnecessary.  Additionally, the
upstream os_horizon role's Apache2 restart handler has no pattern
present.

This commit simply removes pattern, and updates how the handler itself
is defined.

Connects https://github.com/rcbops/u-suk-dev/issues/518

### DIFF
--- a/rpcd/playbooks/roles/horizon_extensions/handlers/main.yml
+++ b/rpcd/playbooks/roles/horizon_extensions/handlers/main.yml
@@ -1,2 +1,4 @@
 - name: Restart apache2
-  service: name=apache2 state=restarted pattern=apache2
+  service:
+    name: apache2
+    state: restarted


rackernexus:
If you want to double-check your formatting, check [the preview](https://staging.developer.rackspace.com/staging.horse/build-c961a79875/docs/private-cloud/release-notes/).

alextricity25:
recheck_ceph

mattt416:
recheck_ceph

rackernexus:
Your content preview [is now ready](https://staging.developer.rackspace.com/staging.horse/build-eb192c26f7/docs/private-cloud/release-notes/). :bow:

mattt416:
recheck_upgrade

mattt416:
recheck_upgrade

git-harry:
recheck_all

git-harry:
👍 

Given that this only fails on upgrades and that the os_horizon handler doesn't appear to fail, it is a reasonable guess that the Ansible module's logic is deficient when a pattern is used.

After rechecking the numbers it seems between 6% and 16% of upgrades fail due to this issue. I would have thought that was sufficiently frequent to test for.

Based on the fact that the fix only changes the handler to the more common format, I am comfortable with merging it and using that as a way to test the change. However, once this has merged we would want to monitor the build summary to validate the success of this change.

alextricity25:
:+1:

PR #1509, opened 2016-10-17
Blacklist tox 2.4.0 which breaks our tests

cloader89:
Version 2.4.0 of tox was released yesterday and broke environment
substitution (which we use in our tox.ini). 2.4.1 was released today
fixing this problem for us. We just need to blacklist the broken version
to ensure it doesn't break us again.

Connects rcbops/u-suk-dev#512

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: python
 
 before_script:
-- pip install tox
 - git fetch origin +refs/heads/*:refs/remotes/origin/*
+- pip install 'tox!=2.4.0,>=2.3'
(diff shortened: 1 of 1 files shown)


git-harry:
👎 The commit message does not reference the original commmit.

To cherry-pick a commit into your current branch you should do:

```
git cherry-pick -x commit
```

Adding '-x' appends a line to the commit message referencing the original commit.

claco:
We should update CONTRIBUTING.md to be verbose about our cherry-pick -x expectation.

Issue #1070, opened 2016-05-12
SSH errors due to the number of forks


git-harry:
##### What version (commit) do you encounter this issue on?

all
##### What type of issue is this (bug, feature, documentation, regression)?

bug
##### What is the expected behavior?

Ansible has the parameter forks [1] that controls the number of parallel processes. For large deployments, increasing the number of forks has the potential to significantly reduce the time taken to complete the deployment.

[1] http://docs.ansible.com/ansible/intro_configuration.html#forks
##### What is actually happening?

The default number of forks used by RPCO is the number of processors on the deployment host. This is generally being overridden by support to a much higher number e.g. 75. This can cause a problem when using delegate_to because the host to which the task is delegated cannot support that number of SSH sessions. The number of sessions allowed for the SSH daemon [2] is set by MaxSessions which defaults to 10. The investigation of this issue has previously been done upstream [3].

On customer deployments the number of processors is likely to exceed 10 which means that by default there is the potential for SSH connection errors and overriding the value to 75 is obviously a potential problem.

[2] http://docs.openstack.org/developer/openstack-ansible/install-guide/app-tips.html#ansible-forks
[3] https://bugs.launchpad.net/openstack-ansible/+bug/1479812
##### How can we reproduce this issue?

Set a high number of forks and run a task that delegates more than MaxSessions connections to a single host.
##### How can we verify your issue is addressed?

This will be addressed when there is no longer confusion about the use of forks. This could be addressed by adjusting the number of MaxSessions, forks or just with documentation.
##### If this is a large issue, what can be done immediately to alleviate the issue while working on a longer-term solution?

It is possible to adjust the number of forks set and so there is no inherent reason that a deployment should be affected by this issue.
##### What is your evaluation on the importance of this issue, and why?

This issue reoccurs but does not prevent deployments happening and so is probably a prio-2.
###### Remove branches (lines) you do not wish to target.
- [ ] Master
- [ ] liberty-12.2
- [ ] liberty-12.1
- [ ] kilo

odyssey4me:
FYI I have patched OSA master to be rid of the ``FORKS`` environment variable and to provide more guidance in the documentation: https://review.openstack.org/319213

prometheanfire:
so, question... shouldn't this be part of the base image?

odyssey4me:
@prometheanfire By 'base image' what do you mean? When it comes to CDC's there is no 'base image' for the hosts.

prometheanfire:
[...]

PR #1516, opened 2016-10-18
Set Host sshd_config MaxSessions option to 100

prometheanfire:
The sshd role imported by OSA was used for this.  I did create a
new role so that we could run this early, as this is host setup.

Fixes: #1070
Connected #1070

### DIFF
--- a/rpcd/etc/openstack_deploy/user_rpco_variables_defaults.yml
+++ b/rpcd/etc/openstack_deploy/user_rpco_variables_defaults.yml
@@ -138,3 +138,6 @@ secure_cluster: true
 # List of secure flags to set on for a pool (options for the list are nodelete, nopgchange, nosizechange - prevents deletion, pg from changing and size from changing respectively).
 secure_cluster_flags:
   - nodelete
+
+# Host options
+sshd_MaxSessions: 100
--- a/rpcd/playbooks/rpc-setup-infrastructure.yml
+++ b/rpcd/playbooks/rpc-setup-infrastructure.yml
@@ -1,0 +1,22 @@
+---
+# Copyright 2016, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+- name: Setup sshd on hosts
+  hosts: "{{ host_group|default('hosts') }}"
+  user: root
+  tags:
+    - rpc-setup-infrastructure
+  roles:
+    - { role: "sshd" }
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -172,6 +172,10 @@ if [[ "${DEPLOY_OA}" == "yes" ]]; then
   # Issue tracking upstream fix: https://github.com/rcbops/rpc-openstack/issues/1028
   ansible repo_all -m file -a 'name=/root/.pip state=absent' 2>/dev/null ||:
 
+  pushd ${RPCD_DIR}/playbooks/
+    run_ansible rpc-setup-infrastructure.yml
+  popd
+
   cd ${OA_DIR}/playbooks/
 
   # setup the haproxy load balancer


rackernexus:
:eyes: [Preview](https://staging.developer.rackspace.com/staging.horse/build-33131df31d/docs/private-cloud/release-notes/)

rackernexus:
:eyes: [Preview](https://staging.developer.rackspace.com/staging.horse/build-f38781233e/docs/private-cloud/release-notes/)

odyssey4me:
:-1: I don't really understand why a new role is required when there is already one that can do the job for you which is already included in the ansible-role-requirements. https://github.com/willshersystems/ansible-sshd

That role is well tested and using it will allow a lot more configurability for support than just the one thing.

rackernexus:
HULK [PREVIEW](https://staging.developer.rackspace.com/staging.horse/build-d1bc66f59b/docs/private-cloud/release-notes/)

prometheanfire:
@odyssey4me ok, updated, but keep in mind this now clobbers osa-security stuff every time it runs.

prometheanfire:
also, link is failing due to that role...

odyssey4me:
@prometheanfire I was simply suggesting an alternative. Note that every single setting you could possibly want in sshd_config is configurable if you give it the vars - so if you want to preserve something, give the role the right vars!

Otherwise, instead of using the heavy-handed approach you first took - just add a lineinfile task into the support role or a playbook. A whole role to edit one line seems a vast overkill.

prometheanfire:
Adding the role is more so when requests that will inevitably come down the line for more host customization do come, we are prepared.  Also, I can't do a simple line in file because of the way sshd_config is formatted (optional sections).  I do think my original way is better and still have the patch if people agree.

cloader89:
recheck_all

cloader89:
recheck_all

rackernexus:
:mag_right: [Preview](https://staging.developer.rackspace.com/staging.horse/build-c76f50e270/docs/private-cloud/release-notes/)

cloader89:
recheck_upgrade

mattt416:
👎 From the parent issue it feels like we were going to remove overriding of ansible forks (which happens in a few places in rpc-openstack) and making relevant doc changes.  I didn't get the impression that we were proceeding with this new role.

odyssey4me:
:-1: +461 LOC just to change a setting I truly believe should be implemented by a kick system and not by us. I am strongly against merging this.

BjoernT:
MaxSessions is important to get the roles working correctly, especially the swift one which seem to fail quite often, if not always, once you increase the forks. From that standpoint I'm in favor of just setting MaxSession inside the SSHd config, nothing else. The SSHD ansible role would maintain the whole config which we don't have control over since customer run their own config management. 
And that conflicts using the SSHD role. So personally I would prefer to add the one liner to fix most out of the box deployments since we only have KIR (kick in the rack) only available on the minority of hosts. Most hosts are managed in customer environments now.
Additionally documentation needs to be provided for customer completely overwriting the config and  we are all aware of it.

Perhaps another way may be to start SSHD with the -o MaxSessions=100 option etc but I'm generally in favor of this PR since if fixes a fundamental problem and I'm personally not seeing this different from what we already do with all the sysctl changes or installing packages.
If the RPC-O needs a operating system condition met to install the product we either check on it or make it happen, i.e. installing packages, maintaining kernel settings etc. etc.

b3rnard0:
@git-harry @mattt416 @odyssey4me Can you please look into @BjoernT response so we can get additional feedback.

odyssey4me:
As previously suggested, and reiterated. The patch can be vastly simplified to a simple lineinfile task in the support role. This will cover what's required and build far less tech debt.

prometheanfire:
@odyssey4me lineinfile doesn't work on formatted sshd config files @major can you explain why, since you have more experience with it.

@odyssey4me also, putting it in the support role will not prevent this issue from occuring during initial deployments, as the support role is one of the last to run, that is why I created an 'early run' version of it.

b3rnard0:
@mancdaz Can you please make sure we get feedback from the UK folks on Monday? Thanks.

evrardjp:
Agreed with @odyssey4me, we can simplify to either:
- own the complete file, because it's an inherent part of our process
- reduce to a few tasks.

Here would be the tasks:
```
  - name: Find for Match in ssh server config
    command: egrep -n '^Match' /etc/ssh/sshd_config
    register: match_in_sshd_config
    failed_when: false
    changed_when: false

  #Set MaxSessions to 100 to allow for more multiplexing
  - name: Ensure MaxSessions is defined to our standards (Match clause)
    lineinfile:
      line: 'MaxSessions 100'
      state: present
      dest: /etc/ssh/sshd_config
      regexp: '^(#)?MaxSessions'
      insertbefore: "^Match"
      validate: '/usr/sbin/sshd -T -f %s'
    when: match_in_sshd_config.rc == 0

  - name: Ensure MaxSessions is defined to our standards (no Match clause)
    lineinfile:
      line: "MaxSessions 100"
      state: present
      dest: /etc/ssh/sshd_config
      regexp: '^(#)?MaxSessions'
      insertafter: "EOF"
      validate: '/usr/sbin/sshd -T -f %s'
    when: match_in_sshd_config.rc == 1

```

prometheanfire:
@evrardjp the reason why `before` does not work with `lineinfile`

https://github.com/rcbops/rpc-openstack/issues/1070#issuecomment-258866368

mattt416:
@prometheanfire OK, but this is a solved problem right?  As far as I understand it, openstack-ansible-security is already handling sshd_config so can we not extract the bits that creates the marker etc. so that multiple playbooks/roles can do it in a consistent way?

prometheanfire:
@mattt416 yes, it's solved in that I am using the same method used in osa-security.  In order to solve the bug though (which is about deployment issues) this needs to run first.  If we want only one marker that's fine, but it'd have to go upstream, I'm not sure @odyssey4me or @andymcc would want that in osa itself (also, given how early this needs to run I'm not sure where to put it).

mattt416:
@prometheanfire Generally speaking this looks OK to me (haven't tested, still need to do).  However, I do still think there may be a need for a separate role/plugin that either lives in OpenStack or elsewhere that can be used by both rpc-openstack and openstack-ansible-security.  If OpenStack, it'll take some time for rpc-openstack to be able to consume that so perhaps outside of OSA would work best in this situation?  Would be good to get everyone else's thoughts here but I don't see any massive issues merging this and then removing it when we have something shared to use.

mattt416:
👎 Until deploy.sh is updated so we can exercise this in the gate.

prometheanfire:
ya, readded that, dunno how that was missed, maybe it was dropped because of a conflict when @cloader89 resubmitted it.

Issue #1512, opened 2016-10-17
Failing to install elasticsearch reindexing plugin


jacobwagner:
###### Please provide the output to 'git describe --all' for  all deployments where you have confirmed it occurs.
r13.1.0rc2
##### What type of issue is this? Please tick the appropriate box.
- [x] bug - existing functionality that does not work for you as documented
- [ ] feature - new functionality that you would like to see added

##### What is the expected behavior?
ElasticSearch installs all necessary plugins

##### What is actually happening?
Failing to install reindexing plugin, unknown command being passed

##### What were you attempting to do when you encountered this error (give as much detail as possible)? 
Upgrade from liberty to mitaka

##### How can we reproduce this issue?
see above

##### How can we verify your issue is addressed?
ELK play finishes and installs / configures elk correctly

##### What is your evaluation on the importance of this issue, and why?
important, as rpc cannont finish install of ELK stack

```
TASK: [elasticsearch | Drop log_test1 template] *******************************
changed: [node101_elasticsearch_container-c989f372]

TASK: [elasticsearch | Place legacy liberty template script] ******************
changed: [node101_elasticsearch_container-c989f372 -> node101_elasticsearch_container-c989f372]

TASK: [elasticsearch | Install legacy liberty template] ***********************
changed: [node101_elasticsearch_container-c989f372 -> node101_elasticsearch_container-c989f372]

TASK: [elasticsearch | Remove legacy liberty template script] *****************
changed: [node101_elasticsearch_container-c989f372 -> node101_elasticsearch_container-c989f372]

TASK: [elasticsearch | Install Reindexing Plugin] *****************************
failed: [node101_elasticsearch_container-c989f372] => (item=org.codelibs/elasticsearch-reindexing/1.7.0) => {"changed": true, "cmd": ["./plugin", "-install", "org.codelibs/elasticsearch-reindexing/1.7.0"], "delta": "0:00:00.512909", "end": "2016-10-17 21:36:51.800972", "item": "org.codelibs/elasticsearch-reindexing/1.7.0", "rc": 64, "start": "2016-10-17 21:36:51.288063", "warnings": []}
stdout: ERROR: unknown command [-install]. Use [-h] option to list available commands

FATAL: all hosts have already failed -- aborting

PLAY RECAP ********************************************************************
elasticsearch | Upgrade elasticsearch to stable 1.7.x version ---------- 14.97s
elasticsearch | Wait for elasticsearch to restart ----------------------- 6.34s
[...]

PR #1538, opened 2016-10-24
Do some verification before starting upgrade tasks

d34dh0r53:
We need to do some verification before starting upgrade tasks in
order to avoid errors on subsequent runs. First off we install a
pre flight check script that gathers version information and
returns codes based on the currently installed version. A pre-flight
task will execute this script and will set a flag to skip the upgrade
if the latest version is installed and exit if the reindexing version
is installed.

I have also created variables elasticsearch_version and
elasticsearch_reindex_version so that we can hard code the versions
of elasticsearch to be used. We were using hardcoded and assumed
versions in several places.

Thirdly I have added a retries to the curl command in order to
skip a transient error that we have occasionally seen.

Connects #1512

### DIFF
--- a/rpcd/playbooks/roles/elasticsearch/defaults/main.yml
+++ b/rpcd/playbooks/roles/elasticsearch/defaults/main.yml
@@ -23,6 +23,9 @@ elasticsearch_legacy_apt_repos:
 # use -e 'logging_upgrade=true' to upgrade logging components
 logging_upgrade: false
 
+# use -e 'force_upgrade=true' for force the upgrade to happen regardless of the pre-flight checks
+force_upgrade: false
+
 elasticsearch_apt_repo_url: "http://packages.elastic.co/elasticsearch/2.x/debian"
 
 elasticsearch_apt_repos:
@@ -31,8 +34,16 @@ elasticsearch_apt_repos:
 elasticsearch_apt_keys:
   - { url: "http://packages.elastic.co/GPG-KEY-elasticsearch", state: "present" }
 
+elasticsearch_version: 2.4.1
+
+elasticsearch_reindex_version: 1.7.5
+
 elasticsearch_apt_packages:
-  - elasticsearch
+  - elasticsearch={{ elasticsearch_version }}
+  - openjdk-7-jre-headless
+
+elasticsearch_reindex_apt_packages:
+  - elasticsearch={{ elasticsearch_reindex_version }}
   - openjdk-7-jre-headless
 
 elasticsearch_pip_packages:
--- a/rpcd/playbooks/roles/elasticsearch/tasks/elasticsearch_install.yml
+++ b/rpcd/playbooks/roles/elasticsearch/tasks/elasticsearch_install.yml
@@ -16,7 +16,7 @@
 - name: Install apt packages
   apt:
     pkg: "{{ item }}"
-    state: latest
+    state: present
     update_cache: yes
     cache_valid_time: 600
   register: install_apt_packages
--- a/rpcd/playbooks/roles/elasticsearch/tasks/main.yml
+++ b/rpcd/playbooks/roles/elasticsearch/tasks/main.yml
@@ -13,9 +13,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+- include: upgrade_pre_flight.yml
+  when:
+    - logging_upgrade | bool
 - include: upgrade_elasticsearch_pre.yml
   when:
     - logging_upgrade | bool
+    - proceed_with_upgrade | bool
 - include: elasticsearch_pre_install.yml
 - include: elasticsearch_install.yml
 - include: elasticsearch_post_install.yml
--- a/rpcd/playbooks/roles/elasticsearch/tasks/upgrade_elasticsearch_pre.yml
+++ b/rpcd/playbooks/roles/elasticsearch/tasks/upgrade_elasticsearch_pre.yml
@@ -30,7 +30,7 @@
 - name: Upgrade elasticsearch to stable 1.7.x version
   apt:
     pkg: "{{ item }}"
-    state: latest
+    state: present
     update_cache: yes
     cache_valid_time: 600
     dpkg_options: "force-confold"
@@ -38,7 +38,7 @@
   until: install_apt_packages|success
   retries: 5
   delay: 2
-  with_items: elasticsearch_apt_packages
+  with_items: elasticsearch_reindex_apt_packages
   tags:
     - logging-upgrade
     - elasticsearch-upgrade
@@ -109,6 +109,7 @@
 
 - name: Drop log_test1 template
   shell: 'curl -sk -XDELETE {{ internal_lb_vip_address }}:{{ elasticsearch_http_port }}/_template/log_test1'
+  retries: 5
   tags:
     - logging-upgrade
     - elasticsearch-upgrade
--- a/rpcd/playbooks/roles/elasticsearch/tasks/upgrade_pre_flight.yml
+++ b/rpcd/playbooks/roles/elasticsearch/tasks/upgrade_pre_flight.yml
@@ -1,0 +1,92 @@
+---
+# Copyright 2016, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+- name: "Set initial values"
+  set_fact:
+    latest_version: false
+    reindex_version: false
+  tags:
+    - logging-upgrade
+    - elasticsearch-upgrade
+    - elasticsearch-pre-flight
+
+- name: "Get ElasticSearch version"
+  command: "dpkg-query --show --showformat='${Version}' elasticsearch"
+  register: elasticsearch_installed_version
+  tags:
+    - logging-upgrade
+    - elasticsearch-upgrade
+    - elasticsearch-pre-flight
+
+- name: "Check for ElasticSearch version > 2.x"
+  set_fact:
+    proceed_with_upgrade: false
+    latest_version: true
+  when: "{{ elasticsearch_installed_version.stdout | version_compare(elasticsearch_version, '>=', strict=True) }}"
+  tags:
+    - logging-upgrade
+    - elasticsearch-upgrade
+    - elasticsearch-pre-flight
+
+- name: "Check for ElasticSearch version 1.7.x"
+  set_fact:
+    proceed_with_upgrade: false
+    reindex_version: true
+  when:
+    - "{{ elasticsearch_installed_version.stdout | version_compare(elasticsearch_reindex_version, '>=', strict=True) }}"
+    - "{{ elasticsearch_installed_version.stdout | version_compare('1.7.9', '<=', strict=True) }}"
+  tags:
+    - logging-upgrade
+    - elasticsearch-upgrade
+    - elasticsearch-pre-flight
+
+- name: "Check for ElasticSEarch version > 1.4.x"
+  set_fact:
+    proceed_with_upgrade: true
+  when:
+    - "{{ elasticsearch_installed_version.stdout | version_compare('1.4.0', '>=', strict=True) }}"
+    - "{{ elasticsearch_installed_version.stdout | version_compare('1.6.9', '<=', strict=True) }}"
+  tags:
+    - logging-upgrade
+    - elasticsearch-upgrade
+    - elasticsearch-pre-flight
+
+- name: "Check for force_upgrade=true"
+  set_fact:
+    proceed_with_upgrade: true
+  when:
+    - "{{ force_upgrade | bool }}"
+  tags:
+    - logging-upgrade
+    - elasticsearch-upgrade
+    - elasticsearch-pre-flight
+
+- debug:
+    msg: "ElasticSearch is at the latest version, skipping reindexing steps."
+  when: "{{ latest_version | bool }}"
+  tags:
+    - logging-upgrade
+    - elasticsearch-upgrade
+    - elasticsearch-pre-flight
+
+- fail:
+    msg: "ElasticSearch is at the reindexing version, there may be a reindexing in progress.  Aborting installation, use -e 'force_upgrade=true' to continue."
+  when:
+    - "{{ reindex_version | bool }}"
+    - "{{ not force_upgrade | bool }}"
+  tags:
+    - logging-upgrade
+    - elasticsearch-upgrade
+    - elasticsearch-pre-flight


rackernexus:
:mag_right: [Preview](https://staging.developer.rackspace.com/staging.horse/build-d3c352d3bf/docs/private-cloud/release-notes/)

rackernexus:
If you want to double-check your formatting, check [the preview](https://staging.developer.rackspace.com/staging.horse/build-f8b5ed7405/docs/private-cloud/release-notes/).

rpcd/playbooks/roles/elasticsearch/tasks/upgrade_elasticsearch_pre.yml
```diff
@@ -13,6 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+- name: Check current elasticsearch version
+  apt:
```
sigmavirus24: Is this actually valid? What does this do?

rpcd/playbooks/roles/elasticsearch/tasks/upgrade_pfc.yml
```diff
@@ -0,0 +1,48 @@
+---
```
sigmavirus24: I'd rather name the file something better than "upgrade_pfc". Perhaps "pre_upgrade_checks"? "PFC" is kind of ... not very future maintainer friendly

d34dh0r53: good point, I'll fix that

rackernexus:
HULK [PREVIEW](https://staging.developer.rackspace.com/staging.horse/build-5659a3382e/docs/private-cloud/release-notes/)

rpcd/playbooks/roles/elasticsearch/tasks/upgrade_pre_flight.yml
```diff
@@ -0,0 +1,48 @@
...
+    - es-pre-flight
+
+- name: Check for latest version
+  set_fact:
+    logging_upgrade: False
+  when:
+    - pre_flight_status.rc == '1'
+
+- name: Check for reindexing in progress
+  command: /bin/false
```
sigmavirus24: So if reindexing is in progress, we just die here, right?

Also, are the return codes really strings? *gross*

rackernexus:
:mag_right: [Preview](https://staging.developer.rackspace.com/staging.horse/build-ca9bb4c900/docs/private-cloud/release-notes/)

rpcd/playbooks/roles/elasticsearch/tasks/upgrade_pre_flight.yml
```diff
@@ -0,0 +1,36 @@
...
+  delegate_to: "{{ groups['elasticsearch_all'][0] }}"
+  tags:
+    - logging-upgrade
+    - elasticsearch-upgrade
+    - es-pre-flight
+
+- name: Run Pre Flight Checks
+  shell: /opt/upgrade-pre-flight-checks.py
+  delegate_to: "{{ groups['elasticsearch_all'][0] }}"
+  register: pre_flight_status
```
sigmavirus24: Are we using ``pre_flight_status`` anywhere now?

rpcd/playbooks/roles/elasticsearch/tasks/upgrade_pre_flight.yml
d34dh0r53: the return codes should be ints but ansible is really funky with type checking.  I just got rid of that as we can't not run the upgrade plays if the version is already at the latest, there isn't a way that I can see to get that fact back up into the main.yml playbook.

rpcd/playbooks/roles/elasticsearch/tasks/main.yml
```diff
@@ -13,6 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+- include: upgrade_pre_flight.yml
+  when:
+    - logging_upgrade | bool
```
sigmavirus24: Should this task file maybe use ``pre_flight_check`` to determine if ``upgrade_elasticsearch_pre`` should run? Maybe give it multiple when conditions? For example, add:

```
when:
  - pre_flight_check.rc == 0
# etc.
```

rpcd/playbooks/roles/elasticsearch/templates/upgrade-pre-flight-checks.py.j2
```diff
@@ -0,0 +1,34 @@
...
+
+import apt
+import sys
+
+cache = apt.Cache()
+es_pkg = cache['elasticsearch']
+
+es_version = es_pkg.installed.version
+
+if es_pkg.installed.version == '{{ elasticsearch_version }}':
```
sigmavirus24: I'm wondering if using exit codes is the best way to manage this. Can't ansible understand it if you print JSON, something like:

```py
if es_pkg.installed.version == '{{ elasticsearch_version }}':
    pkg_status = {
        'status': 'latest-installed',
        'explanation': 'Latest version of elasticsearch installed already. Skip upgrade.',
   }
elif es_pkg.installed_version == '{{ elasticsearch_reindex_version }}':
    pkg_status = {
        'status': 'reindexing',
        'explanation': '...',
    }
else:
    pkg_status = {
        'status': 'needs-upgrade',
        'explanation': 'Elasticsearch is at version {} and needs to be upgraded'.format(es_pkg.installed.version),
    }

print(json.dumps(pkg_status))
```

That said, I'm not sure how ansible detects/parses that, so you may be better off with this method for now and then experimenting on improving this for 13.2

rpcd/playbooks/roles/elasticsearch/tasks/upgrade_pre_flight.yml
d34dh0r53: :/ no

rackernexus:
:eyes: [Preview](https://staging.developer.rackspace.com/staging.horse/build-3fd9337060/docs/private-cloud/release-notes/)

rpcd/playbooks/roles/elasticsearch/tasks/main.yml
d34dh0r53: Ohh, that may work

rpcd/playbooks/roles/elasticsearch/templates/upgrade-pre-flight-checks.py.j2
sigmavirus24: Right, like I'm saying 13.2 is when we plan on getting rid of technical debt, so this might be something to tack on for that.

rpcd/playbooks/roles/elasticsearch/tasks/main.yml
d34dh0r53: Hah, that works, updating now.

rpcd/playbooks/roles/elasticsearch/tasks/upgrade_pre_flight.yml
```diff
@@ -0,0 +1,35 @@
...
+    group: root
+    mode: 0755
+  delegate_to: "{{ groups['elasticsearch_all'][0] }}"
+  tags:
+    - logging-upgrade
+    - elasticsearch-upgrade
+    - es-pre-flight
+
+- name: Run Pre Flight Checks
+  shell: /opt/upgrade-pre-flight-checks.py
```
sigmavirus24: Anything non-zero here would cause this to fail, right? That's the idea?

d34dh0r53: Right, so the pre-flight-check returns 1 if ES is already at 2.x and 2 if it's in the 1.7.x temporary state.  Ideally we would just skip all of the upgrade stuff if the rc is 1 and bail if the rc is 2.  If the rc is 0 we continue with the upgrade.

rackernexus:
Your content preview [is now ready](https://staging.developer.rackspace.com/staging.horse/build-95bfa5ee3e/docs/private-cloud/release-notes/). :bow:

rpcd/playbooks/roles/elasticsearch/tasks/upgrade_pre_flight.yml
sigmavirus24: This old comment showed back up because you readded `register: pre_flight_status`.

To future reviewers, this is no longer a valid conversation, but y'know, GitHub thinks it is.

sigmavirus24:
I'm 👍 just based on 👀 only. I'd like to 🚗 it around and see if I can make it 💩 .

rpcd/playbooks/roles/elasticsearch/templates/upgrade-pre-flight-checks.py.j2
git-harry: I wonder if this could be done more simply with something along the lines of:

```
- name: "Get ElasticSearch APT package version
  command: "dpkg-query --show --showformat='${Version}' elasticsearch"
  register: elasticsearch_installed_version

- name: "Do the thing"
  some_module:
    foo: "bar"
  when: "{{ elasticsearch_installed_version.stdout }} | version_compare('1.2.3', '<=') }}"
```

Not tested, so could be wrong.

rackernexus:
:eyes: [Preview](https://staging.developer.rackspace.com/staging.horse/build-6ce49d00d9/docs/private-cloud/release-notes/)

rackernexus:
:eyes: [Preview](https://staging.developer.rackspace.com/staging.horse/build-63f3780077/docs/private-cloud/release-notes/)

git-harry:
@d34dh0r53 I've added a commit to this patch because a package install task was trying to use the elasticsearch package version for all packages. The greenfield checks have passed and the upgrade is still in progress.

Also been having a play with the code and came up with https://github.com/git-harry/rpc-openstack/tree/issues/1512-2. It's not complete or fully tested but I think it addresses a few things and it was easier to comment with code.
  - currently it appears the 1.7 reindex stuff only happens if logging_upgrade is set to true, it looks like that would mean that if the playbook was run without setting that var ElasticSearch would be upgraded to 2.4.x without doing the reindexing.
  - instead of failing if it looks like the playbook was part run, we could have it check whether or not a migration was in progress and block. Although I'm not sure that's even necessary assuming it's okay to stop elasticsearch during reindexing.
  - simplifed the version checking to use ansible tasks/filters.

rpcd/playbooks/roles/elasticsearch/tasks/upgrade_pre_flight.yml
```diff
@@ -0,0 +1,43 @@
...
+  delegate_to: "{{ groups['elasticsearch_all'][0] }}"
+  register: pre_flight_status
+  ignore_errors: true
+  tags:
+    - logging-upgrade
+    - elasticsearch-upgrade
+    - es-pre-flight
+
+- name: Check for reindexing in progress
+  command: /bin/false
```
sigmavirus24: Good call @git-harry. I like that better. That allows us to explain to support why it has failed.

rpcd/playbooks/roles/elasticsearch/tasks/elasticsearch_install.yml
```diff
@@ -15,7 +15,7 @@
```
sigmavirus24: "Remov" in your commit message.

sigmavirus24:
@git-harry I quite like the simplicity of your approach.

rpcd/playbooks/roles/elasticsearch/defaults/main.yml
```diff
@@ -31,8 +34,16 @@ elasticsearch_apt_repos:
 elasticsearch_apt_keys:
   - { url: "http://packages.elastic.co/GPG-KEY-elasticsearch", state: "present" }
 
+elasticsearch_version: 2.4.1
```
git-harry: Why are you requiring specific versions installed? In general we've not tended to do that. If it is necessary we need to look at tracking that so that it is updated on a regular schedule.

d34dh0r53: It's more for the version check than anything, I'm fine with not installing specific versions.

d34dh0r53:
@git-harry I'm getting this error which is very similar to the errors I was getting last week when trying to do this with ansible.  Hence the move to the python script:

```
TASK: [elasticsearch | Verify ElasticSearch version] **************************
fatal: [d34d-aio-1_elasticsearch_container-1916e3e1] => Failed to template {% if 2.4.1 | version_compare('2.4.1', '<=') %} True {% else %} False {% endif %}: Version comparison: float object has no element 1

FATAL: all hosts have already failed -- aborting
```

Any ideas?

d34dh0r53:
Test task for reference:

```
- name: "Get ElasticSearch APT version"
  command: "dpkg-query --show --showformat='${Version}' elasticsearch"
  register: elasticsearch_installed_version

- debug: var=elasticsearch_installed_version

- name: "Verify ElasticSearch version"
  debug:
    msg: "Version passed"
  when: "{{ elasticsearch_installed_version.stdout }} | version_compare('2.4.1', '<=')"
```

git-harry:
@d34dh0r53 take a look at https://github.com/git-harry/rpc-openstack/tree/issues/1512-2 for an integrated working example.


That following should work for your example, I needed to move '}}' to the end of the line. I also added strict=True for more predictable results.
```
- hosts: elasticsearch_all[0]
  tasks:
    - name: "Get ElasticSearch APT version"
      command: "dpkg-query --show --showformat='${Version}' elasticsearch"
      register: elasticsearch_installed_version

    - debug: var=elasticsearch_installed_version

    - name: "Verify ElasticSearch version"
      debug:
        msg: "Version passed"
      when: "{{ elasticsearch_installed_version.stdout | version_compare('2.4.1', '<=', strict=True) }}"
```

sigmavirus24:
Yeah including the filter inside the jinja2 interpolation makes sense here. Nice catch @git-harry I was just trying to figure out why @d34dh0r53 was seeing that problem

rackernexus:
If you want to double-check your formatting, check [the preview](https://staging.developer.rackspace.com/staging.horse/build-ade28aa0d1/docs/private-cloud/release-notes/).

rackernexus:
If you want to double-check your formatting, check [the preview](https://staging.developer.rackspace.com/staging.horse/build-fe9ea3ec47/docs/private-cloud/release-notes/).

rpcd/playbooks/roles/elasticsearch/tasks/upgrade_pre_flight.yml
```diff
@@ -0,0 +1,52 @@
...
+- name: "Get ElasticSearch version"
+  command: "dpkg-query --show --showformat='{$Version}' elasticsearch"
+  register: elasticsearch_installed_version
+
+- name: "Check for ElasticSearch version > 2.x"
+  set_fact:
+    proceed_with_upgrade: false
+    latest_version: true
+  register: proceed_with_upgrade
+  when: "{{ elasticsearch_installed_version.stdout | version_compare({{ elasticsearch_version }}, '>=', strict=True) }}"
```
sigmavirus24: Does the `{{ elasticsearch_version }}` work inside the `{{ }}`? Is it necessary?

rpcd/playbooks/roles/elasticsearch/tasks/upgrade_pre_flight.yml
```diff
@@ -0,0 +1,52 @@
...
+
+- name: "Check for ElasticSearch version 1.7.x"
+  set_fact:
+    proceed_with_upgrade: false
+    reindex_version: true
+  register: proceed_with_upgrade
+  when:
+    - "{{ elasticsearch_installed_version.stdout | version_compare({{ elasticsearch_reindex_version }}, '>=', strict=True) }}"
```
sigmavirus24: Same question here as above.

rpcd/playbooks/roles/elasticsearch/tasks/upgrade_pre_flight.yml
```diff
@@ -0,0 +1,52 @@
...
+  when:
+    - "{{ elasticsearch_installed_version.stdout | version_compare('1.4.0', '>=', strict=True) }}"
+    - "{{ elasticsearch_installed_version.stdout | version_compare('1.6.9', '<=', strict=True) }}"
+
+- debug:
+  msg: "ElasticSearch is at the latest version, skipping reindexing steps."
+  when: "{{ latest_version | bool }}"
+
+- fail:
+  msg: "ElasticSearch is at the reindexing version, there may be a reindexing in progress.  Aborting installation, use -e 'force_upgrade=true' to continue."
```
sigmavirus24: I think this needs to be:

```yml
- debug:
    msg: "..."
  when: "..."
- fail:
    msg: "..."
  when:
    - ...
```

(Notice the indentation of "msg")

rpcd/playbooks/roles/elasticsearch/tasks/upgrade_pre_flight.yml
d34dh0r53: Fixed, good catch.

rackernexus:
:eyes: [Preview](https://staging.developer.rackspace.com/staging.horse/build-b54a391fbc/docs/private-cloud/release-notes/)

rpcd/playbooks/roles/elasticsearch/tasks/upgrade_pre_flight.yml
d34dh0r53: Done

rackernexus:
Your content preview [is now ready](https://staging.developer.rackspace.com/staging.horse/build-160e1a2805/docs/private-cloud/release-notes/). :bow:

rpcd/playbooks/roles/elasticsearch/tasks/upgrade_pre_flight.yml
```diff
@@ -13,31 +13,38 @@
...
-  ignore_errors: true
-  tags:
-    - logging-upgrade
-    - elasticsearch-upgrade
-    - es-pre-flight
+- name: "Check for ElasticSearch version > 2.x"
+  set_fact:
+    proceed_with_upgrade: false
+    latest_version: true
+  register: proceed_with_upgrade
```
d34dh0r53: @git-harry actually, in this case, I think I do since I need to access that variable from the top level playbook.  In previous testing I was not able to access facts set within tasks from the top level playbook, the only thing accessible are registered variables.

sigmavirus24: `set_fact` and `register` do different things. `set_fact` operates on the host that's running the task. `register` is a global thing (accessible from everywhere). I'm not sure how `proceed_with_upgrade.ansible_facts.proceed_with_upgrade` works though (or how that would work at all).

git-harry: As far as I'm aware Ansible does not have local scopes, I just did a quick test and it seemed to work okay for me.

d34dh0r53: @sigmavirus24 I dumped hostvars and that's is where the boolean value that I set in the task shows up.

rpcd/playbooks/roles/elasticsearch/tasks/upgrade_pre_flight.yml
```diff
@@ -0,0 +1,76 @@
...
+
+- debug:
+    msg: "ElasticSearch is at the latest version, skipping reindexing steps."
+  when: "{{ latest_version | bool }}"
+  tags:
+    - logging-upgrade
+    - elasticsearch-upgrade
+    - elasticsearch-pre-flight
+
+- fail:
```
git-harry: Is there a way we can detect if a reindexing in progress? If there was one, would it matter if we just ran through upgrade_elasticsearch_pre.yml again? I would have thought it would just stop on the monitoring task if there was anything to wait for.

rpcd/playbooks/roles/elasticsearch/defaults/main.yml
```diff
@@ -23,6 +23,9 @@ elasticsearch_legacy_apt_repos:
 # use -e 'logging_upgrade=true' to upgrade logging components
 logging_upgrade: false
```
git-harry: I think this var might be potentially dangerous. If ElasticSearch is 1.5 and this role is run with logging_upgrade set to false, ElasticSearch will be upgraded without the reindexing. Is there a situation where that would be appropriate? If this is going to be kept I think it would make sense to have the role fail in upgrade_pre_flight.yml if logging_upgrade is not true.

rpcd/playbooks/roles/elasticsearch/tasks/upgrade_pre_flight.yml
d34dh0r53: @git-harry there is a way but I'm concerned about the scenario of a very long runing reindexing job.  It took several days to reindex the logs on OSIC1 and that is not a realistic time for the play to monitor.  If the upgrade play is run when elasticsearch is in the 1.7.x branch then the user needs to be absolutely sure what state the environment is in before continuing.

rpcd/playbooks/roles/elasticsearch/defaults/main.yml
d34dh0r53: @git-harry I have that set to false so that we can re-use the plays for a fresh elasticsearch installation.  I suppose we could add some checks for legacy indices in the actuall `elasticsearch_pre_install.yml` to ensure that the upgrade does not happen unintentionally.

rpcd/playbooks/roles/elasticsearch/tasks/upgrade_pre_flight.yml
git-harry: What is anyone going to do to verify the state? Personally I would say we should just fail if there is a reindex in progress otherwise let the role carry on with its job.

Is a backup taken before the upgrade starts?

rackernexus:
:eyes: [Preview](https://staging.developer.rackspace.com/staging.horse/build-7749bcc107/docs/private-cloud/release-notes/)

rackernexus:
:mag_right: [Preview](https://staging.developer.rackspace.com/staging.horse/build-425be64795/docs/private-cloud/release-notes/)

rpcd/playbooks/roles/elasticsearch/tasks/upgrade_pre_flight.yml
d34dh0r53: @git-harry the operator that is in charge of running the reindexing will be able to verify the state, on smaller deployments everything is automated but on long lived clusters with a lot of ES data the operators will want to perform the reindexing prior to the actual upgrade running.  My problem with a deterministic approach is that if for some reason the reindex_in_progress check fails and if the upgrade continues when an actual reindex is happening turns into a much more difficult problem to solve than adding a flag to force the upgrade.

We do not do an automated backup of the data, customers can backup their elasticsearch data but I believe that is up to them to decide.

rpcd/playbooks/roles/elasticsearch/tasks/main.yml
```diff
@@ -13,9 +13,13 @@
...
 - include: upgrade_elasticsearch_pre.yml
   when:
     - logging_upgrade | bool
+    - proceed_with_upgrade.ansible_facts.proceed_with_upgrade | bool
```
sigmavirus24: o_O this is hideous

d34dh0r53: @sigmavirus24 you're welcome?

sigmavirus24: 😭

git-harry: Agreed, I'd be interested to see an example of where set_fact is not sufficient.

rackernexus:
HULK [PREVIEW](https://staging.developer.rackspace.com/staging.horse/build-a9bfef1319/docs/private-cloud/release-notes/)

rackernexus:
HULK [PREVIEW](https://staging.developer.rackspace.com/staging.horse/build-1d2f21f43d/docs/private-cloud/release-notes/)

rpcd/playbooks/roles/elasticsearch/tasks/upgrade_pre_flight.yml
```diff
@@ -0,0 +1,83 @@
...
+  when:
+    - "{{ force_upgrade | bool }}"
+  tags:
+    - logging-upgrade
+    - elasticsearch-upgrade
+    - elasticsearch-pre-flight
+
+- debug:
+    msg: "ElasticSearch is at the latest version, skipping reindexing steps."
+  when: "{{ latest_version | bool }}"
```
mattt416: `latest_version` is only set when we are in fact on the latest version, do you need to set this to false if we're not on the latest?

rackernexus:
:eyes: [Preview](https://staging.developer.rackspace.com/staging.horse/build-a4b5cfd432/docs/private-cloud/release-notes/)

d34dh0r53:
recheck_upgrade - elasticsearch port timeout

d34dh0r53:
recheck_upgrade

git-harry:
recheck_upgrade

git-harry:
👍 while I still have some concerns about this patch/role I think the changes are sufficient to address the issue reported.

@d34dh0r53 would you mind adding a comment with an updated commit message for use when this code is squashed.

sigmavirus24:
I'm 👍 as well. Still waiting on a revised commit message.

d34dh0r53:
@git-harry @sigmavirus24 sure thing, although I think something may have broken with the upgrade test.  Going to check the logs once this current run finishes.

d34dh0r53:
### New Commit Message
Do some verification before starting ES upgrade

We need to do some verification before starting upgrade tasks in order to avoid errors on subsequent runs. This commit creates the upgrade_pre_flight.yml task which does the following:

* Gathers the version information of the installed Elasticsearch package.
* Does the following checks:
  1. If the version is greater than 2.x bypass the reindexing steps entirely
  2. If the version is in the 1.7.x branch abort the run. This is due to the fact that if an installation is sitting on the 1.7.x branch there is a chance that a pre-upgrade reindexing is underway. Continuing with the upgrade tasks or updated package installs at this point could be damaging.
  3. If the version is less that 1.7.0 proceed with the reindex and upgrade.
 
* The force_upgrade variable was added so that in the event of a pre-upgrade reindexing the playbook can be carried forward once the reindexing is completed.

I have also created variables elasticsearch_version and elasticsearch_reindex_version so that we can hard code the versions of elasticsearch to be used. We were using assumed versions in several places.

Thirdly I have added retries to the curl command in order to skip a transient error that we have occasionally seen.

Connects #1512

sigmavirus24:
s/Do some/Perform/ and I'm happy

git-harry:
recheck_upgrade looks to be an unrelated maas issue

d34dh0r53:
Good call @sigmavirus24 "do some" is vague.

git-harry:
@d34dh0r53 @sigmavirus24 I've made the suggested change and split up the lines so that the maximum length isn't exceeded.

```
Perform verification before starting ES upgrade

We need to do some verification before starting upgrade tasks in order
to avoid errors on subsequent runs. This commit creates the 
upgrade_pre_flight.yml task file which does the following:

Gathers the version information of the installed Elasticsearch package.
Does the following checks:

If the version is greater than 2.x bypass the reindexing steps entirely
If the version is in the 1.7.x branch abort the run. This is due to the 
fact that if an installation is sitting on the 1.7.x branch there is a
chance that a pre-upgrade reindexing is underway. Continuing with the 
upgrade tasks or updated package installs at this point could be
damaging.

If the version is less that 1.7.0 proceed with the reindex and upgrade.
The force_upgrade variable was added so that in the event of a
pre-upgrade reindexing the playbook can be carried forward once the 
reindexing is completed.

I have also created variables elasticsearch_version and 
elasticsearch_reindex_version so that we can hard code the versions of
elasticsearch to be used. We were using assumed versions in several
places.

Thirdly I have added retries to the curl command in order to skip a
transient error that we have occasionally seen.

Connects https://github.com/rcbops/rpc-openstack/issues/1512
```

sigmavirus24:
I say :shipit:

rackernexus:
HULK [PREVIEW](https://staging.developer.rackspace.com/staging.horse/build-f8b47b5795/docs/private-cloud/release-notes/)

git-harry:
git diff 9db7752 92079c5 returned no difference so the squash looks good.

PR #1539, opened 2016-10-24
Check status of ceph cluster before rebuilding

mattt416:
This commit checks the status of the cluster before rebuilding on each
iteration of the loop to ensure that the cluster is healthy before
proceeding.  If this isn't done and attention isn't being paid to the
upgrade process then all containers could be torn down and data lost.

Connects https://github.com/rcbops/u-suk-dev/issues/471

(cherry picked from commit 54e5e2360cf6b8dfff8d01a61d06ecec53a3a41f)


hughsaunders:
👍

PR #1537, opened 2016-10-24
Revert "Add MaaS checks for processes status (#1366)"

cloader89:
Reverts rcbops/rpc-openstack#1491

Issue #1236, opened 2016-07-14
Add monitoring for beaver / filebeat and rsyslog processes


JCallicoat:
##### What version (commit) do you encounter this issue on?

N/A
##### What type of issue is this (bug, feature, documentation, regression)?

Feature request
##### What is the expected behavior?

The beaver (or filebeat for r13) and rsyslog processes are monitored to ensure they are running.
##### What is actually happening?

The beaver and/or rsyslog processes sometimes die and logs are not shipped to elasticsearch and/or rsyslog, respectively.
###### Remove branches (lines) you do not wish to target.
- [x] Master
- [x] mitaka-13.1
- [ ] liberty-12.2
- [ ] kilo
##### Does your change require documentation?
- [ ] Yes, and I submitted an issue to [Docs issues](https://github.com/rackerlabs/docs-rpc/issues)
- [ ] Yes, and I contributed a documentation patch to [docs-rpc](https://github.com/rackerlabs/docs-rpc)
- [x] No documentation required for this change

tomc603:
I'm interested in gathering feedback on this. Specifically, how can we supervise processes inside containers without having some kind of "agent" within the container. Perhaps we could attempt to connect to the daemons? Doing that may require external dependencies to prove more than TCP connectivity, though.

tomc603:
As of now, the solution will probably be to create a generic process check function in maas_common, which can then be called from any other check scripts that may need to verify the existence of a running process.

Once that is done, I will create a check script each for beaver, filebeat, and rsyslog. I'm not entirely sure how we would control which script(s) are run where, but I'll ask around once they're done just to confirm I don't need them to fail safely if the utility isn't present or supposed to run.

b3rnard0:
@tomc603 BTW, what size enhancement is this? Can it go in a revision release or is it more appropriate for a minor/major release that will need QE testing, etc.

tomc603:
This is a pretty trivial change- Adding a few functions to maas_common, then writing scripts to utilize those functions. I'd say a revision release is good enough for this.

tomc603:
I've changed tack slightly, but it may work better. I have a generic `process_check.py` script, which accepts one or more process names (the executable portion of a command line) as parameters.

The script reports a `metric_bool` status for each process specified, with the value of that status being the result of searching the running process list for the process.

[...]

PR #1540, opened 2016-10-25
Add MaaS checks for processes status

sigmavirus24:
Reverts rcbops/rpc-openstack#1537
Connects rcbops/rpc-openstack#1236


cloader89:
recheck_upgrade

cloader89:
This looks to be a timeout error on the setup-maas playbook inside the upgrade jenkins job. I saw a few others in the in-review with a similar error. I will try another recheck_upgrade and see.

sigmavirus24:
But @major, this is a revert of a revert. You should be happy!

Issue #1541, opened 2016-10-25
Upgrade release note about Beaver is incorrect now


alextricity25:
###### Please provide the output to 'git describe --all' for  all deployments where you have confirmed it occurs.
```
git describe --all
heads/master
```
##### What type of issue is this? Please tick the appropriate box.
- [x] bug - existing functionality that does not work for you as documented
- [ ] feature - new functionality that you would like to see added

##### What is the expected behavior?
Release notes shouldn't tell operators to [disable Beaver service](https://github.com/rcbops/rpc-openstack/blob/master/releasenotes/notes/add-filebeat-log-shipping-5ed89739de2cf427.yaml#L5). The rpc-pre-upgrade playbook [already removes](https://github.com/rcbops/rpc-openstack/pull/1427) the service from the environment.
##### What is actually happening?
The release notes advise the operator to disable the Beaver service.

##### What were you attempting to do when you encountered this error (give as much detail as possible)? 
NA
##### How can we reproduce this issue?
NA
##### How can we verify your issue is addressed?
NA
##### If you are using or know of a workaround please describe the necessary steps?
NA
##### What is your evaluation on the importance of this issue, and why?
Prio-3, this has no impact on the functionality of the product. It's just a doc issue.
----------------------------------------------------------------------------------------------------------------------------
###### For each pull request, update the branch line to reference it.
- [ ] master
- [ ] mitaka-13.1

##### Upgrade impact. Consider the impact on moving between major, minor and patch releases.
- [ ] This change does not impact upgrades
- [ ] This change addresses a potential upgrade impact
- [ ] The upgrade impact is being tracked by another issue.

The issue tracking the upgrade impact is:

##### Testing impact
- [ ] An issue/pull-request has been create or exists to track the testing gap, and is referenced below
- [ ] This change includes additional testing
- [ ] This change is tested by the existing setup

Additional testing links:

##### Release note
- [ ] This change includes a release note.
- [ ] This change does not require a release note.

##### Does your change require documentation?
- [ ] Yes, and I submitted an issue to [Docs issues](https://github.com/rackerlabs/docs-rpc/issues "Docs issues")
- [ ] Yes, and I contributed a documentation patch to [docs-rpc](https://github.com/rackerlabs/docs-rpc "docs-rpc")
- [x] No documentation required for this change

Additional documentation links (please add links if you have ticked yes):

PR #1542, opened 2016-10-25
Modified releasenote about Beaver

alextricity25:
The behavior described in the current releasenote
for beaver is no longer needed.
The releasenote has been updated to reflect the
new behavior.

Connects #1541


rackernexus:
Your content preview [is now ready](https://staging.developer.rackspace.com/staging.horse/build-3d3812fd7f/docs/private-cloud/release-notes/). :bow:

alextricity25:
There was some concern around changing existing releasenotes, and how that might cause some release notes to disappear from earlier tags, and to reappear in the subsequent one. However, my understanding is that, because of the way we tag our releases, reno doesn't exactly recognize our tags, thus lumping the releasenotes together.
@git-harry Did I explain that correctly?

alextricity25:
Closing this in favor of not changing existing reno release notes

PR #1548, opened 2016-10-26
Perform verification before starting ES upgrade

git-harry:
We need to perform verification before starting upgrade tasks in order
to avoid errors on subsequent runs. This commit creates the
upgrade_pre_flight.yml task file which does the following:

* Gathers the version information of the installed Elasticsearch package.

* Does the following checks:

  1. If the version is greater than 2.x bypass the reindexing steps
  entirely

  2. If the version is in the 1.7.x branch abort the run. This is due
  to the fact that if an installation is sitting on the 1.7.x branch
  there is a chance that a pre-upgrade reindexing is underway. Continuing
  with the upgrade tasks or updated package installs at this point could
  be damaging.

  3. If the version is less that 1.7.0 proceed with the reindex and
  upgrade.

* The force_upgrade variable was added so that in the event of a
pre-upgrade reindexing the playbook can be carried forward once the
reindexing is completed.

I have also created variables elasticsearch_version and
elasticsearch_reindex_version so that we can specify the versions of
elasticsearch to be used. We were using assumed versions in several
places which could lead to issues when packages are updated in the
repositories.

Thirdly I have added retries to the curl command in order to skip a
transient error that we have occasionally seen.

Connects https://github.com/rcbops/rpc-openstack/issues/1512

(cherry picked from commit 92079c55793d8dd984c0f6d47946d46a211553da)

### DIFF
--- a/rpcd/playbooks/roles/elasticsearch/defaults/main.yml
+++ b/rpcd/playbooks/roles/elasticsearch/defaults/main.yml
@@ -23,6 +23,9 @@ elasticsearch_legacy_apt_repos:
 # use -e 'logging_upgrade=true' to upgrade logging components
 logging_upgrade: false
 
+# use -e 'force_upgrade=true' for force the upgrade to happen regardless of the pre-flight checks
+force_upgrade: false
+
 elasticsearch_apt_repo_url: "http://packages.elastic.co/elasticsearch/2.x/debian"
 
 elasticsearch_apt_repos:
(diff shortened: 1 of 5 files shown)


git-harry:
recheck_upgrade
recheck_ceph

sigmavirus24:
👍 :shipit:

d34dh0r53:
:+1:

Issue #1182, opened 2016-06-21
ceph_osd_stats check fails with > 10 osd


npawelek:
##### What version (commit) do you encounter this issue on?

Kilo+
##### What type of issue is this (bug, feature, documentation, regression)?

bug
##### What is the expected behavior?

The ceph_osd_stats check is able to successfully complete with more than 10 osd
##### What is actually happening?

With more than 10 osd on a ceph node, the ceph_monitoring.py check will attempt to build five metrics per osd causing the check to fail due to the metrics limit of 50. It appears that we're building metrics for 'up', 'in', 'kb', 'kb_used', and 'kb_avail'. However, only metric 'up' is actually used by any current alarm criteria that is deployed up to the current Liberty release.
##### How can we reproduce this issue?

Run the check on a node with >10 osd with an updated limits in maas_common.py as outlined here: 
https://github.com/rcbops/rpc-openstack/issues/1180
##### How can we verify your issue is addressed?

Limit number of metrics allowing the check to pass. Potentially only building metrics against the values which are being used in alarm criteria.
##### What is your evaluation on the importance of this issue, and why?

Unable to effectively monitor ceph osd's
###### Remove branches (lines) you do not wish to target.
- [x] Master
- [x] liberty-12.2
- [ ] kilo
##### Does your change require documentation?
- [ ] No documentation required for this change

mancdaz:
Given that we are primarily focused on monitoring with these plugins, rather than gathering arbitrary metrics for capacity planning (appformix is targeted at that), I would advocate removing those things that we are not alerting on.

git-harry:
@npawelek to add further details.

npawelek:
We generate 5 metrics per ceph OSD in the environment and we're only evaluating against 1, whether it's 'up' or not. Due to current metric limit of 50, when we deploy accounts with more than 10 ceph OSD, these checks will fail post-deploy. The deployment workaround is to comment out some of the unused metrics to allow evaluation of the checks once it's handed over to the customer. This typically results in commenting out the following lines where metrics are generated.

#        for key in ('kb', 'kb_used', 'kb_avail'):
#            name = '_'.join((osd_ref, key))
#            maas_common.metric(name, 'uint64', osd[key])

tomc603:
@Frank-ZhangXin worked with @mattt416 yesterday to build an AIO with more than the default number of OSD.

Frank-ZhangXin:
[...]

PR #1555, opened 2016-10-31

Frank-ZhangXin:

### DIFF
--- a/maas/plugins/ceph_monitoring.py
+++ b/maas/plugins/ceph_monitoring.py
@@ -70,17 +70,15 @@ def get_osd_statistics(client=None, keyring=None, osd_ids=None):
         else:
             msg = 'The OSD ID %s does not exist.' % osd_id
             raise maas_common.MaaSException(msg)
-        for key in ('up', 'in'):
-            name = '_'.join((osd_ref, key))
-            maas_common.metric_bool(name, osd[key])
+
+        key = 'up'
+        name = '_'.join((osd_ref, key))
+        maas_common.metric_bool(name, osd[key])
 
         for _osd in pg_osds_dump:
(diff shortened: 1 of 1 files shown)


rackernexus:
HULK [PREVIEW](https://staging.developer.rackspace.com/staging.horse/build-b0e742e850/docs/private-cloud/release-notes/)

rackernexus:
If you want to double-check your formatting, check [the preview](https://staging.developer.rackspace.com/staging.horse/build-bea2a800ee/docs/private-cloud/release-notes/).

mancdaz:
recheck_upgrade

Frank-ZhangXin:
recheck_upgrade

cloader89:
@Frank-ZhangXin Why are you overriding the key value?

Frank-ZhangXin:
@cloader89 it is just a local variable, which is generated for metric function's args. You can name it whatever you want but the important thing is we removed other 'key' except 'up'.

mattt416:
Hi @Frank-ZhangXin, I think this is 👍 based on what was requested (and based on what we currently alarm against), but I do think we need to look at doing a bit more alarming against different OSD situations (up but out, reaching capacity, etc.)  This is difficult based on the max number of metrics a host can have, but an alternative would be to have a single metric called osd_health for example which would encapsulate ALL the OSDs on a host.  That means we can have more than 50 OSDs on a host and still be able to alert on the host if something OSD-related is wrong.

Frank-ZhangXin:
@mattt416 so how about I keep original 5 metrics code as commented and this would give future better monitor system a chance to look after different stat of ceph?

alextricity25:
:+1:

PR #1558, opened 2016-11-03

Frank-ZhangXin:

(cherry picked from commit 50a3959582ee6e994f79296188773dd8c1bb1cd6)


cloader89:
Straightfoward backport

alextricity25:
+1

PR #1559, opened 2016-11-03

Frank-ZhangXin:

(cherry picked from commit 50a3959582ee6e994f79296188773dd8c1bb1cd6)


Frank-ZhangXin:
recheck_upgrade

alextricity25:
We should probably document that these metrics have been removed. Just in case someone has written their own check that actually utilizes these metrics.
@Frank-ZhangXin, could you submit an issue to docs-rpc so we can properly document the removal of theses metrics?

Frank-ZhangXin:
@alextricity25 yeah, no problem

alextricity25:
:+1:

PR #1564, opened 2016-11-08
Set Host sshd_config MaxSessions option to 100

prometheanfire:
The sshd role imported by OSA was used for this. I did create a
new role so that we could run this early, as this is host setup.

See https://github.com/rcbops/rpc-openstack/pull/1516 for more info

Fixes: #1070
Connected #1070

### DIFF
--- a/rpcd/playbooks/roles/rpc_setup_infrastructure/tasks/sshd.yml
+++ b/rpcd/playbooks/roles/rpc_setup_infrastructure/tasks/sshd.yml
@@ -1,0 +1,79 @@
+---
+# Copyright 2016, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Adding additional sshd configuration options is usually easy, but if a
+# # configuration file ends with certain configurations, like a "Match" stanza,
+# # we need a blank line to separate those configurations from the ones that
+# # are added by the security role. For that reason, we check for the existence
+# # of a marker line here and add a marker line to the file if it doesn't exist.
+- name: Check for rpc-openstack role marker in sshd_config
+  command: "grep '^# rpc-openstack configurations' /etc/ssh/sshd_config"
+  register: sshd_marker_check
+  changed_when: False
+  always_run: True
+  failed_when: False
+  tags:
+    - ssh
+
+# Check for "Match" stanzas in the sshd_config.
+- name: Check for Match stanzas in sshd_config
+  command: "grep '^Match' /etc/ssh/sshd_config"
+  register: sshd_match_check
+  changed_when: False
+  always_run: True
+  failed_when: False
+  tags:
+    - ssh
+
+# If the marker is missing, and "Match" stanzas are present, we must carefully
+# add a marker line above any "Match" stanzas in the configuration file. This
+# is done by finding the first match with sed and then adding a marker
+# line above it.
+- name: Add rpc-openstack role marker with sed above Match stanza
+  shell: |
+    sed -i '0,/^Match/s/^Match/\n# rpc-openstack configurations\n\n&/' /etc/ssh/sshd_config
+  when:
+    - sshd_marker_check.rc != 0
+    - sshd_match_check.rc == 0
+  tags:
+    - ssh
+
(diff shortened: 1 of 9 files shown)


rackernexus:
:mag_right: [Preview](https://staging.developer.rackspace.com/staging.horse/build-c003edd61d/docs/private-cloud/release-notes/)

prometheanfire:
@odyssey4me @cloader89 @mattt416 @BjoernT @b3rnard0 @evrardjp sorry about that, needed to remake the branch as we don't allow force pushes anymore.

rackernexus:
:mag_right: [Preview](https://staging.developer.rackspace.com/staging.horse/build-1cc35dc3fc/docs/private-cloud/release-notes/)

prometheanfire:
recheck_all

dunno why no results were shown

rpcd/playbooks/roles/rpc_setup_infrastructure/tasks/sshd.yml
```diff
@@ -0,0 +1,79 @@
+---
```
alextricity25: It would greatly simplify things if these tasks were in a small playbook rather then in it's own role.
Did you have something else in mind for the ``rpc_setup_infrastructure`` role? If not, I'd rather see these tasks in a small playbook. We could even add a variable in deploy.sh to toggle the MaxSessions adjustments. i.e. ``ADJUST_SSH_MAXSESSIONS``.

prometheanfire: They could be.  Initially I was going in that direction but decided against it because I think it likely that support may want something like this in the future (to be able to install before osa is laid down).  I didn't have anything specific in mind for what else the role could do, just setting up the framework for future use.

As far as the maxsessions adjustment goes, ya, that can be done via overrides already.  i.e.  +    `line: 'MaxSessions {{ sshd_maxsessions | default(100) }}' `

odyssey4me: Of course if this was just added to the support role, and support/deploy.sh just was able to override the target hosts for the playbook which implements the support role then you'd have a far simpler solution.

prometheanfire: Possibly, if we were fine with running the support role on it's own, and before anything else is laid down as well.

alextricity25: The support role targets openstack service containers so we won't be able to do it there since the changes being made here need to be performed before anything else.

major:
Could we potentially add this change into OpenStack-Ansible with a configurable variable?

odyssey4me:
@major Yeah, adding a facility into OSA would be a sensible approach - however it's potentially scope creep for the security role which is the only real place to put it.

I still don't think we should be messing with this setting at all. The implementation of RPC does not require it.

If we really want to prevent a large number of forks causing issues with the standard openssh settings, then perhaps a guard should be implemented somehow to ensure that the forks are never larger than 10 for the swift and control plane hosts.

prometheanfire:
True, we could create a function to switch out the FORKS variable with 10 if it's currently over 10, then another one to restore it after a playbook is called.  @BjoernT would that be sufficient?

odyssey4me:
Note that you should switch to using ``ANSIBLE_FORKS`` instead of something custom. ``ANSIBLE_FORKS`` is a well known Ansible environment variable to set for this purpose.

b3rnard0:
@alextricity25 We could use your PTL feedback on this to move forward on it.

alextricity25:
@b3rnard0 I still stand with the comment I made in-line. Pasting it here so it's viewable in waffle:
" It would greatly simplify things if these tasks were in a small playbook rather then in it's own role.
Did you have something else in mind for the rpc_setup_infrastructure role? If not, I'd rather see these tasks in a small playbook. We could even add a variable in deploy.sh to toggle the MaxSessions adjustments. i.e. ADJUST_SSH_MAXSESSIONS."

I don't believe it's necessary for us to carry an extra role at this point. A small playbook that can be run via setting an environment variable will work. 
i.e in deploy.sh we can have something like:
```
if [ "${ADJUST_SSH_MAXSESSIONS}" == "yes" ]; then
   run-ansible [small-playbook-here]
fi
```

prometheanfire:
I first tried to make this a small playbook.  That didn't work because of the need for handlers to restart sshd.  If we could get past that then a small playbook would work (and I'd even prefer that :P)

alextricity25:
@prometheanfire Is that possible by using the [Ansible service module](http://docs.ansible.com/ansible/service_module.html)?

alextricity25:
Hi @BjoernT. Since we have plans to move over the rpc-support role over to openstack-ops, I don't really think it's necessary to have this PR open any longer since it's more of a task that belongs over in openstack-ops. Let me know your thoughts.

b3rnard0:
@alextricity25 Should this issue then be closed if we are moving this to the rpc-support role?

alextricity25:
@b3rnard0 Yes, I think it should be, but let's discuss it a little during our standup today before we close it.

PR #1573, opened 2016-11-14
Adding Magnum support to 'deploy.sh'

chrishultin:
Adding Magnum support to 'deploy.sh'.  This code is NOT idempotent.

Connects https://github.com/rcbops/rpc-openstack/issues/1574

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -197,24 +197,21 @@ heat_policy_overrides:
 magnum_config_overrides:
   certificates:
     cert_manager_type: x509keypair
+# TODO(chris_hultin):
+# Remove this line upon transition to Newton
 ansible_service_mgr: "upstart"
-magnum_rabbitmq_port: "{{ rabbitmq_port }}"
-magnum_rabbitmq_servers: "{{ rabbitmq_servers }}"
-magnum_rabbitmq_use_ssl: "{{ rabbitmq_use_ssl }}"
 EOF
     cp $OA_DIR/playbooks/roles/os_magnum/extras/env.d/magnum.yml /etc/openstack_deploy/env.d/
     cat $OA_DIR/playbooks/roles/os_magnum/extras/haproxy_magnum.yml >> $OA_DIR/playbooks/vars/configs/haproxy_config.yml
     cat $OA_DIR/playbooks/roles/os_magnum/extras/group_vars_magnum.yml >> $OA_DIR/playbooks/inventory/group_vars/magnum_all.yml
-    cat >> $OA_DIR/inventory/group_vars/magnum_all.yml <<'EOF'
+    cat >> $OA_DIR/playbooks/inventory/group_vars/magnum_all.yml <<'EOF'
 magnum_developer_mode: true
 magnum_git_install_branch: stable/newton
 magnum_requirements_git_install_branch: stable/newton
 pip_install_options: "--isolated"
-EOF
-    cat >> $OA_DIR/defaults/repo_packages/openstack_services.yml <<'EOF'
-magnum_git_repo: https://git.openstack.org/openstack/magnum
-magnum_git_install_branch: stable/mitaka
-magnum_git_dest: "/opt/magnum_{{ magnum_git_install_branch | replace('/', '_') }}"
+magnum_rabbitmq_port: "{{ rabbitmq_port }}"
+magnum_rabbitmq_servers: "{{ rabbitmq_servers }}"
+magnum_rabbitmq_use_ssl: "{{ rabbitmq_use_ssl }}"
 EOF
     cp $OA_DIR/playbooks/roles/os_magnum/extras/os-magnum-install.yml $OA_DIR/playbooks/
     sed -i 's/openstack-ansible-magnum/os_magnum/' $OA_DIR/playbooks/os-magnum-install.yml
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -188,12 +188,17 @@ if [[ "${DEPLOY_OA}" == "yes" ]]; then
   #Distribute Magnum configuration files
   if [[ "${DEPLOY_MAGNUM}" == "yes" ]]; then
     cat > $OA_OVERRIDES <<'EOF'
+# These configuration entries for Keystone configure some settings for Trusts to
+# function properly
 keystone_keystone_conf_overrides:
   resource:
     admin_project_name: '{{ keystone_admin_tenant_name }}'
     admin_project_domain_name: default
+# This change to the Heat Policy configuration file allows Magnum to list all
+# stacks, regardless of owner.  This does NOT allow Magnum to have Write access
 heat_policy_overrides:
   "stacks:global_index": "role:admin"
+# This configures Magnum to use the MySQL/Galera database to store certificates
 magnum_config_overrides:
   certificates:
     cert_manager_type: x509keypair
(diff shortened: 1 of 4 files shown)


Apsu:

Apsu:

Apsu:

Apsu:

Apsu:

Apsu:

Apsu:

Apsu:

Apsu:

Apsu:

Apsu:

Apsu:

rackernexus:
:mag_right: [Preview](https://staging.developer.rackspace.com/staging.horse/build-e32e908f29/docs/private-cloud/release-notes/)

alextricity25:
recheck_all

scripts/deploy.sh
```diff
@@ -174,6 +185,42 @@ if [[ "${DEPLOY_OA}" == "yes" ]]; then
 
   cd ${OA_DIR}/playbooks/
 
+  #Distribute Magnum configuration files
+  if [[ "${DEPLOY_MAGNUM}" == "yes" ]]; then
+    cat > $OA_OVERRIDES <<'EOF'
+keystone_keystone_conf_overrides:
```
nrb: Could these variable sections perhaps be moved into config files, rather than inline in the script?

alextricity25:
ok to test

scripts/deploy.sh
chrishultin: What configuration files?

nrb: That's a good question...we perhaps need a place to do this kind of logic where we're extending the OSA-provided files. I can't really think of a good, existing place for these kinds of changes at the moment.

Putting the vars in here requires that support personnel either have to edit the values here or do a deploy then edit the resulting files. Things like the `admin_project_domain_name` may vary across installations, though admittedly I'm not sure most of this would.

alextricity25:
Chris,
I apologize for the lack of direction yesterday. I needed some time to think about this a little and to decide what we are ultimately going to do to make deploy.sh more extend-able.
There is one concern in my mind with the way this is written. 1) The Magnum variables are being added ``user_rpco_variables_overrides.yml`` in a non-AIO scenario - Typically, in a multi-node scenario, the deployer lays down the variable files before running ``deploy.sh``.  If you look closely you'll notice that we don't add anything to the variable files unless ``DEPLOY_AIO`` is set to "yes". Having said that, It would still be nice to have a way for people to pull in these variables. Maybe we could have an example variable file with [these](https://github.com/rcbops/rpc-openstack/pull/1573/files#diff-93518fda10b0403d3c5c20b4df4740a6R191) values in them. We can then document to pull in those variable into ``/etc/openstack_deploy/`` where the deployer can change any values as they wish. We'll still need to perform some actions in ``deploy.sh`` (i.e. clone os_magnum role, add password entry, and everything from [this](https://github.com/rcbops/rpc-openstack/pull/1573/files#diff-93518fda10b0403d3c5c20b4df4740a6R205) line onwards), but this way we aren't dropping any user variables that the deployer won't be aware of.
I would love to get some other thoughts on this.

scripts/deploy.sh
```diff
@@ -174,6 +185,42 @@ if [[ "${DEPLOY_OA}" == "yes" ]]; then
...
+    cp $OA_DIR/playbooks/roles/os_magnum/extras/env.d/magnum.yml /etc/openstack_deploy/env.d/
+    cat $OA_DIR/playbooks/roles/os_magnum/extras/haproxy_magnum.yml >> $OA_DIR/playbooks/vars/configs/haproxy_config.yml
+    cat $OA_DIR/playbooks/roles/os_magnum/extras/group_vars_magnum.yml >> $OA_DIR/playbooks/inventory/group_vars/magnum_all.yml
+    cat >> $OA_DIR/inventory/group_vars/magnum_all.yml <<'EOF'
+magnum_developer_mode: true
+magnum_git_install_branch: stable/newton
+magnum_requirements_git_install_branch: stable/newton
+pip_install_options: "--isolated"
+EOF
+    cat >> $OA_DIR/defaults/repo_packages/openstack_services.yml <<'EOF'
```
odyssey4me: What's the purpose of adding this here if the magnum group is being overridden to use stable/newton and installing in an isolated manner?

odyssey4me: This is causing the repo build process to build the wheels to install the mitaka version of Newton, but the --isolated pip install option above is forcing the magnum install process to ignore the repo.

scripts/deploy.sh
```diff
@@ -174,6 +185,42 @@ if [[ "${DEPLOY_OA}" == "yes" ]]; then
...
+    admin_project_domain_name: default
+heat_policy_overrides:
+  "stacks:global_index": "role:admin"
+magnum_config_overrides:
+  certificates:
+    cert_manager_type: x509keypair
+ansible_service_mgr: "upstart"
+magnum_rabbitmq_port: "{{ rabbitmq_port }}"
+magnum_rabbitmq_servers: "{{ rabbitmq_servers }}"
+magnum_rabbitmq_use_ssl: "{{ rabbitmq_use_ssl }}"
```
odyssey4me: These ``magnum_rabbitmq_*`` vars should be in the groups vars for magnum_all

rackernexus:
:eyes: [Preview](https://staging.developer.rackspace.com/staging.horse/build-7f99641875/docs/private-cloud/release-notes/)

scripts/deploy.sh
```diff
@@ -174,6 +185,39 @@ if [[ "${DEPLOY_OA}" == "yes" ]]; then
...
+magnum_developer_mode: true
+magnum_git_install_branch: stable/newton
+magnum_requirements_git_install_branch: stable/newton
+pip_install_options: "--isolated"
+magnum_rabbitmq_port: "{{ rabbitmq_port }}"
+magnum_rabbitmq_servers: "{{ rabbitmq_servers }}"
+magnum_rabbitmq_use_ssl: "{{ rabbitmq_use_ssl }}"
+EOF
+    cp $OA_DIR/playbooks/roles/os_magnum/extras/os-magnum-install.yml $OA_DIR/playbooks/
+    sed -i 's/openstack-ansible-magnum/os_magnum/' $OA_DIR/playbooks/os-magnum-install.yml
```
odyssey4me: It probably makes more sense to add this playbook in the rpc-openstack repo with a TODO to remove it in Newton.

scripts/deploy.sh
```diff
@@ -51,6 +52,11 @@ ansible-galaxy install --role-file=/opt/rpc-openstack/ansible-role-requirements.
 # Enable playbook callbacks from OSA to display playbook statistics
 grep -q callback_plugins playbooks/ansible.cfg || sed -i '/\[defaults\]/a callback_plugins = plugins/callbacks' playbooks/ansible.cfg
 
+#Clone Magnum repository if required
+if [[ "$DEPLOY_MAGNUM" == "yes" ]]; then
+  git clone https://github.com/openstack/openstack-ansible-os_magnum.git -b stable/newton --single-branch $OA_DIR/playbooks/roles/os_magnum
```
odyssey4me: It's likely better to submodule the role repo to ensure that the SHA is properly fixed and managed.

scripts/deploy.sh
```diff
@@ -138,6 +144,11 @@ if [[ ! -f /etc/openstack_deploy/user_osa_secrets.yml ]] && [[ -f /etc/openstack
   mv /etc/openstack_deploy/user_secrets.yml /etc/openstack_deploy/user_osa_secrets.yml
 fi
 
+#Add Magnum password entries if deploying Magnum
+if [[ "$DEPLOY_MAGNUM" == "yes" ]]; then
+  cat $OA_DIR/playbooks/roles/os_magnum/extras/user_secrets_magnum.yml >> /etc/openstack_deploy/user_osa_secrets.yml
```
odyssey4me: These should probably be added to a file in the rpc-openstack repo rather than copying it in like this. Not sure though @alextricity25 your thoughts?

scripts/deploy.sh
sigmavirus24: I'd rather it be in `ansible-role-requirements.yml`. RPC OpenStack doesn't need to become a  mess of submodules

sigmavirus24: @chrishultin Correct. It'd be no different than what's presently in there.

sigmavirus24: > And where would that role be installed? I'll need to reference the path for some of the later portions.

There's a role path that ansible uses. It's typical /etc/ansible/roles. Why do you think you'll need to reference the path to the role? If it's for the extras and secrets bits, I'd tend to agree with @odyssey4me that those should live in this repository rather than trying to copy them out of the git clone

chrishultin: Makes sense.  But where would we put them?

Issue #1307, opened 2016-08-10
cinder_backup maas check doesn't use venv


mancdaz:
##### On which version(s) do you encounter this issue?

heads/master
##### What type of issue is this? Please tick the appropriate box.
- [x] bug - existing functionality that does not work for you as documented
- [ ] feature - new functionality that you would like to see added
##### What is the expected behavior?

The cinder backup check is created as expected, using the venv wrapper where necessary
##### What is actually happening?

The cinder backup check was added without the 'run_plugin_in_venv' wrapper, which creates the check incorrectly. It actually still runs, but at some point if dependencies are added in to the venv in the future, that don't exist on the host, it will fail.

https://github.com/rcbops/rpc-openstack/blob/e1dd1a4ab74d8de14d6ffedb7d53b3c6d6d913c0/rpcd/playbooks/roles/rpc_maas/templates/cinder_backup_check.yaml.j2#L7-L8
##### What were you attempting to do when you encountered this error (give as much detail as possible)?

Run a multinode deployment, and then run setup-maas and check the cinder_backup template.

Expected content:

```
---snip---
details     :
    file    : run_plugin_in_venv.sh
    args    : ["/usr/lib/rackspace-monitoring-agent/plugins/cinder_service_check.py", "--host", "533811-node15", "192.168.100.2"]
---snip---
```

Actual content:

```
---snip---
details     :
    file    : cinder_service_check.py
    args    : ["--host", "533811-node15", "192.168.100.2"]
---snip---
```
##### How can we reproduce this issue?

Run a multinode deployment, and then run setup-maas and verify-maas - the cinder_backup check template will not have the run_plugin_in_venv wrapper.
##### How can we verify your issue is addressed?

After running setup-maas, check the cinder_backup template and ensure it has the run_plugin_in_venv wrapper configured
##### If you are using or know of a workaround please describe the necessary steps?

At this time, not necessary as the plugin still runs
##### What is your evaluation on the importance of this issue, and why?

Low as the plugin still runs from outside the venv

---
###### For each pull request, update the branch line to reference it.
- [x] master - https://github.com/rcbops/rpc-openstack/pull/1570
- [x] mitaka-13.1 https://github.com/rcbops/rpc-openstack/pull/1578

##### Upgrade impact. Consider the impact on moving between major, minor and patch releases.
- [x] This change does not impact upgrades
- [ ] This change addresses a potential upgrade impact
- [ ] The upgrade impact is being tracked by another issue.

[...]

PR #1578, opened 2016-11-16
Add cinder_backup_check to venv

Frank-ZhangXin:
Once we moved most of MaaS checks to venv, but cinder_backup_check
is left over, though it's working fine with current dependency.
Here's the fix to add it venv in case improper behevior on future
updated dependency.

Connected https://github.com/rcbops/rpc-openstack/issues/1307

(cherry picked from commit 441fb84b2f6052980b4e6a41e930607067a40b78)


git-harry:
👍 
Reviewing rackspace-monitoring-agent.log I can see there are no longer any occurrences of 'plugin=cinder_service_check.py' and the only plugin used is 'run_plugin_in_venv.sh'.

I would have expected to see a release note for this change however, based on the fact that this is a backport and appears to function correctly I will merge it.

PR #1579, opened 2016-11-16
Add cinder_backup_check to venv

Frank-ZhangXin:
Once we moved most of MaaS checks to venv, but cinder_backup_check
is left over, though it's working fine with current dependency.
Here's the fix to add it venv in case improper behevior on future
updated dependency.

Connected https://github.com/rcbops/rpc-openstack/issues/1307

(cherry picked from commit 441fb84b2f6052980b4e6a41e930607067a40b78)


cloader89:
Liberty/kilo don't use venvs

Frank-ZhangXin:
@cloader89 yeah, you're right, my bad. Thanks.

Issue #1328, opened 2016-08-17
deploy.sh doesn't properly check if osa submodule is properly present


evrardjp:
##### On which version(s) do you encounter this issue?

all
##### What type of issue is this? Please tick the appropriate box.
- [x] bug - existing functionality that does not work for you as documented
- [ ] feature - new functionality that you would like to see added
##### What is the expected behavior?

Seemless working deploy.sh
##### What is actually happening?

When you don't clone recursive or don't git submodule init/update, you won't have the osa files.
However scripts/deploy.sh is making use of it. We should ensure the presence of osa files before continuing further.

Moreover the issue you see in the logs is ./scripts/bootstrap-ansible.sh not present. 
So you may think the absent script is in rpc-o/scripts/ where it was about the presence of the script in rpc-o/openstack-ansible/scripts/.
This is fairly trivial issue with a non-obvious error message.
##### What were you attempting to do when you encountered this error (give as much detail as possible)?

Regular work.
##### How can we reproduce this issue?

git clone <rpc-o url>; rpc-openstack/scripts/deploy.sh
##### How can we verify your issue is addressed?

Making sure the error message is correct or alternatively force the git submodule init/update conditionally (env variable).
We could test if the bootstrap-ansible.sh is absent for another conditional, as it's the first step of our deploy.sh script.
##### If you are using or know of a workaround please describe the necessary steps?

Do the stuff properly
##### What is your evaluation on the importance of this issue, and why?

Not important for the devs -- fairly useful for the users/user-friendliness.

---
###### For each pull request, update the branch line to reference it.
- [x] master
- [x] mitaka-13.1
- [x] liberty-12.2
- [ ] kilo

##### Upgrade impact. Consider the impact on moving between major, minor and patch releases.
- [ ] This change does not impact upgrades
- [ ] This change addresses a potential upgrade impact
- [ ] The upgrade impact is being tracked by another issue.

The issue tracking the upgrade impact is:
##### Testing impact
- [ ] An issue/pull-request has been create or exists to track the testing gap, and is referenced below
- [ ] This change includes additional testing
- [ ] This change is tested by the existing setup

Additional testing links:
##### Release note
- [ ] This change includes a release note.
- [ ] This change does not require a release note.
##### Does your change require documentation?
- [ ] Yes, and I submitted an issue to [Docs issues](https://github.com/rackerlabs/docs-rpc/issues)
[...]

PR #1584, opened 2016-11-18
Check openstack-ansible submodule status

tomc603:
When executing the deploy script, check for the presence of a properly
checked out openstack-ansible submodule. This check will not attempt to
update the submodule.

If the submodule is not checked out or has merge conflicts, an error
message will be printed and the deploy script will halt execution. This
is to prevent missing or broken files from causing partial or incorrect
deployments.

If the submodule is checked out, but does not match the expected SHA,
a warning message will be printed but the deploy script will continue
execution. This is to allow local changes or specific versions of OSA
to be tested during deployment.

Connects #1328

(cherry picked from commit 1cd5977f2a20ec6bc4dd53039b62fdff65f9379c)

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -29,6 +29,22 @@ function run_ansible {
   openstack-ansible ${ANSIBLE_PARAMETERS} --forks ${FORKS} $@
 }
 
+# Confirm OA_DIR is properly checked out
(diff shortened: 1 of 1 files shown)


Frank-ZhangXin:
LGTM 👍

PR #1598, opened 2016-11-28
Upaded OSA SHA to tip of stable/mitaka

alextricity25:
Connects https://github.com/rcbops/u-suk-dev/issues/709


sigmavirus24:
👍 the SHA matches the head of stable/mitaka

PR #1602, opened 2016-12-01
Reduce failures due to MaaS issues

mattt416:

- adds retries to tasks that call raxmon-*-list because on very large
  accounts these calls can result in timeouts, 500s, etc. and a single
  API failure should be retried in the event that the failure is
  transient
- removes the failed_when from the `Verify raxmon-entities-list and
  MaaS API authentication does not fail` task as this is specifying the
  standard failure criteria
- adds any_errors_fatal=True to setup-maas.yml so that a failure of a
  task will fail the playbook immediate rather than allow the playbook
  to continue to the end [1]

[1] This is helpful because setup-maas.yml basically operates on all
hosts and containers. If a host fails on (for example) `Verify
raxmon-entities-list and MaaS API authentication does not fail` the
playbook will continue on through the rest of the role because we have
so many hosts included in the hosts line.

Connects https://github.com/rcbops/u-suk-dev/issues/703

### DIFF
--- a/rpcd/playbooks/roles/rpc_maas/tasks/agent_install_remote.yml
+++ b/rpcd/playbooks/roles/rpc_maas/tasks/agent_install_remote.yml
@@ -13,22 +13,19 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-- name: Verify raxmon-entities-list and MaaS API authentication does not fail
-  shell: |
-    {% if maas_venv_enabled | bool %}
-    . {{ maas_venv_bin }}/activate
-    {% endif %}
-    raxmon-entities-list
-  register: raxmon_output
-  failed_when: raxmon_output.rc != 0
-
 - name: Entity {{ entity_name }} count
   shell: |
+    set -o pipefail
     {% if maas_venv_enabled | bool %}
     . {{ maas_venv_bin }}/activate
     {% endif %}
     raxmon-entities-list | grep ' label={{ entity_name }} provider=Rackspace Monitoring ...>$' | wc -l
+  args:
+    executable: "/bin/bash"
   register: entity_count
+  until: entity_count|success
+  retries: 5
+  delay: 2
 
 - name: Zero entities with label {{ entity_name }} exist
   fail:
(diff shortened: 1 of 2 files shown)


rackernexus:
If you want to double-check your formatting, check [the preview](https://staging.developer.rackspace.com/staging.horse/build-b0ac15aaad/docs/private-cloud/release-notes/).

rackernexus:
HULK [PREVIEW](https://staging.developer.rackspace.com/staging.horse/build-abfe77882d/docs/private-cloud/release-notes/)

mattt416:
recheck_swift
recheck_upgrade

mattt416:
recheck_all

mattt416:
recheck_all

mattt416:
recheck_ceph

mattt416:
recheck_all

mattt416:
recheck_all

mattt416:
recheck_swift

mattt416:
recheck_all

rpcd/playbooks/roles/rpc_maas/tasks/agent_install_remote.yml
```diff
@@ -13,22 +13,19 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-- name: Verify raxmon-entities-list and MaaS API authentication does not fail
-  shell: |
```
hughsaunders: Are you removing this task because it runs against a large number of hosts so increases the chance of failure?

mattt416:
recheck_upgrade horizon_extensions apache2 restart failure

rpcd/playbooks/roles/rpc_maas/tasks/agent_install_remote.yml
mattt416: I actually removed that because the next task (which also does a raxmon-entities-list) will now fail correctly due to the pipefail, so it seemed redundant.  Should it get added back in?

hughsaunders: Ah, no that makes sense.

alextricity25:
:+1:

PR #1615, opened 2016-12-08
Revert "Reduce failures due to MaaS issues"

mattt416:
While this change was passing the gate fine, it wasn't thought through correctly as the greps in the pipelines will cause the pipelines to fail if no results are found.  We could create a subshell for the greps and || true but this is probably taking us down the wrong path and we should re-implement properly without going to this length.

This reverts commit 7568bceaf85876983c4669f3b383f23f3a6d584f.


rackernexus:
HULK [PREVIEW](https://staging.developer.rackspace.com/staging.horse/build-c5e77dfd16/docs/private-cloud/release-notes/)

alextricity25:
:+1:

mattt416:
recheck_upgrade

PR #1620, opened 2016-12-12
Magnum support

chrishultin:
Connects https://github.com/rcbops/rpc-openstack/issues/1574


Apsu:

Apsu:

Apsu:

Apsu:

Apsu:

Apsu:

Apsu:

Apsu:

Apsu:

Apsu:

Apsu:

Apsu:

major:
ok to test

major:
add to whitelist

mattt416:
recheck_all

cloader89:
recheck_all

alextricity25:
ok to test

alextricity25:
add to whitelist

major:
We're waiting

alextricity25:
ok to test

alextricity25:
ok to test

alextricity25:
add to whitelist

scripts/deploy.sh
```diff
@@ -122,6 +123,15 @@ if [[ ! -f /etc/openstack_deploy/user_osa_secrets.yml ]] && [[ -f /etc/openstack
   mv /etc/openstack_deploy/user_secrets.yml /etc/openstack_deploy/user_osa_secrets.yml
 fi
 
+if [[ "$DEPLOY_MAGNUM" == "yes" ]]; then
+  cat >> /etc/openstack_deploy/user_osa_secrets.yml <<'EOF'
```
alextricity25: @chrishultin could you move this into a seperate file in tree. You can place it here:
https://github.com/rcbops/rpc-openstack/tree/mitaka-13.1/rpcd/etc/openstack_deploy
This script will move the file to /etc/openstack_deploy automatically. All that is left to do is to use pw-token-gen.py in here to populate the values.

chrishultin:
It looks like there's an issue with the role dependencies now that we're using ansible-role-requirements.yml.  We can either try including the following roles:
```
apt_package_pinning
pip_install
galera_client
```
Or we need to find a different way of doing this.

alextricity25:
The way I see it, we have two options:
1. Add the magnum role to the OSA ansible-role-requirements file and live with the fact that in-tree files will be modified as part of the deployment process.

2. Ignore linting test for the magnum playbook - example: https://github.com/rcbops/rpc-openstack/blob/master/scripts/linting-ansible.sh#L36

I like option #2. It's a shame that the playbook won't run the linting test, but I'd rather not modify in-tree files.

claco:
And this is only a Mitaka/temp issue right?  So in Newton, it's all proper.

chrishultin:
@claco Correct.  Newton has all this baked in, so we don't have to fight it

alextricity25:
recheck_ceph

alextricity25:
recheck_upgrade

alextricity25:
recheck_upgrade

contrib/magnum/os-magnum-install.yml
```diff
@@ -0,0 +1,143 @@
+---
+# Copyright 2016, Ian Cordasco
```
sigmavirus24: Who is that guy? Sounds like a jerk

rpcd/playbooks/roles/filebeat/defaults/main.yml
```diff
@@ -154,6 +154,13 @@ filebeat_logging_paths:
       pattern: "{{ multiline_openstack_pattern }}"
       negate: 'true'
       match: after
+  - paths:
```
sigmavirus24: Does magnum have multiline log messages? If so, does it at least use oslo.log so that they're all uniform?

sigmavirus24: If the answer to those is both "yes", then you should be able to copy-pasta the 

```py
multiline:
  pattern: ...
  negate: 'true'
  match: after
```

Stanza from above

sigmavirus24: Also, as a separate work item, we should probably rework how we put down filebeat files so that they can be disabled if unnecessary.

Either way, since there's no path to these log files there should be little problem to this being laid down everywhere.

chrishultin:
recheck_ceph

chrishultin:
recheck_swift

PR #1641, opened 2016-12-22
Remove unnecessary default variables

alextricity25:
This commit removes the following variables:

* ssl_protocol and ssl_cipher_suite variables. These variables are
  already set to these values by default. Therefore there is no need
  to set them here.

* *_ceilometer_enabled variables. OSA Newton now enables ceilometer
  based on whether the hosts are allocated in the inventory. Since
  we don't carry a ceilometer env.d file, these variables will
  automatically get set to False.

* Removed unnecessary comment regarding apt_pinned_packages

* Removed apply_security_hardening variable. There is no need to set
  this variable to true as that's already it's default value.

* Removed the "cinder_service_backup_program_enabled: true" setting
  from deploy.sh. This variable is now automatically set to true
  when there are swift hosts present

* Renamed the cinder_rpc_thread_pool_size variable to the proper
  name. See:
  https://github.com/rcbops/u-suk-dev/issues/797#issuecomment-268566795

Connects https://github.com/rcbops/u-suk-dev/issues/797

### DIFF
--- a/rpcd/etc/openstack_deploy/user_osa_variables_defaults.yml
+++ b/rpcd/etc/openstack_deploy/user_osa_variables_defaults.yml
@@ -26,7 +26,7 @@ db_max_overflow: 60
 db_max_pool_size: 120
 db_pool_timeout: 60
 
-cinder_rpc_thread_pool_size: "{{ rpc_thread_pool_size }}"
+cinder_rpc_executor_thread_pool_size: "{{ rpc_thread_pool_size }}"
 cinder_rpc_response_timeout: "{{ rpc_response_timeout }}"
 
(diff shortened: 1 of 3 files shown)


alextricity25:
recheck_all

Frank-ZhangXin:
The commit message has minor typo:
2nd section: ...OSA Newton now enables ceilometer based on wheater the hosts are allocated in the inventory. 
'wheater' should be 'whether'

major:
recheck_ceph

A single tempest test failed:
```
Body: {"itemNotFound": {"message": "Volume 5a912f14-a129-4135-80cb-64204c21baca could not be found.", "code": 404}}
```

PR #1643, opened 2016-12-22
Removed solutions tab, horizon-extensions subtree

alextricity25:
This commit removes the horizon-extensions subtree,
as well as the bits in the horizon_extenstions playbook
related to deploying the solutions tab.

Connects https://github.com/rcbops/u-suk-dev/issues/868


alextricity25:
recheck_all

evrardjp:
LGTM. Removed files, removed references in user_variables, should be ok.
We have to be careful upon upgrades that these variables are still respected, but it should be generally good.

👍

Issue #1647, opened 2016-12-29
Maas deprecation warnings and errors in Newton


BjoernT:
##### What type of issue is this? Please tick the appropriate box.
- [x] bug - existing functionality that does not work for you as documented
- [ ] feature - new functionality that you would like to see added

##### What is the expected behavior?

No warnings

##### What is actually happening?

When running setup-maas.yml you run into several warning and errors for

- maas_requires_pip_packages
- plugin_files.stdout_lines
- maas_filesystem_monitors
- maas_filesystem_overrides
- vhost_json
- inventory_hostname == groups['swift_proxy'][0]

Warnings:

```
[DEPRECATION WARNING]: Using bare variables is deprecated. Update your playbooks so that the environment value uses the full variable syntax ('{{maas_requires_pip_packages}}').
This feature
will be removed in a future release. Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
```

Error:

```
fatal: [infra03_nova_conductor_container-ce254a90]: FAILED! => {"failed": true, "msg": "The conditional check 'inventory_hostname == groups['swift_proxy'][0]' failed. The error was: error while evaluating conditional (inventory_hostname == groups['swift_proxy'][0]): list object has no element 0\n\nThe error appears to have been in '/opt/rpc-openstack/rpcd/playbooks/roles/rpc_maas/tasks/main.yml': line 48, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- include: swift_access_check.yml\n  ^ here\n"}
```
##### What were you attempting to do when you encountered this error (give as much detail as possible)? 

See above

##### How can we reproduce this issue?

See above

##### How can we verify your issue is addressed?

No warning

##### If you are using or know of a workaround please describe the necessary steps?

N/A

##### What is your evaluation on the importance of this issue, and why?

Prio 2

----------------------------------------------------------------------------------------------------------------------------
###### For each pull request, update the branch line to reference it.
- [ ] master
- [ ] newton-14.0

##### Upgrade impact. Consider the impact on moving between major, minor and patch releases.
- [ ] This change does not impact upgrades
- [ ] This change addresses a potential upgrade impact
- [ ] The upgrade impact is being tracked by another issue.

The issue tracking the upgrade impact is:

##### Testing impact
- [ ] An issue/pull-request has been create or exists to track the testing gap, and is referenced below
- [ ] This change includes additional testing
[...]

PR #1648, opened 2016-12-29
Prepare rpc_maas role for Ansible 2

BjoernT:
The warnings and errors as a result of using Ansible 2
are fixed for the rpc_maas role.

Closes-Bug: #1647 
Connects #1647

PR #1650, opened 2017-01-04
Fix deprecation warnings & errors in MaaS

major:
This patch fixes some deprecation warnings for bare variables and fixes
an error that stops the playbook when no servers are in the swift_proxy
group.

A few other cleanups of `when:` conditionals are also included.

Connects rcbops/rpc-openstack#1647

### DIFF
--- a/rpcd/playbooks/roles/rpc_maas/tasks/agent_install_local.yml
+++ b/rpcd/playbooks/roles/rpc_maas/tasks/agent_install_local.yml
@@ -44,7 +44,7 @@
   file:
     path: "{{ item }}"
     mode: "0755"
-  with_items: plugin_files.stdout_lines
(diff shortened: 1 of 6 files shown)


d34dh0r53:
This looks good and I prefer the list format over the `and` format that is used in #1648, perhaps we can combine the two PRs.  For now I'll give this one a 👍  as #1648 needs more work.

BjoernT:
@d34dh0r53 I merged the style from this PR into #1648

major:
I'll close this up since #1648 is moving along. I didn't see the other PR originally because it wasn't attached to the waffle card. ;)

BjoernT:
@major no waffles for you, lol. I guess we need to look how to cover PRs as part of the waffle board but also we had no bug tracking meetings due to the holidays. Maybe as simple as that

Issue #1616, opened 2016-12-09
Newton: openstack-ansible-security : V-38660 failure


BjoernT:
##### What type of issue is this? Please tick the appropriate box.
- [x] bug - existing functionality that does not work for you as documented
- [ ] feature - new functionality that you would like to see added

##### What is the expected behavior?

We need to investigate the snmp configuration in regards to HP/Dell montioring tools and our default configuration if managed by kick.
We may need to add migrations steps to fix the snmpd config.
Additionally it seems the egrep is a little to greedy....

##### What is actually happening?

```
TASK [openstack-ansible-security : V-38660 - The snmpd service must only use SNMPv3 or newer] ***
fatal: [storage02]: FAILED! => {"changed": false, "failed": true, "msg": "Insecure SNMP configuration found -- use SNMPv3 only"}
fatal: [storage01]: FAILED! => {"changed": false, "failed": true, "msg": "Insecure SNMP configuration found -- use SNMPv3 only"}
fatal: [storage03]: FAILED! => {"changed": false, "failed": true, "msg": "Insecure SNMP configuration found -- use SNMPv3 only"}
fatal: [infra03]: FAILED! => {"changed": false, "failed": true, "msg": "Insecure SNMP configuration found -- use SNMPv3 only"}
fatal: [infra01]: FAILED! => {"changed": false, "failed": true, "msg": "Insecure SNMP configuration found -- use SNMPv3 only"}
fatal: [infra02]: FAILED! => {"changed": false, "failed": true, "msg": "Insecure SNMP configuration found -- use SNMPv3 only"}
fatal: [logging01]: FAILED! => {"changed": false, "failed": true, "msg": "Insecure SNMP configuration found -- use SNMPv3 only"}
fatal: [compute01]: FAILED! => {"changed": false, "failed": true, "msg": "Insecure SNMP configuration found -- use SNMPv3 only"}
fatal: [compute02]: FAILED! => {"changed": false, "failed": true, "msg": "Insecure SNMP configuration found -- use SNMPv3 only"}

NO MORE HOSTS LEFT *************************************************************




egrep 'v1|v2c|com2sec|community' /etc/snmp/snmpd.conf | grep -v '^\\s*#'
#rocommunity public  localhost
 rocommunity public  default    -V systemonly
                                                 #     settings, change the community string,
#rocommunity secret  10.0.0.0/16
#  It's no longer typically necessary to use the full 'com2sec/group/access' configuration
[...]

PR #1651, opened 2017-01-04
Disable SNMP hardening check

major:
This patch disables the V-38660 hardening check for SNMP and updates
the comments in the `deploy.sh` script to reflect the changes.

V-38660 checks to see if any SNMPv1/v2 configurations exist. If they
exist, the playbook halts. This is highly disruptive for production
deployments.

Connects rcbops/rpc-openstack#1616

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -214,12 +214,15 @@ if [[ "${DEPLOY_OA}" == "yes" ]]; then
     run_ansible haproxy-install.yml
   fi
 
-  # We have to skip V-38462 when using an unauthenticated mirror
-  # V-38660 is skipped for compatibility with Ubuntu Xenial
+  # NOTE(mhayden): V-38642 must be skipped when using an apt repository with
+  # unsigned/untrusted packages.
+  # NOTE(mhayden): V-38660 halts the playbook run when it finds SNMP v1/2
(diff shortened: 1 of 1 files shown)


scripts/deploy.sh
```diff
@@ -214,12 +214,13 @@ if [[ "${DEPLOY_OA}" == "yes" ]]; then
     run_ansible haproxy-install.yml
   fi
 
-  # We have to skip V-38462 when using an unauthenticated mirror
-  # V-38660 is skipped for compatibility with Ubuntu Xenial
+  # NOTE(mhayden): V-38642 must be skipped when using an apt repository with
+  # unsigned/untrusted packages. V-38660 checks for SNMP configs with SNMPv1/2
+  # enabled and it must be skipped all of the time.
```
hughsaunders: if it should be skipped all the time, shouldn't it be removed?

major: @hughsaunders What should be removed? The only reason we skip it here is that we have SNMPv1/2 configs on RPC hosts. We shouldn't do that, but that's a separate problem to fix. ;)

d34dh0r53: @major, I think the comment just needs to be reworded

major:
I'll update the patch to make the comment more explicit.

hughsaunders:
LGTM 👍

Issue #1639, opened 2016-12-19
Ensure ansible.cfg is still needed


alextricity25:
##### On which version(s) do you encounter this issue?
Master on Newton

##### What type of issue is this? Please tick the appropriate box.
- [x] bug - existing functionality that does not work for you as documented
- [ ] feature - new functionality that you would like to see added

##### What is the expected behavior?
TBD

##### What is actually happening?
Currently we override a few ansible behavior configurations here. However, this behavior may overlap with the upstream defaults that are set and might not be needed anymore. We need to investigate to see if any of these configurations are needed. If not, then we can following in the footsteps of OSA to remove the ansible.cfg file as done in this patchset: https://review.openstack.org/#/c/345572/17

##### What were you attempting to do when you encountered this error (give as much detail as possible)? 
This was something that came out of the OSA newton SHA bump work done here: https://github.com/rcbops/rpc-openstack/pull/1577
##### How can we reproduce this issue?
NA
##### How can we verify your issue is addressed?
By the time this issue is ready to be closed, we should have determined whether the ansible.cfg file is still needed and which configurations still need to be overridden.
##### If you are using or know of a workaround please describe the necessary steps?
NA
##### What is your evaluation on the importance of this issue, and why?
Prio-3. It's not needed to continue deployment, but it is necessary to align ourselves with OSA.
----------------------------------------------------------------------------------------------------------------------------
###### For each pull request, update the branch line to reference it.
- [ ] master
- [ ] mitaka-13.1
- [ ] liberty-12.2
- [ ] kilo

##### Upgrade impact. Consider the impact on moving between major, minor and patch releases.
- [ ] This change does not impact upgrades
- [ ] This change addresses a potential upgrade impact
- [ ] The upgrade impact is being tracked by another issue.

The issue tracking the upgrade impact is:

##### Testing impact
- [ ] An issue/pull-request has been create or exists to track the testing gap, and is referenced below
- [ ] This change includes additional testing
- [ ] This change is tested by the existing setup

Additional testing links:

##### Release note
- [ ] This change includes a release note.
- [ ] This change does not require a release note.

##### Does your change require documentation?
- [ ] Yes, and I submitted an issue to [Docs issues](https://github.com/rackerlabs/docs-rpc/issues "Docs issues")
- [ ] Yes, and I contributed a documentation patch to [docs-rpc](https://github.com/rackerlabs/docs-rpc "docs-rpc")
- [ ] No documentation required for this change

[...]

PR #1653, opened 2017-01-04
Remove duplicated ansible.cfg configurations

major:
This patch removes duplicated Ansible configurations that match OSA's
defaults. The ssh timeout was left in the file since it differs from
the OSA default (120 seconds in RPC vs 5 seconds in OSA).

RPC previously set forks at 15 for all deployments, but OSA uses a
safer mechanism to determine the maximum amount of forks. It sets a
hard limit of 10 but it lowers that limit to the maximum amount of
CPUs in the system if the CPU count is under 10.

Connects rcbops/rpc-openstack#1639

### DIFF
--- a/rpcd/playbooks/ansible.cfg
+++ b/rpcd/playbooks/ansible.cfg
@@ -1,25 +1,4 @@
 [defaults]
-gathering = smart
-fact_caching = jsonfile
-fact_caching_connection = /etc/openstack_deploy/ansible_facts
-fact_caching_timeout = 86400
-host_key_checking = False
-
-# Setting forks should be based on your system. The Ansible defaults to 5,
-# the os-lxc-hosts assumes that you have a system that can support
-# OpenStack, thus it has been conservatively been set to 15
-forks = 15
-
-# SSH timeout
+# NOTE: The ssh timeout is increased here to 120 (OSA's default is 5) to
+#       avoid transient ssh errors in some environments.
 timeout = 120
-
-# Set the path to the folder in openstack-ansible which holds the dynamic
-# inventory script - new config setting for ansible v1.9 and above
-inventory = ../../openstack-ansible/playbooks/inventory/
-
-# Set the path to the folder in openstack-ansible which holds the roles
-# that are depended on by the rpc-openstack roles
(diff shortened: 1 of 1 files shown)


alextricity25:
recheck_ceph

major:
That darned tempest test!

```
20:17:37         Response - Headers: {'status': '404', 'content-length': '109', 'content-location': 'http://172.29.236.100:8774/v2.1/6ed426f8fdce4d17954765d93e49009a/os-volumes/13a8a7a3-0e76-4beb-a08c-1941dc6863cd', 'x-compute-request-id': 'req-27a5e158-7836-4b7d-9307-3bdbf2f40a89', 'vary': 'X-OpenStack-Nova-API-Version', 'openstack-api-version': 'compute 2.1', 'connection': 'close', 'x-openstack-nova-api-version': '2.1', 'date': 'Wed, 04 Jan 2017 19:56:11 GMT', 'content-type': 'application/json; charset=UTF-8'}
20:17:37             Body: {"itemNotFound": {"message": "Volume 13a8a7a3-0e76-4beb-a08c-1941dc6863cd could not be found.", "code": 404}}
```

rpcd/playbooks/ansible.cfg
```diff
@@ -1,25 +1,4 @@
...
-
-# Setting forks should be based on your system. The Ansible defaults to 5,
-# the os-lxc-hosts assumes that you have a system that can support
-# OpenStack, thus it has been conservatively been set to 15
-forks = 15
-
-# SSH timeout
+# NOTE: The ssh timeout is increased here to 120 (OSA's default is 5) to
+#       avoid transient ssh errors in some environments.
 timeout = 120
```
prometheanfire: Is there a reason we need this timeout?

prometheanfire: Which transient errors? can they be fixed so we don't have to be 'special'?

major: Not sure exactly what they were, but I remember @andymcc proposed those patches. We had some ssh connectivity issues to newly built containers and increasing the timeouts got us over the hump.

prometheanfire: Ya, looked at the blame/commit for it and it didn't explain it much.

PR #1654, opened 2017-01-05
Bug labels

nitinmehra20:
Added details in CONTRIBUTING.md file around additional labeling for bugs

Connects rcbops/u-suk-dev#977


b3rnard0:
Do not merge this as I'm still reviewing it.

robb-romans:
Admiring this from the sidelines. We're following along in the Docs repos.

b3rnard0:
@alextricity25 I also agree with the bulk of the changes here that @nitinmehra20 is proprosing. The severity labels are good but they will overlap or complicate with the priority labels, for instance, it wouldn't make sense to have a sev-1 issue that's a prio-3. I also agree that we eliminate the priority labels, use the new severity labels to gauge priority as well as to rely on visual prioritization within Waffle, i.e., prioritization is based on the order of the issues that have been approved so the highest priority issue is at the top, followed by next highest issue, etc. This applies to the issues in the other workflow columns.

If an issue does not have a type-bug label, does this mean we now assume it's a bug? This would mean that it must have a type-* something label so that we can tell what type of issue it is.

I think the intent of the origin labels was to show which team found the issue, however, I can see these overlapping with impacts labels somewhat.

b3rnard0:
@nitinmehra20 We need your feedback on some of our initial comments/thoughts.

CONTRIBUTING.md
```diff
@@ -127,6 +143,18 @@ Issues are categorised to simplify the process of identifying the highest priori
   * If an issue is reported by QE, they update an existing issue to say they are impacted or someone else discovers an issue that affects a release candidate add the label 'impacts-qe'.
 * Identify the product features impacted by the issue.
   Any labels relevant to this section will start with 'feature-'.
+* Identity the severity of the issue based on its impact on the stability and/or functionality on the product or end customer
```
nitinmehra20: Will fix this and re-submit :) Thanks for catching this

nitinmehra20:
I generally agree with the review/feedback that is provided. We would still need a bug label if github does not make any distinction between an issue that has been logged as opposed to a task. If the difference does exist, we don't need a type-bug label. Otherwise, I propose just ass a label 'bug' so that it is easy to filter. 

I agree that priority is not needed. Typically, QE used to add priority because the process was quite antiquated and QE acted in role of assuring quality as opposed to engineering it. Nowadays, with bug grooming, sprint planning and waffle board visual priority, priority field is redundant. 

The role of origin label is to identify, where the bug originated from. The comp-type label is more in line with team-affected label. The component that is affected by the bug is the true owner of the bug. Though keep in mind that if the bug is impacting a component but the bug origin is due to an upstream dependency failure, then it is ok to add two comp-type labels to reflect the same.

What are the next steps here?

alextricity25:
I was thinking that the new 'type' labels introduced would imply that an issue is a bug. But now I see the need for a more general "type-bug" label, so please excuse my previous comment about removing that.
@nitinmehra20 I believe all that is left to do is to remove the existing priority labels from the document. We all seem to be in agreement about that.

prometheanfire:
so many labels.  can we even use them all effectivly, that's my question

nitinmehra20:
I have made the changes as requested above as well as consolidated the enhancement label. Let me know your thoughts on the same

alextricity25:
This PR will need to close.
The label definitions were pulled out of the CONTRIBUTING guidelines recently, with the view of only having information specific to contributing to the RPCO project in there.
All of the label definitions, along with anything related to our workflow, are now located in the WORKFLOW document here: https://github.com/rcbops/u-suk-dev/blob/master/WORKFLOW.md

Please make these changes there. Thanks!

b3rnard0:
Closing for now as this is not currently a priority based on upcoming Jira migration.

PR #1658, opened 2017-01-06
Change horizon-extensions git branch version

alextricity25:
This commit specifies the most recent commit on the horizon-extensions
subtree in the mitaka-13.1 branch as the git install branch.

Connects https://github.com/rcbops/u-suk-dev/issues/937


evrardjp:
Definitely better than master.

But, I'd leave it as head of the branch. So probably mitaka-13.1 here.

hughsaunders:
I agree with @evrardjp https://github.com/rcbops/rpc-openstack/compare/mitaka-13.1...bug/939
Liberty will need similar.

evrardjp:
@hughsaunders I'm not sure liberty needs similar.
Liberty used to have horizon extensions from the horizon-extensions repository, which is supposedly fine.
Maybe worth checking the variables (url, fragments, branch) are correct 'though.

Other 13.* branches need checking.

PR #1660, opened 2017-01-10
Skip holland backup check for short lived machines

hughsaunders:
Recently a MaaS check for holland was introduced which checks that
yesterday's backup exists. This makes sense for production but not for
gate runs. In the same PR a mechanism for excluding checks from local
MaaS verification was added, however not all gate runs use local verification
and no mechanism was included for excluding the check from full MaaS API
verification.

This commit attempts to guess the age of a machine by stat-ing
/etc/hostname.  If that file is less than a day old, then the check is
skipped with a successful status returned. This means that gate runs
will pass with no backups, and production deploys will be indicating
backup failures before they are handed over to customers.

Also:
 * Removes local verification check exclusion mechnaism as its no longer used.
 * Removes duplicate state keys in holland_preinstall

Connects rcbops/u-suk-dev#950

### DIFF
--- a/maas/plugins/holland_local_check.py
+++ b/maas/plugins/holland_local_check.py
@@ -15,6 +15,7 @@
 # limitations under the License.
 import argparse
 import datetime
+import os
 import shlex
 import subprocess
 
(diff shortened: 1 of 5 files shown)


hughsaunders:
This is MaaS Related, so needs a gate check with DEPLOY_MAAS=yes :  http://jenkins.propter.net/job/JJB-RPC-AIO_master-swift-periodic/18/

hughsaunders:
Post lint fix check job: http://jenkins.propter.net/job/JJB-RPC-AIO_master-swift-periodic/19/

alextricity25:
@hughsaunders Commit messages must be under 50 characters :/

hughsaunders:
@alextricity25 I was about to get super mad, but thankfully its only the title 😅

alextricity25:
+1

BjoernT:
@hughsaunders @alextricity25  I would have welcomed not to back out all of my changes, since there maybe reasons to exclude checks from verification in the future

PR #1662, opened 2017-01-10
Fixing incorrect path

chrishultin:
Replacing the non-existant $OA_OVERRIDES variable
with the full path to the OSA Variable Overrides file.


chrishultin:
recheck_ceph

Issue #1626, opened 2016-12-14
f5 scripts needs to generate configs for pacakge manager cache


alextricity25:
##### On which version(s) do you encounter this issue?
This is on master

##### What type of issue is this? Please tick the appropriate box.
- [x] bug - existing functionality that does not work for you as documented
- [ ] feature - new functionality that you would like to see added

##### What is the expected behavior?
The package manager cache server has it's own respective virtual server, pools, and monitors when an f5 LB is configured using our f5 scripts. This server was added in newton per https://github.com/openstack/openstack-ansible/commit/c9925bee814dd1ee25ac4f6d29d31922c43eb0f3

##### What is actually happening?
The new endpoint does not have any virtual farms, pools, or monitors created when the f5 LB is configured using our f5 scripts

##### What were you attempting to do when you encountered this error (give as much detail as possible)? 
multi-node lab build with a hardware LB

##### How can we reproduce this issue?
Run the f5-config script. Notice how it's missing the configurations for the package manger cache server.

##### How can we verify your issue is addressed?
The f5 scripts should create configurations for the package manage cache server.

##### If you are using or know of a workaround please describe the necessary steps?
Create the virtual servers, pools, and monitors manually.

##### What is your evaluation on the importance of this issue, and why?
Prio-2. This does not effect development, but does need to be addressed before 14.0.0rc1 is released.
----------------------------------------------------------------------------------------------------------------------------
###### For each pull request, update the branch line to reference it.
- [ ] master

##### Upgrade impact. Consider the impact on moving between major, minor and patch releases.
- [ ] This change does not impact upgrades
- [ ] This change addresses a potential upgrade impact
- [ ] The upgrade impact is being tracked by another issue.

The issue tracking the upgrade impact is:

##### Testing impact
- [ ] An issue/pull-request has been create or exists to track the testing gap, and is referenced below
- [ ] This change includes additional testing
- [ ] This change is tested by the existing setup

Additional testing links:

##### Release note
- [ ] This change includes a release note.
- [ ] This change does not require a release note.

##### Does your change require documentation?
- [ ] Yes, and I submitted an issue to [Docs issues](https://github.com/rackerlabs/docs-rpc/issues "Docs issues")
- [ ] Yes, and I contributed a documentation patch to [docs-rpc](https://github.com/rackerlabs/docs-rpc "docs-rpc")
- [ ] No documentation required for this change

Additional documentation links (please add links if you have ticked yes):

PR #1664, opened 2017-01-10
Add repo_cache to f5-config pool parts and monitor

alextricity25:
Per https://github.com/openstack/openstack-ansible/commit/c9925bee814dd1ee25ac4f6d29d31922c43eb0f3, 
OSA by default implements a package manager cache on the repo servers. This commit makes the changes necessary to the f5-config.py script so that the cache can be utilized.

Connects https://github.com/rcbops/rpc-openstack/issues/1626

### DIFF
--- a/scripts/f5-config.py
+++ b/scripts/f5-config.py
@@ -346,12 +346,12 @@ POOL_PARTS = {
         'hosts': []
     },
     'repo_cache':{
-        'port': 3142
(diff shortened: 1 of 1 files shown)


alextricity25:
For those reviewing, I ran this code in one of the LEM lab f5s and it gave me a healthy virtual server and monitor:
screenshot from 2017-01-10 14 04 31

Issue #1632, opened 2016-12-15
f5-config.py creates incorrect horizon monitor and node configs after new default behavior introduced in Newton


alextricity25:
##### On which version(s) do you encounter this issue?
Master running newton

##### What type of issue is this? Please tick the appropriate box.
- [x] bug - existing functionality that does not work for you as documented
- [ ] feature - new functionality that you would like to see added

##### What is the expected behavior?
The f5-config.py script should create a monitor for the horizon pools that poll horizon on 443 using HTTP, since this is the new default behavior introduced in Newton per https://review.openstack.org/#/c/277199/.
An alternative is to look into a way for the f5-config.py script to discover this information and dynamically build out the monitors according to how the user is deploying their service.
The short term fix would be to change the monitor [here](https://github.com/rcbops/rpc-openstack/blob/master/scripts/f5-config.py#L71-L72) to poll using HTTP, and the node configuration [here](https://github.com/rcbops/rpc-openstack/blob/master/scripts/f5-config.py#L294) to use the backend port 80.

##### What is actually happening?
As of now, the f5-config.py script will create a montior to poll on port 443 using HTTPS, which is no longer valid for a default newton environment. It will also configure the nodes in the horizon_ssl pool to send traffic to port 443, which is also invalid for a default newton environment since the virtualhost configurations for the openstack dashboard are on are for port 80 now.

##### What were you attempting to do when you encountered this error (give as much detail as possible)? 
Newton multi-node lab build with hardware LB

##### How can we reproduce this issue?
Use the f5_deploy.sh script in the LEM labs to generate f5 configurations.

##### How can we verify your issue is addressed?
An alternative is to look into a way for the f5-config.py script to discover this information and dynamically build out the monitors according to how the user is deploying their service.
The short term fix would be to change the monitor [here](https://github.com/rcbops/rpc-openstack/blob/master/scripts/f5-config.py#L71-L72) to poll using HTTP, and the node configuration [here](https://github.com/rcbops/rpc-openstack/blob/master/scripts/f5-config.py#L294) to use the backend port 80. However this assumes that the deployer is using the new default behavior.

##### If you are using or know of a workaround please describe the necessary steps?
The monitors can be changed manually via the LB UI dashboard. The node backend ports can also be changed via the LB dashboard.

##### What is your evaluation on the importance of this issue, and why?
prio-2. This fix is not needed to continue development, but will need to be addressed before 14.0.0rc1 release.
[...]

PR #1668, opened 2017-01-12
Change horizon SSL monitor to HTTP_443

alextricity25:
Also changes backend port to use port 80.
This is kind of a strange one. Due to the new behavior
introduced in https://review.openstack.org/#/c/277199/
and https://review.openstack.org/#/c/288786, horizon listens
on port 443 and 80 on http. The LB monitor needs to be changed
accordingly. Curling http://internal-vip:443 results in a
200 OK now. Curling https://internal-vip:443 will no longer
work since certs no longer exist on the container

### DIFF
--- a/scripts/f5-config.py
+++ b/scripts/f5-config.py
@@ -68,8 +68,8 @@ MONITORS = [
     r'create ltm monitor http /' + PART + '/' + PREFIX_NAME + '_MON_HTTP_NOVA_SPICE_CONSOLE {'
     r' defaults-from http destination *:6082 recv "200 OK" send "HEAD /'
     r' HTTP/1.1\r\nHost: rpc\r\n\r\n" }',
-    r'create ltm monitor https /' + PART + '/' + PREFIX_NAME + '_MON_HTTPS_HORIZON_SSL { defaults-from'
-    r' https destination *:443 recv "302 FOUND" send "HEAD / HTTP/1.1\r\nHost:'
+    r'create ltm monitor http /' + PART + '/' + PREFIX_NAME + '_MON_HTTP_HORIZON_443 { defaults-from'
+    r' http destination *:443 recv "200 OK" send "HEAD / HTTP/1.1\r\nHost:'
     r' rpc\r\n\r\n" }',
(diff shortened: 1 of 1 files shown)


alextricity25:
recheck_all

cloader89:
This looks good to me, will test it.

cloader89:
recheck_all

Frank-ZhangXin:
working with a lab now but found commit title has small typo "hoirzon" which should be "horizon"

busterswt:
In my opinion, if the backend port is changing to port 80, then the monitor needs to be changed to check the pool member on port 80. Checking 443 using cleartext (non SSL) will simply serve to confuse folks down the road. I would rather add some logic to the script to configure http or https Horizon VIP based on runtime flag or something.

alextricity25:
@busterswt That makes more sense to me too. I'm going to wait to see if any of the other folks feel differently before making those changes. If I haven't heard anything by tomorrow then I'll make the change. cc @dpham-rs

Sidenote: We are looking to create an rpc_f5 ansible role that will generate all of these commands based on what is in the ansible inventory. That work is tracked here: https://github.com/rcbops/u-suk-dev/issues/836
If you have any feedback, please, we'll take it!

dpham-rs:
I've seen this in lab and HTTP/80 gave a 302 redirect to HTTPS/443.   If we're going to change the backend to :80, it must respect the request and not try to redirect to HTTPS://<horizon>:443 (which won't work anyway since it's HTTP).

Having HTTP/443 is going to trip up traditional network admins as it goes against common knowledge of well-known ports.   My preference is to use :80 throughout the pool and monitor.

alextricity25:
@dpham-rs, @busterswt: I've created a new PR that uses port 80 for the HORIZON_SSL pool, and changes it's corresponding monitor to HTTP_HORIZON(which uses port 80). These changes worked in my lab. Please let me know what you think: https://github.com/rcbops/rpc-openstack/pull/1682

dpham-rs:
I think connections to the public VIP will barf, we need to remove the 'backend_ssl' attribute from that too or else the F5 will attempt to initiate an HTTPS connection to :80.

Issue #1671, opened 2017-01-13
[f5-config.py] nova_console f5 monitor should be sending heartbeats to /spice_auto.html


alextricity25:
##### What type of issue is this? Please tick the appropriate box.
- [x] bug - existing functionality that does not work for you as documented
- [ ] feature - new functionality that you would like to see added

##### What is the expected behavior?
The f5-config.py script generates commands to create a working f5 monitor for the nova_console service

##### What is actually happening?
The f5-config.py script is generating commands that creates a monitor using HTTP with path "/", which results in an unhealthy monitor(see screenshot):
![screenshot from 2017-01-13 11 11 18](https://cloud.githubusercontent.com/assets/4964964/21938455/07d5b11a-d981-11e6-90a7-77c92c725abf.png)

##### What were you attempting to do when you encountered this error (give as much detail as possible)? 
Run LEM's f5_deploy.sh script and bring up a healthy virtual server for the nova_console service.

##### How can we reproduce this issue?
Clone RPCO master branch and configure it. Execute the dynamic_inventory.py script and feed the output file(openstack_inventory.json) into the LEM's f5_deploy.sh script.
The f5_deploy.sh script will not create a working nova_console virtual server due to a faulty monitor.

##### How can we verify your issue is addressed?
The nova_console service will have a healthy virtual server after the f5_deploy.sh script is invoked.

##### If you are using or know of a workaround please describe the necessary steps?
Change the monitor manually to send heartbeats to "/spice_auto.html" via the BIGIP UI.

##### What is your evaluation on the importance of this issue, and why?
prio-3. There is an easy workaround nor is development halted.
----------------------------------------------------------------------------------------------------------------------------
###### For each pull request, update the branch line to reference it.
- [ ] master

##### Upgrade impact. Consider the impact on moving between major, minor and patch releases.
- [x] This change does not impact upgrades
- [ ] This change addresses a potential upgrade impact
- [ ] The upgrade impact is being tracked by another issue.

The issue tracking the upgrade impact is:

##### Testing impact
- [ ] An issue/pull-request has been create or exists to track the testing gap, and is referenced below
- [ ] This change includes additional testing
- [x] This change is tested by the existing setup

Additional testing links:

##### Release note
- [ ] This change includes a release note.
- [x] This change does not require a release note.

##### Does your change require documentation?
- [ ] Yes, and I submitted an issue to [Docs issues](https://github.com/rackerlabs/docs-rpc/issues "Docs issues")
[...]

PR #1672, opened 2017-01-13
Change console monitor path to /spice_auto.html

alextricity25:
The way it's now, sending a request to the root path "/" will
respond with a 404. The proper base url to use is http://<host>:6082/spice_auto.html.
This commit changes the nova console monitor to send packets to this URL.

This commit also removes the HTTPS nova console monitor, as it's unused.

Connects https://github.com/rcbops/rpc-openstack/issues/1671

### DIFF
--- a/scripts/f5-config.py
+++ b/scripts/f5-config.py
@@ -66,14 +66,11 @@ MONITORS = [
     r' destination *:80 recv "302 Found" send "HEAD / HTTP/1.1\r\nHost:'
     r' rpc\r\n\r\n" }',
     r'create ltm monitor http /' + PART + '/' + PREFIX_NAME + '_MON_HTTP_NOVA_SPICE_CONSOLE {'
-    r' defaults-from http destination *:6082 recv "200 OK" send "HEAD /'
+    r' defaults-from http destination *:6082 recv "200 OK" send "HEAD /spice_auto.html'
     r' HTTP/1.1\r\nHost: rpc\r\n\r\n" }',
     r'create ltm monitor https /' + PART + '/' + PREFIX_NAME + '_MON_HTTPS_HORIZON_SSL { defaults-from'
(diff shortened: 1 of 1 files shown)


alextricity25:
Running this code produced a healthy monitor:
screenshot from 2017-01-13 10 56 18

alextricity25:
FYI - This will need to be backported to the newton-14.0 branch before it's closed.

BjoernT:
@alextricity25  👎  What is the reason for this, we should never run TCP checks in gating

alextricity25:
In my lab(fcfs-iad3-4), the http check didn't come up green for this monitor.
I just curl'ed http://internal_vip:6082, and it returns a "connection refused" message.
Thanks for bringing this up, this might not be a problem with the monitor configurations, but rather some network device blocking 6082.

devx:
I will have to echo @BjoernT 👎 

One should not use TCP unless we are just making sure that there exists 3 way hand shake.  Additionally if you are monitoring an HTTP service you should expect an OK 200.  So if port is not coming up it could indicate a change in the service, a dead server, bad network update, etc. Not to mention the case that we change the service to only respond to HTTPS traffic.

alextricity25:
@BjoernT @devx I've updated the title and description of this fix to account for your feedback. The problem was that the monitor was sending heartbeats to http://URL:6082/ and responding with a 404 when the monitor was expecting a 200 OK. The proper base URL is http://URL:6082/spice_auto.html. This URL responds with a 200 OK. 
I've changed the monitor to send heartbeats to this URL and keep using HTTP.

BjoernT:
👍 Yes that is good. We internally fixed that monitor to use spice_auto.html, I guess that never made it back into this script :-(

alextricity25:
recheck_swift

PR #1676, opened 2017-01-16
Use to_json filter to properly format unicode

d34dh0r53:
For some reason in Newton the filebeat paths and tags are being
populated into the jinja2 template as unicode strings.  The leading
u is causing filebeat to treat the path as not found, hence it is
not sending any logs.  Using the to_json filter causes jinja2 to
properly format the string allowing the path to be read and the
correct tag(s) to be applied.

Connects https://github.com/rcbops/u-suk-dev/issues/863


hughsaunders:
recheck_ceph

hughsaunders:
recheck_ceph failed on resolving pub cloud endpoint

andymcc:
LGTM

PR #1682, opened 2017-01-17
Revert changes to HORIZON_SSL monitor

alextricity25:
And change what monitor the HORIZON_SSL pool is using.
This commit configures the HORIZON_SSL pool to use the
HTTP monitor which checks the backend nodes using port 80.
This commit fulfills the feedback given in:
https://github.com/rcbops/rpc-openstack/pull/1668#issuecomment-272914892.

The reason this is done is due to the new behavior introduced
in OSA Newton. SSL is expected to be terminated at the LB, and
is no longer implemented on the horizon containers. For more
information on that change, please see:
https://review.openstack.org/#/c/277199/


dpham-rs:
I think connections to the public VIP will barf, we need to remove the 'backend_ssl' attribute from that too or else the F5 will attempt to initiate an HTTPS connection to :80.  The other changes look good.

`Line 305:        'backend_ssl': True`

alextricity25:
@dpham-rs Done. Thanks for the feedback!

alextricity25:
recheck_ceph

PR #1683, opened 2017-01-17
Prepare rpc_maas role for Ansible 2

BjoernT:
The warnings and errors as a result of using Ansible 2
are fixed for the rpc_maas role.

Connects #1647
(cherry picked from commit 4277d77cf81a568ee6d6e786acdfd088cc3ee9e9)


alextricity25:
recheck_ceph

PR #1684, opened 2017-01-18
[backport] Change hoirzon SSL monitor to HTTP_443

alextricity25:
Also changes backend port to use port 80.
This is kind of a strange one. Due to the new behavior
introduced in https://review.openstack.org/#/c/277199/
and https://review.openstack.org/#/c/288786, horizon listens
on port 443 and 80 on http. The LB monitor needs to be changed
accordingly. Curling http://<container-ip>:443 results in a
200 OK now. Curling https//<container-ip>:443 will no longer
work since certs no longer exist on the container

(cherry picked from commit eb59440ad5cce2a33f87ae06d2281fbcf14b2d67)


alextricity25:
**This is the wrong commit I cherry picked from. Please excuse this**

PR #1691, opened 2017-01-19
MaaS: Disable alarms/checks with regex

major:

Connects rcbops/u-suk-dev#1019

### DIFF
--- a/rpcd/playbooks/roles/rpc_maas/templates/cpu_check.yaml.j2
+++ b/rpcd/playbooks/roles/rpc_maas/templates/cpu_check.yaml.j2
@@ -7,6 +7,7 @@ alarms            :
     idle_percent_average        :
         label                   : idle_percent_average--{{ inventory_hostname|quote }}
         notification_plan_id    : "{{ maas_notification_plan }}"
+        disabled                : {{ (('idle_percent_average--'+inventory_hostname|quote) | match(maas_excluded_alarms_regex)) | ternary('true', 'false') }}
         criteria                : |
             :set consecutiveCount={{ maas_alarm_local_consecutive_count }}
             if (metric["idle_percent_average"] <= {{ cpu_idle_percent_avg_critical_threshold }}) {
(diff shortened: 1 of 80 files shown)


major:
This patch is up as a RFC. Obviously, the conditional belongs in each check file, but I've added it to just one of them to generate some discussion.

@npawelek and @BjoernT -- let me know if this is going in the right direction.

BjoernT:
It seems we should probably support regex so we can disable alarms per subset of host or very specifically rather than just disabling all alarms of a given name

major:
That's a decent idea, @BjoernT. I guess we could make `maas_excluded_alarms` into a list of regexes perhaps? I'd have to loop through and apply the regex to the name of the alarm each time.

npawelek:
I like that idea @BjoernT. 

@major For alarms that are included in the list, will the alarm still be included in the server-side yaml or will it iterate over and be excluded? If excluded, one thing that comes to mind is if a yaml were to include only check and no alarm definition, it could be portrayed that it's not in-use and may be subject to be deleted by a member of support (this is an education piece). It may be worth noting that alarms section within the yaml definition can also have a `disabled` boolean, it's just that none of our templates include it. It would probably be a good idea to include the alarm, but mark it as disabled. This also ties back to https://github.com/rcbops/u-suk-dev/issues/948, "Agent-based check/alarm configuration (yamls) are missing ‘disabled’ flag under each alarm section."

prometheanfire:
@npawelek  ya, that does sound like the preferred way of disabling an alarm.

@major the regex solution sounds like what we need, but I'm not sure how much that that will complicate the logic in the template though.

major:
Let me know how this patch looks. I've only applied it to one of the checks.

rpcd/playbooks/roles/rpc_maas/templates/cpu_check.yaml.j2
```diff
@@ -6,6 +6,7 @@ timeout           : "{{ maas_check_timeout }}"
 alarms            :
     idle_percent_average        :
         label                   : idle_percent_average--{{ inventory_hostname|quote }}
+        disabled                : {{ idle_percent_average | match(maas_excluded_alarms_regex) | ternary('true', 'false') }}
```
prometheanfire: Really wish the maas agent didn't make us use double negatives.

major: I don't understand why we can't fix this. :trollface:

major:
Oof, I have some "order of operations" issues going on here.

major:
recheck_ceph

Ceph mirrors have some shenanigans.

major:
I've heavily tested this on a AIO without errors.

BjoernT:
I take your word that you have heavily tested, I don't have the cycles for it now. So thanks and 👍

PR #1692, opened 2017-01-19
Fix item.group bug in process.yml

major:
This patch fixes a bug where the `setup-maas.yml` playbook would
fail when checking for `item.group` in the host groups.

Connects rcbops/u-suk-dev#1025


prometheanfire:
looks like another ansible2 bug, probably should have changed in that commit

major:
recheck_ceph

Looks like the Ceph repo servers had a temporary issue.

major:
I'll go ahead and merge since this has two positive reviews.

PR #1704, opened 2017-01-20
Document scripts in scripts directory

alextricity25:
This is only part of the work that is being done here:
Connects https://github.com/rcbops/u-suk-dev/issues/833


alextricity25:
recheck_ceph

alextricity25:
TODO: convert to RST

scripts/deploy.sh
```diff
@@ -168,9 +168,6 @@ if [[ ! -f /etc/openstack_deploy/user_osa_secrets.yml ]] && [[ -f /etc/openstack
   mv /etc/openstack_deploy/user_secrets.yml /etc/openstack_deploy/user_osa_secrets.yml
 fi
 
-# update the RPC-O secrets
```
prometheanfire: what's with this stuff, if this is just docs

scripts/README.rst
```diff
@@ -0,0 +1,145 @@
...
+toggled on or off. Please follow the same pattern outlined in this
+script, and enclose any instructions specific to deploying the new
+feature with a conditional similar to what is done
+`here <https://github.com/rcbops/rpc-openstack/blob/master/scripts/deploy.sh#L226>`__
+for ceph.
+
+f5-config.py
+------------
+
+NOTE: The f5-config.py script is currently being reworked as part of
```
prometheanfire: Not sure if we should include info like this that is rapidly changing

scripts/README.rst
```diff
@@ -0,0 +1,145 @@
...
+---------------
+
+Is used by our tox tests, which are run by the travis ci, to check the
+python scripts used in RPCO are pep8 compliant. This script is invoked
+`here <https://github.com/rcbops/rpc-openstack/blob/master/tox.ini#L39>`__
+
+migrate-yaml.py
+---------------
+
+TODO: (alextricity25) Remove this in the master and newton-14.0 branches
```
prometheanfire: Don't think this is the right place for a TODO, should make a backlogged issue instead

scripts/README.rst
alextricity25: I'll make this a part of https://github.com/rcbops/u-suk-dev/issues/836 and add a note there

scripts/README.rst
alextricity25: Good point. I'm going to bring this up in https://github.com/rcbops/u-suk-dev/issues/802 as it seems fitting for that card.

alextricity25:
@prometheanfire I have addressed your feedback. Thanks for the review!

alextricity25:
Backport for this is https://github.com/rcbops/rpc-openstack/pull/1735

PR #1706, opened 2017-01-20
MaaS: Do not deploy filebeat alarms w/o ELK

major:
The `rpc_maas` role currently deploys filebeat checks and alarms to
all hosts and containers, but this causes alarms if ELK isn't
deployed in the environment.

This patch checks to see if the ELK group exists and only deploys
filebeat checks and alarms if it exists.

Connects rcbops/u-suk-dev#1032

### DIFF
--- a/rpcd/playbooks/roles/rpc_maas/defaults/main.yml
+++ b/rpcd/playbooks/roles/rpc_maas/defaults/main.yml
@@ -577,3 +577,9 @@ swift_accesscheck_file: "index.html"
 swift_accesscheck_user_name: "accesscheck"
 swift_operator_role: "swiftoperator"
 swift_service_project_name: "service"
+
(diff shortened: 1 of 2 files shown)


BjoernT:
Hmm how about situations where the customer wants to use his own elk and we still need some log forwarding ?

major:
That's a good question. Perhaps there could be an override that forces the filebeat checks/alarms to be deployed even if ELK isn't deployed in the RPC-O environment. I'll give that a try.

major:
recheck_swift
recheck_xenial

major:
Fixed a logic bug in my `when:` section.

major:
recheck_swift

major:
This is much easier with the updated MaaS template files patch that already merged.

PR #1716, opened 2017-01-23
[Backport - Mitaka] Disable MaaS alarms/checks with regex

major:

Manual edits were needed to solve a conflict in `rpcd/playbooks/roles/rpc_maas/templates/lb_api_check_swift_proxy.yaml.j2` since the label of the check was different in Mitaka/Liberty. Also, `rpcd/playbooks/roles/rpc_maas/templates/holland_local_check.yaml.j2` and `rpcd/playbooks/roles/rpc_maas/templates/lb_api_check_swift_access.yaml.j2` didn't exist in Mitaka/Liberty, so they were left out of the backport.

Interdiff: https://gist.github.com/major/1e9af9c38b0fc68923f1b476d9e882a7

(manual backport from https://github.com/rcbops/rpc-openstack/commit/d9ec02cd4f941558725892293a3e44130f0813d2)
Connects rcbops/u-suk-dev#1019


major:
Hold off on merging until #1722  makes it in. That one will need to be added to this patch.

PR #1721, opened 2017-01-23
MaaS: Fix excluded checks removal

major:
This patch updates the tasks in `maas_exclude.yml` to handle regular
expressions rather than strings.

Connects rcbops/u-suk-dev#1052

### DIFF
--- a/rpcd/playbooks/roles/rpc_maas/tasks/maas_exclude.yml
+++ b/rpcd/playbooks/roles/rpc_maas/tasks/maas_exclude.yml
@@ -13,18 +13,19 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-- name: Ensure local checks are removed
-  file:
-    path: "/etc/rackspace-monitoring-agent.conf.d/{{ item }}--{{ inventory_hostname }}.yaml"
-    state: absent
-  with_items:
-    - "{{ maas_excluded_checks }}"
+- name: Get a list of all checks that exist on the system
+  find:
+    paths: /etc/rackspace-monitoring-agent.conf.d/
+    patterns: "*.yaml"
+  register: maas_check_files
   delegate_to: "{{ physical_host }}"
 
-- name: Ensure remote checks are removed
+- name: Ensure excluded checks are removed
   file:
-    path: "/etc/rackspace-monitoring-agent.conf.d/{{ item }}.yaml"
+    path: "{{ item.path }}"
     state: absent
   with_items:
-    - "{{ maas_excluded_checks }}"
+    - "{{ maas_check_files.files }}"
+  when:
+    - item.path | basename | splitext | first | match(maas_excluded_checks_regex)
   delegate_to: "{{ physical_host }}"
--- a/rpcd/playbooks/roles/rpc_maas/tasks/main.yml
+++ b/rpcd/playbooks/roles/rpc_maas/tasks/main.yml
@@ -87,6 +87,8 @@
     - groups['ceph_all']|length > 0
     - inventory_hostname in groups['ceph_all']
 
+# Remove any checks that match the regular expressions provided in the
+# `maas_excluded_checks` list.
 - include: maas_exclude.yml
 
 - include: restart_raxmon.yml


rpcd/playbooks/roles/rpc_maas/tasks/maas_exclude.yml
```diff
@@ -13,18 +13,17 @@
...
   file:
-    path: "/etc/rackspace-monitoring-agent.conf.d/{{ item }}.yaml"
+    path: "{{ item.path }}"
     state: absent
   with_items:
-    - "{{ maas_excluded_checks }}"
-  delegate_to: "{{ physical_host }}"
+    - "{{ maas_check_files.files }}"
+  when:
+    - item.path | basename | match(maas_excluded_checks_regex)
```
mattt416: This didn't clean up any checks for me -- do we not need the `delegate_to` here and above so the operations dealing w/ `/etc/rackspace-monitoring-agent.conf.d/` are executed on the physical host, where the agent is installed?

major: We do need to get the `delegate_to` back in there. I'll fix that.

Can you give me the value of `maas_excluded_checks` so I can test on my AIO?

mattt416: maas_excluded_checks: ['^nova.*$']

rpcd/playbooks/roles/rpc_maas/tasks/maas_exclude.yml
```diff
@@ -13,18 +13,18 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-- name: Ensure local checks are removed
-  file:
-    path: "/etc/rackspace-monitoring-agent.conf.d/{{ item }}--{{ inventory_hostname }}.yaml"
-    state: absent
-  with_items:
-    - "{{ maas_excluded_checks }}"
-  delegate_to: "{{ physical_host }}"
```
mattt416: ^^^ Think this needs to be added back in also.

major: This seems redundant with the second task below.  The task below looks through all of the files in `/etc/rackspace-monitoring-agent.conf.d/` to find any which match the regexes in `maas_excluded_checks`.

mattt416: But this new task is what is gathering the list of files, right?  And that needs to run on the physical_host.

rpcd/playbooks/roles/rpc_maas/tasks/maas_exclude.yml
```diff
@@ -13,18 +13,18 @@
...
+- name: Ensure excluded checks are removed
   file:
-    path: "/etc/rackspace-monitoring-agent.conf.d/{{ item }}.yaml"
+    path: "{{ item.path }}"
     state: absent
   with_items:
-    - "{{ maas_excluded_checks }}"
+    - "{{ maas_check_files.files }}"
+  when:
+    - item.path | basename | match(maas_excluded_checks_regex)
```
mattt416: Not sure if it's of concern, but this match is actually matching against a filename, which may not be what people expect (since it's including `.yaml`).  I _think_ that's differing behaviour to how this worked before.

rpcd/playbooks/roles/rpc_maas/tasks/maas_exclude.yml
major: You're right on that. I'll fix it.

rpcd/playbooks/roles/rpc_maas/tasks/maas_exclude.yml
major: Good point. Do you have any suggestions on how to work around that? I'll think on it.

major:
Rebased.

major:
Fixing this in #1718 instead.

PR #1727, opened 2017-01-24
MaaS: Add additional HP checks

antonym:
This patch adds three additional HP checks for MaaS:

- HP Controller Check
- HP Controller Cache Check
- HP Controller RAID Battery Check

Connects rcbops/u-suk-dev#1054


npawelek:
@antonym This looks good to me.

On a side note, I wonder what the reasoning behind the initial creation being separate yamls (each a separate check)? Wouldn't it make more sense to have a single check with all the different metrics as their own alarm? This way the check is run once a minute, not 6 independent times.

maas/plugins/hp_monitoring.py
```diff
@@ -59,7 +59,7 @@ def get_controller_status():
 
 def get_controller_cache_status():
     return check_command(('hpssacli', 'ctrl', 'all', 'show', 'status'),
-                         'Cache Status', 'OK')
+                         'Cache Status', ('OK', 'Not Configured'))
```
BjoernT: I have not come across a unconfigured cache yet, we may remove that until we come across this

antonym:
@npawelek my guess is to have more control over which checks are on and off in the defaults file?  Different hardware types might have different checks needed.  If the server has SSDs, cache and battery might not be required for instance.  We can always open a card to combine the checks if you think that might be better.

Also it should be pretty easy to add any additional checks to the hp_hardware.py file as it's just screen scraping the various HP utilities.  If you think there are some other checks that would make supports life easier, let me know and I can look at adding them.

maas/plugins/hp_monitoring.py
antonym: Yeah, I was testing on my DL380 Gen8 and got this (I think you are using Gen9s):

root@rpc-openstack:~# hpssacli ctrl all show status

Smart Array P420i in Slot 0 (Embedded)
   Controller Status: OK
   Cache Status: Not Configured
   Battery/Capacitor Status: OK

Usually you don't configure cache if you're using SSDs as it actually slows down the array.  I can remove though if you think it'll be an issue.

npawelek:
@antonym #1716 adds the granularity to disable alarms, if necessary. @BjoernT what do you think?

maas/plugins/hp_monitoring.py
BjoernT: OK lets add an override so we can allow unconfigured cache of alternatively we can check for installed drives with 'hpssacli ctrl slot=3 pd all show' and search for 'Solid State' turning off the cache

antonym:
I tested them collapsed into a single hp_check.yaml which works, but I did notice that re-running the playbook leaves the existing checks in place in the /etc/rackspace-monitoring-agent.conf.d directory.  In an upgrade situation it might just simply make duplicate checks.  Any concerns there or would a cleanup occur on deployment?

npawelek:
If you delete the yamls for the old hp checks and restart the agent, the checks will disappear. In most upgrade scenarios we re-deploy maas which would solve this issue.

antonym:
Ok, consolidated into one check.  Also merged the dell_check and hp_check variables into a host_check variable which abstracts the hardware type from the user.  I can also consolidate the openmanage checks the same way either here or in another commit.

major:
@antonym Go ahead and rebase this one now that the master gate is happy again.

antonym:
recheck_xenial

antonym:
Putting a block on this until #1718 lands so I can refactor on top of it.

antonym:
Rebased to align with #1718.

antonym:
recheck_ceph

BjoernT:
Please back port into newton branch

PR #1722, opened 2017-01-23
MaaS template fixes

mattt416:
This commit makes the following template updates:

- in ceph_osd_stats.yaml.j2, osd_id is cast to a string otherwise
  ansible will throw a template error since it cannot contact an int
  and a string
- in rabbitmq_status.yaml.j2, we update the rabbitmq_mem_alarm_status
  alarm to reference rabbitmq_mem_alarm_status rathe than
  rabbitmq_disk_free_alarm_status

Connects https://github.com/rcbops/u-suk-dev/issues/1057

PR #1731, opened 2017-01-24
Update OSA SHA to tip of stable/newton

alextricity25:
FYI - There was a need to bring in another fix that is necessary for the ceph job to pass:
#1722
That patch has been applied here so we can get all of the jobs to pass.

Connects https://github.com/rcbops/u-suk-dev/issues/1060
Connects rcbops/u-suk-dev#1057


major:
We need to get https://github.com/rcbops/rpc-openstack/pull/1722 in before this one to fix that MaaS template issue.

alextricity25:
FYI - There was a need to bring in another fix that is necessary for the ceph job to pass:
https://github.com/rcbops/rpc-openstack/pull/1722
That patch has been applied here so we can get all of the jobs to pass.

major:
Yeah, @alextricity25. Sorry about that broken build. ;)

Toot

PR #1732, opened 2017-01-24
Update OSA SHA to tip of stable/newton

alextricity25:
Connects https://github.com/rcbops/u-suk-dev/issues/1060


major:
We need to get https://github.com/rcbops/rpc-openstack/pull/1722 in before this one to fix that MaaS template issue.

alextricity25:
recheck_xenial

PR #1733, opened 2017-01-24
Modularise AIO bootstrap portion of deploy.sh

odyssey4me:
This PR moves the AIO bootstrap out of deploy.sh into its own script.


rpcd/etc/openstack_deploy/user_rpco_variables_overrides.yml.aio
```diff
@@ -0,0 +1,35 @@
...
+#
+# This file is laid down only once and is used for overriding variables based on the file name.
+# This file will not be overwritten by upgrades.
+
+elasticsearch_heap_size_mb: 1024
+lb_name: "{{ ansible_hostname }}"
+maas_notification_plan: "npTechnicalContactsEmail"
+net_max_speed: 1000
+
+{% if (rpco_deploy_swift | default(False)) | bool %}
```
odyssey4me: This file can't just be copied - it must be interpreted. Also, this needs to be ``rpco_deploy_ceph``.

scripts/bootstrap-aio.sh
```diff
@@ -0,0 +1,57 @@
...
+## Shell Opts ----------------------------------------------------------------
+set -e -u -x
+
+## Variables -----------------------------------------------------------------
+# Extra options to pass to the AIO bootstrap process
+export BOOTSTRAP_OPTS=${BOOTSTRAP_OPTS:-''}
+
+export BASE_DIR=${BASE_DIR:-"/opt/rpc-openstack"}
+export OA_DIR="${BASE_DIR}/openstack-ansible"
+export RPCD_DIR="${BASE_DIR}/rpcd"
```
alextricity25: Is there anyway we could centralize these environment variables? I believe they are used in deploy-rpc-playbooks.sh as well. Could we move them into functions.sh, then source functions.sh anywhere we need them? You seem to already be doing this for BASE_DIR and OA_DIR. could you add RPCD_DIR in there too?

scripts/bootstrap-aio.yml
```diff
@@ -0,0 +1,134 @@
...
+        dest: "/etc/openstack_deploy/env.d/{{ item.name }}"
+      when: "{{ rpco_deploy_elk | bool }}"
+      with_items:
+        - name: "elasticsearch.yml"
+        - name: "kibana.yml"
+        - name: "logstash.yml"
+
+  vars:
+    osa_cfg_src_path: "{{ lookup('env', 'OA_DIR') | default('/opt/rpc-openstack/openstack-ansible') ~ '/etc/openstack_deploy' }}"
+    rpco_cfg_src_path: "{{ lookup('env', 'RPCD_DIR') | default('/opt/rpc-openstack') ~ '/etc/openstack_deploy' }}"
```
alextricity25: The default path is missing the rpcd directory. It should be:
```
... | default('/opt/rpc-openstack/rpcd') ~ '/etc/openstack_deploy' }}"
```

scripts/deploy.sh
```diff
@@ -57,121 +57,24 @@ if [[ "$DEPLOY_AIO" != "yes" ]] && [[ "$DEPLOY_HARDENING" != "yes" ]]; then
...
-  DATA_DISK_MIN_SIZE="$((1024**3 * $(awk '/bootstrap_host_data_disk_min_size/{print $2}' ${OA_DIR}/tests/roles/bootstrap-host/defaults/main.yml) ))"
-  # Determine the largest secondary disk device available for repartitioning which meets the minimum size requirements
-  DATA_DISK_DEVICE=$(lsblk -brndo NAME,TYPE,RO,SIZE | \
-                     awk '/d[b-z]+ disk 0/{ if ($4>m && $4>='$DATA_DISK_MIN_SIZE'){m=$4; d=$1}}; END{print d}')
-  # Only set the secondary disk device option if there is one
-  if [ -n "${DATA_DISK_DEVICE}" ]; then
-    export BOOTSTRAP_OPTS="${BOOTSTRAP_OPTS} bootstrap_host_data_disk_device=${DATA_DISK_DEVICE}"
-  fi
+  # execute the AIO configuration bootstrap
+  source "$(dirname "${0}")/bootstrap-ansible.sh"
```
alextricity25: I think you meant ``bootstrap-aio.sh`` here, not ``bootstrap-ansible.sh``

scripts/bootstrap-aio.sh
odyssey4me: @alextricity25 I would recommend against that. Right now they will work if one of the scripts, or the user overrides them but the scripts will operate on their own too. If you're trying to be DRY I guess it could be done, but I don't really see much value from it.

scripts/bootstrap-aio.yml
odyssey4me: Good catch - will address that.

scripts/deploy.sh
odyssey4me: Good catch - will address that.

odyssey4me:
Once https://github.com/rcbops/rpc-openstack/pull/1731 merges this will require a rebase on top of that.

odyssey4me:
recheck_all

alextricity25:
```
Connects: rcbops/u-suk-dev#834
```
Waffle is kind of picky here. It won't link the PR to the bug until you remove the colon.
```
Connects rcbops/u-suk-dev#834
```

odyssey4me:
@alextricity25 Thanks - fixed up now. I found a bunch of bugginess that I think is now solved. I'd like to see whether it passes gates before cleaning it up as I think it's a bit messy right now. Reviews are welcome though!

rpcd/etc/openstack_deploy/user_rpco_variables_overrides.yml.aio
```diff
@@ -0,0 +1,35 @@
...
+#
+# This file is laid down only once and is used for overriding variables based on the file name.
+# This file will not be overwritten by upgrades.
+
+elasticsearch_heap_size_mb: 1024
+lb_name: "{{ ansible_hostname }}"
+maas_notification_plan: "npTechnicalContactsEmail"
+net_max_speed: 1000
+
+{% if (groups['ceph_all'] is defined) and (groups['ceph_all'] | length > 0) %}
```
odyssey4me: This won't work as the dynamic inventory is not being used when this file is templated out.

scripts/bootstrap-aio.sh
```diff
@@ -0,0 +1,57 @@
...
+DATA_DISK_MIN_SIZE="$((1024**3 * $(awk '/bootstrap_host_data_disk_min_size/{print $2}' ${OA_DIR}/tests/roles/bootstrap-host/defaults/main.yml) ))"
+# Determine the largest secondary disk device available for repartitioning which meets the minimum size requirements
+DATA_DISK_DEVICE=$(lsblk -brndo NAME,TYPE,RO,SIZE | \
+                   awk '/d[b-z]+ disk 0/{ if ($4>m && $4>='$DATA_DISK_MIN_SIZE'){m=$4; d=$1}}; END{print d}')
+# Only set the secondary disk device option if there is one
+if [ -n "${DATA_DISK_DEVICE}" ]; then
+  export BOOTSTRAP_OPTS="${BOOTSTRAP_OPTS} bootstrap_host_data_disk_device=${DATA_DISK_DEVICE}"
+fi
+
+pushd ${OA_DIR}/tests
```
odyssey4me: I don't think we need to change into that directory here any more.

scripts/bootstrap-aio.sh
```diff
@@ -0,0 +1,57 @@
...
+  # Run AIO bootstrap playbook
+  if [ -z "${BOOTSTRAP_OPTS}" ]; then
+    ansible-playbook -vvv ${BASE_DIR}/scripts/bootstrap-aio.yml \
+                     -i test-inventory.ini
```
odyssey4me: I'm not convinced that we need to make use of the test-inventory.ini here. We could just use ``-i "localhost,"`` I think.

scripts/bootstrap-aio.yml
```diff
@@ -0,0 +1,147 @@
...
+        config_overrides:
+          container_skel:
+            ceph_osd_container:
+              properties:
+                is_metal: false
+        config_type: "yaml"
+      when: "{{ rpco_deploy_ceph | bool }}"
+
+    - name: Deploy the RPC-O ceph conf.d file
+      copy:
```
odyssey4me: move this into the task at line 77 and add a conditional to the dict for each item

scripts/bootstrap-aio.yml
```diff
@@ -0,0 +1,147 @@
...
+        mode: "{{ item.mode | default(omit) }}"
+      with_items:
+        - name: "user_osa_variables_defaults.yml"
+          mode: "0440"
+        - name: "user_rpco_variables_defaults.yml"
+          mode: "0440"
+        - name: "user_rpco_secrets.yml"
+        - name: "env.d/nova.yml"
+
+    - name: Deploy the RPC-O variable override file
```
odyssey4me: merge this into the config_template task at line 95 and add a conditional into the dict

scripts/bootstrap-aio.yml
```diff
@@ -0,0 +1,147 @@
...
+        - "ceilometer.yml"
+        - "gnocchi.yml"
+
+    - name: Remove the swift config file
+      file:
+        path: "/etc/openstack_deploy/conf.d/swift.yml"
+        state: absent
+      when: "{{ not rpco_deploy_swift | bool }}"
+
+    - name: Deploy the RPC-O ELK config files
```
odyssey4me: move this into the task at line 77 and add a conditional to the dict for each item

scripts/bootstrap-aio.yml
```diff
@@ -0,0 +1,147 @@
...
+    - name: Remove the telemetry config files
+      file:
+        path: "/etc/openstack_deploy/conf.d/{{ item }}"
+        state: absent
+      with_items:
+        - "aodh.yml"
+        - "ceilometer.yml"
+        - "gnocchi.yml"
+
+    - name: Remove the swift config file
```
odyssey4me: merge this into the task at line 113 through the use of a conditional

scripts/deploy.sh
```diff
@@ -57,121 +57,24 @@ if [[ "$DEPLOY_AIO" != "yes" ]] && [[ "$DEPLOY_HARDENING" != "yes" ]]; then
...
-                     awk '/d[b-z]+ disk 0/{ if ($4>m && $4>='$DATA_DISK_MIN_SIZE'){m=$4; d=$1}}; END{print d}')
-  # Only set the secondary disk device option if there is one
-  if [ -n "${DATA_DISK_DEVICE}" ]; then
-    export BOOTSTRAP_OPTS="${BOOTSTRAP_OPTS} bootstrap_host_data_disk_device=${DATA_DISK_DEVICE}"
-  fi
+  # execute the AIO configuration bootstrap
+  source "$(dirname "${0}")/bootstrap-aio.sh"
+
   # force the deployment of haproxy for an AIO
   export DEPLOY_HAPROXY="yes"
```
odyssey4me: The ``DEPLOY_HAPROXY`` no longer carries any meaning. Remove it.

scripts/deploy.sh
```diff
@@ -57,121 +57,24 @@ if [[ "$DEPLOY_AIO" != "yes" ]] && [[ "$DEPLOY_HARDENING" != "yes" ]]; then
...
-    else
-      if [[ "$DEPLOY_SWIFT" == "yes" ]]; then
-        echo "glance_default_store: swift" >> $OA_OVERRIDES
-      else
-        echo "glance_default_store: file" >> $OA_OVERRIDES
-      fi
-    fi
-
-    if [[ "$DEPLOY_HARDENING" != "yes" ]]; then
-      echo "apply_security_hardening: false" >> $OA_OVERRIDES
```
odyssey4me: This needs to be added to the override config.

scripts/bootstrap-aio.sh
odyssey4me: @alextricity25 OK, it looks like we have a bunch of vars being set all over the place. They're not consistently being set either, so it seems that perhaps it is a good idea.

Do you think that all the ``DEPLOY_`` vars should move over into the functions too?

alextricity25: Yes. The more we can centralize, the better IMO.

odyssey4me: Alright, I'll work that up in a subsequent commit as I think it belongs in one of its own.

scripts/deploy.sh
odyssey4me: Done.

scripts/bootstrap-aio.yml
```diff
@@ -0,0 +1,157 @@
...
+        - name: "aodh.yml"
+        - name: "ceilometer.yml"
+        - name: "gnocchi.yml"
+        - name: "swift.yml"
+          condition: "{{ not rpco_deploy_swift | bool }}"
+
+    - name: Update the RPC-O secrets
+      shell: "{{ rpco_base_dir }}/scripts/update-secrets.sh"
+
+    - name: Check for the existance of a gating settings file
```
hughsaunders: We should be able to move the generation of the gating settings file to after /etc/openstack_deploy has been created so that it no longer needs to be written into the repo. Also the placeholder creds should probably be removed.

evrardjp:
This looks good.

scripts/bootstrap-aio.yml
odyssey4me: @hughsaunders Good call - can you create cards for any follow on ideas that can be based on this patch?

odyssey4me:
Note that now that the OSA SHA has been updated I'll be pushing up another patch to improve the idempotence of the AIO bootstrap. But for now this PR as-is achieves the goal I set out to achieve, which was to allow the Ansible and AIO bootstrap to be executed without being forced to run all the playbooks.

scripts/deploy.sh
```diff
@@ -57,121 +35,24 @@ if [[ "$DEPLOY_AIO" != "yes" ]] && [[ "$DEPLOY_HARDENING" != "yes" ]]; then
...
-
-    # set the necessary bits for ceph
-    if [[ "$DEPLOY_CEPH" == "yes" ]]; then
-      cp -a ${RPCD_DIR}/etc/openstack_deploy/conf.d/ceph.yml.aio /etc/openstack_deploy/conf.d/ceph.yml
-
-      # In production, the OSDs will run on bare metal however in the AIO we'll put them in containers
-      # so the MONs think we have 3 OSDs on different hosts.
-      sed -i 's/is_metal: true/is_metal: false/' /etc/openstack_deploy/env.d/ceph.yml
+  # set the kibana admin password
+  sed -i "s/kibana_password:.*/kibana_password: ${ADMIN_PASSWORD}/" $RPCD_SECRETS
```
evrardjp: This could be included above, replacing it from copy to config_template.
This, or I don't see the spacing correctly.

odyssey4me: @evrardjp Yeah, it could. I can do this in further refactoring rather than in this patch. I'd prefer this set of patches to go in sooner to unblock the artifact work which prompted this being done in the first place. If you can create a card on our board for this I'd appreciate it.

scripts/bootstrap-aio.yml
hughsaunders: Yep, I thought overnight that there is a chicken-egg problem, this patch should go in as-is to maintain compatibility with the jenkins-rpc and jenkins.propter, then the process can be refactored for cit jenkins. I'm tracking that in rcbops/u-suk-dev#1069

scripts/bootstrap-aio.sh
```diff
@@ -0,0 +1,48 @@
...
+# Determine the largest secondary disk device available for repartitioning which meets the minimum size requirements
+DATA_DISK_DEVICE=$(lsblk -brndo NAME,TYPE,RO,SIZE | \
+                   awk '/d[b-z]+ disk 0/{ if ($4>m && $4>='$DATA_DISK_MIN_SIZE'){m=$4; d=$1}}; END{print d}')
+# Only set the secondary disk device option if there is one
+if [ -n "${DATA_DISK_DEVICE}" ]; then
+  export BOOTSTRAP_OPTS="${BOOTSTRAP_OPTS} bootstrap_host_data_disk_device=${DATA_DISK_DEVICE}"
+fi
+
+# Run AIO bootstrap playbook
+if [ -z "${BOOTSTRAP_OPTS}" ]; then
```
hughsaunders: -e "" is fine, so this if isn't necessary.

scripts/bootstrap-aio.yml
```diff
@@ -0,0 +1,157 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+- name: Set facts for the AIO bootstrap process
+  hosts: localhost
+  user: root
+  tasks:
+    - name: Set ceph user_variables override
```
hughsaunders: s/ceph/security hardening/?

scripts/bootstrap-aio.yml
```diff
@@ -0,0 +1,157 @@
...
+        - name: "user_osa_variables_defaults.yml"
+          mode: "0440"
+        - name: "user_rpco_variables_defaults.yml"
+          mode: "0440"
+        - name: "user_rpco_secrets.yml"
+        - name: "env.d/nova.yml"
+        - name: "conf.d/ceph.yml.aio"
+          condition: "{{ rpco_deploy_ceph | bool }}"
+
+    - name: Template the RPC-O config files
```
hughsaunders: 😍

scripts/bootstrap-aio.sh
odyssey4me: @hughsaunders Yeah this is just matching the bootstrap-aio.sh script in OSA. This check was introduced because of issues seen when executing ``-e ""`` which may have been temporary. I can clean this up in a follow on commit.

hughsaunders: ok

PR #1736, opened 2017-01-24
search through all pids and their children for process checks

prometheanfire:
cgroup.procs only displays one layer, sometimes containers will have
child cgroups that we want to check.  This change allows us to switch
to a check that checks all descendants of a parent as well.

Connects https://github.com/rcbops/u-suk-dev/issues/1001

### DIFF
--- a/maas/plugins/process_check.py
+++ b/maas/plugins/process_check.py
@@ -15,6 +15,8 @@
 # limitations under the License.
 import argparse
 import os
+from subprocess import CalledProcessError
+from subprocess import check_output
 
 from maas_common import metric_bool
 from maas_common import print_output
@@ -48,7 +50,7 @@ def get_processes(procs_path):
     process_names = set()
     try:
         with open(procs_path, mode='r') as f:
-            for pid in f.read().splitlines():
+            for pid in f.read().split():
                 process_names.update(get_process_name(pid))
     except OSError as ex:
         # Could not read the PID list file, probably due to a permission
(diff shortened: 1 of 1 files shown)


alextricity25:
FYI - We have to ensure this has a passing xenial job with maas deployed. PR triggered jobs don't do this by default, so we'll have to trigger it manually off this branch.
I've already fired one off: http://jenkins.propter.net/job/JJB-RPC-AIO_master-xenial-periodic/34/

prometheanfire:
ah, thanks

prometheanfire:
recheck_all

I think we need to recheck because the branches were broken...  Just to be safe I'm going to rebase on master.

alextricity25:
@prometheanfire You're going to have to install the lxc module in the maas venv to move forward with this. The error I'm seeing in the gate jobs is:
```
Traceback (most recent call last):\\n  File \"/usr/lib/rackspace-monitoring-agent/plugins/process_check.py\", line 17, in <module>\\n    import lxc\\nImportError: No module named lxc\\n'), ('filebeat_process_check--jramcp-148-4562-memcached-container-dae3f8f5', 'Traceback (most recent call last):\\n  File \"/usr/lib/rackspace-monitoring-agent/plugins/process_check.py\", line 17, in <module>\\n    import lxc\\nImportError: No module named lxc\\n')
```

alextricity25:
The xenial maas job is here:
http://jenkins.propter.net/job/JJB-RPC-AIO_master-xenial-periodic/37/console
I'm going to block this until we see a passing xenial job with maas enabled.

PR #1737, opened 2017-01-24
MaaS: Enable agent auto upgrade

antonym:
Enables auto upgrade in the rackspace-monitoring-agent.  It appears that the key is monitoring_upgrade, but oddly, if this is set to true or false or not even set, the logs always show:

Tue Jan 24 05:39:33 PM 2017 INF: Upgrades are enabled

It might be hard coded to be enabled here:

https://github.com/virgo-agent-toolkit/rackspace-monitoring-agent/blob/4d20ce09493b75fdbb3140ca15c89ea420836920/agent.lua#L58

Connects rcbops/u-suk-dev#1064

### DIFF
--- a/rpcd/playbooks/roles/rpc_maas/defaults/main.yml
+++ b/rpcd/playbooks/roles/rpc_maas/defaults/main.yml
@@ -560,6 +560,12 @@ verify_maas_retries: 5
 #
 maas_use_api: true
 
+#
+# maas_agent_upgrade: Allow for automatic MaaS agent upgrades
+#
+#
+maas_agent_upgrade: true
+
 # Swift object access check
 swift_accesscheck_enabled: true
 swift_accesscheck_container: "accesscheck"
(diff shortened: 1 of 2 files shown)


major:
Might want to ask @devx since he's been knee deep in the agent code lately.

jayofdoom:
I'm reasonably sure new agents automatically default to upgrades being enabled, but this appears to be correct to force it.

nrb:
https://github.com/virgo-agent-toolkit/rackspace-monitoring-agent/blob/c3ae6f16b06820182203c70467202ecc2eed6437/agent.lua#L58 is the auto upgrade setting in the code at the 2.6.7 commit SHA; looks like it was forced on then, too.

antonym:
Yeah, it's hard coded by default, so this would ensure it never gets shut off if if the hard code is ever removed. :)

alextricity25:
Thanks @antonym! Is the backport PR to the newton-14.0 branch already up?

PR #1743, opened 2017-01-25
MaaS: Search all pids and children for process checks

prometheanfire:
cgroup.procs only displays one layer, sometimes containers will have
child cgroups that we want to check.  This change allows us to switch
to a check that checks all descendants of a parent as well.

We do not use the lxc module here for two reasons.  One, it is built
only for python3, and two, it is not available in pypi (and is not
fun to build out of tree).

Connects https://github.com/rcbops/u-suk-dev/issues/1001
(cherry picked from commit b8468b8695cd474483541cf06a33404a01419312)

### DIFF
--- a/maas/plugins/process_check.py
+++ b/maas/plugins/process_check.py
@@ -15,6 +15,8 @@
 # limitations under the License.
 import argparse
 import os
+from subprocess import CalledProcessError
+from subprocess import check_output
 
 from maas_common import metric_bool
 from maas_common import print_output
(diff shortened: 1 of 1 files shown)


alextricity25:
xenial+newton+maas job has been triggered for this PR: http://jenkins.propter.net/job/JJB-RPC-AIO_newton-xenial-periodic/
Blocking this PR until it passes.

alextricity25:
recheck_ceph

alextricity25:
xenial+newton+maas job passed. Once the ceph jobs passes this can merge.

alextricity25:
recheck_ceph

PR #1749, opened 2017-01-26
[Backport - Newton] MaaS improvements

major:
This backport includes the following four patches to master:

- 63369c2 MaaS: Improve period/timeout/disabling checks
- f501904 MaaS: Enable agent auto upgrade
- ef1bbae MaaS template fixes
- 20d4ba9 Disable MaaS alarms/checks with regex

Connects rcbops/u-suk-dev#1033
Connects rcbops/u-suk-dev#1050
Connects rcbops/u-suk-dev#1052

### DIFF
--- a/rpcd/playbooks/roles/rpc_maas/defaults/main.yml
+++ b/rpcd/playbooks/roles/rpc_maas/defaults/main.yml
@@ -524,6 +524,20 @@ maas_rpc_scripts_dir: /opt/rpc-openstack/scripts
 maas_excluded_checks: []
 maas_excluded_alarms: []
 
+# Customizing MaaS check period and timeout.
+# The following two variables allow deployers to provide custom check periods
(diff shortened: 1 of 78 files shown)


antonym:
recheck_swift
recheck_xenial

major:
recheck_xenial

cloudnull:
Changes LGTM however waiting the gate to approve. In a follow on patch I'd love to see if there's a way to limit all of the repetition though the check yaml files. even though Ansible isn't a programming language and shouldn't be treated as such I feel like we may be able to leverage something like a jinja2 include (http://jinja.pocoo.org/docs/2.9/templates/#include) which would cut down on all of the ``rpcd/playbooks/roles/rpc_maas/templates/*.yaml.j2`` file repetition.

major:
Good point, @cloudnull. That would make things much easier.

PR #1750, opened 2017-01-26
MaaS: Set rabbitmq_qgrowth_excl_notifications to WARNING

antonym:
rabbitmq_qgrowth_excl_notifications identifies the rate of messages in
the queue but in most cases will also trigger
rabbitmq_msgs_excl_notifications resulting in two tickets for the same
issue.  Flipping rabbitmq_qgrowth_excl_notifications to WARNING will
prevent a ticket from being opened but still capture metrics.

Connects rcbops/u-suk-dev#1075


antonym:
recheck_ceph
recheck_swift

antonym:
Ok, I'll leave it at critical.  The current default threshold is 15 for rabbitmq_queue_growth_rate_threshold so bumping it to 100 would help alleviate noise?

npawelek:
The rate is based on the period of the check, so this would greatly increase it. It should cut out most of the noise and still allow us to see potential issues.

BjoernT:
@antonym Put it on hold until I have verified if that is a source for false alerts and which threshold might be suitable. As said this is the only type of alarm to identify Rabbit consumer issues, hence we need to keep alarming on it

antonym:
Sounds good, I'll close this one out for now.  Thanks!

PR #1752, opened 2017-01-27
Tidy up the nits from PR 1733

odyssey4me:
This PR just tidies up the nits from https://github.com/rcbops/rpc-openstack/pull/1733 now that the PR has merged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,15 +1,12 @@
 # rpc-openstack
-Optional add-ons for Rackspace Private Cloud
+Rackspace Private Cloud (RPC)
 
 # openstack-ansible integration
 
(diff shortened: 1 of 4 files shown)


evrardjp:
I'd be happy to see the documentation change in the basic README.md too. I guessed you'll do another patch for that to separate the functional of the documentation bits.

odyssey4me:
@evrardjp Wow, the README is woefully out of date. Yeah, I'll try and do some clean-up there.

odyssey4me:
@evrardjp So the procedure described in the README does not need changing, but there was some other stuff out of date and that needed to be removed. Once the follow on work for this is done, we can change the whole procedure because using ``deploy.sh`` won't be required any more. For now, though, it's still the optimal method.

PR #1753, opened 2017-01-27
[backport - newton] Modularise AIO bootstrap portion of deploy.sh

odyssey4me:
This PR moves the AIO bootstrap out of deploy.sh into its own script.


odyssey4me:
recheck_ceph
recheck_xenial

evrardjp:
I didn't fully test this, but it seems good. It passes gates too, so I think we're good to go.
👍

PR #1757, opened 2017-01-27
Make RPC Deployment optional

hughsaunders:
This allows deploy.sh to be used to prepare a deployment without
starting the deploy. This is necessary as Jenkins jobs are now being
divided into stages such as prepare and deploy, so that common stages
can be shared between jobs.

Connects rcbops/u-suk-dev#1069


hughsaunders:
recheck_upgrade failed on waiting for elasticsearch to come up.

cloader89:
recheck_upgrade

PR #1759, opened 2017-01-27
MaaS: Set default alarm/check params

major:
This patch sets the default alarm exclusions and check periods as
specified in rcbops/u-suk-dev#948.

Connects rcbops/u-suk-dev#1081

### DIFF
--- a/rpcd/etc/openstack_deploy/user_rpco_variables_defaults.yml
+++ b/rpcd/etc/openstack_deploy/user_rpco_variables_defaults.yml
@@ -82,7 +82,20 @@ maas_alarm_remote_consecutive_count: 1
 # horizon_local_check is disabled until https://github.com/rcbops/u-suk-dev/issues/781
 # is resolved
 maas_excluded_checks:
-  - horizon_local_check
+  - 'horizon_local_check.*'
+
+# Disable the following MaaS alarms
+# Issue: https://github.com/rcbops/u-suk-dev/issues/1081
+maas_excluded_alarms:
+  - '^idle_percent_average.*'
+  - '^memory_used.*'
+  - '^alarm-network-receive.*'
+  - '^alarm-network-transmit.*'
+
+# Set overrides for check periods
+# Issue: https://github.com/rcbops/u-suk-dev/issues/1081
+maas_check_period_override:
+  disk_utilisation: 900
 
 # Set the threshold for filesystem monitoring when you are not specifying specific filesystems.
 maas_filesystem_warning_threshold: 80.0


BjoernT:
I'm on the fence with nova_cloud_memory_status. I would prefer to inform customers when they are about to hit quota rather them hitting that unattended. I don't think this alarm create any sizable false alerts.

npawelek:
@major I think that's fine, we can leave this one enabled for now. Some customers may find use and it's really not something that will continually re-alert for support. I've noticed in the past that it will sit in a critical status after it alerts. Many instances I've found that the checks are also disabled by support as there's nothing we can do. 

@BjoernT We probably need to define the proper mechanisms for "troubleshooting" these alerts and disseminate that to the team. Whether that be to escalate to the account team or otherwise.

npawelek:
@BjoernT On a side note. Do you think there would be use in having CPU idle percentage become load average? That was a push we made back in ER to help weed out false positives and continually re-alerting noise. I think with a proper threshold like 600/1200s with consecutiveCount=3 would probably be a good place to start.

major:
Sounds good. I'll remove `'^nova_cloud_memory_status.*'` and push a new patch.

major:
Any concerns with the current state of the patch, @BjoernT and @npawelek?

rpcd/etc/openstack_deploy/user_rpco_variables_defaults.yml
```diff
@@ -82,7 +82,20 @@ maas_alarm_remote_consecutive_count: 1
...
+maas_excluded_alarms:
+  - '^idle_percent_average.*'
+  - '^memory_used.*'
+  - '^alarm-network-receive.*'
+  - '^alarm-network-transmit.*'
+
+# Set overrides for check periods
+# Issue: https://github.com/rcbops/u-suk-dev/issues/1081
+maas_check_period_override:
+  disk_utilisation: 1800
```
npawelek: @BjoernT are you okay with this value? This will result in an alarm/ticket after 90 minutes with consecutiveCount=3.

major:
Nice catch on that, @npawelek. I forgot that 30 minutes x consecutiveCount would be 90 minutes. We can easily back that down if needed.

rpcd/etc/openstack_deploy/user_rpco_variables_defaults.yml
BjoernT: It depends in the end on how much free disk space you have vs how fast it is filling up. 90 minutes sounds pretty high currently we have 3 minutes. I would say let's do 900

major:
Checking in with @npawelek and @BjoernT prior to merging...

PR #1761, opened 2017-01-30
Avoid curl return '0' in 'Drop log_test1 template' task

Frank-ZhangXin:
As before, curl always return '0' when met HTTP 503 error, which
lead the task wouldn't retry times until dropping successfully.
By adding '-f' will let curl return '22' when doc doesn't exists,
and would trigger more retries.


Frank-ZhangXin:
wrong branch to commit

PR #1762, opened 2017-01-30
Add -f (--fail) flag to curl to return non-zero

Frank-ZhangXin:
Adding -f to the curl command will make curl return a non-zero
exit code when the DELETE request fails.  This will allow Ansible
to retry the task.

Connects https://github.com/rcbops/u-suk-dev/issues/1065

### DIFF
--- a/rpcd/playbooks/roles/elasticsearch/tasks/upgrade_elasticsearch_pre.yml
+++ b/rpcd/playbooks/roles/elasticsearch/tasks/upgrade_elasticsearch_pre.yml
@@ -108,7 +108,7 @@
     - reindex-wrapper
 
 - name: Drop log_test1 template
-  shell: 'curl -sk -XDELETE {{ internal_lb_vip_address }}:{{ elasticsearch_http_port }}/_template/log_test1'
+  shell: 'curl -fsk -XDELETE {{ internal_lb_vip_address }}:{{ elasticsearch_http_port }}/_template/log_test1'
   retries: 5
   tags:
     - logging-upgrade
--- a/rpcd/playbooks/roles/elasticsearch/tasks/upgrade_elasticsearch_pre.yml
+++ b/rpcd/playbooks/roles/elasticsearch/tasks/upgrade_elasticsearch_pre.yml
@@ -109,7 +109,12 @@
 
 - name: Drop log_test1 template
   shell: 'curl -fsk -XDELETE {{ internal_lb_vip_address }}:{{ elasticsearch_http_port }}/_template/log_test1'
-  retries: 5
+  args:
+     warn: no
+  register: curl_result
+  retries: 12
+  delay: 5
+  until: curl_result.rc == 0
   tags:
     - logging-upgrade
     - elasticsearch-upgrade


Frank-ZhangXin:
@major Added necessary components to the task, let's see what happens next.

hughsaunders:
recheck_upgrade failed on cloning ceph mon

Frank-ZhangXin:
It looks good to me with uri module and we've consider using it yesterday. As concern without breaking the uniform format of this script, we gave up changing the module.

Frank-ZhangXin:
@alextricity25 Actually the shell module succeed done the job:

-----------------------------------------------------------------
13:43:03 TASK: [elasticsearch | Drop log_test1 template] ******************************* 
13:43:13 changed: [jramup-42-13001_elasticsearch_container-c656e63d]
-----------------------------------------------------------------
13:43:16 TASK: [elasticsearch | Wait for elasticsearch port to close] ****************** 
13:43:16 ok: [jramup-42-13001_elasticsearch_container-c656e63d]
-----------------------------------------------------------------

alextricity25:
@Frank-ZhangXin Are you saying that you tried using the URI module yesterday, and it didn't work? I still prefer that we use the URI module where ever we can. I feel like it's more idiomatic Ansible, as suppose to just using shell with the curl command.

Frank-ZhangXin:
@alextricity25 hey Alex, what I said is 'shell' module with 'curl' can do the job which prevents elasticsearch rebuild fail. And I recommend this PR dedicated to fix our Mitaka upgrade bug of elasticsearch. Switching module to 'uri' is OK for me, and shell either can do the job, so which way to use depends on your preference.

Frank-ZhangXin:
@alextricity25 And one more thing, 12 times retry and 5 s delay is made by assumption which meant to give a large time period to let dropping complete. So it might need optimized later with shorter time.

cloader89:
@Frank-ZhangXin @alextricity25 It may be best to use curl for now, and then add a card to tech-debt to optimize the fix at a later time. This solution seems to be a reasonable fix since it is a blocker for other cards.

Frank-ZhangXin:
@cloader89 Could you specify the possibility to block other card? I personally intend to use uniform style and keep a consistent format of original code.

cloader89:
@Frank-ZhangXin I am not saying to block a card, rather perform a simple fix, and then get this merged so that other issues can continue. Then we put a card to optimize later.

rpcd/playbooks/roles/elasticsearch/tasks/upgrade_elasticsearch_pre.yml
```diff
@@ -108,13 +108,13 @@
     - reindex-wrapper
 
 - name: Drop log_test1 template
-  shell: 'curl -fsk -XDELETE {{ internal_lb_vip_address }}:{{ elasticsearch_http_port }}/_template/log_test1'
-  args:
-     warn: no
-  register: curl_result
+  uri:
+    url: "{{ internal_lb_vip_address }}:{{ elasticsearch_http_port }}/_template/log_test1"
+    method: DELETE
```
Frank-ZhangXin: Recommend to add 'return_content: yes', source: http://docs.ansible.com/ansible/uri_module.html

alextricity25:
@hughsaunders: The uri module for ansible 1.9 requires the httplib2 packages, which from the looks of it isn't installed by default.

PR #1763, opened 2017-01-31
[Backport - Newton] Make RPC Deployment optional

hughsaunders:
This allows deploy.sh to be used to prepare a deployment without
starting the deploy. This is necessary as Jenkins jobs are now being
divided into stages such as prepare and deploy, so that common stages
can be shared between jobs.

Connects rcbops/u-suk-dev#1069

(cherry picked from commit f463dd7fbda0b43dc77b351f24f9670464215c6d)


major:
recheck_xenial

PR #1767, opened 2017-01-31
MaaS: Move MaaS plugins into MaaS role

antonym:
Consolidates MaaS plugins into rpc_maas roles where they belong.

connects rcbops/u-suk-dev#1106


prometheanfire:
Doesn't rpc-r want them where they are now?  That's the reason it was not moved to it's own repo or the location you are proposing, I think.

nrb:
I think this looks fine for rpc-openstack's usage, but I seem to recall the RDO-based product consumed the plugins from where they were. This was so their puppet-based deploy didn't have to rely too much on an Ansible role.

However, I was not involved in writing the plugins as they are now, nor do I know who on the RDO team would the best source of information.

PR #1772, opened 2017-02-01
Removed an unnecessary requirement

cloudnull:
This removes the contributing guideline that requires development to
happen within the ``rpc-openstack`` repo which does nothing but limit
contributions.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -32,8 +32,8 @@
 ## Commits, release notes and pull requests<a name="commits-release-notes-pull-requests"></a>
 ### Fixing a bug<a name="fixing-a-bug"></a>
 
-1. After the bug is triaged and prioritised, make the required changes. New features, breaking changes and other patches of note must include a release note generated using the `reno` tool. Please see `Release Notes` for more information.
-2. Push changes directly to a branch on ```rpc-openstack``` in the ```rcbops``` github namespace (rather than a developer's own fork). A pull request is then made from this branch to the ```master``` branch.
+1. After the bug is triaged and prioritized, make the required changes. New features, breaking changes and other patches of note must include a release note generated using the `reno` tool. Please see `Release Notes` for more information.
+2. Make the needed changes following a git workflow that works best for you. [Example Branching workflow](https://git-scm.com/book/en/v2/Git-Branching-Branching-Workflows)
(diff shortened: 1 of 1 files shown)


odyssey4me:
@claco Surely this is a RPC engineering specific workflow. Being exclusive in this requirement completely excludes anyone that does not have the rights to branch in this repo, which means pretty much anyone outside of RPC engineering.

Considering the need to cater for other teams within RAX and possibly outside contributors I would think that this does need a revision and that perhaps the current documentation should be specific to RPC engineering only.

alextricity25:
We shouldn't tell external contributers that their contributions aren't valid until they follow our way(which requires write access to our repo). I'm okay with loosening the leash **for people outside our team**. If they want to use their own fork and make a PR off that, fine. They will have no other way of contributing otherwise until someone gives them write access to rpc-openstack.
However, as a team(rpc dev team), we need to have the understanding that we [**indeed have rules, processes, and regulations**](https://github.com/rcbops/u-suk-dev/blob/master/WORKFLOW.md) that need to be followed (i.e. make a branch within RPCO then make a PR off that branch so we can take over the work need be). So long as that is communicated internally, I'm okay with removing these bits from CONTRIBUTING.

claco:
Looking at the perms, everyone from the outside teams are added into contributors, w/ RW perms, so aside from specifically protected branches, anyone in these teams contributing shouldn't have any problems pushing a bug fix, or feature branch into the repo. In fact, to me, this doubles the need as an organization. Their team will need to solve the same problem, or worst case, we can move a PR of theirs along as well.

I'm open to a discussion around the change, but it doesn't appear we're chatted about this in email, nor is there an associated github issue that has gone through the triage/prioritization process for a wider audience to weigh in on.

```The PR description should reference the original issue.```

I'll leave it up to @alextricity25 on that front for what to do next.

alextricity25:
I don't think it's fair to ask external contributers(feature teams, or even one-off PRs from support to fix a small bug) to conform to our work flow.
Yes, we prefer that they do, but my hunch is that they won't care enough to use the "Connects" keyword on their PRs, or will most likely not know about it. They also won't care about our labels, waffle board, or u-suk-dev agendas.

Let's make it as easy as possible for external folks to contribute, but still hold our team accountable for the processes we have in place.

claco:
```Let's make it as easy as possible for external folks to contribute, _but still hold our team accountable for the processes we have in place_.```

I can get behind that sentiment. I'm here only to point out the irony that we just let our team push a PR, from a fork, without an issue.

PR #1773, opened 2017-02-01
 Set neutron_legacy_ha_tool_enabled to true by default

cloader89:
This will allow for neutron_legacy_ha_tool_enabled to be enabled
by default. This tool was providing HA capabilities for networks
and routers that were not using the native Neutron L3HA.

Connects rcbops/u-suk-dev#865


alextricity25:
@rcbops/support-feedback Please give us the sign off on this change to enable the legacy neutron ha tool in the Newton release.
Last I heard, we are still announcing the neutron l3ha failover feature as too unstable for production deployments. Please correct me if I'm wrong.

PR #1777, opened 2017-02-01
Switch keystone UUID to Fernet

Frank-ZhangXin:
Old UUID tonken would consume space of DB and cause authentication
 slow when heavy traffic access DB. Fernet tokens use shared
private keys to avoid having to store or replicate tokens in your
DB. This change will make authentication super fast and big plus
for keystone. And it's already in upstream.

Connects https://github.com/rcbops/u-suk-dev/issues/134

### DIFF
--- a/rpcd/etc/openstack_deploy/user_osa_variables_defaults.yml
+++ b/rpcd/etc/openstack_deploy/user_osa_variables_defaults.yml
@@ -69,7 +69,6 @@ rabbitmq_ulimit: 65535
 memcached_connections: 16384
 
 # Keystone overrides
-keystone_token_provider: "uuid"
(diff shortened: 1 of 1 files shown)


alextricity25:
@rcbops/support-feedback Please provide your sign off on this change to use the fernet token provider.
This was investigated and discussed in the past on these cards:
https://github.com/rcbops/u-suk-dev/issues/134
https://github.com/rcbops/u-suk-dev/issues/576

Frank-ZhangXin:
Note:
@odyssey4me commented "it's probably best to also remove the 'keystone_token_driver' setting too - right now it will have no effect, but things may change in the future"

PR #1785, opened 2017-02-02
Use new bootstrap role variables

alextricity25:
https://review.openstack.org/#/c/424810/1 introduced the
bootstrap_host_user_variables_filename and
bootstrap_host_user_secrets_filename variable that allow us
to specificy any naming convention for these variable files.

This commit utilizes these variables by setting them to the
naming convention used in RPCO

Connects https://github.com/rcbops/u-suk-dev/issues/1076

### DIFF
--- a/scripts/bootstrap-aio.yml
+++ b/scripts/bootstrap-aio.yml
@@ -36,6 +36,8 @@
 - name: Execute the OSA AIO bootstrap
   include: "{{ lookup('env', 'OA_DIR') }}/tests/bootstrap-aio.yml"
   vars:
+    bootstrap_host_user_variables_filename: "user_osa_variables_overrides.yml"
+    bootstrap_host_user_secrets_filename: "user_osa_secrets.yml"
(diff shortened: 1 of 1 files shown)


alextricity25:
@odyssey4me I believe this is all you had in mind when writing: https://github.com/rcbops/u-suk-dev/issues/1076, but please correct me if I'm wrong.

odyssey4me:
@alextricity25 That's one part of it. The other part is that we should be able to override the confid_overrides var https://github.com/openstack/openstack-ansible/blob/master/tests/bootstrap-aio.yml#L26-L48

That should allow us to set the confd files we want to deploy during the bootstrap phase, instead of removing/adding them afterwards.

Using this we can also introduce the use of the 'scenario' mechanism to implement different files for different scenario's. For example the Swift vs Ceph scenario.

Note that the conf file can have an extra attribute called 'path' to allow a completely different source path (for example a conf.d file from rpcd) to be used as a source template.

PR #1791, opened 2017-02-02
[Backport - Newton] MaaS: Move MaaS plugins into MaaS role

antonym:
Consolidates MaaS plugins into rpc_maas roles where they belong.

connects rcbops/u-suk-dev#1106


antonym:
recheck_ceph

PR #1795, opened 2017-02-03
Implement scenarios for RPCO AIOs

alextricity25:
This commit introduces the use of scenarios for RPCO AIO
deployments, much like how OSA is implmenting them here:
https://github.com/openstack/openstack-ansible/blob/master/tests/bootstrap-aio.yml#L26-L48

Note:
I had to remove a couple of not needed overrides in our
ceph.yml.aio file in order to make this work. These
variables carried no significant value in ceph.yml.aio
and caused variable scope issues when used with this new
scenario implementation

Connects https://github.com/rcbops/u-suk-dev/issues/1076

### DIFF
--- a/rpcd/etc/openstack_deploy/conf.d/ceph.yml.aio
+++ b/rpcd/etc/openstack_deploy/conf.d/ceph.yml.aio
@@ -25,5 +25,3 @@ storage_hosts:
           rados_connect_timeout: -1
           glance_api_version: 2
           volume_backend_name: ceph
-          rbd_user: "{{ cinder_ceph_client }}"
-          rbd_secret_uuid: "{{ cinder_ceph_client_uuid }}"
--- a/scripts/bootstrap-aio.yml
+++ b/scripts/bootstrap-aio.yml
@@ -36,8 +36,30 @@
 - name: Execute the OSA AIO bootstrap
   include: "{{ lookup('env', 'OA_DIR') }}/tests/bootstrap-aio.yml"
   vars:
+    scenario: "{{ lookup('env', 'DEPLOY_CEPH') | bool | ternary('ceph', 'swift') }}"
     bootstrap_host_user_variables_filename: "user_osa_variables_overrides.yml"
     bootstrap_host_user_secrets_filename: "user_osa_secrets.yml"
+    confd_overrides:
+      swift:
+        - name: cinder.yml.aio
+        - name: glance.yml.aio
+        - name: heat.yml.aio
+        - name: horizon.yml.aio
+        - name: keystone.yml.aio
+        - name: neutron.yml.aio
+        - name: nova.yml.aio
+        - name: swift.yml.aio
+      ceph:
+        - name: cinder.yml.aio
+        - name: glance.yml.aio
+        - name: heat.yml.aio
+        - name: horizon.yml.aio
+        - name: keystone.yml.aio
+        - name: neutron.yml.aio
+        - name: nova.yml.aio
+        - name: swift.yml.aio
+        - name: ceph.yml.aio
+          path: "{{ lookup('env', 'RPCD_DIR') ~ '/etc/openstack_deploy/conf.d' }}"
     openstack_user_config_overrides:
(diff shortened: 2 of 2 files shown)


scripts/bootstrap-aio.yml
```diff
@@ -36,6 +36,28 @@
 - name: Execute the OSA AIO bootstrap
   include: "{{ lookup('env', 'OA_DIR') }}/tests/bootstrap-aio.yml"
   vars:
```
alextricity25: Hmm..It seems as if the bootstrap role still uses the values specified in the OSA bootstrap-aio.yml playbook.
https://github.com/openstack/openstack-ansible/blob/stable/newton/tests/bootstrap-aio.yml#L25
@odyssey4me Where else could I specify these overrides so they take precedence?

scripts/bootstrap-aio.yml
```diff
@@ -36,6 +36,28 @@
...
+        - name: swift.yml.aio
+      ceph:
+        - name: cinder.yml.aio
+        - name: glance.yml.aio
+        - name: heat.yml.aio
+        - name: horizon.yml.aio
+        - name: keystone.yml.aio
+        - name: neutron.yml.aio
+        - name: nova.yml.aio
+        - name: swift.yml.aio
```
odyssey4me: Doesn't swift need to be removed from this one?

alextricity25: I thought so too, but it turns out that our ceph AIO jobs also deploy swift.

scripts/bootstrap-aio.yml
alextricity25: I've proposed an upstream change to move these variables to the ``vars`` section:
https://review.openstack.org/#/c/428820/1

rpcd/etc/openstack_deploy/conf.d/ceph.yml.aio
```diff
@@ -25,5 +25,3 @@ storage_hosts:
           rados_connect_timeout: -1
           glance_api_version: 2
           volume_backend_name: ceph
-          rbd_user: "{{ cinder_ceph_client }}"
```
odyssey4me: You could escape these: http://jinja.pocoo.org/docs/2.9/templates/#escaping

alextricity25: Actually, I prefer to remove them since they are essentially cruft at this point.

scripts/bootstrap-aio.yml
alextricity25: Done.

scripts/bootstrap-aio.yml
```diff
@@ -60,15 +82,12 @@
         src: "{{ rpco_cfg_src_path }}/{{ item.name }}"
         dest: "/etc/openstack_deploy/{{ item.name | regex_replace('.aio$', '') }}"
```
odyssey4me: not essential, but `` | regex_replace('.aio$', '')`` can come out now seeing as none of the files have that extension

scripts/bootstrap-aio.yml
```diff
@@ -36,6 +36,28 @@
...
+        - name: glance.yml.aio
+        - name: heat.yml.aio
+        - name: horizon.yml.aio
+        - name: keystone.yml.aio
+        - name: neutron.yml.aio
+        - name: nova.yml.aio
+        - name: swift.yml.aio
+        - name: ceph.yml.aio
+          path: "{{ lookup('env', 'RPCD_DIR') ~ '/etc/openstack_deploy/conf.d' }}"
+    scenario: "{{ lookup('env', 'DEPLOY_CEPH') | bool | ternary('ceph', 'swift') }}"
```
odyssey4me: if lines 60-62 were at line 39 (ie if the non dict vars were above the dict vars) it would be slightly easier to read

alextricity25:
recheck_all

alextricity25:
recheck_all

alextricity25:
recheck_swift

PR #1806, opened 2017-02-07
Ensure the apt artifact build derives the version

odyssey4me:
Currently the apt artifact build process uses a static
version. This patch ensures that it uses the appropriate
script to derive the artifact version in the same way
as the python artifact building job.

Connects https://github.com/rcbops/u-suk-dev/issues/1133

### DIFF
--- a/scripts/artifacts-building/apt/artifacting-vars.yml
+++ b/scripts/artifacts-building/apt/artifacting-vars.yml
@@ -20,10 +20,9 @@ artifact_extradownloads_dest_folder: "{{ artifacts_root_folder }}/downloads"
 ansible_roles_folder: "/etc/ansible/roles"
 
 # Generic details
-rpc_branch: "master"
-osa_branch: 'stable/newton' #can be automated later by lookup
-roles_details: "{{ lookup('url','https://raw.githubusercontent.com/openstack/openstack-ansible/'+osa_branch+'/ansible-role-requirements.yml',wantlist=True) | join('\n') | from_yaml  }}"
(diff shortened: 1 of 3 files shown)


scripts/artifacts-building/apt/artifacting-vars.yml
```diff
@@ -20,10 +20,9 @@ artifact_extradownloads_dest_folder: "{{ artifacts_root_folder }}/downloads"
 ansible_roles_folder: "/etc/ansible/roles"
 
 # Generic details
-rpc_branch: "master"
-osa_branch: 'stable/newton' #can be automated later by lookup
-roles_details: "{{ lookup('url','https://raw.githubusercontent.com/openstack/openstack-ansible/'+osa_branch+'/ansible-role-requirements.yml',wantlist=True) | join('\n') | from_yaml  }}"
-artifacts_version: "rpc-14.0.0rc1" #Move to reading variable from here: "https://github.com/rcbops/rpc-openstack/blob/master/rpcd/playbooks/group_vars/all.yml"
+role_requirements_file: "{{ lookup('env', 'BASE_DIR') ~ 'openstack-ansible/ansible-role-requirements.yml') }}"
```
evrardjp: /openstack-ansible ?

scripts/artifacts-building/apt/artifacting-vars.yml
```diff
@@ -20,10 +20,9 @@ artifact_extradownloads_dest_folder: "{{ artifacts_root_folder }}/downloads"
...
+roles_details: "{{ lookup('file', role_requirements_file, wantlist=True) | join('\n') | from_yaml }}"
```
evrardjp: the lookup will probably behave differently, so it's probably worth a rewrite.

Maybe lookup('pipe', 'cat ' ~ ...)

Issue #1790, opened 2017-02-02
RabbitMQ should be upgraded to 3.6.6 or greater, Erlang upgraded to 19.2 or greater


brewc:
##### On which version(s) do you encounter this issue?
Liberty +

###### Please provide the output to 'git describe --all' for  all deployments where you have confirmed it occurs.
r12.2.8
r12.2.2

Only change needed would be to update defaults/main.yml for rabbit on these fields:
rabbitmq_package_url: ""
rabbitmq_package_sha256: ""

##### What type of issue is this? Please tick the appropriate box.
- [X ] bug - existing functionality that does not work for you as documented
- [ ] feature - new functionality that you would like to see added

##### What is the expected behavior?
RabbitMQ automagically recovers from a partitioning event.

##### What is actually happening?
RabbitMQ is not consistently recovering from partition events.

##### What were you attempting to do when you encountered this error (give as much detail as possible)? 
This error can occur during any partition event within RabbitMQ.  It is usually evident under high load (large message ingestion)

##### How can we reproduce this issue?
Within a RPC-O environment, simulate a partition event by blocking all outbound traffic from one rabbit node to another.  This will force the cluster to see the node as "down".

node1 - ip : 10.10.0.1
node2 - ip : 10.10.0.2
node3 - ip : 10.10.0.3

on node2:
iptables -A OUTPUT -d 10.10.0.1 -j DROP
rabbitmqctl cluster_status <see alarm and node down>

to recover:
Within same environment used to reproduce the issue, remove the iptables rules on node2 and ensure the cluster recovers.
on node2:
iptables -F
rabbitmqctl cluster_status <see no alarms and all nodes reported?

##### How can we verify your issue is addressed?
Upgrade RabbitMQ to 3.6.6 or later, and Erlang to 19.2 or later.  Repeat steps to cause a partition plus steps to resolve.  Note in logs that RabbitMQ handled the partition event and restored the cluster

##### If you are using or know of a workaround please describe the necessary steps?
Only resolution is a manual restart of rabbit services on the affected node.

##### What is your evaluation on the importance of this issue, and why?
URGENT.  The current version of rabbit in Liberty (3.5.7) does not handle partition events well, consistently resulting in some nodes never being brought online and a degradation of service.

----------------------------------------------------------------------------------------------------------------------------
###### For each pull request, update the branch line to reference it.
- [ ] master
- [ ] newton-14.0
- [ ] mitaka-13.1
- [ ] liberty-12.2

[...]

PR #1816, opened 2017-02-07
Bump RabbitMQ version to 3.6.6-1

d34dh0r53:
Due to some performance and clustering issues we have determined that
RabbitMQ should be at version 3.6.6-1 at a minimum. This overrides the
rabbitmq_package_url and the rabbitmq_package_sha256 variables within
RPC-O.

Connects #1790


odyssey4me:
@d34dh0r53 Is there a reason why this hasn't been done upstream?

odyssey4me:
Pushed review upstream: https://review.openstack.org/430604

PR #1817, opened 2017-02-08
Change default console to novnc

alextricity25:

* Add ``nova_console_type: novnc`` into the osa default variable file
* Make the necessary changes in maas for the console check to support
  both novnc and spice
* Change the rpc_support role to conditionally include the spice-specific
  bit
* Add an f5 'pool part' in the f5-config.py script for the novnc console
  type

Connects https://github.com/rcbops/u-suk-dev/issues/131


alextricity25:
recheck_all

alextricity25:
recheck_all

alextricity25:
I tested this with MaaS in my own AIO, but I've also fired up some maas jobs to make sure this check is implemented correctly:
http://jenkins.propter.net/job/JJB-RPC-AIO_master-swift-periodic/91/
http://jenkins.propter.net/job/JJB-RPC-AIO_master-xenial-periodic/

PR #1833, opened 2017-02-09
[Kilo] Disable MaaS alarms/checks with regex

major:

(cherry picked from commit d9ec02cd4f941558725892293a3e44130f0813d2)


major:
Tests fine in Kilo AIO. Two small adjustments needed to be made in galera checks (master had a bunch more) as well as the swift proxy check rename.

PR #1834, opened 2017-02-09
[Kilo] MaaS template fixes

major:
This commit makes the following template updates:

- in ceph_osd_stats.yaml.j2, osd_id is cast to a string otherwise
  ansible will throw a template error since it cannot contact an int
  and a string
- in rabbitmq_status.yaml.j2, we update the rabbitmq_mem_alarm_status
  alarm to reference rabbitmq_mem_alarm_status rathe than
  rabbitmq_disk_free_alarm_status

(cherry picked from commit 70a1538418004b59262729809caa4cad0bae8c01)


major:
Works fine in Kilo AIO.

PR #1836, opened 2017-02-09
MaaS: Improve period/timeout/disabling checks

major:
This patch adds two new variables that allow deployers to customize
the check period and the check timeout for each MaaS check:

    maas_check_period_override:
      - disk_utilization: 60
      - conntrack_count: 45
    maas_check_timeout_override:
      - cinder_backup_check: 120

The patch makes some adjustments to templates to improve ease of use:

- the check label is set as a variable so that it can be re-used
  in multiple places within the template
- the check period is checked against the list of overrides
- the check timeout is checked against the list of overrides
- the full check name is checked against the list of excluded checks

If a check is disabled, the YAML is left in place on the server, but
the `disabled` option is set to `true`.

(cherry picked from commit a987481baeaf5f7a247a1a18e338470ddce8755f)


major:
Works fine in Kilo AIO.

PR #1837, opened 2017-02-09
[Kilo] MaaS plugins relocation + HP checks

major:


major:
Tests fine in Kilo AIO.

PR #1857, opened 2017-02-10
fix generated f5 config to actually work

prometheanfire:
Without the missing line the two rules are joined and there is a
dangling ".


prometheanfire:
@alextricity25 you touched this last, so your ack would be nice

Frank-ZhangXin:
#1861 Backport to Newton

PR #1859, opened 2017-02-13
MaaS: Remove excluded checks logic from tasks

major:
Checks are now excluded using the `disabled` attribute in the YAML
file for the check (see #1691). The original conditionals in each
MaaS task is no longer needed.

Connects rcbops/u-suk-dev#1224


major:
This will need to be backported through Kilo.

major:
recheck_swift

PR #1860, opened 2017-02-13
Re-add ceph variables in cinder_backends

alextricity25:
These variables were incorrectly removed as part of
e80a7a5
This commit re-adds the rdb_* variables using the lineinfile ansible module.
This is because the bootstrap role runs this
file through config_template, which attempts to resolve the ceph
variables. However, since the variables are not avaialbe when the
bootstrap role is invoked, it will fail. Adding these variables
via lineinfile makes it possible to write out raw jinja, whereas
using config_template makes this impossible.

Connects https://github.com/rcbops/u-suk-dev/issues/1188

### DIFF
--- a/rpcd/etc/openstack_deploy/conf.d/ceph.yml.aio
+++ b/rpcd/etc/openstack_deploy/conf.d/ceph.yml.aio
@@ -8,20 +8,3 @@ osds_hosts:
     ip: 172.29.236.100
     affinity:
       ceph_osd_container: 3
-
-storage_hosts:
-  aio1:
-    ip: 172.29.236.100
-    container_vars:
-      cinder_backends:
-        limit_container_types: cinder_volume
-        ceph:
-          volume_driver: cinder.volume.drivers.rbd.RBDDriver
-          rbd_pool: volumes
-          rbd_ceph_conf: /etc/ceph/ceph.conf
-          rbd_flatten_volume_from_snapshot: 'false'
-          rbd_max_clone_depth: 5
-          rbd_store_chunk_size: 4
-          rados_connect_timeout: -1
-          glance_api_version: 2
-          volume_backend_name: ceph
--- a/scripts/bootstrap-aio.yml
+++ b/scripts/bootstrap-aio.yml
@@ -72,6 +72,33 @@
   hosts: localhost
   user: root
   tasks:
+    # We need to write out the cinder_backend block here, so we avoid
+    # running this block through the config_template module when the
+    # upstream bootstrap role puts the conf.d files into place.
+    - name: Write out storage hosts and cinder backends to conf.d/ceph.yml
+      lineinfile:
+        name: "/etc/openstack_deploy/conf.d/ceph.yml"
+        line: |
+          storage_hosts:
+            aio1:
+              ip: 172.29.236.100
+              container_vars:
+                cinder_backends:
+                  limit_container_types: cinder_volume
+                  ceph:
+                    volume_driver: cinder.volume.drivers.rbd.RBDDriver
+                    rbd_pool: volumes
+                    rbd_ceph_conf: /etc/ceph/ceph.conf
+                    rbd_flatten_volume_from_snapshot: 'false'
+                    rbd_max_clone_depth: 5
+                    rbd_store_chunk_size: 4
+                    rados_connect_timeout: -1
+                    glance_api_version: 2
+                    volume_backend_name: ceph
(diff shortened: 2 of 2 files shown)


alextricity25:
@mattt416 Yes. The nova config template doesn't use the same kind of mechanism the cinder conf file template uses to populate the ``rbd_secret_uuid`` directive. [It explicitly defines ``rbd_secrete_uuid``](https://github.com/openstack/openstack-ansible-os_nova/blob/stable/newton/templates/nova.conf.j2#L280-L282), unlike the cinder config file.The cinder config template iterates through the user-defined attributes in cinder_backends, [then builds directives from that](https://github.com/openstack/openstack-ansible-os_cinder/blob/stable/newton/templates/cinder.conf.j2#L100-L103). That's why it's necessary to define ``rbd_user`` here in cinder_backends.
In addition, ``rbd_secret_uuid`` isn't listed as a directive in the [cinder configuration reference](http://docs.openstack.org/newton/config-reference/image/backends.html), so I believe it's safe to remove it.

> Is it absolutely necessary to run these files through config_template? What is the benefit of doing this?

That's something that the upstream bootstrap role does [here](https://github.com/openstack/openstack-ansible/blob/stable/newton/tests/roles/bootstrap-host/tasks/prepare_aio_config.yml#L28-L36). If we continue supplying this role with the list of conf.d files, as we do [here](https://github.com/rcbops/rpc-openstack/blob/master/scripts/bootstrap-aio.yml#L42), then we can't avoid it. The idea here was to move to the same model OSA is using for AIO bootstrap deployments.
I wish there was a way to tell it not to interpolate these variables, but I couldn't find a way. I tried jinja escaping, using ``raw`` blocks, and different variables of single quotes and double quotes.

alextricity25:
UPDATE: I'm testing this now on my AIO just to make sure the ceph boot volume backend still functions correctly with ``rbd_secret_uuid`` removed from ``ceph.yml.aio``.

I've also created a couple of newton+ceph periodic jobs off this branch:
http://jenkins.propter.net/job/JJB-RPC-AIO_newton-ceph-periodic/63/
http://jenkins.propter.net/job/JJB-RPC-AIO_newton-ceph-periodic/62/

mattt416:
@alextricity25, it looks like `rbd_secret_uuid` is a config option for cinder.  See https://github.com/openstack/cinder/blob/stable/newton/cinder/volume/drivers/rbd.py#L65-L67.

alextricity25:
maas job: http://jenkins.propter.net/job/JJB-RPC-AIO_newton-ceph-periodic/64/

alextricity25:
new maas job: http://jenkins.propter.net/job/JJB-RPC-AIO_newton-ceph-periodic/66/

rpcd/etc/openstack_deploy/conf.d/ceph.yml.aio
```diff
@@ -8,20 +8,3 @@ osds_hosts:
...
-        ceph:
-          volume_driver: cinder.volume.drivers.rbd.RBDDriver
-          rbd_pool: volumes
-          rbd_ceph_conf: /etc/ceph/ceph.conf
-          rbd_flatten_volume_from_snapshot: 'false'
-          rbd_max_clone_depth: 5
-          rbd_store_chunk_size: 4
-          rados_connect_timeout: -1
-          glance_api_version: 2
-          volume_backend_name: ceph
```
mattt416: Is there any point carrying half the file?  Would it be better to just embed the whole lot in bootstrap-aio.yml?

alextricity25: Well, I wanted to keep the ceph.yml file under the confd_overrides dictionary here: https://github.com/rcbops/rpc-openstack/blob/master/scripts/bootstrap-aio.yml#L61 so that we keep this "scenario" structure. If you think that's silly, then I could write out the entire file in a task.

alextricity25:
recheck_swift

mattt416:
@alextricity25 nice, all passed!  The last question I have is ... we define `storage_hosts` in `conf.d/ceph.yml` (after it gets appended) but it also gets defined in `conf.d/cinder.yml` (from openstack-ansible).  Is that going to be a problem?

nrb:
This has been merged, but https://github.com/openstack/openstack-ansible/blob/stable/newton/playbooks/inventory/dynamic_inventory.py#L982-L990 is where ``conf.d`` entries are processed. They should merge under the key, and in the case of a key collision with different values, the last one read will win.

That said, I don't think ``os.walkdir`` provides a consistent ordering, so we may have to fix that upstream, which is unrelated to this change.

PR #1865, opened 2017-02-14
MaaS: Add Service Checks for Swift Processes

antonym:
Adds service checks for Swift processes.  Will add swift-object-expirer (mitaka+) as a separate commit as it will not need to be backported to kilo and liberty.

Connects rcbops/u-suk-dev#1215

Depends on https://github.com/rcbops/rpc-openstack/pull/1869 being merged.


antonym:
@npawelek Do those swift groupings look correct for a normal deployment?  Just wanted to make sure the processes checks are on the right containers.

PR #1868, opened 2017-02-14
Add git and python artifact staging

odyssey4me:
This patch adds git and python artifact staging as an option
for rpc-openstack.

Connects https://github.com/rcbops/u-suk-dev/issues/1235

### DIFF
--- a/rpcd/playbooks/stage-python-artifacts.yml
+++ b/rpcd/playbooks/stage-python-artifacts.yml
@@ -1,0 +1,193 @@
+---
+# Copyright 2017, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+- name: Stage the git and python artifacts
+  hosts: "{{ staging_host | default('localhost') }}"
+  user: root
+  vars:
+    debug: no
+    rpco_mirror_base_url: "https://rpc-repo.rackspace.com"
+    http_proxy_env: "{{ lookup('env', 'http_proxy') | default('not_set', true) }}"
+    https_proxy_env: "{{ lookup('env', 'https_proxy') | default('not_set', true) }}"
+    https_validate_certs: yes
+  tasks:
+
+    - name: Set the staging path
+      set_fact:
+        staging_path: |-
+          {%- if (groups['repo_all'] is defined) and (groups['repo_all'] | length > 0) -%}
+          /openstack/{{ hostvars[groups['repo_all'][0]]['inventory_hostname'] }}/repo
+          {%- else -%}
+          /openstack/stage
+          {%- endif -%}
+      tags:
+        - always
+
+    - name: Fetch the upstream manifest file
+      uri:
+        url: "{{ rpco_mirror_base_url }}/os-releases/{{ rpc_release }}/MANIFEST.in"
+        return_content: yes
+        validate_certs: "{{ https_validate_certs | bool }}"
+      register: manifest
+      tags:
+        - always
+
+    # TODO(odyssey4me)
+    # Remove this and adjust the fact setting once
+    # https://review.openstack.org/433152 is available
+    - name: Fetch the upstream venv list
+      uri:
+        url: "{{ rpco_mirror_base_url }}/venvs/{{ rpc_release }}/ubuntu/"
+        return_content: yes
+        validate_certs: "{{ https_validate_certs | bool }}"
+      register: venv_index
+      tags:
+        - always
+
+    - name: Derive the lists of artifacts to download
+      set_fact:
+        python_wheel_list: |
+          {%- set content_list = manifest.content.split('\n') -%}
+          {%- set result_list = [] -%}
+          {%- for item in content_list -%}
+          {%-   if item.split('/')[0] == 'pools' -%}
+          {%-     set _ = result_list.append(item) -%}
+          {%-   endif -%}
+          {%- endfor -%}
+          {{- result_list -}}
+        git_folder_list: |
+          {%- set content_list = manifest.content.split('\n') -%}
+          {%- set result_list = [] -%}
+          {%- for item in content_list -%}
+          {%-   if item.split('/')[0] == 'openstackgit' -%}
+          {%-     set _ = result_list.append(item) -%}
+          {%-   endif -%}
+          {%- endfor -%}
+          {{- result_list -}}
+        python_venv_list: |
+          {%- set content_list = venv_index.content.split('\r\n') -%}
+          {%- set result_list = [] -%}
+          {%- for item in content_list -%}
+          {%-   if item | search("<a href.*>.*</a>") -%}
+          {%-     set item_clean = item | regex_replace(".*<a href.*>(.*)</a>.*", "\\1") -%}
+          {%-     if item_clean != "../" -%}
+          {%-       set _ = result_list.append(item_clean) -%}
+          {%-     endif -%}
+          {%-   endif -%}
+          {%- endfor -%}
+          {{- result_list -}}
+      tags:
+        - always
+
+    - name: Staging folders setup
+      file:
+        path: "{{ item }}"
+        state: "directory"
+      with_items:
+        - "{{ staging_path }}/links"
+        - "{{ staging_path }}/openstackgit"
+        - "{{ staging_path }}/os-releases/{{ rpc_release }}"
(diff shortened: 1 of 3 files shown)


rpcd/playbooks/stage-python-artifacts.yml
```diff
@@ -0,0 +1,184 @@
...
+    - name: Fetch the upstream manifest file
+      uri:
+        url: "{{ upstream_artifact_base_url }}/os-releases/{{ rpc_release }}/MANIFEST.in"
+        return_content: yes
+        validate_certs: yes
+      register: manifest
+      tags:
+        - always
+
+    - name: Fetch the upstream venv list
```
cloudnull: With recent changes in the repo-build process found here: [ https://review.openstack.org/#/c/433152/ ] this should no longer be a needed task.

rpcd/playbooks/stage-python-artifacts.yml
```diff
@@ -0,0 +1,184 @@
...
+        git_folder_list: |
+          {%- set content_list = manifest.content.split('\n') -%}
+          {%- set result_list = [] -%}
+          {%- for item in content_list -%}
+          {%-   if item.split('/')[0] == 'openstackgit' -%}
+          {%-     set _ = result_list.append(item) -%}
+          {%-   endif -%}
+          {%- endfor -%}
+          {{- result_list -}}
+        python_venv_list: |
```
cloudnull: Same as above. This should not be needed once the venvs in the manifest file is merged.

rpcd/playbooks/stage-python-artifacts.yml
```diff
@@ -0,0 +1,184 @@
...
+
+    - name: Clone git repositories asynchronously
+      git:
+        repo: "{{ upstream_artifact_base_url }}/{{ item }}"
+        dest: "{{ staging_path }}/{{ item }}"
+        version: master
+        force: yes
+      with_items: "{{ git_folder_list }}"
+      register: _git_clone
+      async: 1800
```
cloudnull: this should be fine should be move to ansible 2.1+ I dont believe async w/ 1.9 works. This is no blocker but something we should be aware of.

rpcd/playbooks/stage-python-artifacts.yml
```diff
@@ -0,0 +1,184 @@
...
+      poll: 0
+      tags:
+        - git
+
+    - name: Wait for git clones to complete
+      async_status:
```
cloudnull: same as above.

rpcd/playbooks/stage-python-artifacts.yml
odyssey4me: @cloudnull Sure, although that patch is targeted at master whereas this is targeting Newton. If that patch merges and is backported, then a subsequent patch can remove this.

rpcd/playbooks/stage-python-artifacts.yml
odyssey4me: Newton is running Ansible 2.1.x so we're OK here.

rpcd/playbooks/stage-python-artifacts.yml
evrardjp: @odyssey4me I'd be happy to see @cloudnull patch backported to master.
If you are waiting for this, please mark this with a `#TODO(odyssey4me): Remove this when  https://review.openstack.org/#/c/433152/  is backported to N`

rpcd/playbooks/stage-python-artifacts.yml
```diff
@@ -0,0 +1,184 @@
...
+          {%-     if item_clean != "../" -%}
+          {%-       set _ = result_list.append(item_clean) -%}
+          {%-     endif -%}
+          {%-   endif -%}
+          {%- endfor -%}
+          {{- result_list -}}
+      tags:
+        - always
+
+    - name: Git staging folders setup
```
evrardjp: "staging folders setup" instead of "Git staging folders setup"

rpcd/playbooks/stage-python-artifacts.yml
```diff
@@ -0,0 +1,184 @@
...
+      file:
+        path: "{{ item }}"
+        state: "directory"
+      with_items:
+        - "{{ staging_path }}/links"
+        - "{{ staging_path }}/openstackgit"
+        - "{{ staging_path }}/os-releases/{{ rpc_release }}"
+        - "{{ staging_path }}/venvs/{{ rpc_release }}/ubuntu"
```
evrardjp: Worth adding the lxc-container images?

rpcd/playbooks/stage-python-artifacts.yml
```diff
@@ -0,0 +1,184 @@
...
+      tags:
+        - always
+
+    - name: Clone git repositories asynchronously
+      git:
+        repo: "{{ upstream_artifact_base_url }}/{{ item }}"
+        dest: "{{ staging_path }}/{{ item }}"
+        version: master
```
evrardjp: version: master? Shouldn't we remove this line?

rpcd/playbooks/stage-python-artifacts.yml
```diff
@@ -0,0 +1,184 @@
...
+        force: yes
+      with_items: "{{ git_folder_list }}"
+      register: _git_clone
+      async: 1800
+      poll: 0
+      tags:
+        - git
+
+    - name: Wait for git clones to complete
```
evrardjp: Why do you want to wait for git clones here? Just do it as last task?

evrardjp:
I have not tested it 'though.

rpcd/playbooks/stage-python-artifacts.yml
```diff
@@ -0,0 +1,184 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+- name: Stage the git and python artifacts
+  hosts: "{{ staging_host | default('localhost') }}"
+  user: root
+  vars:
+    upstream_artifact_base_url: "{{ rpco_mirror_base_url | default('https://rpc-repo.rackspace.com') }}"
```
odyssey4me: The intent with this playbook is to be able to execute it regardless of whether any configuration has been bootstrapped. This allows the staging to be done, for example, onto a laptop or something. That's why I set it to use ``rpco_mirror_base_url`` if it's available, but provide a default. Once the artifacts are staged to a laptop/USB key the expectation is that the deployer would copy the contents over to the target using rsync or whatever.

rpcd/playbooks/stage-python-artifacts.yml
```diff
@@ -0,0 +1,184 @@
...
+    http_proxy_env: "{{ lookup('env', 'http_proxy') | default('not_set', true) }}"
+    https_proxy_env: "{{ lookup('env', 'https_proxy') | default('not_set', true) }}"
+  tasks:
+
+    - name: Set the staging path
+      set_fact:
+        staging_path: |-
+          {%- if groups['repo_all'] is defined -%}
```
odyssey4me: Sure, I can do that.

rpcd/playbooks/stage-python-artifacts.yml
```diff
@@ -0,0 +1,184 @@
...
+          /openstack/stage
+          {%- endif -%}
+      tags:
+        - always
+
+    - name: Fetch the upstream manifest file
+      uri:
+        url: "{{ upstream_artifact_base_url }}/os-releases/{{ rpc_release }}/MANIFEST.in"
+        return_content: yes
+        validate_certs: yes
```
odyssey4me: I was considering this, but thought of doing it later. I'll work up a change similar to this.

rpcd/playbooks/stage-python-artifacts.yml
odyssey4me: In a subsequent patch, yes.

rpcd/playbooks/stage-python-artifacts.yml
odyssey4me: No, it must just pull the git repositories as they are. The git checkout to the right SHA happens during the repo build. The only git repositories that will actually be used are the novnc/spice repositories anyway - and maybe some of the RPC ones like holland.

rpcd/playbooks/stage-python-artifacts.yml
odyssey4me: Sure, will move it.

rpcd/playbooks/stage-python-artifacts.yml
evrardjp: By setting version: master you are NOT pulling as they are, you are forcing the checkout master.

Remove that and it will just use HEAD.

evrardjp:
See my comment on master.

rpcd/playbooks/stage-python-artifacts.yml
odyssey4me: Ah ok, I'll adjust now.

odyssey4me: Unfortunately not setting a version, or setting it to HEAD causes ansible-lint to fail. Ultimately the repo checkout doesn't matter because the only tasks that use these git sources after the wheels are built are the novnc/spice tasks and they do their own version checkout.

PR #1877, opened 2017-02-16
[Mitaka] MaaS: Updated process.yml

major:
This patch updates process.yml to work well for Xenial and Trusty
and makes use of the existing psutil and lxc python modules.

Connects rcbops/u-suk-dev#1233
(cherry picked from commit 44453e475f0d3486d27947750dadc7c6df201260)


antonym:
recheck_ceph

PR #1880, opened 2017-02-16
Add apt artifact staging

odyssey4me:
This patch adds apt mirror staging as an option for
rpc-openstack.

Connects https://github.com/rcbops/u-suk-dev/issues/1234

### DIFF
--- a/rpcd/playbooks/stage-apt-artifacts.yml
+++ b/rpcd/playbooks/stage-apt-artifacts.yml
@@ -1,0 +1,88 @@
+---
+# Copyright 2017, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


rpcd/playbooks/stage-apt-artifacts.yml
```diff
@@ -0,0 +1,88 @@
...
+        name: "apt-mirror"
+        state: present
+
+    - name: Write the mirror configuration
+      template:
+        src: "stage-apt-artifacts-mirror.list.j2"
+        dest: "/etc/apt/mirror.list"
+
+    - name: Execute apt-mirror
+      command: "apt-mirror"
```
evrardjp: nit: You should probably use full path here.

I'm no expert of apt-mirror cli and its return code, could you tell me more about it?

odyssey4me: What benefit does using a full path give me?

What would you like to know about the return code? I expect that it's 0 for success and 1 for failure, much like most other scripts.

evrardjp:
👍

cloudnull:
Looks good, it'd be great to get a few reps on this to find out the nasty bits but otherwise LGTM at the moment.

nrb:
Merged already, but the playbook made sense mentally stepping through it.

PR #1882, opened 2017-02-16
Add container artifact staging

odyssey4me:
This patch adds container artifact staging as an option
for rpc-openstack.

Connects https://github.com/rcbops/u-suk-dev/issues/1250

### DIFF
--- a/rpcd/playbooks/stage-container-artifacts.yml
+++ b/rpcd/playbooks/stage-container-artifacts.yml
@@ -1,0 +1,111 @@
+---
+# Copyright 2017, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+- name: Stage the container artifacts
+  hosts: "{{ staging_host | default('localhost') }}"
+  user: root
+  vars:
+    aria_input_file: "/tmp/container_artifact_list.txt"
+    rpco_mirror_base_url: "https://rpc-repo.rackspace.com"
+    http_proxy_env: "{{ lookup('env', 'http_proxy') | default('not_set', true) }}"
+    https_proxy_env: "{{ lookup('env', 'https_proxy') | default('not_set', true) }}"
+    https_validate_certs: yes
+  tasks:
+
+    - name: Set the staging path
+      set_fact:
+        staging_path: |-
+          {%- if (groups['repo_all'] is defined) and (groups['repo_all'] | length > 0) -%}
+          /openstack/{{ hostvars[groups['repo_all'][0]]['inventory_hostname'] }}/repo
+          {%- else -%}
+          /openstack/stage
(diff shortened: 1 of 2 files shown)


rpcd/playbooks/stage-container-artifacts.yml
```diff
@@ -0,0 +1,119 @@
...
+          {%- set result_list = [] -%}
+          {%- for item in content_list -%}
+          {%-   if item | search('-' ~ rpc_release  ~ ';') -%}
+          {%-     set _ = result_list.append(item) -%}
+          {%-   endif -%}
+          {%- endfor -%}
+          {{- result_list -}}
+      tags:
+        - always
+
```
evrardjp: Or the more concise way:

```
    - name: set list
      set_fact:
        lxc_cache_index_list: "{{ lxc_cache_index_list | default([]) }} + [ '{{ item }}' ] "
      when: item.find(rpc_release) != -1
      with_url:
        - http://rpc-repo.rackspace.com/meta/1.0/index-system

```

rpcd/playbooks/stage-container-artifacts.yml
```diff
@@ -0,0 +1,119 @@
...
+        state: present
+      tags:
+        - python
+
+    - name: Write container artifact URL list
+      copy:
+        content: |
+          {% for item in lxc_cache_index_list %}
+          {% set item_path = item.split(';') | last %}
+          {{ rpco_mirror_base_url }}/{{ item_path }}/SHA256SUMS
```
evrardjp: you could have thrown another loop there but that's really nitpicking.
Don't edit :p

rpcd/playbooks/stage-container-artifacts.yml
```diff
@@ -0,0 +1,119 @@
...
+      tags:
+        - python
+
+    - name: Download container artifacts
+      command: >-
+        aria2c --input-file={{ staging_path }}/container_artifact_list.txt
+        --allow-overwrite=true
+        --conditional-get=true
+        --quiet
+        --check-certificate={{ (https_validate_certs | bool) | lower }}
```
evrardjp: Are you sure you want | bool | lower?

rpcd/playbooks/stage-container-artifacts.yml
```diff
@@ -0,0 +1,119 @@
...
+      lineinfile:
+        line: "{{ item }}"
+        dest: "{{ staging_path }}/meta/1.0/index-system"
+        create: yes
+        state: present
+      with_items: "{{ lxc_cache_index_list }}"
+      tags:
+        - python
+
+    - name: Remove container artifact URL list
```
evrardjp: You don't need to do this task, as you ensure content is always set to what you want with the copy module.
This breaks idempotency for no real thing.

odyssey4me: The intent is to remove it from the repo so that it's not served. I guess I could change the location to some sort of temp location instead.

rpcd/playbooks/stage-container-artifacts.yml
odyssey4me: Yes, aria's documentation shows that it uses the values of 'true' or 'false', so I'm ensuring that's what's put there.

rpcd/playbooks/stage-container-artifacts.yml
odyssey4me: Yeah, I know. I got lazy. In a way this is more readable, but meh.

rpcd/playbooks/stage-container-artifacts.yml
odyssey4me: While this is more concise, I find it less readable. Also I don't see how to disable validate_certs with this method.

PR #1884, opened 2017-02-16
[Mitaka] MaaS: Remove excluded checks from tasks

major:
Checks are now excluded using the `disabled` attribute in the YAML
file for the check (see #1691). The original conditionals in each
MaaS task is no longer needed.

Connects rcbops/u-suk-dev#1224

(cherry picked from commit d89652448e134d0f627829180f57b5106cf27273)


antonym:
recheck_ceph
recheck_upgrade

PR #1888, opened 2017-02-16
MaaS: Restrict swift-recon checks to two proxies

antonym:
Restrict swift-recon checks to proxy hosts.

Connects rcbops/u-suk-dev#1212

### DIFF
--- a/rpcd/playbooks/roles/rpc_maas/defaults/main.yml
+++ b/rpcd/playbooks/roles/rpc_maas/defaults/main.yml
@@ -423,16 +423,18 @@ cinder_backup_checks_list:
   - { name: "cinder_backup_check", group: "cinder_backup" }
 
 swift_checks_list:
-  - { name: "swift_object_server_check", group: "swift_obj" }
   - { name: "swift_account_server_check", group: "swift_acc" }
   - { name: "swift_container_server_check", group: "swift_cont" }
+  - { name: "swift_object_server_check", group: "swift_obj" }
+
+swift_recon_checks_list:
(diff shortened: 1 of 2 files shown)


antonym:
@npawelek Do the swift-recon checks also still need to be restricted to just two swift proxy hosts?  If so we might need to just deploy to the first two proxy servers if that's acceptable.

npawelek:
@antonym: Sorry for the delay. Yes, the restriction to limit the nodes is due to the all-or-nothing behavior of swift-recon and the plugin. i.e. if one system alerts, all nodes with the check will pop tickets because they're seeing the same output. The particular environment this has stemmed from has ~16 proxy nodes, therefore, would generate 16 tickets for each blip (much better than the current 60+ that it was generating for each swift node. I was thinking we deploy the checks to all proxy nodes, but disable most of them by default. Similarly to the logic for deploying the lb_api_checks on the infra nodes. Thoughts?

antonym:
@npawelek Yeah, that makes sense.  I pushed an updated which I think should take care of that situation.  I have it only installing the proxy checks if it's the first or second server in the swift_proxy list and all swift_recon checks are now on the proxies only.  I've only tested it on AIO so far but I'll see if I can get an multi node put together to make sure it'll work with 2+ nodes.

npawelek:
@antonym Sweet, that might be doable, although, I wonder about multi-region setups? Maybe an override would be good to have as well? Then we'll be able to specify one in each region.

antonym:
Splitting up the work and moving the region specific stuff to rcbops/u-suk-dev#1279

PR #1910, opened 2017-02-22
MaaS: Increase default growth rate check

antonym:
Increases swift_object_replication_growth_rate_threshold

Connects rcbops/u-suk-dev#1202


antonym:
recheck_ceph

PR #1912, opened 2017-02-22
Adding Magnum AIO scenario

chrishultin:


scripts/bootstrap-aio.yml
```diff
@@ -36,7 +36,7 @@
 - name: Execute the OSA AIO bootstrap
   include: "{{ lookup('env', 'OA_DIR') }}/tests/bootstrap-aio.yml"
   vars:
-    scenario: "{{ lookup('env', 'DEPLOY_CEPH') | bool | ternary('ceph', 'swift') }}"
+    scenario: "{% if lookup('env', 'DEPLOY_MAGNUM') == 'yes' %}magnum{% else %}{{ lookup('env', 'DEPLOY_CEPH') |bool |ternary('ceph','swift') }}{%endif%}"
```
d34dh0r53: So it's either `magnum` or `ceph`/`swift`? Not both? Is there a scenario where you would want `magnum` with either `ceph` or `swift`?

chrishultin: Magnum has no integration or reliance on either Ceph or Swift, so we really don't care whether or not they're installed.

d34dh0r53: Cool, just wanted to clarify

PR #1926, opened 2017-02-27
Use recent OSA key/repo management

evrardjp:
The pip_install role added the key/repo management to be
consistent everywhere (when doing RDO/UCA). A patch was also
included in OSA to be able to define an alternative repo to UCA.

We can now in RPC leverage this feature, and therefore we need
to wire the variables to point to our mirror that holds the UCA
packages.

This should do the trick.

Note: This commit should wait for this commit to merge back into newton: https://review.openstack.org/#/c/438456/

Connected rcbops/u-suk-dev#1252


evrardjp:
Updated with latest osa patchset changes.

PR #1933, opened 2017-02-27
Enable trusty-backports repo by default

cloader89:
This change enables the trusty-backports repo by default. Some
images did not have this enabled. This way, we can standardize
the repos needed.

Connects rcbops/u-suk-dev#841

### DIFF
--- a/scripts/bootstrap-aio.yml
+++ b/scripts/bootstrap-aio.yml
@@ -33,6 +33,20 @@
     rpco_deploy_ceph: "{{ lookup('env', 'DEPLOY_CEPH') }}"
     rpco_deploy_hardening: "{{ lookup('env', 'DEPLOY_HARDENING') }}"
 
+- name: Enable trusty-backports
+  hosts: all
+  tasks:
+    - name: Modify sources.list
+      lineinfile:
+        dest: /etc/apt/sources.list.d/trusty-backports.list
+        create: yes
+        line: "{{ item }}"
+        state: present
+      with_items:
+        - "deb http://rackspace.clouds.archive.ubuntu.com/ubuntu/ trusty-backports main restricted universe multiverse"
+        - "deb-src http://rackspace.clouds.archive.ubuntu.com/ubuntu/ trusty-backports main restricted universe multiverse"
+      when: ansible_distribution_version == "14.04"
+
 - name: Execute the OSA AIO bootstrap
   include: "{{ lookup('env', 'OA_DIR') }}/tests/bootstrap-aio.yml"
   vars:


alextricity25:
Review and testing this now.

alextricity25:
I just realized, this fixes the problem for AIO deployments, but not multi-nodes. The bootstrap-aio.yml playbook isn't run for multi-node deployments, so placing this task in here will not ensure the trusty backports repo is enabled for multi-node deployments.

Only the physical hosts will need to have the trusty backports repo enabled. I personally think it would be better to have this task in the openstack_hosts role: https://github.com/openstack/openstack-ansible-openstack_hosts

odyssey4me:
@major I'd agreed - in fact it should be enabled on swift hosts and lxc hosts. It should also only be enabled in the following conditions:

1. The host is running trusty
2. The host does not already have the backports repo enabled

Instead of doing it in the ``os_swift`` and ``lxc_hosts`` role, we could instead put it into the more universal ``openstack_hosts``... but that might be less obvious and will then add backports to the compute hosts which is probably not the best idea.

major:
@odyssey4me Why not put it in `openstack_hosts` and only enable it with a variable? We could toss that variable into `group_vars`.

odyssey4me:
@major Yeah, that's a good idea actually. It's probably best to push up a patch and discuss it in review.

major:
Sounds good, @odyssey4me. Let us know if you need help, @cloader89.
Do it

odyssey4me:
@major I see what you did there.

scripts/bootstrap-aio.yml
```diff
@@ -33,6 +33,20 @@
     rpco_deploy_ceph: "{{ lookup('env', 'DEPLOY_CEPH') }}"
     rpco_deploy_hardening: "{{ lookup('env', 'DEPLOY_HARDENING') }}"
 
+- name: Enable trusty-backports
+  hosts: all
```
odyssey4me: This should only target 'hosts'. If you target 'all' it will also execute against containers which is unnecessary.

scripts/bootstrap-aio.yml
```diff
@@ -33,6 +33,20 @@
...
+  tasks:
+    - name: Modify sources.list
+      lineinfile:
```
odyssey4me: I expect that using the copy module with the content argument will be more efficient than running a task using with_items. Each item results in a script copied to the host to be executed, whereas the copy module with content argument will only result in one action against the target host.

scripts/bootstrap-aio.yml
```diff
@@ -33,6 +33,20 @@
...
+        dest: /etc/apt/sources.list.d/trusty-backports.list
+        create: yes
+        line: "{{ item }}"
+        state: present
+      with_items:
+        - "deb http://rackspace.clouds.archive.ubuntu.com/ubuntu/ trusty-backports main restricted universe multiverse"
```
odyssey4me: The url ``http://rackspace.clouds.archive.ubuntu.com/ubuntu/`` will only work inside a rackspace instance. The URL is not universal. Also, ``main`` and ``multiverse`` are unnecessary.

odyssey4me:
@major After a quick discussion with @andymcc we realised that the backports repo is *not* required for swift hosts any more. The UCA repo provides the required liberasure packages. I expect some RPC docs will need changing for Newton to that effect.

cloader89:
@odyssey4me This is mostly an issue in lab environments. That is why @major and @alextricity25 are saying to put it in a different location other than bootstrap-aio.yml

scripts/bootstrap-aio.yml
```diff
@@ -33,6 +33,20 @@
...
+        - "deb-src http://rackspace.clouds.archive.ubuntu.com/ubuntu/ trusty-backports main restricted universe multiverse"
```
odyssey4me: I don't think that source packages are necessary for anything we do.

odyssey4me:
@major As per your suggesiton, I've pushed up a patch: https://review.openstack.org/439053

cloader89:
Closing this PR, for upstream has a alternative solution.

PR #1945, opened 2017-02-28
Make magnum.sh more idempotent

BjoernT:
This fix allows for running deploy.sh multiple times without overwriting  user_osa_variables_defaults.yml or adding over and over the same config values


chrishultin:
We're not targeting Mitaka anymore, so we can simply remove the Magnum code entirely from this branch.

PR #1960, opened 2017-03-01
Build repo, memcache, rsyslog artifacts

evrardjp:
Repo_server need patching to remove the ssh key generation, but
the other roles will just make use of the strategy.

Connected rcbops/u-suk-dev#1298
Connected rcbops/u-suk-dev#1207
Connected rcbops/u-suk-dev#1206


evrardjp:
Now with more env.d pleasure.

PR #1967, opened 2017-03-02
MaaS: Add Adaptec JBOD checks

major:
This patch adds a new check for JBODs connected via Adaptec HBAs.
Any disks with bad SMART data will trigger a CRITICAL alarm. There
is only one alarm per physical server, not one alarm per disk.

Connects rcbops/u-suk-dev#1191
Connects rcbops/u-suk-dev#1214

### DIFF
--- a/rpcd/playbooks/roles/rpc_maas/files/plugins/disk_status.py
+++ b/rpcd/playbooks/roles/rpc_maas/files/plugins/disk_status.py
@@ -1,0 +1,165 @@
+#!/usr/bin/env python
+# Copyright 2017, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Check for failed (or failing) disks."""
+import argparse
+import os
+import re
+import shlex
+import subprocess
+
+
+from maas_common import metric_bool
+from maas_common import print_output
+from maas_common import status_err
+from maas_common import status_ok
+
+
+def run_cmd(command, command_input=None):
+    """Run a command and get the output."""
+    if isinstance(command, str):
+        command = shlex.split(command)
+
+    try:
+        output = subprocess.check_output(command)
+    except Exception:
+        return False
+
+    return output
+
+
+def adaptec_check():
+    """Query SMART data for JBODs connected to Adaptec HBAs."""
+    script_path = os.path.abspath(
+        os.path.join(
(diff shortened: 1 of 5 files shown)


rpcd/playbooks/roles/rpc_maas/templates/adaptec_jbod_check.yaml.j2
```diff
@@ -0,0 +1,20 @@
+{% set label = "adaptec-jbod-check" %}
```
antonym: Does the filename need to be adaptec-jbod-check.yaml.j2 instead with dashes?

major: Good catch.

major:
@npawelek They are redundant, but the Dell one checks for disks with pending failures. We could switch over to use this later or just stick with our existing Dell/HP checks.

rpcd/playbooks/roles/rpc_maas/files/plugins/jbod_checker.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
```
antonym: jbod_checker.sh probably needs to be marked executable to keep it in line with the other plugins.

rpcd/playbooks/roles/rpc_maas/files/plugins/disk_status.py
```diff
@@ -0,0 +1,165 @@
+#!/usr/bin/env python
```
antonym: Tried this on an HP and got:

```
root@rpc-openstack:/usr/lib/rackspace-monitoring-agent/plugins# ./disk_status.py --raid-type hp
Traceback (most recent call last):
  File "./disk_status.py", line 145, in <module>
    hp_raid = hpssacli_check()
  File "./disk_status.py", line 81, in hpssacli_check
    'failing_disks': failing_disks,
NameError: global name 'failing_disks' is not defined
```

Don't have an Adaptec card to test on yet though or a Dell.

major: Oof, dumb error on my part. ``failing_disks`` should be set to 0 since HP can't show predictive failures easily.

major:
@npawelek I roped your changes into the single commit.

npawelek:
@major My worry about the redundant checks is that it will cause confusion on the support side and it will generate two separate alarms and ticket updates for the same problem. 

With Dell chassis using OMSA, failed and failing disks report different statuses already, critical and non-critical, if I'm not mistaken. What is the difference between the failing_disks functionality and openmanage.py generating an alarm based on the conditional '!= Ok'?

I'm all for adding efficiency, but the idea of multiple alarms for the same issue doesn't sit well.

major:
@npawelek We don't plan to set up alarms/checks for the Dell/HP stuff here. It was mainly added so we could gather everything we needed with one script later if needed.

npawelek:
@major Perfect, that works for me.

cloudnull:
MaaS in master has been moved to the rpc-maas repo. Can this change be ported to the new repo?

alextricity25:
Closing due to inactivity. Please re-open this PR in the new rpc-maas repo: https://github.com/rcbops/rpc-maas

PR #1969, opened 2017-03-02
[Backport - Newton] MaaS: Remove older legacy HP checks

antonym:
Ensures previously used HP checks are no longer present.

Connects rcbops/u-suk-dev#1320


claco:
Maybe impact rc1 timeline, depending on devqe/testing coordination.

PR #1970, opened 2017-03-02
[Backport - Mitaka] MaaS: Remove older legacy HP checks

antonym:
Ensures previously used HP checks are no longer present.

Connects rcbops/u-suk-dev#1320


antonym:
recheck_upgrade

Issue #1977, opened 2017-03-03
Remove ssl_impossible from console services in f5-config.py


alextricity25:
##### What type of issue is this? Please tick the appropriate box.
- [x] bug - existing functionality that does not work for you as documented
- [ ] feature - new functionality that you would like to see added

##### What is the expected behavior?
The NoVNC virtual server configured via the ``f5-config.py`` script should have a public SSL equivalent.

##### What is actually happening?
 Right now, ``ssl_impossible`` [is being set to true](https://github.com/rcbops/rpc-openstack/blob/master/scripts/f5-config.py#L278) for these virtual servers, which doesn't match the current configurations that expect SSL to be terminated at the load balancer.

##### What were you attempting to do when you encountered this error (give as much detail as possible)? 
This issue was found as a result of the work being done here: https://github.com/rcbops/u-suk-dev/issues/1221#issuecomment-283774006

##### How can we reproduce this issue?
Run the ``f5-config.py`` script, then try to access the NoVNC console via the external_lb_vip_address.

##### How can we verify your issue is addressed?
The console is available via the external_lb_vip_address

##### If you are using or know of a workaround please describe the necessary steps?
The virtual server can be configured manually via the f5 UI.

##### What is your evaluation on the importance of this issue, and why?
prio-1 - Console access is critical for an OpenStack deployment.
----------------------------------------------------------------------------------------------------------------------------
###### For each pull request, update the branch line to reference it.
- [x] master
- [x] newton-14.0
- [ ] mitaka-13.1
- [ ] liberty-12.2

##### Upgrade impact. Consider the impact on moving between major, minor and patch releases.
- [ ] This change does not impact upgrades
- [ ] This change addresses a potential upgrade impact
- [x] The upgrade impact is being tracked by another issue.

The issue tracking the upgrade impact is:

##### Testing impact
- [ ] An issue/pull-request has been create or exists to track the testing gap, and is referenced below
- [ ] This change includes additional testing
- [x] This change is tested by the existing setup

Additional testing links:

##### Release note
- [ ] This change includes a release note.
- [x] This change does not require a release note.

##### Does your change require documentation?
- [ ] Yes, and I submitted an issue to [Docs issues](https://github.com/rackerlabs/docs-rpc/issues "Docs issues")
- [ ] Yes, and I contributed a documentation patch to [docs-rpc](https://github.com/rackerlabs/docs-rpc "docs-rpc")
- [x] No documentation required for this change

[...]

PR #1979, opened 2017-03-03
[Backport - newton-14.0] Remove ssl_impossible from console services

alextricity25:
Traffic to this service is now expected to be SSL-terminated
at the load balancer, which means a public, ssl virtual server
is needed.
This commit removes ``ssl_impossible`` from the console service's
POOL_PARTS. This will tell the script to create a public ssl
virtual server for the services.

Connects https://github.com/rcbops/rpc-openstack/issues/1977

(cherry picked from commit 62b07d58d4f48e171c3b205be3f43b47745496f3)


Rydor:
Alex and I attempted this in IAD3-1 and were successful. I believe this should resolve a defect reported in relation to NoVNC console being accessible over https.

PR #1980, opened 2017-03-06
[WIP] Add Jenkinsfile and Dockerfile for AIO Jobs

hughsaunders:
This Jenkinsfile will be called by CIT Jenkins for PRs and Branch builds.
It will in turn call the RPC-AIO job to test swift/ceph/xenial/upgrades.
The results of the AIO builds will be added to the PR, with direct links.

Lint Jobs are executed in Docker, to isolate them from the CIT slave.
The Lint jobs are executed before AIOs to prevent wasting resources on
AIO builds when lint fails.

Connects rcbops/u-suk-dev#1313

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,15 @@
+FROM ubuntu
+RUN apt-get update
+RUN apt-get install -y python2.7
+RUN apt-get install -y python-pip
+RUN apt-get install -y build-essential
+RUN apt-get install -y python-dev
+RUN apt-get install -y libssl-dev
+RUN apt-get install -y libffi-dev
+RUN apt-get install -y sudo
+RUN apt-get install -y git
+RUN pip install virtualenv
+RUN pip install 'tox!=2.4.0,>=2.3'
+RUN pip install 'jenkinsapi'
+RUN useradd jenkins --shell /bin/bash --create-home --uid 500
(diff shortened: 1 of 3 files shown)


Jenkinsfile
```diff
@@ -0,0 +1,273 @@
...
+
+    /* Determine series (mitaka, newton, master etc) by finiding the newest
+       branch that contains the first common commit. The problem is that some
+       branches/PRs contain multiple commits, so we start of looking at
+       HEAD^, then HEAD^^, HEAD^^^ etc until a branch match is found.
+    */
+
+    // Determine the series by finding the latest branch
+    // that contains the parent commit
+    containing_branches = sh(
```
hughsaunders: This block should be removed, as it was duplicated to L94. Will wait for other review comments before pushing another revision.

Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM ubuntu
+RUN apt-get update
```
melissa-kam: It's a best practice to run `apt-get update` and `apt-get install` in the same RUN command so that if you add another package to install, Docker reruns the update instead of using the old cache.

https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#run

hughsaunders: Hah, I deliberately split them to build on the old cache rather the rerunning update each time. But I can see how that may cause problems in the long run.

PR #1987, opened 2017-03-06
Adding Magnum to F5 Configuration

chrishultin:


scripts/f5-config.py
```diff
@@ -356,6 +356,13 @@
         'mon_type': '/' + PART + '/' + PREFIX_NAME + '_MON_TCP_REPO_GIT',
         'group': 'pkg_repo',
         'hosts': []
+    },
+    'magnum': {
+        'port': 9511,
+        'backend_port': 9511,
+        'mon_type': '/' + PART + '/RPC-MON-EXT-ENDPOINT',
```
prometheanfire: Shouldn't this have included PREFIX_NAME?

prometheanfire: guess it's not always used

how was this tested?

Issue #1988, opened 2017-03-06
Join confusing MaaS Python package variables


major:
There are two lists of Python packages needed in the MaaS virtual environment:
* https://github.com/rcbops/rpc-openstack/blob/master/rpcd/playbooks/roles/rpc_maas/defaults/main.yml#L510-L535

However, both of these lists are installed in the same task:
* https://github.com/rcbops/rpc-openstack/blob/master/rpcd/playbooks/roles/rpc_maas/tasks/package_install.yml#L41-L51

This is confusing and difficult to debug.

PR #1989, opened 2017-03-06
MaaS: Join Python packages lists

major:
Both lists of Python packages are installed at the same time and
they should be consolidated into one to remove confusion.

Connects rcbops/rpc-openstack#1988

### DIFF
--- a/rpcd/playbooks/roles/rpc_maas/defaults/main.yml
+++ b/rpcd/playbooks/roles/rpc_maas/defaults/main.yml
@@ -508,33 +508,28 @@ maas_requires_pip_packages:
   - virtualenv
 
 #
-# maas_pip_packages: Packages we need but are not built by openstack-ansible
+# maas_pip_packages: These packages are installed inside the virtualenv.
 #
-maas_pip_packages:
-  - ipaddr
-  - lxc-python2
-  - psutil
-  - rackspace-monitoring-cli
-  - python-magnumclient
-  - swift
-  - waxeye
-
-#
-# maas_pip_dependencies: These are pip packages we depend on, but should already be built in
-#                        openstack-ansible
 # NOTE: python-novaclient 7.x changes how the list() method works and requires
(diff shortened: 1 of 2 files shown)


major:
@alextricity25 Since you requested this one. ;)

rpcd/playbooks/roles/rpc_maas/defaults/main.yml
```diff
@@ -508,31 +508,25 @@ maas_requires_pip_packages:
...
-  - python-magnumclient
-  - swift
-  - waxeye
-
-#
-# maas_pip_dependencies: These are pip packages we depend on, but should already be built in
-#                        openstack-ansible
-maas_pip_dependencies:
-  - cryptography
-  - requests
```
antonym: Don't need requests any more?  Just making sure it wasn't left out.

major: Oops, if that was removed, it wasn't intentional. ;)

major:
Added requests, still need to fix the merge conflict. :(

major:
Should be in better shape.

rpcd/playbooks/roles/rpc_maas/defaults/main.yml
```diff
@@ -508,33 +508,26 @@ maas_requires_pip_packages:
...
   - psutil
   - rackspace-monitoring-cli
-  - python-magnumclient
-  - swift
-  - waxeye
-
-#
-# maas_pip_dependencies: These are pip packages we depend on, but should already be built in
-#                        openstack-ansible
-# NOTE: python-novaclient 7.x changes how the list() method works and requires
```
alextricity25: Should we keep this not in somewhere?

major: Oh, the comment? I can add it back in.

PR #1991, opened 2017-03-07
Add Jenkinsfile and Dockerfile for AIO Jobs

hughsaunders:
This Jenkinsfile will be called by CIT Jenkins for PRs and Branch builds.
It will in turn call the RPC-AIO job to test swift/ceph/xenial/upgrades.
The results of the AIO builds will be added to the PR, with direct links.

Lint Jobs are executed in Docker, to isolate them from the CIT slave.
The Lint jobs are executed before AIOs to prevent wasting resources on
AIO builds when lint fails.

Connects rcbops/u-suk-dev#1313

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,5 @@
+FROM ubuntu:16.04
+RUN apt-get update && apt-get install -y python2.7 python-pip build-essential python-dev libssl-dev libffi-dev sudo git
+RUN pip install virtualenv 'tox!=2.4.0,>=2.3' jenkinsapi
+RUN useradd jenkins --shell /bin/bash --create-home --uid 500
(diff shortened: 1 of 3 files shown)


hughsaunders:
recheck_ceph

Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM ubuntu
+RUN apt-get update
```
mkam: Did we want to stay with apt-get update and install as separate commands?

hughsaunders:
CIT ceph and xenial jobs failed on downloading ruby gems, retrying.

hughsaunders:
skip ci

hughsaunders:
recheck_ceph recheck_xenial

Issue #2008, opened 2017-03-08
[MaaS] swift-recon.py plugin runs swift-recon locally


alextricity25:
##### What type of issue is this? Please tick the appropriate box.
- [x] bug - existing functionality that does not work for you as documented
- [ ] feature - new functionality that you would like to see added

##### What is the expected behavior?
The swift-recon.py plugin doesn't try to invoke ``swift-recon`` on the infra hosts.

##### What is actually happening?

After running the ``setup-maas.yml`` playbook, I noticed that the swift-recon checks are being dropped in my infra hosts:
```
root@568575-infra01:/opt/rpc-openstack/rpcd/playbooks# ls /etc/rackspace-monitoring-agent.conf.d/ | grep swift
filebeat_process_check-568575-infra01_swift_proxy_container-65b24edd.yaml
lb_api_check_swift_access.yaml
lb_api_check_swift_healthcheck.yaml
swift_account_replication_check--568575-infra01_swift_proxy_container-65b24edd.yaml
swift_async_check--568575-infra01_swift_proxy_container-65b24edd.yaml
swift_container_replication_check--568575-infra01_swift_proxy_container-65b24edd.yaml
swift_md5_check--568575-infra01_swift_proxy_container-65b24edd.yaml
swift_object_replication_check--568575-infra01_swift_proxy_container-65b24edd.yaml
swift_proxy_server_check--568575-infra01_swift_proxy_container-65b24edd.yaml
swift_quarantine_check--568575-infra01_swift_proxy_container-65b24edd.yaml
swift_time_sync_check--568575-infra01_swift_proxy_container-65b24edd.yaml
```
The current swift_recon.py maas plugin is trying to invoke the swift-recon executable on the host it's running on. That can be seen here:
https://github.com/rcbops/rpc-openstack/blob/newton-14.0/rpcd/playbooks/roles/rpc_maas/files/plugins/swift-recon.py#L65-L67

The result of this is that we now have failing checks on the infra hosts since the infra hosts aren't running swift binaries. cc @xeregin

##### What were you attempting to do when you encountered this error (give as much detail as possible)? 
We ran ``setup-maas.yml`` on our multi-node lab for QE testing after blowing away all of the content in ``/etc/rackspace-monitoring-agent.conf/*``. We noticed that the swift recon checks are also being dropped into this directory on our infra hosts.

##### How can we reproduce this issue?
See above.

##### How can we verify your issue is addressed?
[...]

PR #2012, opened 2017-03-09
MaaS: Run swift-recon checks inside swift-proxy container

antonym:
With the consolidation of the swift-recon checks to the swift-proxy hosts, the checks need to run within the swift-proxy container.  This patch runs the swift-recon checks inside the container instead of on physical host.

### DIFF
--- a/rpcd/playbooks/roles/rpc_maas/files/plugins/swift-recon.py
+++ b/rpcd/playbooks/roles/rpc_maas/files/plugins/swift-recon.py
@@ -24,9 +24,11 @@
 
 import argparse
 import re
+import shlex
 import subprocess
 
 import maas_common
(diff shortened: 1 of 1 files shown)


alextricity25:
I tested this in the QE lab, and the fix worked fine!
```
(maas-r13.1.0rc1) root@568575-infra01:/usr/lib/rackspace-monitoring-agent/plugins# ./run_plugin_in_venv.sh swift-recon.py time
status okay
metric time_sync_total uint64 3
metric time_sync_errors uint64 0
metric time_sync_success uint64 1
metric time_sync_time_differ uint64 0
```

We reployed maas with all of it's checks, and everything came up green, including all of the other swift recon checks.

PR #2014, opened 2017-03-09
Restrict Ubuntu updates for build/deploy

odyssey4me:
For the deployment to be repeatable it is crucial that
the apt sources do not include -security, -updates,
-backports, etc. All this must be provided through the
source added for the apt artifacts repo. For the build
and deployment to match in packages used, the build must
also be restricted.

Hosts and containers must only be enabled for the
'trusty' / 'xenial' repository which should, by default,
use the rackspace mirror. This mirror should be
unchanging after the initial distribution release.

For deployments in CDC's or in other environments with
a local mirror, we detect the existing host repository
and use it for the deployment.

The pip-install bit is required to make the build pass.

Connects https://github.com/rcbops/u-suk-dev/issues/1310

### DIFF
--- a/rpcd/playbooks/configure-apt-sources.yml
+++ b/rpcd/playbooks/configure-apt-sources.yml
@@ -1,0 +1,40 @@
+---
+# Copyright 2017, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+- name: Configure the default apt sources for RPC-O
+  hosts: "{{ apt_target_group | default('hosts') }}"
+  user: root
+  tasks:
+
+    - name: Determine the existing Ubuntu repo configuration
+      shell: 'grep -oP "^deb \K(\[?.*\]?.*ubuntu\S*\/?)(?= {{ ansible_distribution_release }} main)" /etc/apt/sources.list'
+      register: _ubuntu_repo
+      when:
+        - host_ubuntu_repo is not defined
+      changed_when: false
+      delegate_to: "{{ physical_host | default(omit) }}"
+
+    - name: Replace the apt sources file with our content
+      copy:
+        content: |
+          # {{ ansible_managed }}
+
+          # Base repository
+          deb {{ host_ubuntu_repo | default(_ubuntu_repo.stdout_lines[0]) }} {{ ansible_distribution_release }} main universe
+        dest: "/etc/apt/sources.list"
+        backup: yes
+      tags:
+        - always
+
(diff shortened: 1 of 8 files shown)


rpcd/playbooks/configure-apt-sources.yml
```diff
@@ -0,0 +1,39 @@
...
+      changed_when: false
+      delegate_to: "{{ physical_host | default(omit) }}"
+
+    - name: Replace the apt sources file with our content
+      copy:
+        content: |
+          # {{ ansible_managed }}
+
+          # Base repository
+          deb {{ host_ubuntu_repo | default(_ubuntu_repo.stdout_lines[0]) }} {{ ansible_distribution_release }} main universe
```
evrardjp: dest ?

scripts/artifacts-building/containers/build-process.sh
```diff
@@ -77,6 +77,12 @@ patch_all_roles
 
 # Run playbooks
 cd /opt/rpc-openstack/openstack-ansible/playbooks
+
+# The host must only have the base Ubuntu repository configured.
+# All updates (security and otherwise) must come from the RPC-O apt artifacting.
+openstack-ansible ${RPCD_DIR}/playbooks/configure-apt-sources.yml -e "host_ubuntu_repo=http://mirror.rackspace.com/ubuntu"
```
evrardjp: This would work for the host, but then the containers don't inherit from this, we are not using lxc_container_create.

I think we should ensure this in artifact-build-chroot.

scripts/deploy.sh
```diff
@@ -153,6 +157,10 @@ if [[ "${DEPLOY_OA}" == "yes" ]]; then
     run_ansible setup-hosts.yml --skip-tags=V-38660
   fi
 
+  # The containers must only have the base Ubuntu repository configured.
+  # All updates (security and otherwise) must come from the RPC-O apt artifacting.
+  run_ansible ${RPCD_DIR}/playbooks/configure-apt-sources.yml -e apt_target_group=all_containers
```
evrardjp: Why do we need this? 
If the container was properly cleaned up (which is what we should do), we shouldn't have to enforce it.

lxc_hosts is the place we copy the apt sources from host for the default container, so if we have our artifacted default that is clean, or any other clean artifacts, it should be alright.
Less lines in deploy.sh the better.

scripts/artifacts-building/containers/build-process.sh
odyssey4me: The container artifacts will use mirror.rackspace.com as the default. See my comment below regarding the containers.

rpcd/playbooks/configure-apt-sources.yml
odyssey4me: Good catch - missed that.

scripts/deploy.sh
evrardjp: Then please add a comment there.

scripts/artifacts-building/containers/build-process.sh
evrardjp: Oh yeah, we don't destroy the cache anymore like you used to require :p 
I'd modify the comment there to add something like:

```
# If we modify this now, the apt list will be ready for lxc_hosts,
# who will copy the apt sources from the host into the default
# container. This default container cache (built in
# lxc_hosts role) is then used in artifact building process.

```
Remember, not everyone has a deep knowledge of the roles in OSA and this avoid unnecessary loookup for ppl.

PR #2015, opened 2017-03-09
[Backport - Newton] MaaS: Run swift-recon in swift-proxy container

antonym:
With the consolidation of the swift-recon checks to the swift-proxy hosts, the checks need to run within the swift-proxy container. This patch runs the swift-recon checks inside the container instead of on physical host.


cloader89:
recheck_ceph

PR #2018, opened 2017-03-09
[Backport - Kilo] MaaS: Run swift-recon in swift-proxy container

antonym:
With the consolidation of the swift-recon checks to the swift-proxy hosts, the checks need to run within the swift-proxy container. This patch runs the swift-recon checks inside the container instead of on physical host.


antonym:
No venv's here, location is different in container, need to fix: /usr/local/bin/swift-recon

antonym:
Cherry-picked https://github.com/rcbops/rpc-openstack/pull/2023 to find the correct path, if kilo tag found, it looks for binary in /usr/local/bin instead.

PR #2023, opened 2017-03-10
MaaS: Use openstack-release instead of rpc for swift-recon

antonym:
Use openstack-release instead of rpc-release for DISTRIB_RELEASE as it's more consistent across releases:

For more details: https://github.com/rcbops/rpc-openstack/issues/2008#issuecomment-285747230

Connects rcbops/u-suk-dev#1388


antonym:
Will backport to newton 14.0 and 14.1 post release.

PR #2042, opened 2017-03-15
Fix async job

evrardjp:
Ansible needs some time to build the job dict.
This ensures at least one period of time has passed before forcing
the review of results.


claco:
MOAR SHEPS!

PR #2043, opened 2017-03-15
MaaS: Modify ceph alarm evaluations

antonym:
Changes ceph_cluster_stats and ceph_mon_stats WARN alarms
to critical.

Connects rcbops/u-suk-dev#1403

### DIFF
--- a/rpcd/playbooks/roles/rpc_maas/templates/ceph_cluster_stats.yaml.j2
+++ b/rpcd/playbooks/roles/rpc_maas/templates/ceph_cluster_stats.yaml.j2
@@ -26,5 +26,5 @@ alarms      :
         criteria                : |
             :set consecutiveCount={{ maas_alarm_local_consecutive_count }}
             if (metric["cluster_health"] == 1) {
-                return new AlarmStatus(WARNING, "Ceph cluster warning.");
+                return new AlarmStatus(CRITICAL, "Ceph cluster warning.");
             }
--- a/rpcd/playbooks/roles/rpc_maas/templates/ceph_mon_stats.yaml.j2
+++ b/rpcd/playbooks/roles/rpc_maas/templates/ceph_mon_stats.yaml.j2
@@ -26,5 +26,5 @@ alarms      :
         criteria                : |
             :set consecutiveCount={{ maas_alarm_local_consecutive_count }}
             if (metric["mon_health"] == 1) {
-                return new AlarmStatus(WARNING, "Ceph mon warning.");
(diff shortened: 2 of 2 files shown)


rpcd/playbooks/roles/rpc_maas/defaults/main.yml
```diff
@@ -214,8 +214,8 @@ maas_fqdn_extension:
 # maas_filesystem_threshold: Sets the threshold (%) for filesystem monitoring when you are not
 #                            specifying specific filesystems.
 #
-maas_filesystem_warning_threshold: 80.0
-maas_filesystem_critical_threshold: 90.0
+maas_filesystem_warning_threshold: 70.0
+maas_filesystem_critical_threshold: 80.0
```
npawelek: Is it possible to make this change only for ceph osd? Not sure if it's a good idea across the board.

antonym: Yeah, wasn't sure if those were the ones you were referring to or not, didn't see any ceph thresholds set anywhere at first glance.

npawelek: Ah, sorry. We'd want to limit these threshold changes only to ceph osd nodes.

antonym:
@npawelek Ok, think I found them, updated PR.

npawelek:
Hrmm. I think we want to leave these variables alone as this is what dictates percentages when ceph will set health to warn and additionally stop writes to the osds. It does seem that these directly correlate to the metrics we receive from ceph_monitoring.py. Maybe the change to CRITICAL for the ceph warn will be sufficient? 

It turns out the customer that had an issue last week was missing both cluster_health and mon_health checks, so they had no monitoring, even when the cluster stopped writing. In the instance of changing the evaluations to critical, we should at least have a ticket associated when the clusters hits 80% and changes the cluster health to warning.

antonym:
@npawelek 

Did a bit more digging and I think we'd need to set these values in the ceph.conf to do so:

mon data avail warn = 30
mon data avail crit = 20

That would then trigger the health check if the available space dropped before those levels.  It appears though that since we're using upstream roles, it doesn't have the settings for those in the template:

https://github.com/ceph/ansible-ceph-common/blob/1c623611a1d43b4c9116d1a0c21f0bdbdd87a0e7/templates/ceph.conf.j2

So lets at least get these flipped to critical and then we can always evaluate if that's we want to see how we get upstream later on or get this in front of our Ceph SMEs to see if that's something we could pursue.

PR #2054, opened 2017-03-17
Ensure an appropriately named branch checkout

odyssey4me:
For the script which guesses the version to work
it requires the checkout to be a local branch
with a name of the form:
  <whatever>-<major version>.<minor version>

We therefore ensure that we have such a branch so
that we can test PR's against the artifacts-14.0
branch.

Connects https://github.com/rcbops/u-suk-dev/issues/1336

### DIFF
--- a/scripts/functions.sh
+++ b/scripts/functions.sh
@@ -76,11 +76,38 @@ function apt_sources_back_to_stock {
 }
 
 function apt_sources_use_rpc_apt_artifacts {
- source /etc/lsb-release
- # TODO(evrardjp): Replace the derive-artifact-version.py
- # when we have a single source of truth
- RPCO_VERSION=${RPCO_VERSION:-$(/opt/rpc-openstack/scripts/artifacts-building/derive-artifact-version.py)}
- echo "deb $HOST_RCBOPS_REPO/integrated/ ${RPCO_VERSION}-$DISTRIB_CODENAME main" \
-   > /etc/apt/sources.list.d/rpco.list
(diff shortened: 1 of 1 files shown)


scripts/artifacts-building/containers/build-process.sh
```diff
@@ -66,8 +66,24 @@ sed -i.bak '/lxc_cache_download_template_extra_options: /d' /etc/openstack_deplo
...
+#
+# For the script which guesses the version to work
+# it requires the checkout to be a local branch
+# with a name of the form:
+#   <whatever>-<major version>.<minor version>
+# We therefore ensure that we have such a branch so
+# that we can test PR's against the artifacts-14.0
+# branch.
+#
+if ! git branch -v | grep -q ' artifacts-14.0 '; then
```
hughsaunders: If the artifacts-14.0 branch does already exist, do you want to check it out? Maybe rm the branch ignoring failure, then always checkout -b?

odyssey4me: If it already exists, it may be the current checkout so we want to leave it alone.

We have three situations to cater for here:

1. A developer working on his own patch. The git checkout may already be artifacts-14.0 but dirty.
2. A periodic which will already have this branch checked out, but perhaps not locally.
3. A PR which will have some random branch name.

hughsaunders: If current branch matches .*-\d+\.\d+ continue; else checkout -b current_branch_name-14.0

Issue #2026, opened 2017-03-10
setup-maas.yml fails on hosts that are not part of lxc_hosts group


mattt416:
###### Please provide the output to 'git describe --all' for  all deployments where you have confirmed it occurs.

Encountered on r14.0.0rc1, but will affect master also.

##### What type of issue is this? Please tick the appropriate box.
- [x] bug - existing functionality that does not work for you as documented
- [ ] feature - new functionality that you would like to see added

##### What is the expected behavior?

setup-maas.yml should complete successfully

##### What is actually happening?

When deploying the openstack-ansible-ops multi-node-aio, setup-maas.yml fails with the following error:

```
TASK [rpc_maas : Install pip packages to venv] *********************************
...
[...]

PR #2057, opened 2017-03-17
[WIP] Add lxc to all hosts

cloader89:
This will add the lxc package to all of the hosts, because some of
the checks require lxc package.

Connects rcbops/rpc-openstack#2026

### DIFF
--- a/rpcd/playbooks/roles/rpc_maas/defaults/main.yml
+++ b/rpcd/playbooks/roles/rpc_maas/defaults/main.yml
@@ -506,6 +506,7 @@ maas_apt_packages:
 #
 maas_requires_pip_packages:
   - virtualenv
+  - lxc-python2
 
 #
 # maas_pip_packages: These packages are installed inside the virtualenv.


nrb:
Would you mind adding a bit more detail about why this approach was chosen into the commit message? Or are you testing this as a possible solution amongst others?

I realize reworking the tasks to be more specific might be more work, but I feel like that's a better solution. Willing to be outvoted, though, and curious as to what other feedback you've gotten.

cloader89:
@nrb @antonym and I went through the possible changes to make this work. We came up with these:

revert the process_check and all the swift process checks PR
add the lxc package to all hosts
rework the script to drop the lxc library (LXC makes easier to get pid)

@BjoernT seemed to be okay with the lxc package itself, instead of the whole LXC cache.


Still waiting on other feedback, although wanted to propose a solution and see if this could be a fix.


We are also welcoming any suggestions, because the script that is failing runs on all hosts, not just lxc hosts. For example, filebeat is on every container/host. A more challenging solution but duplicate stuff would be to split the process_check script into lxc and non lxc, but we don't think it'd work.

nrb:
> the script that is failing runs on all hosts

Right, but does it fail on all hosts?

> A more challenging solution but duplicate stuff would be to split the process_check script into lxc and non lxc

Yeah, this is what I was thinking about. I acknowledge it's more overhead and more work. If installing the package everywhere fixes it and others are fine with that, then cool!

cloader89:
We are getting feedback from Bjoerrn, Nathan and others. That's why I put WIP until I got a consensus.

cloader89:
And no it only fails on the non-lxc hosts.

odyssey4me:
This python package won't install unless the accompanying LXC packages are installed on the same host. This is a very slippery slope. Please see my comments in https://github.com/rcbops/rpc-openstack/issues/2026

nrb:
https://github.com/rcbops/rpc-openstack/pull/2058 appears to be a more targeted approach.

alextricity25:
Closing this off in favor of https://github.com/rcbops/rpc-openstack/pull/2058

PR #2059, opened 2017-03-18
Fix key error in rpc-maas-tool

hughsaunders:
Simple bug.

Connects rcbops/u-suk-dev#1418


alextricity25:
Does this need to go into any of the other branches?

mattt416:
@alextricity25 this bug appears to exist in liberty-12.2 [1], so I do believe it needs to go all the way back.

[1] https://github.com/rcbops/rpc-openstack/blob/liberty-12.2/scripts/rpc-maas-tool.py#L465-L466

PR #2062, opened 2017-03-20
[Backport - Newton] MaaS: Split process_checks for host and container

antonym:
Splits process_checks.py into process_check_host.py and
process_check_container.py to get around the lxc package
requirement on hosts without lxc. Also changes the checks
using process_checks to choose one or the other depending
on the situation.

process_check_container.py is a copy of the original
process_check.py script, and process_check_host.py has
had the container things related to needing the lxc library
stripped out.

Connects rcbops/rpc-openstack#2026


alextricity25:
recheck_ceph

Issue #2048, opened 2017-03-16
MaaS holland_local_check alerts in greenfield


xeregin:
##### What type of issue is this? Please tick the appropriate box.
- [ ] bug - existing functionality that does not work for you as documented
- [ ] feature - new functionality that you would like to see added

##### What is the expected behavior?
The holland check should pass on a new build.

##### What is actually happening?
The holland check is alerting because it cannot find a backup from the previous day.

##### What were you attempting to do when you encountered this error (give as much detail as possible)? 
Validate an installation of MaaS on Newton Trusty

##### How can we reproduce this issue?
Rekick a lab

##### How can we verify your issue is addressed?
Verify that the alarm doesn't alert on a "day one" lab

##### What is your evaluation on the importance of this issue, and why?
Moderately important because it will cause any new labs to be in alert status for 1 day

----------------------------------------------------------------------------------------------------------------------------
###### For each pull request, update the branch line to reference it.
- [ ] master
- [ ] newton-14.0
- [ ] mitaka-13.1
- [ ] liberty-12.2

##### Upgrade impact. Consider the impact on moving between major, minor and patch releases.
- [ ] This change does not impact upgrades
- [ ] This change addresses a potential upgrade impact
- [ ] The upgrade impact is being tracked by another issue.

The issue tracking the upgrade impact is:

##### Testing impact
- [ ] An issue/pull-request has been create or exists to track the testing gap, and is referenced below
- [ ] This change includes additional testing
- [ ] This change is tested by the existing setup

Additional testing links:

##### Release note
- [ ] This change includes a release note.
- [ ] This change does not require a release note.

##### Does your change require documentation?
- [ ] Yes, and I submitted an issue to [Docs issues](https://github.com/rackerlabs/docs-rpc/issues "Docs issues")
- [ ] Yes, and I contributed a documentation patch to [docs-rpc](https://github.com/rackerlabs/docs-rpc "docs-rpc")
- [ ] No documentation required for this change

Additional documentation links (please add links if you have ticked yes):
This is a relevant PR: https://github.com/rcbops/rpc-openstack/pull/1543

b3rnard0:
This is issue to fix and cherry pick for rc2. 

cc: @drea3 @claco @cloader89 @jacobwagner @xeregin @Rydor

b3rnard0:
From @claco:

>The first one, we said, let’s ensure we document that behaviour but maybe not make a change
[...]

PR #2063, opened 2017-03-21
Fix holland backup monitor for greenfield

BjoernT:
This fix will create a holland backup when not present.
In addition, the holland backup monitor is checking for one
successful backup within 48 hours in order to go into an
OK state. This fixes specifically the greenfield
installation where no backup is present from 'yesterday'

Closes-Bug: #2048

### DIFF
--- a/rpcd/playbooks/roles/rpc_maas/files/plugins/holland_local_check.py
+++ b/rpcd/playbooks/roles/rpc_maas/files/plugins/holland_local_check.py
@@ -15,7 +15,6 @@
 # limitations under the License.
 import argparse
 import datetime
-import os
 import shlex
 import subprocess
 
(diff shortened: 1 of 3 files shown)


rpcd/playbooks/roles/rpc_maas/files/plugins/holland_local_check.py
```diff
@@ -90,27 +90,21 @@ def main():
...
+    yesterday = (datetime.date.today() -
+                 datetime.timedelta(days=1)).strftime('%Y%m%d')
 
     # Get completed Holland backup set
     backupsets = \
         container_holland_lb_check(galera_container, holland_bin, holland_bs)
 
-    if len([backup for backup in backupsets if yesterday in backup[0]]) > 0:
+    if len([backup for backup in backupsets
+            if yesterday or today in backup[0]]) > 0:
```
alextricity25: In the commit description, you say:
> In addition, the holland backup monitor is checking two days
for successful backups now, yesterday **and** today in order to
go into an OK state

However, this logic seems to be checking if the first backup in the ``backupsets`` list was done today *or* yesterday, instead of checking to see if the two backups exist. Is that intended?

BjoernT: Actually I ment that only one backup needs to be there to succeed the check, today or yesterday. I will change the commit message to reflect that

BjoernT: The intention is to be OK if one backup was found from yesterday or today so yes that is intended

alextricity25:
This change looks good to me. I'll circle back around to this PR once the gate jobs finish successfully

alextricity25:
This will also need a backport to the newton-14.0 branch.

BjoernT:
@alextricity25 https://github.com/rcbops/rpc-openstack/pull/2070

PR #2067, opened 2017-03-22
[Mitaka] Fix key error in rpc-maas-tool

hughsaunders:
Simple bug.

Connects rcbops/u-suk-dev#1418

(cherry picked from commit 5e316dbaebe8ffb3c27f447b8dc8cc40aa6e8cbb)

### DIFF
--- a/scripts/rpc-maas-tool.py
+++ b/scripts/rpc-maas-tool.py
@@ -671,8 +671,9 @@ class RpcMassCli(object):
                 try:
                     result = self.rpcm.conn.test_existing_check(check)
                 except rackspace.RackspaceMonitoringValidationError as e:
-                    validation_error = (" Validation Error: %(s):"
-                                        % {'e': e.message})
+                    validation_error = " Validation Error: {s}:".format(
+                        s=e.message
+                    )
                     break
 
                 status = result[0]['status']


major:
recheck_all

cloader89:
recheck_all

hughsaunders:
recheck_swift recheck_cit_ceph recheck_cit_upgrade

hughsaunders:
recheck_upgrade

mancdaz:
recheck_all

hughsaunders:
recheck_upgrade maas issues should be fixed now..

hughsaunders:
recheck_cit_upgrade

hughsaunders:
All required checks have now passed, so this can merge.

scripts/rpc-maas-tool.py
```diff
@@ -671,8 +671,9 @@ def _get_failed_checks(self, checks=None):
                 try:
                     result = self.rpcm.conn.test_existing_check(check)
                 except rackspace.RackspaceMonitoringValidationError as e:
-                    validation_error = (" Validation Error: %(s):"
-                                        % {'e': e.message})
+                    validation_error = " Validation Error: {s}:".format(
+                        s=e.message
+                    )
                     break
```
git-harry: Fixing the key error means this `break` line gets executed, which in turn means the for-loop ends and failed_checks is never updated. At least with the key error the script will fail, this patch is going to cause _get_failed_checks to misreport the status of the checks.

mancdaz: I believe this is the final backport of three (plus the original master patch), so you might want to go back and look at fixing the patch where it has already landed in master/newton/liberty if this patch is broken

git-harry: @mancdaz yes, I noticed that, I was leaving that to @hughsaunders given this is his patch. I'm not sure if you are implying this should be merged because it is a backport, assuming I'm reading the change correctly, I'd prefer it was abandoned than merged. This patch causes verify-maas to report success even when it's failed, I don't think the bug should be propagated for the sake of consistency.

mancdaz: no, just saying fix it here (as in, don't merge brokenness), but then [someone] remember to go back and fix the other branches

git-harry: Yeah, this definitely shouldn't be left as is.

hughsaunders: Added rcbops/u-suk-dev#1532 to address the wider logic issues. This can still be merged to fix the key error.

git-harry: I'm still against merging this given that it causes failure to be hidden.

hughsaunders:
Closing as will be fixed by backports of rcbops/u-suk-dev#1532

PR #2072, opened 2017-03-23
Add newton-14.0 to toxtree and sphinx template

alextricity25:
I hope I worded that right. Anywho, newton-14.0 has been added
into the source directory, so that sphinx will now build release
notes for the latest release, newton-14.0.

NOTE: The newton-14.0 reno release notes shouldn't be taken that
seriously, since we have not been diligent to create a release
file for most of the work done in newton-14.0. Reno releasenotes
should only be relied on for versions newton-14.1 and greater

An additional step is added into .travis.yml so that the
rpc-openstack clone is converted to an unshallow clone. This is
is reno can traverse all the needed refs in order for it to
function.

Connects https://github.com/rcbops/u-suk-dev/issues/216

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,17 @@ language: python
 before_script:
 - pip install 'tox!=2.4.0,>=2.3'
 
+# By default travis clones the rpc-openstack repo
+# with --depth set to 50, creating a shallow clone. 
+# However, reno typically needs more (or all) of the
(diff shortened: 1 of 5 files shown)


robb-romans:
cc: for awareness @kallimachos

mattt416:
recheck_cit_ceph

Issue #2037, opened 2017-03-14
RPC14/OSA: Keepalived check scripts not working when >1s response time LP#1672857


BjoernT:
###### Please provide the output to 'git describe --all' for  all deployments where you have confirmed it occurs.

All OSA versions using ansible-keepalived role

##### What type of issue is this? Please tick the appropriate box.
- [x] bug - existing functionality that does not work for you as documented
- [ ] feature - new functionality that you would like to see added

##### What is the expected behavior?

Keepalived works when using 2 ICMP packets for ping

##### What is actually happening?

See https://bugs.launchpad.net/openstack-ansible/+bug/1672857

##### What were you attempting to do when you encountered this error (give as much detail as possible)? 

Testing a more mature config like:
```
keepalived_scripts:
  haproxy_check_script:
    check_script: "killall -0 haproxy"
    timeout: 2
  pingable_check_script:
    check_script: "/bin/ping -c {{ keepalived_ping_count }} {{ keepalived_ping_address }} 1>&2"
    interval: "{{ keepalived_ping_interval }}"
    fall: 2
    rise: 4
    timeout: 5
```

and I noticed the role does not support timeouts attributes nor are the values correctly set

##### How can we reproduce this issue?

See above

##### How can we verify your issue is addressed?

Fix already in progress

##### If you are using or know of a workaround please describe the necessary steps?

None

##### What is your evaluation on the importance of this issue, and why?

Prio 2

----------------------------------------------------------------------------------------------------------------------------
###### For each pull request, update the branch line to reference it.
- [ ] master
- [ ] newton-14.0
- [ ] mitaka-13.1

##### Upgrade impact. Consider the impact on moving between major, minor and patch releases.
- [ ] This change does not impact upgrades
- [ ] This change addresses a potential upgrade impact
- [ ] The upgrade impact is being tracked by another issue.

The issue tracking the upgrade impact is:

##### Testing impact
- [ ] An issue/pull-request has been create or exists to track the testing gap, and is referenced below
- [ ] This change includes additional testing
- [ ] This change is tested by the existing setup

Additional testing links:

##### Release note
- [ ] This change includes a release note.
- [ ] This change does not require a release note.

##### Does your change require documentation?
- [ ] Yes, and I submitted an issue to [Docs issues](https://github.com/rackerlabs/docs-rpc/issues "Docs issues")
[...]

PR #2080, opened 2017-03-24
Bump keepalived role

BjoernT:
The keepalived role prior to version 2.6.0 is defect in how it
configured keepalived to execute the VRRP check scripts, leading
to undefined results or in worst case failing keepalived instances.
This change increases the version required for this product to 2.6.0

connects #2037

### DIFF
--- a/ansible-role-requirements.yml
+++ b/ansible-role-requirements.yml
@@ -14,3 +14,7 @@
   scm: git
   src: https://github.com/rsoprivatecloud/openstack-ops
   version: 1383d0db302c6fc400a50fae4f85e77bf1633581
+- name: keepalived
+  scm: git
+  src: https://github.com/evrardjp/ansible-keepalived
+  version: 2.6.0


BjoernT:
@alextricity25 It seems like we need to upgrade keepalived due to the concerns voiced at https://review.openstack.org/#/c/449305/ preventing newton from including a functional role. At this point I don't think the community is willing to merge the upstream PR

cloader89:
looks good, will approve after gate passes

BjoernT:
@odyssey4me I'm testing this exact version under Newton and it does not break, the upstream comments actually refer to the change inside haproxy-install.yml not in the role by itself.

cloudnull:
@odyssey4me @BjoernT Is this something that we still need?

BjoernT:
Yes I still want it, we likely will sell more haproxy over F5 soon

cloudnull:
@BjoernT It doesn't look like this is actually needed anymore. In OSA the default in newton is to use version 2.6.0: [ https://github.com/openstack/openstack-ansible/commit/d4089f1621a9a075830c916c8cf64253b2821bd1 ]. I think we can close this patch as it stands.

cloudnull:
Here's the link into stable/newton https://github.com/openstack/openstack-ansible/blob/stable/newton/ansible-role-requirements.yml#L25-L28 I think we simply need to ensure we're running on the correct stable release and we should be good to go.

cjloader:
master bumped. can you see if it's resolved @BjoernT?

cloudnull:
Sha bumped closing this issue.

PR #2082, opened 2017-03-24
Add testing for the container artifacts

evrardjp:
- get-pip, virtualenv, and apt source presence
  testing for an artifact after its building
- Metadata and integrity testing after upload

Connects https://github.com/rcbops/u-suk-dev/issues/1412

### DIFF
--- a/scripts/artifacts-building/containers/build-process.sh
+++ b/scripts/artifacts-building/containers/build-process.sh
@@ -123,6 +123,10 @@ ansible_tag_filter "openstack-ansible containers/artifact-build-chroot.yml -e ro
 ansible_tag_filter "openstack-ansible containers/artifact-build-chroot.yml -e role_name=repo_server -v" "install" "config"
 ansible_tag_filter "openstack-ansible containers/artifact-build-chroot.yml -e role_name=rsyslog_server -v" "install" "config"
 
+# test one container build contents
+openstack-ansible containers/test-built-container.yml
+openstack-ansible containers/test-built-container-idempotency-test.yml | tee /tmp/output.txt; grep -q 'changed=0.*failed=0' /tmp/output.txt && { echo 'Idempotence test: pass';  } || { echo 'Idempotence test: fail' && exit 1; }
+
 # Only push to the mirror if PUSH_TO_MIRROR is set to "YES"
 # This enables PR-based tests which do not change the artifacts
 if [[ "$(echo ${PUSH_TO_MIRROR} | tr [a-z] [A-Z])" == "YES" ]]; then
@@ -152,7 +156,12 @@ if [[ "$(echo ${PUSH_TO_MIRROR} | tr [a-z] [A-Z])" == "YES" ]]; then
 
     # Ship it!
     openstack-ansible containers/artifact-upload.yml -i /opt/inventory -v
+
+    # test the uploaded metadata: fetching the metadata file, fetching a
+    # container, and checking integrity of the downloaded artifact.
+    openstack-ansible containers/test-uploaded-container-metadata.yml -v
   fi
 else
   echo "Skipping upload to rpc-repo as the PUSH_TO_MIRROR env var is not set to 'YES'."
 fi
+
--- a/scripts/artifacts-building/containers/test-built-container-idempotency-test.yml
+++ b/scripts/artifacts-building/containers/test-built-container-idempotency-test.yml
@@ -1,0 +1,47 @@
+---
+- name: Dynamic inclusion of container
+  hosts: localhost
+  connection: local
+  gather_facts: no
+  vars_files:
+    - container-vars.yml
+  vars:
+    image_name: "{{ lookup('env', 'CONTAINER_TYPE') | default('keystone', True) }}"
+  tasks:
+    - name: prepare group for integrity checking
+      add_host:
+        name: "/tmp/{{ image_name }}/"
+        groups: "lxc_container_artifact"
+        container_name: "LXC_NAME"
+        physical_host: "localhost"
+      changed_when: false
+
+- name: Check integrity
+  hosts: lxc_container_artifact
+  connection: chroot
+  # Gether facts is necessary
+  vars_files:
+    - container-vars.yml
+  vars:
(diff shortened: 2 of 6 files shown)


scripts/artifacts-building/containers/build-process.sh
```diff
@@ -123,6 +123,10 @@ ansible_tag_filter "openstack-ansible containers/artifact-build-chroot.yml -e ro
 ansible_tag_filter "openstack-ansible containers/artifact-build-chroot.yml -e role_name=repo_server -v" "install" "config"
 ansible_tag_filter "openstack-ansible containers/artifact-build-chroot.yml -e role_name=rsyslog_server -v" "install" "config"
 
+# test one container build contents
```
odyssey4me: I suggest only doing this if ``PUSH_TO_MIRROR`` != 'yes'`` (include converting its case, just in case).

Also, to reduce the execution time, I'd suggest that the containers not required for this test do not build if  ``PUSH_TO_MIRROR`` != 'yes'``.

scripts/artifacts-building/containers/test-built-container-idempotency-test.yml
```diff
@@ -0,0 +1,47 @@
...
+  # Gether facts is necessary
+  vars_files:
+    - container-vars.yml
+  vars:
+    image_name: "{{ lookup('env', 'CONTAINER_TYPE') | default('keystone', True) }}"
+  tasks:
+    - name: Fetch get-pip for comparison
+      get_url:
+        url: "{{ rpco_mirror_base_url }}/os-releases/{{ rpc_release }}/get-pip.py"
+        dest: /opt/get-pip.py
```
odyssey4me: Why not just fetch it to a different location?

scripts/artifacts-building/containers/test-built-container-idempotency-test.yml
```diff
@@ -0,0 +1,47 @@
...
+        force: yes #Ensure it's redownloaded for testing purposes, not directly ok.
+    - name: Check apt source is pointing to rax mirror
```
odyssey4me: We should also verify that there are no other sources configured.

scripts/artifacts-building/containers/test-built-container.yml
```diff
@@ -0,0 +1,30 @@
...
+    - container-vars.yml
+  vars:
+    image_name: "{{ lookup('env', 'CONTAINER_TYPE') | default('keystone', True) }}"
+    image_path: "{{ lxc_image_folder }}/{{ lxc_index_path }}/{{ image_name }}-{{ rpc_release }}/{{ build_id }}/"
+  tasks:
+    - name: Ensure test directory is created
+      file:
+        path: "/tmp/{{ image_name }}"
+        state: directory
+    - name: Copy rootfs into test directory
```
odyssey4me: nit: Why bother copying when you can extract it directly?

scripts/artifacts-building/containers/test-built-container-idempotency-test.yml
evrardjp: Well if I fetch to a different location, it's gonna be marked as changed. Here everything is done naturally.

scripts/artifacts-building/containers/test-built-container-idempotency-test.yml
evrardjp: Sure we could:
- Add a task to check that files in /etc/apt contains updates or backports, registering resulting files,
- Asserting that the register doesn't hold files.

But that would be broken if we split into different repos and use the same name as upstream. Are you sure you want to do that?

Maybe we could output a list of the repos, and compare to a "known list".

scripts/artifacts-building/containers/test-built-container.yml
evrardjp: Fair point, I was afraid of changes in the "built" rootfs (mtime, or anything that the unarchive module would do).
Because I didn't review unarchive module, I took the safe path. I could have a look.

scripts/artifacts-building/containers/build-process.sh
evrardjp: Well I think the idea is to test if everything can build properly first.

Then we can catch up extra issues there by testing idempotency for a container x. We don't need to test all of these indeed.

Doing it only when push to mirror doesn't bring anything, it only reduces our coverage of testing. I'd be enclined to put all the tests before the ``PUSH_TO_MIRROR`` but some rely on pushes to be done (see below), because they are testing the upload.

scripts/artifacts-building/containers/test-built-container-idempotency-test.yml
evrardjp: Of the two options, I think the compare with known list seems a better option.

scripts/artifacts-building/containers/build-process.sh
odyssey4me: OK, it will mean that this test takes ages - but I guess it takes less time than a deploy so it should be OK. We can also try to optimise the build later.

scripts/artifacts-building/containers/test-built-container-idempotency-test.yml
odyssey4me: Fair enough. So you're only executing this once then?

scripts/artifacts-building/containers/test-built-container-idempotency-test.yml
odyssey4me: If it's a known list of things we want, then yes - sounds good.

scripts/artifacts-building/containers/test-built-container.yml
odyssey4me: Meh.

evrardjp: Will be updated in the next patchset.

scripts/artifacts-building/containers/test-built-container-idempotency-test.yml
evrardjp: yes. and checking idempotency directly.

PR #2083, opened 2017-03-27
Downgrade LXC cache packages when necessary

odyssey4me:
When we build python/container artifacts we pull the default
LXC cache variant from LXC upstream which may have packages
installed which are more recent than those we have available
in our artifacted apt repository. Here we figure out which
packages are installed, but not available in a configured
source, then we downgrade them to the latest available in
the configured sources.

Connects https://github.com/rcbops/u-suk-dev/issues/1449

### DIFF
--- a/scripts/artifacts-building/user_rcbops_artifacts_building.yml
+++ b/scripts/artifacts-building/user_rcbops_artifacts_building.yml
@@ -31,3 +31,66 @@ openstack_repo_url: "{{ rpco_mirror_base_url }}"
 # default variant.
 lxc_image_cache_server: images.linuxcontainers.org
 
+# When we build python/container artifacts we pull the default LXC cache
+# variant from LXC upstream which may have packages installed which are more
+# recent than those we have available in our artifacted apt repository. Here
+# we figure out which packages are installed, but not available in a
+# configured source, then we downgrade them to the latest available in the
+# configured sources.
+#
+# By way of reference, typically the output of 'apt list --installed' will
+# produce output such as the following:
+#
+# adduser/trusty,now 3.113+nmu3ubuntu3 all [installed]
+# ca-certificates/r14.0.0rc1-trusty,now 20160104ubuntu0.14.04.1 all [installed]
+# libc6/now 2.19-0ubuntu6.11 amd64 [installed,local]
+# libglib2.0-0/r14.0.0rc1-trusty,now 2.40.2-0ubuntu1 amd64 [installed,automatic]
+#
+# The format of the output is the following:
+# <package name>/<list of dist sources> <version> <arch> <flags>
+#
+# If a package is flagged as 'installed' and 'local' then it indicates that
+# the package has been installed from an apt source which is not currently
+# configured. These are the packages which need to be downgraded to a version
+# in an available apt source.
+#
+lxc_cache_prep_pre_commands: |
+    # If there is a configured resolver, save it.
+    if [ -a /etc/resolv.conf ]; then
+      mv /etc/resolv.conf /etc/resolv.conf.org
+    fi
+    # Use the LXC host's dnsmasq service as a resolver
+    echo "nameserver {{ lxc_net_address | default('10.0.3.1') }}" > /etc/resolv.conf
+    # Add the host's repository keys, including the RPC-O keys
+    apt-key add /root/repo.keys
+    # Update the apt cache
+    export DEBIAN_FRONTEND=noninteractive
+    apt-get update
+    # Check whether there are any installed packages which
+    # are not available in a configured source.
+    if apt list --installed 2>/dev/null | egrep '\[.*local.*\]'; then
+      # Create a list of those packages.
+      pkg_downgrade_list=$(apt list --installed 2>/dev/null | egrep '\[.*local.*\]' | cut -d/ -f1)
+      # Work through the list, checking for the latest available version of
+      # each package in the configured sources. Put together a list of the
+      # packages and their versions in the format that 'apt-get install'
+      # expects it.
+      pkg_downgrade_list_versioned=""
+      for pkg_name in ${pkg_downgrade_list}; do
+        # 'apt-cache madison' provides an easy to parse format:
+        #   libc-bin | 2.19-0ubuntu6.9 | http://rpc-repo.rackspace.com/apt-mirror/integrated/ r14.0.0rc1-trusty/main amd64 Packages
+        #   libc-bin | 2.19-0ubuntu6 | http://mirror.rackspace.com/ubuntu/ trusty/main amd64 Packages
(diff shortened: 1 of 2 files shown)


scripts/artifacts-building/user_rcbops_artifacts_building.yml
```diff
@@ -31,3 +31,30 @@ openstack_repo_url: "{{ rpco_mirror_base_url }}"
...
+# installed which are more recent than those we have available
+# in our artifacted apt repository. Here we figure out which
+# packages are installed, but not available in a configured
+# source, then we downgrade them to the latest available in
+# the configured sources.
+lxc_cache_prep_pre_commands: |
+    if [ -a /etc/resolv.conf ]; then
+      mv /etc/resolv.conf /etc/resolv.conf.org
+    fi
+    {% for resolver in lxc_cache_prep_dns %}
```
evrardjp: Do you ensure existence of lxc_cache_prep_dns ?
Where was it defined beforehand? If it's inside a role, pay attention to variable scoping.

scripts/artifacts-building/user_rcbops_artifacts_building.yml
```diff
@@ -31,3 +31,30 @@ openstack_repo_url: "{{ rpco_mirror_base_url }}"
...
+    echo "nameserver {{ resolver }}" >> /etc/resolv.conf
+    {% endfor %}
+    apt-key add /root/repo.keys
+    export DEBIAN_FRONTEND=noninteractive
+    apt-get update
+    if apt list --installed 2>/dev/null | grep '\/now'; then
```
evrardjp: I'd put a reference of what /now means somewhere.
I see on my machine /repo,now etc... The format would be good to explain to make sure it's doing what you're explaining it does.

scripts/artifacts-building/user_rcbops_artifacts_building.yml
```diff
@@ -31,3 +31,30 @@ openstack_repo_url: "{{ rpco_mirror_base_url }}"
...
+      pkg_downgrade_list=$(apt list --installed 2>/dev/null | awk -F / '/\/now/ {print $1}')
+      pkg_downgrade_list_versioned=$(for pkg in ${pkg_downgrade_list}; do echo -n "${pkg}=$(apt-cache madison ${pkg} | head -n 1 | awk '{ print $3 }') "; done)
+      apt-get install -y --force-yes ${pkg_downgrade_list_versioned}
```
evrardjp: While I think it's a generally good idea to force what we are using, this forces packages (some packages that we don't need) to be marked as explicitly installed (and downgraded) with theirs deps. 

Because I don't know if we are opening pandora's box, I'd rather let someone else (with more apt experience than me) review this.

scripts/artifacts-building/user_rcbops_artifacts_building.yml
odyssey4me: It's in the role, but this var is only evaluated when the role executes so the default is used. If you feel it's better we could hard code it seeing as it's only used in CI.

scripts/artifacts-building/user_rcbops_artifacts_building.yml
odyssey4me: What's the concern with it being marked as explicitly installed? I don't understand what risk this presents in your mind.

odyssey4me: By the way, the original state is as follows:

```
libc-bin/now 2.19-0ubuntu6.11 amd64 [installed,local]
libc6/now 2.19-0ubuntu6.11 amd64 [installed,local]
libgnutls-openssl27/now 2.12.23-12ubuntu2.7 amd64 [installed,local]
libgnutls26/now 2.12.23-12ubuntu2.7 amd64 [installed,local]
multiarch-support/now 2.19-0ubuntu6.11 amd64 [installed,local]
```

They're therefore all marked as explicitly installed anyway.

scripts/artifacts-building/user_rcbops_artifacts_building.yml
```diff
@@ -31,3 +31,66 @@ openstack_repo_url: "{{ rpco_mirror_base_url }}"
...
+    # Use the LXC host's dnsmasq service as a resolver
+    echo "nameserver {{ lxc_net_address | default('10.0.3.1') }}" > /etc/resolv.conf
+    # Add the host's repository keys, including the RPC-O keys
+    apt-key add /root/repo.keys
+    # Update the apt cache
+    export DEBIAN_FRONTEND=noninteractive
+    apt-get update
+    # Check whether there are any installed packages which
+    # are not available in a configured source.
+    if apt list --installed 2>/dev/null | egrep '\[.*local.*\]'; then
```
evrardjp: nit: to avoid querying twice we could

list_local_installed_pkg = $(apt list --installed 2>/dev/null | egrep '\[.*local.*\]')
if -z list_local_installed_pkg (or something like that); then
    pkg_downgrade_list = $( echo $list_local_installed_pkg | cut -d "/" -f 1  ) ...

scripts/artifacts-building/user_rcbops_artifacts_building.yml
```diff
@@ -31,3 +31,66 @@ openstack_repo_url: "{{ rpco_mirror_base_url }}"
...
+      pkg_downgrade_list=$(apt list --installed 2>/dev/null | egrep '\[.*local.*\]' | cut -d/ -f1)
+      # Work through the list, checking for the latest available version of
+      # each package in the configured sources. Put together a list of the
+      # packages and their versions in the format that 'apt-get install'
+      # expects it.
+      pkg_downgrade_list_versioned=""
+      for pkg_name in ${pkg_downgrade_list}; do
+        # 'apt-cache madison' provides an easy to parse format:
+        #   libc-bin | 2.19-0ubuntu6.9 | http://rpc-repo.rackspace.com/apt-mirror/integrated/ r14.0.0rc1-trusty/main amd64 Packages
+        #   libc-bin | 2.19-0ubuntu6 | http://mirror.rackspace.com/ubuntu/ trusty/main amd64 Packages
```
evrardjp: So the local packages are not listed there?

scripts/artifacts-building/user_rcbops_artifacts_building.yml
odyssey4me: Yeah, I thought of that but the query is cheap.

scripts/artifacts-building/user_rcbops_artifacts_building.yml
odyssey4me: Nope - that is the output verbatim from a system with the package locally installed.

evrardjp:
Let's merge and see. It looks generally good, but we'll have to face our death/heaven.

Issue #2064, opened 2017-03-21
Gate job timing out trying to compress logstash.log


mattt416:
##### On which version(s) do you encounter this issue?

mitaka-13.1

###### Please provide the output to 'git describe --all' for  all deployments where you have confirmed it occurs.

```
root@ramcp-18-41c0:/opt/rpc-openstack# git describe --all
remotes/origin/pr/2010/merge
root@ramcp-18-41c0:/opt/rpc-openstack#
```

This job was triggered by PR 2010 on rpc-openstack.

##### What type of issue is this? Please tick the appropriate box.
- [x] bug - existing functionality that does not work for you as documented
- [ ] feature - new functionality that you would like to see added

##### What is the expected behavior?

We're hitting an error where a gate job periodically cannot finish because it spins trying to compress logstash.log (/openstack/log/ramcp-18-41c0_logstash_container-f58fb259/logstash.log for example).  The log should be a manageable size such that it can be compressed and the job finish successfully.

##### What is actually happening?

The log has numerous errors in it and gets appended so fast that the compress cannot finish.  An error includes:

```
"error"=>{"type"=>"mapper_parsing_exception", "reason
"=>"failed to parse [timestamp]", "caused_by"=>{"type"=>"illegal_argument_exception", "reason"=>"Invalid format: \"21/Mar/2017:13:16:59 +0000\" is malformed at \"Mar/2017:13:16:59 +0000\""}}}}, :level=>:warn
```

This seems to be associated with the `"source"=>"/var/log/nginx/openstack-slushee.access.log"`.

Note that these errors are so large that it's difficult to post an entire error message in this bug report.

##### What were you attempting to do when you encountered this error (give as much detail as possible)? 

This is from a standard gate job in new CIT jenkins instance.  We've seen this failure a couple of times and have disregarded thinking it was transient, but it seems to have happened enough that it's worth reporting.

##### How can we reproduce this issue?

We don't hit this issue every time, so I'm not sure it happens every time.  However, building an AIO running mitaka-13.1 will be a good start.

##### How can we verify your issue is addressed?

The logstash.log should not have the error listed above.

##### What is your evaluation on the importance of this issue, and why?

Medium, since I don't believe this error is thrown on every build.

----------------------------------------------------------------------------------------------------------------------------
###### For each pull request, update the branch line to reference it.
- [ ] master
- [ ] newton-14.0
- [ ] mitaka-13.1
- [ ] liberty-12.2

[...]

PR #2085, opened 2017-03-27
Remove logstash log from filebeat

d34dh0r53:
Having logstash process it's own log can lead to a "feedback" loop
wherein an error that is logged into the logstash log will be
re-logged ad-infinitum, eventually filling the disk and crashing
logstash.  This removes the filebeat shipping of the logstash.log
file to logstash.

connects #2064

### DIFF
--- a/releasenotes/notes/remove-logstash-from-filebeat-dd09be25d7f1c37f.yaml
+++ b/releasenotes/notes/remove-logstash-from-filebeat-dd09be25d7f1c37f.yaml
@@ -1,0 +1,7 @@
+---
+fixes:
+  - Having logstash process it's own log can lead to a "feedback" loop
+    wherein an error that is logged into the logstash log will be
+    re-logged ad-infinitum, eventually filling the disk and crashing
+    logstash. This removes the filebeat shipping of the logstash.log
+    file to logstash.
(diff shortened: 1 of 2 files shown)


alextricity25:
This looks like something we should fix in newton-14.0, huh?

d34dh0r53:
@alextricity25 It's intermittent, it's something to look out for in 14.0 but I don't think this alone necessitates another RC.

alextricity25:
Thanks for the info @d34dh0r53. We should definitely get this into 14.0.1. Once the gate jobs pass here, could you create a backport to the newton-14.0 branch? We'll then merge that backport once we tag 14.0.0.

alextricity25:
recheck_cit_ceph

d34dh0r53:
There may be more to come on this one, noticing some strange logs on my xenial instance, investigating now.  I'm going to hold off on the backport until I'm sure.

alextricity25:
Is this still an issue, @d34dh0r53 ?

d34dh0r53:
I think this has the potential to randomly break the gate at any time, so I think it still is an issue, but not urgent.

alextricity25:
@d34dh0r53 Gotcha. Thanks. In an earlier comment you mentioned that they would be more to come for this one. Is this PR is in progress?

d34dh0r53:
@alextricity25 At this point let's merge this one, I think there are some Xenial things that I need to work out with systemd handling the logging.  Getting this merged should eliminate some failed gates.

d34dh0r53:
recheck_cit_swift

PR #2097, opened 2017-03-30
Update git artifact build scripts for rpc-gating

odyssey4me:
This patch properly splits the download/upload playbooks
so that the artifact build can be executed by hand, in
PR's or through the pipeline. It also ensures that it
can be executed on a temporal node instead of a long
lived node.

Connects https://github.com/rcbops/u-suk-dev/issues/1454

### DIFF
--- a/scripts/artifacts-building/git/openstackgit-pull-from-mirror.yml
+++ b/scripts/artifacts-building/git/openstackgit-pull-from-mirror.yml
@@ -1,0 +1,35 @@
+---
+# Copyright 2017, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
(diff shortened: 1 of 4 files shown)


scripts/artifacts-building/git/openstackgit-pull-from-mirror.yml
```diff
@@ -0,0 +1,35 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This playbook's function is to download all existing git clones
+# from rpc-repo.
+
+- name: Pull the current openstackgit folder from rpc-repo
+  hosts: mirrors
```
evrardjp: should probably be mirrors[0] for safety..

Issue #2095, opened 2017-03-30
Block Device Mapping is Invalid: failed to get volume - multi-node deployments


alextricity25:
###### Please provide the output to 'git describe --all' for  all deployments where you have confirmed it occurs.

tag r14.0.0rc2

##### What type of issue is this? Please tick the appropriate box.
- [x] bug - existing functionality that does not work for you as documented
- [ ] feature - new functionality that you would like to see added

##### What is the expected behavior?
The user is able to create an instance off a bootable cinder volume successfully

##### What is actually happening?
When trying to create an instance from a volume using the CLI, I get the following message:
```
root@605055-infra01-utility-container-1b2de33d:~# openstack server create --flavor $FLAVOR --volume boot_volume3 --key-name alex_key --nic net-id=$NETWORK
 test_volume_instance2                                                                                                                                    
  
WARNING: openstackclient.common.utils is deprecated and will be removed after Jun 2017. Please use osc_lib.utils
Block Device Mapping is Invalid: failed to get volume 43ada3f3-0af3-42ff-a8dc-b961a0543f2a. (HTTP 400) (Request-ID: req-43e456c2-9446-4cdc-8a91-bf0e431915
8b)
```
I get the same message when trying to boot an instance from a volume via the UI:
![screenshot from 2017-03-30 08 29 39](https://cloud.githubusercontent.com/assets/4964964/24508882/76b59c00-152a-11e7-9142-91b8e30446a2.png)

A full log, with the req-id grep'ed, can be viewed here:
https://gist.github.com/alextricity25/4b3199a2c489466d66e203f64996e528

##### What were you attempting to do when you encountered this error (give as much detail as possible)? 
I'm deploying RPC off of the r14.0.0rc2 tag using a multi-node architecture. For details on the deployment, please see https://github.com/rcbops/u-suk-dev/issues/1461.

I uploaded an image to glance and created a bootable volume using this command:
```
openstack volume create --image cirros --size 10 boot_volume3
```

I then tried to create an instance off that volume. See error message above.

I also tried doing this in an AIO, but this issue does not come up in our AIOs.
##### How can we reproduce this issue?
Note: This issue does not come up in the AIOs
* Deploy RPCO r14.0.0rc2 on a multi-node environment
* configure the f5 using our tooling
* Upload an image
* Create a bootable volume off that image
* Try to create an instance from that volume

##### How can we verify your issue is addressed?
Instances will be able to spawn using a volume as the source.

[...]

PR #2098, opened 2017-03-30
Remove nova_cross_az_attach default override

alextricity25:
Per nova documentation:
"If False, volumes attached to an instance must be in the same availability
zone in Cinder as the instance availability zone in Nova. This also means
care should be taken when booting an instance from a volume where source
is not “volume” because Nova will attempt to create a volume using the same
availability zone as what is assigned to the instance. If that AZ is not in
Cinder (or allow_availability_zone_fallback=False in cinder.conf), the volume
create request will fail and the instance will fail the build request. By
default there is no availability zone restriction on volume attach."[1]

Adding this override means that creating instances from volumes will not
work with an RPCO default deployment, which doesn't implement cinder AZs.
For more information, please see the Connected card and it's comments.

[1] https://docs.openstack.org/newton/config-reference/compute/config-options.html

Connects https://github.com/rcbops/rpc-openstack/issues/2095

### DIFF
--- a/rpcd/etc/openstack_deploy/user_osa_variables_defaults.yml
+++ b/rpcd/etc/openstack_deploy/user_osa_variables_defaults.yml
@@ -56,7 +56,6 @@ nova_cpu_allocation_ratio: 2.0
 nova_ram_allocation_ratio: 1.0
 
 # Nova config overrides
-nova_cross_az_attach: False
 nova_console_type: novnc
 
 # RabbitMQ overrides


BjoernT:
That was added on purpose so we need to configure the appropriate setting.
Ideally cinder and nova use the same AZs which are related to locations, since the nova api is passing this through

odyssey4me:
Unfortunately the original commit [1] and the original issue [2] don't really provide an explanation of why this was added in the first place.

[1] https://github.com/rcbops/rpc-openstack/commit/61247446b5646d6c2e0b1f8d34cfad2fb7769476
[2] https://github.com/rcbops/rpc-openstack/issues/724

I suggest that we try to clarify *why* changes are being made in both so that we don't end up in this situation again.

alextricity25:
@BjoernT :
>That was added on purpose so we need to configure the appropriate setting.
Ideally cinder and nova use the same AZs which are related to locations, since the nova api is passing this through

That's cool. Let me know if you want to keep it or remove it by approving or disapproving of this PR. I'm okay with either TBH, so long as support is aware.

BjoernT:
If it is not hurting gating we should keep it

alextricity25:
Sounds good. Closing this off per @BjoernT's request to keep the default override.

PR #2100, opened 2017-03-31
Simplify issue template

major:
The current issue template asks a lot from the reporter, especially for simple bugs. This suggested template simplifies the issue template to make it easier to complete and to open the door to a better conversation.

This is not intended to be snarky. This is intended to start a conversation around lowering the barrier to entry for submitting issues.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,58 +1,37 @@
-#### Do NOT include any confidential information, this issue will be created on a public repo
-#### This section is completed by the issue reporter
-##### On which version(s) do you encounter this issue?
-###### Please provide the output to 'git describe --all' for  all deployments where you have confirmed it occurs.
+> **THIS IS A PUBLIC FACING ISSUE.**
+>
+> Do not include private or confidential information in this issue.
 
-##### What type of issue is this? Please tick the appropriate box.
-- [ ] bug - existing functionality that does not work for you as documented
-- [ ] feature - new functionality that you would like to see added
+## Is this a feature or a bug?
 
-##### What is the expected behavior?
+- [ ] Feature _(request for enhancement, new functionality)_
+- [ ] Bug _(something is broken)_
 
-##### What is actually happening?
-
-##### What were you attempting to do when you encountered this error (give as much detail as possible)? 
-
-##### How can we reproduce this issue?
-
-##### How can we verify your issue is addressed?
-
-##### If you are using or know of a workaround please describe the necessary steps?
-
-##### What is your evaluation on the importance of this issue, and why?
-
-----------------------------------------------------------------------------------------------------------------------------
-#### This section is completed during the issue resolution process. If you are reporting an issue, please do not modify the section below.
-##### Which branches are impacted by this issue?
-###### Remove branches that are not impacted.
-###### For each pull request, update the branch line to reference it.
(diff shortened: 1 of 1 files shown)


antonym:
Would it be worth getting rid of the check boxes, as they are usually used for milestone progress and end up junking up the tickets?  

Maybe something like Ansible's template (https://raw.githubusercontent.com/ansible/ansible/devel/.github/ISSUE_TEMPLATE.md):


```

```

antonym:
Might also be good to leave:

#### Do NOT include any confidential information, this issue will be created on a public repo

And also increase the size of the subtitles.

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,58 +1,20 @@
-#### Do NOT include any confidential information, this issue will be created on a public repo
```
cloudnull: I think we should leave this in.

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,58 +1,20 @@
...
-##### If you are using or know of a workaround please describe the necessary steps?
-
-##### What is your evaluation on the importance of this issue, and why?
-
-----------------------------------------------------------------------------------------------------------------------------
-#### This section is completed during the issue resolution process. If you are reporting an issue, please do not modify the section below.
-##### Which branches are impacted by this issue?
-###### Remove branches that are not impacted.
-###### For each pull request, update the branch line to reference it.
+#### Which version of rpc-openstack does this issue relate to?
```
cloudnull: I'd remove these check boxes and simply replace them with a required text field denoting the actual SHA they're on. This can be easily achieved with the following:

``` bash
git log --pretty=oneline | head -n 1
```

Or maybe just:
``` bash
git status | head -n 1
```

This output would give us the exact point in time the bug reporter is starting from. Regardless of branch or tag we'd be able recreate the issues from the exact point in time.

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,58 +1,20 @@
-#### Do NOT include any confidential information, this issue will be created on a public repo
-#### This section is completed by the issue reporter
-##### On which version(s) do you encounter this issue?
-###### Please provide the output to 'git describe --all' for  all deployments where you have confirmed it occurs.
-
-##### What type of issue is this? Please tick the appropriate box.
```
cloudnull: Would this be worth leaving in if we want folks to be able to log a feature request? I'm keen on moving feature requests to a specs repository elsewhere (or something else) which could have it's own ``ISSUE_TEMPLATE.md``.

alextricity25: I prefer the former. Features and bugs would be better tracked within rpc-openstack since it already connects to the waffle board.

.github/ISSUE_TEMPLATE.md
alextricity25: This section is intended "for office use only", hehe. That's one of the reasons it states: "This section is completed during the issue resolution process. If you are reporting an issue, please do not modify the section below.". This section has personally helped me in the past when working on issues. It has helped me know if a doc bug is needed, or to see if the fix needs to be backported.
It might not be used by many today, so I understand if you guys still think it should be removed.

cloudnull:
I think we should leave LN:1 otherwise looks good

evrardjp:
👍

PR #2101, opened 2017-03-31
Normalise the execution of all artifact builds

odyssey4me:
All artifact builds have three options that are necessary:

1. Should it download the existing data from rpc-repo?
2. Should it replace the existing artifacts for the rpc_release it's building for?
3. Should it publish the artifacts it builds?

These options should be set out in all the scripts and in the jobs.
This patch makes that happen.

Connects https://github.com/rcbops/u-suk-dev/issues/1501


odyssey4me:
recheck_cit_all

PR #2115, opened 2017-04-05
Fix failed check validation in rpc-maas-tool.py

hughsaunders:
The function `_get_failed_checks` uses the Rackspace monitoring API to
confirm that the defined checks are working. The function always reports
there are no failed checks due to the use of `break` in the for-loop
resulting in it being exited when encountering the first
`RackspaceMonitoringValidationError`. The result of this is that
`failed_checks` is not updated with the failure and, unless the failure
happens on the last check, not all the checks are validated.

This commit simplifies the logic and removes the use of `break` so that
all the checks are validated and all failed checks are added to
`failed_checks`.

Connects rcbops/u-suk-dev#1532

### DIFF
--- a/scripts/rpc-maas-tool.py
+++ b/scripts/rpc-maas-tool.py
@@ -667,22 +667,19 @@ class RpcMassCli(object):
                         continue
                     checks.append(check)
         for check in checks:
-                validation_error = ""
-                try:
-                    result = self.rpcm.conn.test_existing_check(check)
-                except rackspace.RackspaceMonitoringValidationError as e:
-                    validation_error = " Validation Error: {s}:".format(
-                        s=e.message
-                    )
-                    break
-
+            try:
+                result = self.rpcm.conn.test_existing_check(check)
+            except rackspace.RackspaceMonitoringValidationError as e:
+                completed = False
+            else:
                 status = result[0]['status']
                 completed = result[0]['available']
                 check.state = (" Completed:%(completed)s Status:%(status)s"
                                % {'completed': completed, 'status': status})
-                if completed is False or validation_error != "":
-                    check.bullet = "!"
-                    failed_checks.append(check)
+
+            if not completed:
+                check.bullet = "!"
+                failed_checks.append(check)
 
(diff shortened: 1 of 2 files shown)


scripts/rpc-maas-tool.py
```diff
@@ -668,18 +668,19 @@ def _get_failed_checks(self, checks=None):
...
+                    status = result[0]['status']
+                    completed = result[0]['available']
+                    check.state = (" Completed:%(completed)s Status:%(status)s"
+                                   % {'completed': completed,
+                                      'status': status})
                 except rackspace.RackspaceMonitoringValidationError as e:
                     validation_error = " Validation Error: {s}:".format(
                         s=e.message
                     )
-                    break
```
git-harry: Instead of moving the following three lines, my preference would be to use `else` to make it clearer which code might throw the exception and what happens whether or not the exception is raised. E.g.

```python
try:
    result = self.rpcm.conn.test_existing_check(check)
except rackspace.RackspaceMonitoringValidationError as e:
    validation_error = " Validation Error: {s}:".format(
        s=e.message
    )
    completed = False
else:
    validation_error = ""
    status = result[0]['status']
    completed = result[0]['available']
    check.state = (" Completed:%(completed)s Status:%(status)s"
                            % {'completed': completed, 'status': status})
```

hughsaunders: That does look better, have incorporated your suggestions.

alextricity25:
Nice catch on the ``break`` line, @git-harry!
It would be nice to get into the habit of including a releasenote for a fix, feature, upgrade impact, or deprecation.
@hughsaunders Could you include one here? Thanks!

hughsaunders:
Hah, I tried to review this as @git-harry  wrote the current version, but github thinks I'm the author so won't let me.. 
screen shot 2017-04-06 at 14 06 04

but 👍  anyway.

git-harry:
recheck_xenial

git-harry:
recheck_xenial TASK [rpc_maas : Install local checks]

PR #2120, opened 2017-04-06
Fix undefined variable

evrardjp:
Last refactor (using a shell loop instead of every role separately)
introduced a regression.

Connected  https://github.com/rcbops/u-suk-dev/issues/1535


evrardjp:
Approved by @cloudnull , merging.

PR #2122, opened 2017-04-06
Stage python artifacts properly

evrardjp:
With new repo build (due to the OSA SHA bump),
the location of the python manifest changes
and the code must change too.

Connected https://github.com/rcbops/u-suk-dev/issues/1538


evrardjp:
This needs to be rebased.

evrardjp:
This is blocked by the python building artifact failure. Dealed with another card.

evrardjp:
Unblocked.

evrardjp:
Removed [WIP] from title.

evrardjp:
To test this, the artifact building process should be fixed. This will be happening in another card.

evrardjp:
It's maybe better to split that out in another card, as it's something that is the follow up of the SHA bump card, not the SHA bump itself.

evrardjp:
Split done.

PR #2130, opened 2017-04-10
Bump RPC

evrardjp:
Connected rcbops/u-suk-dev#1544


evrardjp:
Merging this myself, this way the periodic job can run this night.

PR #2136, opened 2017-04-11
Fix Apt Artifact building

evrardjp:
Typos appeared when moving to new jenkins infra, but wasn't caught
up because of the jobs were not ready.

Connected https://github.com/rcbops/u-suk-dev/issues/1555


evrardjp:
It looks like the aprt artifact building went through with these patches. Letting the CI job running to see how it goes for the rest of the pipeline.

PR #2137, opened 2017-04-11
Update ceph.ceph-common to v2.2.0

cloader89:
This updates our ceph playbooks to the latest version.

Connects rcbops/u-suk-dev#1552


alextricity25:
It's probably a good idea to update the ceph-mon and ceph-osd versions to v2.2 too.
Also, is it possible to put a comment on that line simliar to what is done here: https://github.com/openstack/openstack-ansible/blob/stable/newton/playbooks/defaults/repo_packages/openstack_services.yml#L40

You could say "SHA of tag v2.2.0"

PR #2138, opened 2017-04-11
Ansible best practices

evrardjp:
This cleans up the Ansible plays with best practices.
This is to avoid deprecation warning in the logs.

### DIFF
--- a/rpcd/playbooks/ceph-mon.yml
+++ b/rpcd/playbooks/ceph-mon.yml
@@ -15,7 +15,6 @@
 
 - name: Deploy mons
   hosts: mons
-  user: root
   max_fail_percentage: 0
   gather_facts: True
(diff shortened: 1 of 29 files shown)


evrardjp:
FYI:

user: root should be avoided because it should be in the inventory, and it's the current user by default (which is root generally in our case).

localhost is default term existing in ansible when no inventory is given. example:
```
ansible -m ping -c local -i '127.0.0.1' all 
```
won't work, but 
```
ansible -m ping -c local -i '127.0.0.1,' all will or ansible -m ping -c local localhost would too
```

```| default``` should be avoided, and a role default should be used instead.

evrardjp:
Fixed a stupid typo.

evrardjp:
recheck_travis

PR #2139, opened 2017-04-11
Fix failed check validation in rpc-maas-tool.py

mattt416:
The function `_get_failed_checks` uses the Rackspace monitoring API to
confirm that the defined checks are working. The function always reports
there are no failed checks due to the use of `break` in the for-loop
resulting in it being exited when encountering the first
`RackspaceMonitoringValidationError`. The result of this is that
`failed_checks` is not updated with the failure and, unless the failure
happens on the last check, not all the checks are validated.

This commit simplifies the logic and removes the use of `break` so that
all the checks are validated and all failed checks are added to
`failed_checks`.

Connects rcbops/u-suk-dev#1532

(cherry picked from commit 2f3a0a76689122da4741b284200d3c594900e408)

### DIFF
--- a/releasenotes/notes/fix-rpc-maas-tool-check-validation-a5190a72be5830f6.yaml
+++ b/releasenotes/notes/fix-rpc-maas-tool-check-validation-a5190a72be5830f6.yaml
@@ -1,0 +1,6 @@
+---
+fixes:
+  - | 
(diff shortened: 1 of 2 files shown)


mattt416:
This didn't generate a flake8 error gating on master, so something here has to be askew.

mattt416:
Closing, will re-open when a fix to address lint failure has merged.

Issue #2142, opened 2017-04-11
Travis jobs not linting code changes


alextricity25:
###### Please provide the output to 'git describe --all' for  all deployments where you have confirmed it occurs.
master branch, untagged.

##### What type of issue is this? Please tick the appropriate box.
- [x] bug - existing functionality that does not work for you as documented
- [ ] feature - new functionality that you would like to see added

##### What is the expected behavior?
Ansible lint tests are run against code changes made against the master branch as part of the travis jobs

##### What is actually happening?
Due to the changes introduced here: https://github.com/rcbops/rpc-openstack/blob/master/.travis.yml#L6-L15, the travis jobs are running the linting tests on the cloned rpc-openstack repo, but not the code changes introduced in the commit travis is testing.

##### What were you attempting to do when you encountered this error (give as much detail as possible)? 
https://github.com/rcbops/rpc-openstack/pull/2115 passed linting tests, but it actually has a linting error that was missed by the current travis tests.

##### How can we reproduce this issue?
Make a PR to master with pep8 linting issues. The travis tests will still pass.

##### How can we verify your issue is addressed?
The travis tests will fail when a pep8 linting issue is found
##### If you are using or know of a workaround please describe the necessary steps?
NA
##### What is your evaluation on the importance of this issue, and why?
prio-1.
----------------------------------------------------------------------------------------------------------------------------
###### For each pull request, update the branch line to reference it.
- [x] master

##### Upgrade impact. Consider the impact on moving between major, minor and patch releases.
- [ ] This change does not impact upgrades
- [ ] This change addresses a potential upgrade impact
- [ ] The upgrade impact is being tracked by another issue.

The issue tracking the upgrade impact is:

##### Testing impact
- [ ] An issue/pull-request has been create or exists to track the testing gap, and is referenced below
- [ ] This change includes additional testing
- [ ] This change is tested by the existing setup

Additional testing links:

##### Release note
- [ ] This change includes a release note.
- [ ] This change does not require a release note.

##### Does your change require documentation?
- [ ] Yes, and I submitted an issue to [Docs issues](https://github.com/rackerlabs/docs-rpc/issues "Docs issues")
- [ ] Yes, and I contributed a documentation patch to [docs-rpc](https://github.com/rackerlabs/docs-rpc "docs-rpc")
- [ ] No documentation required for this change

Additional documentation links (please add links if you have ticked yes):

PR #2144, opened 2017-04-11
Fetch PR refspec, checkout commit to be tested

alextricity25:
This commit fetchs the PR refspec after doing a full
clone of rpc-openstack, then checks out the commit that
the current build is testing.

This commit also fixes a linting issue that was hidden
because tests were not being run.

Connects https://github.com/rcbops/rpc-openstack/issues/2142


evrardjp:
Looks good. The flake8 should be fixed that way.

evrardjp:
recheck_cit_swift
recheck_cit_xenial

PR #2148, opened 2017-04-12
Fixing the venv fetch path

evrardjp:
The variable to download the venv is overriden in a group_var
per role. Because we don't add temp containers to group and
setup to get their vars, we override the vars ourselves.

This commit ensure we override the url for the venv fetch to
http://rpc-repo.rackspace.com/venvs/r14.0.0rc1/ubuntu-14.04-x86_64/
instead of
http://rpc-repo.rackspace.com/venvs/r14.0.0rc1/ubuntu/ like it
was before the SHA bump that changed the repo build behavior.

Connected https://github.com/rcbops/u-suk-dev/issues/1562

### DIFF
--- a/scripts/artifacts-building/containers/container-vars.yml
+++ b/scripts/artifacts-building/containers/container-vars.yml
@@ -27,8 +27,10 @@ xz_bin:
 architecture_mapping:
   x86_64: amd64
   ppc64le: ppc64el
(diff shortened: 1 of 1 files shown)


evrardjp:
alternative would be to use a way to include host into a group dynamically based on image name (like adding the LXC_NAME container to "{{ image_name }}_all" group)

Default container build would need trick but that's doable too.

Issue #2150, opened 2017-04-13
Ansible 2.1.5 no longer skips task due to undefined variable, fails on setup-maas.yml


alextricity25:
##### What type of issue is this? Please tick the appropriate box.
- [x] bug - existing functionality that does not work for you as documented
- [ ] feature - new functionality that you would like to see added

##### What is the expected behavior?
The setup-maas.yml playbook completes successfully with Ansible version 2.1.5.

##### What is actually happening?
On the OSA SHA bump PR, the Ansible version is bumped to 2.1.5. This is causing the setup-maas.yml playbook to fail on the following task due to an undefined variable:
```
[DEPRECATION WARNING]: Skipping task due to undefined Error, in the future this
 will be a fatal error.: {{ cinder_vg_checks_list }}: [{u'cinder_vg_name': u'{{
 cinder_vg_name }}', u'group': u'cinder_volume', u'name': u'cinder_vg_check'}]:
 {{ cinder_backends['lvm']['volume_group'] | default('cinder-volumes') }}: 
'cinder_backends' is undefined.
This feature will be removed in a future 
release. Deprecation warnings can be disabled by setting 
deprecation_warnings=False in ansible.cfg.
fatal: [ramsp-108-96c5_nova_api_metadata_container-3756e7e9]: FAILED! => {"failed": true, "msg": "The conditional check 'cinder_backends['lvm'] is defined' failed. The error was: error while evaluating conditional (cinder_backends['lvm'] is defined): 'cinder_backends' is undefined\n\nThe error appears to have been in '/opt/rpc-openstack/rpcd/playbooks/roles/rpc_maas/tasks/main.yml': line 71, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- include: ensure_local_checks.yml\n  ^ here\n"}
```
##### What were you attempting to do when you encountered this error (give as much detail as possible)? 
Bump the OSA SHA: https://github.com/rcbops/rpc-openstack/pull/2126

##### How can we reproduce this issue?
Checkout the PR mentioned above, then run setup-maas.yml

##### How can we verify your issue is addressed?
setup-maas.yml completes successfully with Ansible version 2.1.5 installed.

##### If you are using or know of a workaround please describe the necessary steps?
NA

##### What is your evaluation on the importance of this issue, and why?
[...]

PR #2154, opened 2017-04-13
Add an extra conditonal to check cinder_backends

alextricity25:
Not having this conditional causes failures with ansible 2.1.5
due to the dictionary cinder_backends not being defined.

This commit addes an extra conditional statement to check
if cinder_backends is defined.

Connects https://github.com/rcbops/rpc-openstack/issues/2150


cjloader:
recheck_cit_xenial

alextricity25:
go xenial go!

PR #2166, opened 2017-04-17
[IRR] Move the rpc_maas into an independent role repository

cloudnull:
This change removes the rpc_maas role from the rpc-openstack integrated
project. The intention here is to allow the rpc_maas role to function as
a stand alone role giving us the ability to gate and develop MaaS
independently from the rest of the RPC-O product.

Depends-On: https://github.com/rcbops/rpc-maas/pull/231/commits/36f6827d5b425c2d1740699df30819ed7522e9cc


claco:
If you haven't already, please sync up with the RPC-R team, as they reuse these, and I'm sure will have some thoughts/impacts as well.

major:
recheck_all

major:
recheck_cit_all

claco:
@major @cloudnull Is this something that can wait until Newton -> Pike, where it's moved to Artifact delivery, and so doesn't matter where it resides since it's delivered in a completely different manner?

cloudnull:
@major The new role placement will drop the rpc_maas role in /etc/ansible/roles/ and when folks upgrade from one release to another we won't have a conflict as the old role is removed from the tree.

cloudnull:
@claco I'm sure this could wait however there's nothing in the new artifact work that changes the delivery mechanism and I'd like to attack the MaaS fragmentation across releases as support is ramping up to reinstall rpc-maas on installed versions across the fleet.

cloudnull:
recheck_all

cloudnull:
recheck_cit_swift

cloudnull:
recheck_cit_swift

cloudnull:
@alextricity25 I have been tracking these changes and have already merged them into the rpc-maas repo. https://github.com/rcbops/rpc-maas/commit/1701d34fcb106cec11ba00284ac462e78bef14e0

cloudnull:
recheck_cit_swift

cloudnull:
recheck_travis-ci

cloudnull:
Closed and re-opened seems the gates were wedged.

claco:
@alextricity25 what is old is new again. :-)

image

alextricity25:
"icehouse", "kilo", "juno"...yup. checks out.

claco:
In addition to the original commit message (have I mentioned there's no issue filed for this?), there is likely some additional context we're missing. @cloudnull @major can likely elaborate to some of that. Specifically I'm thinking of some discussions we've had in the past around slimming down RPC-O to really be OSA + some extra artifacts rather than two separate things.

cloudnull:
@claco @alextricity25 I'd be happy to talk more about the intentions. is this something I should write here/in a specs repo somewhere?

alextricity25:
@cloudnull The only question that comes to mind right now is the one regarding the rpc-maas role versions. I know you mentioned yesterday how this would help support keep environments at the latest maas playbooks and plugins, without having to also upgrade the openstack version. I imagine that means we are going to have the same branches in both repos, right? i.e. the rpc-maas repo will also have the mitaka-13.1 and newton-14.0 branch.

cloudnull:
@alextricity25 We certainly could though it'd be my preference to have named branches instead of defining specific versions. So something like Mitaka-13, Newton-14. I think there's a lot of prior art that we can follow within upstreams branching strategy.

cloudnull:
recheck_cit_all

cloudnull:
recheck_cit_all

alextricity25:
Are all the pieces in place for this to go? It looks like it!

cloudnull:
@alextricity25 yes this should be all good to go.

galstrom21:
So would it not be better to have the rpc-maas repo have a non-related version number (as it is really meant to not map 1:1 with the openstack version of RPCO)? It would follow that there would not be 'icehouse' or 'whatever' branches.. but rather just a list of releases.

I am assuming here that the point of splitting this out into its own repo is to have it function as its own thing. Which means that any single version of rpc-maas will have some list of versions of RPCO/openstack that it is capable of monitoring. This would allow us to move the monitoring forward across a heterogenous fleet without asking the fleet to move their versions of openstack/RPCO.

alextricity25:
> So would it not be better to have the rpc-maas repo have a non-related version number (as it is really meant to not map 1:1 with the openstack version of RPCO)? It would follow that there would not be 'icehouse' or 'whatever' branches.. but rather just a list of releases.

That makes sense so long as the rpc-maas playbooks are backwards compatible with earlier ansible versions. 
If we have two different openstack releases consuming the same rpc-maas role version, that essentially means that the rpc-maas role will need to support both Ansible versions that OSA deploys across those two openstack versions. I'm sure there are ways around this, and I'm not saying it's impossible to do, but in my experience Ansible isn't very backwards compatible i.e. tasks written for ansible 2.1(newton) might not run on ansible 1.9(mitaka).

Issue #2107, opened 2017-04-03
LXC issue during verify-maas.yml


mkam:
##### What type of issue is this? Please tick the appropriate box.
- [x] bug - existing functionality that does not work for you as documented
- [ ] feature - new functionality that you would like to see added

##### What is the expected behavior?
The verify-maas playbook runs without issues.

##### What is actually happening?
With multi-node AIO and master, the loadbalancer1 node fails because lxc is not installed, similar to what was seen in https://github.com/rcbops/rpc-openstack/issues/2026.

```
TASK [Local MaaS Checks] *******************************************************
FAILED - RETRYING: TASK: Local MaaS Checks (5 retries left).
ok: [swift2]
ok: [swift1]
ok: [swift3]
ok: [logging1]
ok: [cinder2]
ok: [cinder1]
ok: [infra2]
ok: [infra1]
ok: [infra3]
ok: [compute1]
ok: [compute2]
FAILED - RETRYING: TASK: Local MaaS Checks (4 retries left).
FAILED - RETRYING: TASK: Local MaaS Checks (3 retries left).
FAILED - RETRYING: TASK: Local MaaS Checks (2 retries left).
FAILED - RETRYING: TASK: Local MaaS Checks (1 retries left).
[...]

Issue #2164, opened 2017-04-17
setup-maas.yml, doesn't have pip_install meta dependency


alextricity25:
> **THIS IS A PUBLIC FACING ISSUE.**
>
> Do not include private or confidential information in this issue.

## Is this a feature or a bug?

- [ ] Feature _(request for enhancement, new functionality)_
- [x] Bug _(something is broken)_

## Which version of rpc-openstack does this issue relate to?
- [x] master - https://github.com/rcbops/rpc-openstack/pull/2165
- [ ] newton-14.0

Run `git log --pretty=oneline | head -n 1` and provide the output below between
the backtick lines:

```
90dd72f973390eae06ccefd8cb14a4f6cc807464 Merge pull request #2126 from rcbops/cantu/issue/2125/master
```

## What happened?
When the setup-maas.yml playbook is invoked, pip packages are installed as part of [this task](https://github.com/rcbops/rpc-openstack/blob/master/rpcd/playbooks/roles/rpc_maas/tasks/package_install.yml#L41-L51). However, because the ``rpc_maas`` role does not depend on the ``pip_install`` role, the proper ``pip.conf`` file isn't dropped into the physical hosts, causing discrepancies between package sources across RPCO's playbooks. In this case, the incorrect source was used to download the ``apache-libcloud`` pacakge, resulting in an incorrect version.

## What were you doing when it happened?
We noticed this when the rackspace monitoring libraries all of the sudden started pulling down ``libcloud==2.0.0rc2`` as a dependency, breaking things as a result. [1] When trying to pin libcloud to "<2.0.0", some hosts still had successfully downloaded libcloud==2.0.0rc2 due to a race condition. After further investigation, it was found out that the pip.conf file was not pointing to the repo server as it should.

[1] https://github.com/rcbops/rpc-openstack/issues/2156

## What did you expect to happen instead of what actually happened?
The physical hosts should download python packages from the repo server when the setup-maas.yml playbook is run.

**This is currently blocking our multi-node-aio jobs.**

alextricity25:
UPDATE:
I'm confident that the PR I have up now for this fix will fix the setup-maas.yml issues that the multi-nodes are seeing. However, keep in mind that it won't fix the verify-maas.yml issues. Those issues are being addressed in this PR:
https://github.com/rcbops/rpc-openstack/pull/2175

Once both PRs merge, we should have a working setup-maas and verify-maas playbook on the multinodes.

[...]

PR #2175, opened 2017-04-18
Specify hosts for verify_maas

cjloader:
The host targeting for the verify-maas playbook has been
changed to be more explicit. Instead of using the general group
hosts, it is now using a set of groups that constitutes all
of the hosts that verify-maas operates on.

Connects #2107

### DIFF
--- a/rpcd/playbooks/verify-maas.yml
+++ b/rpcd/playbooks/verify-maas.yml
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 - name: Verify MaaS setup
-  hosts: hosts
+  hosts: shared-infra_hosts:compute_hosts:storage_hosts:swift_hosts:log_hosts
   tasks:
     - name: ensure remote directory
       file:


rpcd/playbooks/verify-maas.yml
```diff
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 - name: Verify MaaS setup
-  hosts: hosts
+  hosts: shared-infra_hosts:compute_hosts:storage_hosts:swift_hosts:log_hosts:all_containers
```
alextricity25: ``verfiy-maas.yml`` doesn't include the all_containers group like ``setup-maas.yml`` does. It can be omitted here.

alextricity25:
https://github.com/rcbops/rpc-openstack/issues/2164 Will have to merge first, then this branch will need to be rebased.

rpcd/playbooks/verify-maas.yml
cjloader: addressed.

alextricity25:
recheck_cit_swift

mattt416:
I'm not sure this is correct, see my comment [here](https://github.com/rcbops/rpc-openstack/pull/2177#issuecomment-295556602).

cjloader:
@mattt416 good catch. Thanks.

cjloader:
taking different approach

PR #2176, opened 2017-04-19
Deploy ARA

git-harry:
Install the Ansible Run Analyser callback. This is used to gather task
failure information in gate jobs. A junit stream is created from failed
tasks, this is reported to jenkins so that failed tasks are reported as
failed unit tests. This allows a user to quickly see a list of failed
tasks rather than digging through logs.

Connects rcbops/u-suk-dev#318

(cherry picked from commit 61ab045c8c824f9358d862383805e3bc8c6dda73)


mattt416:
The swift/ceph job never seemed to kick off for whatever reason.

mattt416:
recheck_cit_ceph
recheck_cit_swift

Issue #1603, opened 2016-12-01
neutron_service_check.py not recognizing fqdn after change from short reporting linuxbridge-agent down


ihavegerms:
##### On which version(s) do you encounter this issue?
Kilo - r11.1.6

###### Please provide the output to 'git describe --all' for  all deployments where you have confirmed it occurs.

root@infra-node1:/opt/rpc-openstack# git describe --all
tags/r11.1.6

##### What type of issue is this? Please tick the appropriate box.
- [x] bug - existing functionality that does not work for you as documented
- [ ] feature - new functionality that you would like to see added

##### What is the expected behavior?
That no alert would be created when the service is indeed responding/ok. 

##### What is actually happening?
When the maas plugin attempts to verify it can find itself with the following piece of code;

    # gather nova service states
    if args.host:
        services = nova.services.list(host=args.host)
    else:
        services = nova.services.list()

    if len(services) == 0:
        status_err("No host(s) found in the service list")

It is not finding the hosts that have changed from short, to a fqdn.

To fix the issue, I modified the plugin as follows;

agents = neutron.list_agents(host=args.host + '.l42scl.hortonworks.com')['agents']

... and the check was successful.

##### What were you attempting to do when you encountered this error (give as much detail as possible)? 
Verify an alert detailing that neutron_linuxbridge_agent was failing/down. When I ran the check manually, it was claiming that "No host(s) found in the agents list". So I looked at the code and realized that it couldn't find itself after the customer changed to fqdn's.

##### How can we reproduce this issue?

Create a host with a short name, install maas, then change the hostname to a fqdn. If you run the neutron_service_check.py maas plugin after, it should fail even though within the environment it is checking in without issue.

##### How can we verify your issue is addressed?

We should be able to work with our customer and re-install/update maas on the hosts effected and ensure all checks are good to go. (via lab you should be able to reproduce and verify as well)

##### If you are using or know of a workaround please describe the necessary steps?

As previously mentioned, if you add a short bit to the specified line in the plugin code;

[from]

agents = neutron.list_agents(host=args.host)['agents']

[to]

agents = neutron.list_agents(host=args.host + '.l42scl.hortonworks.com')['agents']

The error goes away.

##### What is your evaluation on the importance of this issue, and why?

[...]

PR #2180, opened 2017-04-19
Allow the use of fqdn in neutron_service_check.py

cjloader:
This change allows for use of either hostname or fqdns when running
the neutron_service_check.py plugin. It did not find the host when
fqdn was used.

Connects #1603

### DIFF
--- a/rpcd/playbooks/roles/rpc_maas/files/plugins/neutron_service_check.py
+++ b/rpcd/playbooks/roles/rpc_maas/files/plugins/neutron_service_check.py
@@ -36,6 +36,8 @@ def check(args):
     # gather nova service states
     if args.host:
         agents = neutron.list_agents(host=args.host)['agents']
+    elif args.fqdn:
+        agents = neutron.list_agents(host=args.fqdn)['agents']
     else:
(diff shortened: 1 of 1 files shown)


rpcd/playbooks/roles/rpc_maas/files/plugins/neutron_service_check.py
```diff
@@ -36,6 +36,8 @@ def check(args):
     # gather nova service states
     if args.host:
         agents = neutron.list_agents(host=args.host)['agents']
+    elif args.fqdn:
+        agents = neutron.list_agents(host=args.fqdn)['agents']
```
alextricity25: @cjloader the ``fqdn`` argument isn't part of the ``args`` ``NameSpace`` object, and will need to be added using the ``add_argument`` method. Similar to what is done for the ``--host`` flag below. Once that's done, the corresponding MaaS check for this plugin will need to be modified to utilize the argument.

cjloader:
https://github.com/rcbops/rpc-maas/pull/233

alextricity25:
Closing in favor of https://github.com/rcbops/rpc-maas/pull/233

Issue #2179, opened 2017-04-19
Revenge of the InvalidBDM when nova_cross_az_attach is set to False


alextricity25:
> **THIS IS A PUBLIC FACING ISSUE.**
>
> Do not include private or confidential information in this issue.

## Is this a feature or a bug?

- [ ] Feature _(request for enhancement, new functionality)_
- [x] Bug _(something is broken)_

## Which version of rpc-openstack does this issue relate to?
- [x] master - https://github.com/rcbops/rpc-openstack/pull/2182
- [x] newton-14.0 - https://github.com/rcbops/rpc-openstack/pull/2217

Run `git log --pretty=oneline | head -n 1` and provide the output below between
the backtick lines:

```
c10c556e5b2f2c897b6cff773b89adc7cb8e0183 Remove RabbitMQ default version (#2168)
```

## What happened?
In the card https://github.com/rcbops/rpc-openstack/issues/2095, some investigation was done to see why we were getting the error ``Block Device Mapping is Invalid: failed to get volume 43ada3f3-0af3-42ff-a8dc-b961a0543f2a. (HTTP 400) (Request-ID: req-43e456c2-9446-4cdc-8a91-bf0e431915
8b)`` when trying to create an instance from a bootable volume. Turns out that this error had nothing to do with inconsistant AZs across cinder and nova, but is a legitimate upstream bug:
https://bugs.launchpad.net/nova/+bug/1648324

## What were you doing when it happened?
This was brought to our attention when @mkam and @mattt416 where investigating some tempest failures around creating instances from a volume. After further investigation, we found https://bugs.launchpad.net/nova/+bug/1648324

## What did you expect to happen instead of what actually happened?

I expected an instance booted from volume to be successfully created.

alextricity25:
Because of the timing of things, it's going to be a little difficult to get the default ``nova_cross_az_attach`` removed in the r14.0.0 tag. The fix for this won't make it until the r14.0.1 tag.
In the meantime, we should create a doc issue under "known issues" for the Newton release.

claco:
"get the default nova_cross_az_attach removed "   Making it what, True? Would that do more harm than good? We should get some direction from support on this one.

Wr also didn't see boot from volume thing when we were testing backed by Ceph did we?

alextricity25:
> "get the default nova_cross_az_attach removed " Making it what, True?

That's right.

> Would that do more harm than good? We should get some direction from support on this one.

[...]

PR #2182, opened 2017-04-20
Comment out nova_cross_az_attach

alextricity25:
This will cause the value to take it's default value of True.
This value cannot be false when trying to boot an instance from
a volume. See connected card for more details.

Connects https://github.com/rcbops/rpc-openstack/issues/2179

### DIFF
--- a/rpcd/etc/openstack_deploy/user_osa_variables_defaults.yml
+++ b/rpcd/etc/openstack_deploy/user_osa_variables_defaults.yml
@@ -56,7 +56,11 @@ nova_cpu_allocation_ratio: 2.0
 nova_ram_allocation_ratio: 1.0
 
 # Nova config overrides
-nova_cross_az_attach: False
+# NOTE: Due to a bug with nova, this variable cannot be set to true
+# if the environment needs to be able to boot an instance from
+# a volume
+# https://bugs.launchpad.net/nova/+bug/1648324
+#nova_cross_az_attach: False
 nova_console_type: novnc
 
 # RabbitMQ overrides


claco:
@alextricity25, @BjoernT 

So, I'm going to ask some questions because curious and don't remember. :-)

Historically, we set it to False for a reason. I assume that reason was to prevent people from doing wrong things, like booting/connecting to a volume in another AZ, because of performance.

So, if we set this to true now just to get things to pass, are we not putting people back into the position of shooting themselves in the foot?

How many greenfields would be impacted if we left if to False? All? Majority? Few?

Do we, and when do we, set it back to False? IF we will, see first 2 questions.  :-)

BjoernT:
@claco Yes the reason is correct but we have to set it to true because of a nova bug where it false would lead to nova not being able to boot from cinder volumes. It may lead to the situation that we misconfigure greenfields correctly but I don't see any other alternative right now.
I will discuss this internally that everyone is aware why that parameter exist, perhaps we add some release notes around that

mattt416:
recheck_cit_ceph

mattt416:
recheck_cit_swift

alextricity25:
recheck_cit_swift

cloudnull:
recheck_cit_ceph

cloudnull:
recheck_cit_ceph

alextricity25:
recheck_cit_swift
recheck_cit_xenial

alextricity25:
recheck_cit_xenial

alextricity25:
recheck_cit_xenial

PR #2186, opened 2017-04-21
remove tasks that are no longer of any value

cloudnull:
The beaver processes were removed from the cluster in mitaka so in
newton+ these tasks no longer server any purpose.

Additionally, the run-upgrade.yml playbook does nothing.


prometheanfire:
Why isn't this merged with all your other stuff into a megapatch, are you sick?

PR #2188, opened 2017-04-21
Removed the gen-facts playbook

cloudnull:
This playbook is slowing down the fact gathering process for targetted
plays. This has been replaced with the playbook keyword, "gather_facts"
which has a value of "true".


cloudnull:
approved by 2 cores, merging.

PR #2189, opened 2017-04-21
Remove the patcher role and playbook

cloudnull:
This role has no purposes at this point. If we need to patch something
in OpenStack then we should do so using a fork. This brings clarity to
any patches that we may need in the future.


cloudnull:
recheck_cit_swift

cloudnull:
recheck_cit_ceph

odyssey4me:
The only time this was used IIRC was for the kilo release in order to patch nova.

alextricity25:
I think it would be good to have a release note here saying that this is being deprecated.

PR #2191, opened 2017-04-21
Remove the ansible.cfg file

cloudnull:
This file creates a global override when playbooks are executed within
this path. Within upstream OSA we use an rc file to control pathing and
options within the ansible stack. This change simply exports our needed
option within deploy.sh instead of carrying the more heavy handed
ansible.cfg file.


cjloader:
recheck_cit_xenial

Issue #1996, opened 2017-03-07
horizon extentions is pulling down master branch breaking horizon 


BjoernT:
###### Please provide the output to 'git describe --all' for  all deployments where you have confirmed it occurs.

Liberty 12.2.2 

##### What type of issue is this? Please tick the appropriate box.
- [x] bug - existing functionality that does not work for you as documented
- [ ] feature - new functionality that you would like to see added

##### What is the expected behavior?

Locking SHA to a version supported by horizon and liberty

##### What is actually happening?

Per https://github.com/rcbops/rpc-openstack/blob/liberty-12.2/rpcd/playbooks/roles/horizon_extensions/defaults/main.yml#L17

we are not setting any upper constraint and including, depending when you install RPC, the file ui-bootstrap-0.9.0.js which breaks horizon

```
root@infra01_horizon_container-a45802a8:/usr/local/lib/python2.7/dist-packages/horizon_extensions-1.0.0-py2.7.egg/rackspace/static/rackspace/js# ll
total 164
drwxr-sr-x 2 root staff   4096 Mar  6 14:53 ./
drwxr-sr-x 5 root staff   4096 Mar  6 14:53 ../
-rw-r--r-- 1 root staff   7687 Mar  6 14:53 heat.js
-rw-r--r-- 1 root staff    147 Mar  6 14:53 login.js
-rw-r--r-- 1 root staff   1551 Mar  6 14:53 template_filter.js
-rw-r--r-- 1 root staff 142791 Mar  6 14:53 ui-bootstrap-0.9.0.js
```

Since we installed the liberty back on March 6, the commit  https://github.com/rcbops/horizon-extensions/commit/1970910f4b756adafeda1b64ecb325a3133aab7e is not included

##### What were you attempting to do when you encountered this error (give as much detail as possible)? 

N/A

##### How can we reproduce this issue?

N/A

##### How can we verify your issue is addressed?

N/A

##### What is your evaluation on the importance of this issue, and why?

----------------------------------------------------------------------------------------------------------------------------
###### For each pull request, update the branch line to reference it.
- [ ] master
- [ ] mitaka-13.1
- [ ] liberty-12.2

##### Upgrade impact. Consider the impact on moving between major, minor and patch releases.
- [ ] This change does not impact upgrades
- [ ] This change addresses a potential upgrade impact
- [ ] The upgrade impact is being tracked by another issue.

The issue tracking the upgrade impact is:

##### Testing impact
[...]

PR #2195, opened 2017-04-22
Removed solutions tab, horizon_extensions subtree

cloudnull:
This commit removes the horizon-extensions subtree,
as well as the bits in the horizon_extenstions playbook
related to deploying the solutions tab.

Closes-Bug: #1996
(cherry picked from commit 88859c2a9e005cf53a331ce5de7c0be2582a76ba)
(cherry picked from commit a26b3ec3c3f73b5517778f378e96ab3fed8a1438)


cloudnull:
recheck_cit_all

BjoernT:
Thanks but we would have to tag a release to make this useful, so we could consume it. But it is unlikely that we install/update another liberty environment ? But a tag might be useful nonetheless to include the MAAS changes?

alextricity25:
@BjoernT I've been battling with this question myself for the past month or so. I'll coordinate with the tiger team that worked on the MaaS changes to see when it would be best to create the next patch release.

cloudnull:
@BjoernT You could checkout a given SHA, assuming we're not able to create a new tag.

cloudnull:
recheck_cit_all

scripts/deploy-rpc-playbooks.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
```
major: Should this file be in the cherry pick?

odyssey4me:
It looks like this may need a rebase.

cloudnull:
recheck_cit_all

hughsaunders:
This must be re-proposed against the `mitaka` branch if still required.

Issue #1587, opened 2016-11-21
rpc_pre_upgrade can not run with self signed cert apis


BjoernT:
##### What type of issue is this? Please tick the appropriate box.
- [x] bug - existing functionality that does not work for you as documented
- [ ] feature - new functionality that you would like to see added

##### What is the expected behavior?

Allow ostack api calls to be made against self signed cert apis

##### What is actually happening?

```
TASK: [rpc_pre_upgrade | Gather nova service-list] ****************************
failed: [localhost ->infra03_utility_container-1588bacf] => {"changed": true, "cmd": "source ~/openrc; nova service-list", "delta": "0:00:00.581531", "end": "2016-11-21 20:46:21.218308", "rc": 1, "start": "2016-11-21 20:46:20.636777", "warnings": []}
stderr: /usr/local/lib/python2.7/dist-packages/requests/packages/urllib3/util/ssl_.py:100: InsecurePlatformWarning: A true SSLContext object is not available. This prevents urllib3 from configuring SSL appropriately and may cause certain SSL connections to fail. For more information, see https://urllib3.readthedocs.org/en/latest/security.html#insecureplatformwarning.
  InsecurePlatformWarning
ERROR (SSLError): SSL exception connecting to https://xxxx.net:5000/v3/auth/tokens: [Errno 1] _ssl.c:510: error:14090086:SSL routines:SSL3_GET_SERVER_CERTIFICATE:certificate verify failed

FATAL: all hosts have already failed -- aborting
```
##### How can we reproduce this issue?

Run rpc_pre_upgrade play against self signed cert protected API

##### How can we verify your issue is addressed?

No more error

##### If you are using or know of a workaround please describe the necessary steps?

Just adding --insecure to the roles/rpc_pre_upgrade/tasks/gather_stats.yml play

##### What is your evaluation on the importance of this issue, and why?

Prio 2 since we have a lot of public api running with self signed certs

----------------------------------------------------------------------------------------------------------------------------
###### For each pull request, update the branch line to reference it.
- [ ] master
- [ ] mitaka-13.1

##### Upgrade impact. Consider the impact on moving between major, minor and patch releases.
- [ ] This change does not impact upgrades
- [ ] This change addresses a potential upgrade impact
- [ ] The upgrade impact is being tracked by another issue.

The issue tracking the upgrade impact is:

##### Testing impact
- [ ] An issue/pull-request has been create or exists to track the testing gap, and is referenced below
[...]

PR #2198, opened 2017-04-22
Add the --insecure flag to gather_stats task file

cloudnull:
This change adds the "--insecure" to the needed OpenStack CLI commands
so that should it encounter an API using a self signed certificate it
can continue to function as normal. Note this does not have a negative
impact if no SSL be found or if there's a valid certificate.

Closes-Bug: #1587


alextricity25:
This will also need to be backported to the newton-14.0 branch. Thanks @cloudnull!

BjoernT:
@cloudnull I'm not a fan of hard coding `--insecure` and prefer to make it conditional to openrc_insecure flag, if this still exists. What do you think

PR #2201, opened 2017-04-23
[newton-14.0] Update OSA SHA

alextricity25:
Connects https://github.com/rcbops/rpc-openstack/issues/2200


alextricity25:
recheck_cit_swift

PR #2202, opened 2017-04-24
Cleanup artifact_version

evrardjp:
Apt artifact building was using artifacting vars, which was not
rpc_release, and rpc_release was used for testing, which lead to
failures.  Using our (now standard) rpc_release everywhere,
we should be able to test and build artifacts.

To avoid duplication of work for the definition of rpc_release,
I kept the derive artifact version.

However, because of the derive artifact version shouldn't read
the git details because it's now set in a static file, we should
simplify its content.

That's what's done here.

Connected https://github.com/rcbops/u-suk-dev/issues/1588


scripts/functions.sh
```diff
@@ -102,7 +102,7 @@ function apt_sources_use_rpc_apt_artifacts {
   fi
 
   # Derive the rpc_release version from the branch name
-  RPCO_VERSION=${RPCO_VERSION:-$(/opt/rpc-openstack/scripts/artifacts-building/derive-artifact-version.py)}
+  RPCO_VERSION=${RPCO_VERSION:-$(/opt/rpc-openstack/scripts/artifacts-building/derive-artifact-version.sh)}
```
odyssey4me: lines 80-102 can be removed if we're not using a branch name for the derivation

evrardjp: Yup Further cleanup to be done in other commits: One for ansible vars, One for shell scripts.

evrardjp:
recheck_cit_apt

evrardjp:
recheck_cit_artifact-trusty-Apt

evrardjp:
recheck_cit_artifact_trusty_Apt

evrardjp:
recheck_cit_artifact_trusty_Apt

evrardjp:
recheck_cit_artifact_trusty_Apt

PR #2210, opened 2017-04-24
Corrects variable name in use

dankolbrs:
Corrects neutron_venv_bin to use the correct neutron variable. This
has no change in functionality and is instead used to match previous
variable convention and use correct variable name.

Backport pull requests already include this change.

Connects rcbops/u-suk-dev#857


major:
Looks like there's a merge conflict now, @dankolbrs.

dankolbrs:
Got it fixed up @major

PR #2220, opened 2017-04-26
Remove RPC-O unused/old variables

evrardjp:
This is a branch were we cleanup clutter accumulated on RPC-O.

Please have a look at each commit for changes and their reason.

### DIFF
--- a/rpcd/etc/openstack_deploy/user_osa_variables_defaults.yml
+++ b/rpcd/etc/openstack_deploy/user_osa_variables_defaults.yml
@@ -73,6 +73,7 @@ rabbitmq_ulimit: 65535
 # since memory utilization is primarily driven by the objects placed in memcached.
 memcached_connections: 16384
 
+#TODO (evrardjp): Remove this when ocata
 # Keystone overrides
 keystone_token_driver: "sql"
 
@@ -95,8 +96,6 @@ pkg_locations:
   - /etc/ansible/roles
   - /etc/openstack_deploy
   - /opt/rpc-openstack/rpcd
-
-horizon_venv_tag: "{{ openstack_release }}"
 
 # List of files to override using the OSA horizon role
 rackspace_static_files_folder: "/opt/rpc-openstack/logo_files"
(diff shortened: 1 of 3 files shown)


rpcd/etc/openstack_deploy/user_osa_variables_defaults.yml
```diff
@@ -73,6 +73,7 @@ rabbitmq_ulimit: 65535
 # since memory utilization is primarily driven by the objects placed in memcached.
 memcached_connections: 16384
 
+#TODO (evrardjp): Remove this when ocata
 # Keystone overrides
 keystone_token_driver: "sql"
```
odyssey4me: I thought that we'd switched to fernet tokens for Newton?

evrardjp: Fernet is token provider.
If something else is used as token driver, you can define token persistence driver. We force it to sql. Which is fine until this is gonna be deprecated.

odyssey4me: Hmm, I'm not entirely sure that you can persist a fernet token considering that the token is provided by the client for every interaction.

mancdaz: wouldn't the persistence driver just be ignored if you are using fernet provider...which is fine?

evrardjp: @mancdaz yes indeed. So no need to ship the variable by default, except if the official stance currently (until O is released) is "When not using fernet, you must be using UUID + SQL tokens, else you'd be alone".

evrardjp: I can remove this altogether, but keep in mind we are still in N.

odyssey4me: Ah, so I see what's happening here. Why don't we make OSA smarter instead of carrying this? In OSA there's a note to use the sql token driver if you're using sql for token storage, so why don't we adjust it so that this happens automatically instead of requiring the user to set it?

evrardjp: OSA was doing something different, if not fernet (read: UUID nowadays) it was using memcache as backend. Here RPC forced to have sql persistence instead. 

I'd rather move to fernet tokens, and keep this as long as we have a fallback to uuid, because we (knowingly) already took a different path than OSA.

evrardjp: The alternative would be to have, in OSA, a sql fallback in all cases, and drop the memcache. I have a commit somewhere for that.

PR #2222, opened 2017-04-26
Fix python artifact build for ceilometer/octavia

odyssey4me:
The python artifact build for octavia fails as there is no
git source information provided to match the presence of
the ansible role inclusion in ansible-role-requirements.

The python artifact build for ceilometer fails as the wheel
build process does not properly understand how to process
the optional dependency requirements.

The Mitaka upper constraint for elasticsearch is removed
so that the constraint from Newton is applied instead.

(cherry picked from commit f3595ae5ce30d4b54967e9d08ccf83b14e667558)


major:
@odyssey4me Does this need to go back to newton-14.0?

PR #2223, opened 2017-04-26
[kilo] Correct ansible-role-requirements

odyssey4me:
The openstack-ansible ansible-role-requirements.yml file
is specified in galaxy format and no longer works. In this
patch we ensure that the same role is specified using the
git format so that it's always retrievable. This ensures
that the ansible bootstrap works properly.

Also, the ceph roles which were specified in this file are
removed as this file is never used. The ceph roles are
instead included using submodules.

Connects https://github.com/rcbops/u-suk-dev/issues/1515

### DIFF
--- a/ansible-role-requirements.yml
+++ b/ansible-role-requirements.yml
@@ -1,9 +1,5 @@
-- name: ceph-common
-  src: https://github.com/ceph/ansible-ceph-common.git
-  version: 1c623611a1d43b4c9116d1a0c21f0bdbdd87a0e7
-- name: ceph-mon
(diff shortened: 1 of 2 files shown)


evrardjp:
3 birds one stone:
- Fixing the inconsistency of a-r-r with in tree submodule
- Fix the absence of sshd role by default (by adding it in a-r-r and using this in the bootstrap process by exporting)
- Fixing the absence of sshd role SHA (absence of role version in openstack-ansible a-r-r).

Good job.

evrardjp:
Verified sha, var name, all that jazz. Should be good.

PR #2224, opened 2017-04-26
maas_target_alias is already the default

evrardjp:
This override don't need to exist:
- Its value is already the group default
- It's not overriden by any other mechanism
- It's not re-use outside the role scope.

Connected https://github.com/rcbops/u-suk-dev/issues/1620

### DIFF
--- a/rpcd/etc/openstack_deploy/user_rpco_variables_defaults.yml
+++ b/rpcd/etc/openstack_deploy/user_rpco_variables_defaults.yml
@@ -49,7 +49,6 @@ rpc_repo_path: /opt/rpc-openstack/openstack-ansible
 # By default we will create an agent token for each entity, however if you'd
 # prefer to use the same agent token for all entities then specify it here
 #maas_agent_token: some_token
-maas_target_alias: public0_v4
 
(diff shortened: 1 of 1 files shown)


evrardjp:
the CIT/ceph job got an issue with rpc_support role, not rpc_maas role. I'll dig deeper into this to see what needs fixing.

evrardjp:
how this https://github.com/rcbops/rpc-openstack/commit/d49c4a266fb00f853ace69051a3a6e526d605c68 get through the door, that's a good question.

evrardjp:
Instead of continuing to cherry pick all what's wrong in rpc-o, I'll just wait for the other patches to merge, this one should be auto-rebased on top.

i.e.: read this message as a "Don't workflow this unless the rest is merged".

hughsaunders:
recheck_cit_xenial

evrardjp:
@hughsaunders This can wait for other patches to merge before rechecking.

evrardjp:
recheck_cit_ceph
recheck_cit_swift
recheck_cit_xenial

Issue #2227, opened 2017-04-26
rpc-support.yml failing trying to create default flavors


alextricity25:
> **THIS IS A PUBLIC FACING ISSUE.**
>
> Do not include private or confidential information in this issue.

## Is this a feature or a bug?

- [ ] Feature _(request for enhancement, new functionality)_
- [x] Bug _(something is broken)_

## Which version of rpc-openstack does this issue relate to?
- [ ] master
- [ ] newton-14.0

Run `git log --pretty=oneline | head -n 1` and provide the output below between
the backtick lines:

```
9f2812e0cb89c8b561d6b546d7fee3789ec68cdd Merge pull request #2222 from rcbops/odyssey4me/octavia-ceilometer-build-fix
```

## What happened?

The `rpc-support.yml` playbook is failing on the task ``TASK [rpc_support : Add default flavors]`` with the following error:
failed: [ramcp-209-7c99_utility_container-c5118c3b] (item={u'vcpus': u'1', u'disk': u'1', u'ram': u'512', u'name': u'm1.tiny'}) => {"failed": true, "item": {"disk": "1", "name": "m1.tiny", "ram": "512", "vcpus": "1"}, "msg": "Error fetching flavor list (Inner Exception: __init__() got an unexpected keyword argument 'app_name')"}

The error can be seen on the following jenkins jobs:
https://rpc.jenkins.cit.rackspace.net/job/RPC-AIO_master-ceph-pr/209/consoleFull
https://rpc.jenkins.cit.rackspace.net/job/RPC-AIO_master-swift-pr/205/console

## What were you doing when it happened?
This failure was noticed in the gate jobs referenced above.

## What did you expect to happen instead of what actually happened?

The task runs successfully

alextricity25:
I've compared the version of shade between a successfully job and a job that failed on this tasks, and they were both 1.16.0. We can rule out shade version inconsistencies as the problem.

alextricity25:
Looking into this now

evrardjp:
I know how to fix it.

evrardjp:
#2229 Is including the flavor fix, but also the flake8 fix.

evrardjp:
Should be fixed.

alextricity25:
Thanks @evrardjp. The changes to the flavor creation task will also need to go in the newton-14.0 branch.

evrardjp:
I can backport it for you.

alextricity25:
@evrardjp Can this close?

[...]

PR #2229, opened 2017-04-27
Use CLI to create flavors

evrardjp:
There is no good version (read: with compatible requirements)
for together shade, pbr, os client tools, and another conflicting
package.

This removes the dependency to shade and works out of the box.
Idempotency is ensured for existing/non existing flavors, without
checking their values.

This can be rolled back when the requirements will be bumped (i.e.
when we are in an upper branch, meant for the usage of shade and
these modules)

Connected https://github.com/rcbops/rpc-openstack/issues/2227

### DIFF
--- a/rpcd/playbooks/roles/rpc_support/tasks/create_default_flavors.yml
+++ b/rpcd/playbooks/roles/rpc_support/tasks/create_default_flavors.yml
@@ -13,33 +13,21 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-- name: Install shade and its dependencies
-  pip:
-    name: "{{ item.name }}"
-    version: "{{ item.version }}"
-    extra_args: "--isolated"
-  with_items:
-    - { name: "shade", version: "1.16.0"}
-
+# TODO(evrardjp): Move back to the previous commit with os_ module in
+# a more recent branch, with compatible requirements AND playbook
+# will be targetting to good hosts directly.
 - name: Add default flavors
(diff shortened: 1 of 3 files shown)


rpcd/playbooks/roles/rpc_support/tasks/create_default_flavors.yml
```diff
@@ -13,33 +13,21 @@
...
-    ram: "{{ item.ram }}"
-    auth:
-      auth_url: "{{ keystone_service_internalurl }}"
-      username: "admin"
-      password: "{{ keystone_auth_admin_password }}"
-      project_name: "admin"
-      domain_name: "Default"
-    validate_certs: false
+  shell: |
+    openstack --os-cloud default flavor create --ram {{ item.RAM }} --disk {{ item.Disk }} --vcpus {{ item.VCPUs }} --public {{ item.Name }}
```
odyssey4me: I think you may need to source the openrc file, unless the clouds.yaml file is automatically discovered?

evrardjp: yeah clouds.yaml always auto discovered. I think it's the os_client lib that takes care of this.

evrardjp:
Also, good point, I changed the dict variable names to match upstream column headers, if we ever need to fetch results and do a | difference.

alextricity25:
@evrardjp Could we get the flavor CLI change backported to the newton-14.0 branch?

PR #2230, opened 2017-04-27
[kilo] Allow builds to be executed on new kernels

odyssey4me:
This change conditionally includes the "scsi_dh" module which
is no longer available when using the 4.4 kernel.

This is a patcher role implementation of:
https://github.com/openstack/openstack-ansible/commit/8d40626145c566f9e5c1a0ca40f4121b78e1b18f

Connects https://github.com/rcbops/u-suk-dev/issues/1629

### DIFF
--- a/rpcd/patches/openstack-hosts-u-suk-dev-issues-1629.patch
+++ b/rpcd/patches/openstack-hosts-u-suk-dev-issues-1629.patch
@@ -1,0 +1,13 @@
+diff --git a/playbooks/roles/openstack_hosts/defaults/main.yml b/playbooks/roles/openstack_hosts/defaults/main.yml
+index 2c305552..30d6a238 100644
+--- a/playbooks/roles/openstack_hosts/defaults/main.yml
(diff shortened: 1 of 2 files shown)


rpcd/patches/openstack-hosts-u-suk-dev-issues-1629.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/playbooks/roles/openstack_hosts/defaults/main.yml b/playbooks/roles/openstack_hosts/defaults/main.yml
```
cloudnull: I kinda hate this, can we not just fork the role in rcbops and make the change?

evrardjp: I hate it too, but kilo. Forking would encourage ppl to do the wrong thing and patch things there. 

Having an ugly solution is maybe a good solution in some cases.

cloudnull: OK, I'm good with that.

PR #2232, opened 2017-04-27
Fix PEP8/Flake8 errors for release.py

evrardjp:

PR #2234, opened 2017-04-27
RI-19 Bump OSA SHA to tip of stable/ocata

alextricity25:
OpenStack-Ansible Ocata Release Notes:
https://docs.openstack.org/releasenotes/openstack-ansible/ocata.html

The following changes have been made to comply with upstream
changes:

* The bootstrap scenario variable ``scenario`` has been changed per
  openstack/openstack-ansible@cd8f931
  (release note provided)

* OSA no longer supports the use of custom scenarios due to the way
  it builds out the user_variables template filename. An upstream
  patch has been submitted to change this: https://review.openstack.org/#/c/459329/1
  (release note provided)

* Move elasticsearch version variables to user_variable files.
  The elasticsearch version variables needs to be available to
  the repo_build role due to the way it builds out the distro
  packages dataset. Distro packages within
  a role are indexed using their raw form(jinja is not rendered).
  The repo_build role then attempts to render these datasets,
  and will fail if it cannot resolve a variable. Putting these
  variables in user_variables ensures the repo_build
  role can interpolate the variables correctly.
  (release note provided)

* Change references of `ansible_ssh_host` to `ansible_host`
  (release note provided)

* Remove elasticsearch constraints in role's ``defaults/main.yml``
  The constraint is now respected in OpenStack's upper-constraints.
  Additionally, the elasticsearch role has been modified to respect
  these upper-constraints when install pip packages.
  (release note provided)

* Put quotes and brakets around the "logstash_apt_packages" variable

* Bump the rpc-maas role SHA to fix some issues with python module
  versions: rcbops/rpc-maas#246

* Bump the openstack-ops role SHA to bring in some fixes for ansible 2.2.2

* The ansible version in tox.ini has been changed to ansible 2.2.2

* Removed a work-around in rpcd/playbooks/setup-maas.yml that is no longer
  required as a result of the rpc-maas role SHA bump.

* Changed tempest execution so that the tasks in the os_tempest role are used

### DIFF
--- a/group_vars/all/rpc-o.yml
+++ b/group_vars/all/rpc-o.yml
@@ -29,11 +29,3 @@ horizon_custom_uploads:
 # Use RPC python package index
 repo_build_pip_extra_indexes:
   - "https://rpc-repo.rackspace.com/pools"
-
-# Tempest test execution options
-tempest_run_tempest_opts:
-  - "--serial"
-# Tempest testr options
-tempest_testr_opts: []
-# Tempest tests to run. A one-line, space-delimited string
-tempest_test_sets: "scenario defcore cinder_backup"
--- a/releasenotes/notes/add-upper-constraints-to-es-c2eb0b51c83e30e2.yaml
+++ b/releasenotes/notes/add-upper-constraints-to-es-c2eb0b51c83e30e2.yaml
@@ -1,0 +1,5 @@
+---
+features:
+  - |
+    The elasticsearch role now respects upper constraints when installing
+    pip packages.
--- a/releasenotes/notes/move-es-version-vars-17c22cfd558a68d5.yaml
+++ b/releasenotes/notes/move-es-version-vars-17c22cfd558a68d5.yaml
@@ -1,0 +1,6 @@
+---
+fixes:
+  - |
+    The ``elasticsearch_version`` and ``elasticsearch_reindex_version``
+    variables have been moved to ``group_vars/all/rpc-o.yml``
+    so that all OSA roles can have access to the variables.
--- a/releasenotes/notes/remove-elasticsearch-constraint-9dae50a1b4aeb7e6.yaml
+++ b/releasenotes/notes/remove-elasticsearch-constraint-9dae50a1b4aeb7e6.yaml
@@ -1,0 +1,7 @@
+---
+fixes:
+  - |
+    The ``elasticsearch>=2.0.0,<3.0.0`` constraint has been removed from the
+    ElasticSearch role, and is now controlled via OpenStack upper-constraints
+    placed here:
+    https://github.com/openstack/requirements/blob/b6c588574ed23e7e5b8c1b788c7117f7c3aa4cff/upper-constraints.txt#L124
--- a/rpcd/playbooks/roles/elasticsearch/defaults/main.yml
+++ b/rpcd/playbooks/roles/elasticsearch/defaults/main.yml
@@ -42,7 +42,7 @@ elasticsearch_version: 2.4.1
 elasticsearch_reindex_version: 1.7.5
 
 elasticsearch_pip_packages:
-  - elasticsearch>=2.0.0,<3.0.0
+  - elasticsearch
   - elasticsearch-curator==4.0.4
   - httplib2
 
--- a/rpcd/playbooks/roles/elasticsearch/tasks/elasticsearch_install.yml
+++ b/rpcd/playbooks/roles/elasticsearch/tasks/elasticsearch_install.yml
@@ -31,7 +31,9 @@
   pip:
     name: "{{ item }}"
     state: latest
-    extra_args: "{{ pip_install_options | default('') }}"
+    extra_args: >-
+      {{ (pip_install_upper_constraints is defined) | ternary('--constraint ' + pip_install_upper_constraints | default(''),'') }}
+      {{ pip_install_options | default('') }}
   register: install_pip_packages
   until: install_pip_packages | success
   retries: 5
--- a/rpcd/playbooks/roles/filebeat/defaults/main.yml
+++ b/rpcd/playbooks/roles/filebeat/defaults/main.yml
@@ -31,7 +31,7 @@ filebeat_logstash_port: "{{ logstash_beats_port |default(5044) }}"
 # NOTE(sigmavirus24): Once we upgrade to Ansible 2.1, we can use the extract
 # filter
 #"{{ groups['logstash_all'] | map('extract', hostvars, ['container_address']) | list }}"
-filebeat_logstash_hosts: "{% for host in groups['logstash_all'] %}{{ hostvars[host]['ansible_ssh_host'] }}:{{ filebeat_logstash_port }}{% if not loop.last %},{% endif %}{% endfor %}"
+filebeat_logstash_hosts: "{% for host in groups['logstash_all'] %}{{ hostvars[host]['ansible_host'] }}:{{ filebeat_logstash_port }}{% if not loop.last %},{% endif %}{% endfor %}"
 
 # NOTE(sigmavirus24): Explanation of the regular expression:
 # - The Date is formatted by oslo.log as YYYY-MM-DD
--- a/scripts/bootstrap-aio.yml
+++ b/scripts/bootstrap-aio.yml
@@ -21,6 +21,9 @@
       set_fact:
         user_variables_overrides:
           apply_security_hardening: "{{ rpco_deploy_hardening }}"
+          # Tempest is turned off to prevent the tests from running by default
+          tempest_run: no
+          tempest_install: no
       when: "{{ not rpco_deploy_ceph | bool }}"
     - name: Set ceph user_variables override
       set_fact:
@@ -28,6 +31,9 @@
           apply_security_hardening: "{{ rpco_deploy_hardening }}"
           glance_default_store: rbd
           nova_libvirt_images_rbd_pool: vms
+          # Tempest is turned off to prevent the tests from running by default
+          tempest_run: no
+          tempest_install: no
       when: "{{ rpco_deploy_ceph | bool }}"
   vars:
     rpco_deploy_ceph: "{{ lookup('env', 'DEPLOY_CEPH') }}"
@@ -38,6 +44,8 @@
   vars:
     bootstrap_host_apt_distribution_suffix_list: "{{ (lookup('env', 'RPCO_APT_ARTIFACTS_AVAILABLE') | bool) | ternary([], ['updates', 'backports']) }}"
     scenario: "{% if lookup('env', 'DEPLOY_MAGNUM') == 'yes' %}magnum{% else %}{{ lookup('env', 'DEPLOY_CEPH') |bool |ternary('ceph','swift') }}{%endif%}"
+    bootstrap_host_scenario: "{% if lookup('env', 'DEPLOY_MAGNUM') == 'yes' %}magnum{% else %}{{ lookup('env', 'DEPLOY_CEPH') |bool |ternary('ceph','swift') }}{%endif%}"
+    bootstrap_user_variables_template: "user_variables.aio.yml.j2"
     bootstrap_host_user_variables_filename: "user_osa_variables_overrides.yml"
     bootstrap_host_user_secrets_filename: "user_osa_secrets.yml"
     uca_enable: no
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -171,7 +171,7 @@ if [[ "${DEPLOY_OA}" == "yes" ]]; then
 
   if [[ "${DEPLOY_TEMPEST}" == "yes" ]]; then
     # Deploy tempest
-    run_ansible os-tempest-install.yml
+    run_ansible ${BASE_DIR}/scripts/run_tempest.yml
   fi
 
   if [[ "${DEPLOY_RALLY}" == "yes" ]]; then
--- a/scripts/run_tempest.yml
+++ b/scripts/run_tempest.yml
@@ -13,23 +13,53 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Get defcore tests and set tempest_test_whitelist fact
+- name: Get defcore tests and set tempest_test_whitelist fact
+  hosts: utility_all[0]
+  gather_facts: no
+  user: root
+  tasks:
+    - name: Get defcore test list from defcore repo
+      get_url:
+        url: "{{ defcore_url }}"
+        dest: "/opt/defcore_tempest_list_{{ version_number }}"
+      delegate_to: localhost
+
+    - name: Set fact for defcore json file
+      set_fact:
+        defcore_json_file: "{{ lookup('file', '/opt/defcore_tempest_list_'~version_number) | from_json }}"
+
+    - name: Set fact for defcore tests
+      set_fact:
+        defcore_tests: "{{ defcore_tests|default([]) + item.tests.keys() }}"
+      with_items: "{{ defcore_json_file['capabilities'].values() }}"
+
+    - name: Set tempest_test_whitelist fact
+      set_fact:
+        tempest_test_whitelist: "{{ tempest_test_whitelist|default([]) + defcore_tests }}"
+  tags:
+    - tempest_install
+  vars:
+    # 2017.01 corresponds to Ocata
+    version_number: "2017.01"
+    defcore_url: "https://raw.githubusercontent.com/openstack/defcore/master/{{ version_number }}.json"
 
 # Run the tempest playbook to install tempest
 - include: ../openstack-ansible/playbooks/os-tempest-install.yml
   tags:
     - tempest_install
+  vars:
+    tempest_run: no
 
-- name: Execute Tempest Tests
-  hosts: utility[0]
+# Run the tempest tests from the os_tempest role
+- name: Run tempest tests from the os_tempest role
+  hosts: utility_all[0]
+  gather_facts: no
   user: root
-  tasks:
-    - name: Execute tempest tests
-      shell: |
-        export RUN_TEMPEST_OPTS='{{ tempest_run_tempest_opts | join(' ') }}'
-        export TESTR_OPTS='{{ tempest_testr_opts | join(' ') }}'
-        bash /opt/openstack_tempest_gate.sh {{ tempest_test_sets }}
-      changed_when: false
   tags:
     - tempest_execute_tests
-
-
+  tasks:
+    - name: Run Tempest Tests
+      include_role:
+        name: os_tempest
+        tasks_from: tempest_run
--- a/tox.ini
+++ b/tox.ini
@@ -12,7 +12,7 @@ whitelist_externals =
     sed
 deps =
     -rtest-requirements.txt
-    ansible{env:ANSIBLE_VERSION:==2.1.0}
+    ansible{env:ANSIBLE_VERSION:==2.2.2}
     -e./hacking
 setenv =
     ANSIBLE_ACTION_PLUGINS = {homedir}/.ansible/roles/plugins/action
--- a/ansible-role-requirements.yml
+++ b/ansible-role-requirements.yml
@@ -13,19 +13,15 @@
 - name: rcbops_openstack-ops
   scm: git
   src: https://github.com/rsoprivatecloud/openstack-ops
-  version: de006ade2f88790c43ec06a104abc895d98fe9b6
+  version: f87406880f67584b5fe562d2711734bbd63c0fdb
 - name: os_octavia
   scm: git
   src: https://github.com/openstack/openstack-ansible-os_octavia.git
   version: master
-#TODO(odyssey4me):
-# When the rpc_maas SHA is updated to include the below SHA,
-# remove the workaround from rpcd/playbooks/setup-maas.yml
-# https://github.com/rcbops/rpc-maas/commit/be5a9f82592e5dfd835aa76d8a6cff2d7ea53d3f
 - name: rpc_maas
   scm: git
   src: https://github.com/rcbops/rpc-maas
-  version: 1701d34fcb106cec11ba00284ac462e78bef14e0
+  version: 52d9084c8eca07c3bcb66bbb8d7f5ab93a300bc0
 # Override the current lxc_hosts role until
 # https://github.com/openstack/openstack-ansible-lxc_hosts/commit/df227381ff3d318fcb8d7ab925f1ee66ed161ec8
 # gets taken into a SHA bump in OSA and RPC.
@@ -36,4 +32,4 @@
 - name: rpc-role-logstash
   scm: git
   src: https://github.com/rcbops/rpc-role-logstash.git
-  version: e887f843b623c08ba5b0298df4df8ecd6b5303fc 
+  version: 3f5e7a6044629e76acb1933af19f09a89e55917b
--- a/group_vars/all/elasticsearch.yml
+++ b/group_vars/all/elasticsearch.yml
@@ -1,0 +1,10 @@
+# Elasticsearch versions
+# NOTE (alextricity25):
+# The elasticsearch version variables needs to be available to the repo_build role
+# due to the way it builds out the distro packages dataset. Distro packages within
+# a role are indexed using their raw form(jinja is not rendered). The repo_build
+# role then attempts to render these datasets, and will fail if it cannot resolve
+# a variable. Putting these variables in here ensures the repo_build
+# role can interpolate the variables correctly.
+elasticsearch_version: 2.4.1
+elasticsearch_reindex_version: 1.7.5
--- a/group_vars/all/osa.yml
+++ b/group_vars/all/osa.yml
@@ -122,3 +122,9 @@ haproxy_extra_services:
       haproxy_balance_type: tcp
       haproxy_backend_options:
         - "ssl-hello-chk"
+
+# Do not run tempest as part of setup-openstack. Instead, tempest is run in deploy.sh
+# so we can control when to run it apart from setup-openstack.yml
+# See https://github.com/openstack/openstack-ansible/blob/485100fbd36adbacd6f300a34a6acd9df3fd5932/playbooks/setup-openstack.yml#L40
+tempest_install: no
+tempest_run: no
--- a/group_vars/all/release.yml
+++ b/group_vars/all/release.yml
@@ -16,5 +16,5 @@
 # The release tag to use
 # Note that this is set here so that the openstack_release override in osa_defaults
 # is able to use it.
-rpc_release: "r14.1.0rc1"
+rpc_release: "r15.0.0rc1"
 
--- a/releasenotes/notes/rename-ansible-ssh-host-7332cc551f494495.yaml
+++ b/releasenotes/notes/rename-ansible-ssh-host-7332cc551f494495.yaml
@@ -1,0 +1,6 @@
+---
+deprecations:
+  - |
+    ``ansible_ssh_host`` has been removed from Ansible 2.2 and renamed to
+    ``ansible_host``. All references of ``ansible_ssh_host`` in RPCO have
+    been renamed to ``ansible_host``.
--- a/rpcd/playbooks/roles/elasticsearch/tasks/elasticsearch_post_install.yml
+++ b/rpcd/playbooks/roles/elasticsearch/tasks/elasticsearch_post_install.yml
@@ -44,7 +44,7 @@
 
 - name: Wait for ElasticSearch port
   wait_for:
-    host: "{{ ansible_ssh_host }}"
+    host: "{{ ansible_host }}"
     port: "9200"
 
 - name: Setup the ElasticSearch Curator cron job
--- a/rpcd/playbooks/roles/kibana/tasks/kibana_install.yml
+++ b/rpcd/playbooks/roles/kibana/tasks/kibana_install.yml
@@ -39,4 +39,3 @@
   with_items: "{{ kibana_pip_packages }}"
   tags:
     - kibana-pip-packages
-
--- a/rpcd/playbooks/setup-maas.yml
+++ b/rpcd/playbooks/setup-maas.yml
@@ -24,11 +24,5 @@
   gather_facts: true
   roles:
     - role: "rpc_maas"
-      #TODO(odyssey4me):
-      # This is a workaround until the below commit is included in the
-      # rpc-maas SHA we consume.
-      # https://github.com/rcbops/rpc-maas/commit/be5a9f82592e5dfd835aa76d8a6cff2d7ea53d3f
-      holland_venv_enabled: true
-      holland_venv_bin: "/openstack/venvs/holland-{{ rpc_release }}/bin"
   tags:
     - maas-setup


alextricity25:
Blocked by: rcbops/rpc-openstack#2232

rpcd/etc/openstack_deploy/user_osa_variables_defaults.yml
```diff
@@ -142,6 +142,12 @@ repo_build_upper_constraints_overrides:
   # the optional dependencies specfied here:
   # https://github.com/openstack/ceilometer/blob/stable/newton/setup.cfg#L36
   - "gnocchiclient>=2.2.0,<3.0.0"
+  # The elasticsearch version is pinned here due to a bug within the Ansible
+  # pip module. The Ansible pip module can no longer accept constraints as
+  # part of the 'name' parameter. To work around this, the constraint is added
+  # here then passed to pip using the "extra_args" parameter.
+  # https://github.com/ansible/ansible/issues/22399
+  - "elasticsearch>=2.0.0,<3.0.0"
```
odyssey4me: This is an entirely unnecessary pin as OpenStack already has a pin for elasticsearch: https://github.com/openstack/requirements/blob/stable/ocata/upper-constraints.txt#L129

rpcd/playbooks/roles/elasticsearch/tasks/elasticsearch_install.yml
```diff
@@ -31,7 +31,9 @@
   pip:
     name: "{{ item }}"
     state: latest
-    extra_args: "{{ pip_install_options | default('') }}"
+    extra_args: >-
+      {{ (pip_install_upper_constraints is defined) | ternary('--constraint ' + pip_install_upper_constraints | default(''),'') }}
```
odyssey4me: The upper constraints are never defined for the rpcd/playbooks, but this should be OK as the containers should be configured to be restricted by the pip install role. If you'd like to be extra safe, add https://github.com/openstack/openstack-ansible/blob/stable/ocata/playbooks/inventory/group_vars/all.yml#L64 to the rpcd group_vars/all.yml

rpcd/etc/openstack_deploy/user_osa_variables_defaults.yml
evrardjp: And a quite specific one indeed!
Agreed with its removal.

rpcd/playbooks/roles/elasticsearch/tasks/elasticsearch_install.yml
evrardjp: pip install is a dependency indeed.

For the extra safety, we could have the group vars all in rpc, but I will have a commit soon to remove the need for it.
In the meantime, we can follow what @odyssey4me said.

rpcd/playbooks/roles/filebeat/defaults/main.yml
```diff
@@ -31,7 +31,7 @@ filebeat_logstash_port: "{{ logstash_beats_port |default(5044) }}"
 # NOTE(sigmavirus24): Once we upgrade to Ansible 2.1, we can use the extract
 # filter
 #"{{ groups['logstash_all'] | map('extract', hostvars, ['container_address']) | list }}"
-filebeat_logstash_hosts: "{% for host in groups['logstash_all'] %}{{ hostvars[host]['ansible_ssh_host'] }}:{{ filebeat_logstash_port }}{% if not loop.last %},{% endif %}{% endfor %}"
+filebeat_logstash_hosts: "{% for host in groups['logstash_all'] %}{{ hostvars[host]['ansible_host'] }}:{{ filebeat_logstash_port }}{% if not loop.last %},{% endif %}{% endfor %}"
```
evrardjp: This kind of code should be in a group var instead, IMO.  But because we have no re-usability it should be fine.

scripts/bootstrap-aio.yml
```diff
@@ -36,7 +36,8 @@
 - name: Execute the OSA AIO bootstrap
   include: "{{ lookup('env', 'OA_DIR') }}/tests/bootstrap-aio.yml"
   vars:
-    scenario: "{% if lookup('env', 'DEPLOY_MAGNUM') == 'yes' %}magnum{% else %}{{ lookup('env', 'DEPLOY_CEPH') |bool |ternary('ceph','swift') }}{%endif%}"
+    bootstrap_host_scenario: "{% if lookup('env', 'DEPLOY_MAGNUM') == 'yes' %}magnum{% else %}{{ lookup('env', 'DEPLOY_CEPH') |bool |ternary('ceph','swift') }}{%endif%}"
```
evrardjp: Nit: 
This could be made more clear right by ensuring the shell script is properly outputing the right scenario instead of re doing the logic? One should export the scenario, validate the scenario, and here we should stick to the DEPLOY_SCENARIO.

evrardjp:
Don't we need to pull the reno for that? Or you're 100% relying on the differ?

rpcd/etc/openstack_deploy/user_osa_variables_defaults.yml
alextricity25: Nice catch. I'll remove it.

scripts/bootstrap-aio.yml
alextricity25: That's a very good suggestion! Would it be alright if I save this for another PR?

alextricity25:
> Don't we need to pull the reno for that? Or you're 100% relying on the differ?

@evrardjp Are you talking about putting the releasenotes on this PR?

rpcd/playbooks/roles/elasticsearch/tasks/elasticsearch_install.yml
alextricity25: Thanks! I'll put that in.

evrardjp:
Yeah I think we should have a lean approach on release notes, and write them as soon as we think the introduced change needs to documented for anyone. A more "continuous" approach of the release notes. It could be a good exercise.

Basically, when we sha bump in OSA, we pull from each of the roles their release notes. That could be the same here: You pull OSA release notes when bumping.

I know it's quite different from what we are currently doing (i.e. publishing release notes when publishing a release). What do you think?

claco:
Re @evrardjp  and adding the upstream changes in the SHA bump release notes. +1  The more information, we communicate sooner, the better IMO. Plus it gives the reader of the commit even more information and context to quickly identify possible issues.

scripts/bootstrap-aio.yml
evrardjp: Yes, ofc! It was because you changed the var name that I allowed myself to ask this :p

alextricity25:
@evrardjp:
> Basically, when we sha bump in OSA, we pull from each of the roles their release notes. That could be the same here: You pull OSA release notes when bumping.

I'm all for including the releasenotes on the PR GitHub description, but pulling the OSA releasenotes into the RPCO repo? Wouldn't that mean that RPCO release notes will get mingled with OSA notes? Maybe we should keep those separate.

> I know it's quite different from what we are currently doing (i.e. publishing release notes when publishing a release). What do you think?

For Newton, we didn't use reno at all to compile our release notes that were made in the rpc-docs repo. Those release notes were hand-picked by the individuals who worked on the "Review <service>" cards during the Newton SPIKE effort. **I think it's a great idea to use reno instead, and that is definitely our plan for upcoming releases.**

rpcd/playbooks/roles/elasticsearch/tasks/elasticsearch_install.yml
alextricity25: Done.

rpcd/playbooks/roles/filebeat/defaults/main.yml
odyssey4me: Agreed - the role should default to having no hosts, then the group_var should set the host list.

alextricity25: Thanks for the feedback! I've created a bug report for the request change here: https://github.com/rcbops/rpc-openstack/issues/2237

scripts/bootstrap-aio.yml
alextricity25: Thanks JP! A bug report has been submitted for this request: https://github.com/rcbops/rpc-openstack/issues/2238

alextricity25:
Blocked by: https://github.com/rcbops/u-suk-dev/issues/1642

alextricity25:
UPDATE:
Blocked by: rcbops/u-suk-dev#1642

alextricity25:
Waiting on https://github.com/rcbops/rpc-maas/pull/246 to merge. Once that merges, we need to bump the rpc-maas role sha here.

cjloader:
Once https://github.com/rcbops/rpc-maas/pull/246 merges, will recheck gate

alextricity25:
UPDATE:
There are two PRs out there that need to be merged before we can continue on with this. As of now, they are still pending reviews:
rcbops/rpc-maas#246
rcbops/rpc-openstack#2250

cjloader:
recheck_cit_ceph
recheck_cit_swift

alextricity25:
UPDATE:
There is a [small bug](https://github.com/rcbops/rpc-openstack/issues/2255) in rpc-maas that caused a small disturbance in progress here. That bug is only happening on my local environment due to QEMU DVD-ROM device being mounted.
Right now it seems like the tempest tests are having a hard time running in gate when the SHA is bumped to Ocata:
https://rpc.jenkins.cit.rackspace.net/job/RPC-AIO_master-xenial-pr/256/console

My initial response is that the execution method changed in Ocata, and needs to be adjusted accordingly in gate. I'm currently investigating.

alextricity25:
UPDATE:
I'm currently resolving merge conflicts and investigation the best possible solution for execution tempest in the gate.

Sidenote: I'll be on PTO from Wed - Fri this week. There doesn't seem to be much work left, but perhaps it would be a good idea to have someone look after this PR while I'm gone. @cjloader Would you be available?

cjloader:
Yes I can, @alextricity25.

cjloader:
recheck_cit_all

alextricity25:
UPDATE:
Blocked by: https://github.com/rcbops/rpc-gating/pull/182

It's vital that PR merges before this one, and before we start working on gating jobs for an ocata branch. Note that the aforementioned PR has dependencies, and hence might take a little longer to get through.

alextricity25:
UPDATE:
Still waiting on https://github.com/rcbops/rpc-openstack/pull/2262 and https://github.com/rcbops/rpc-openstack/pull/2261 to merge so we can move forward to unblocking this.

alextricity25:
UPDATE:
As a note to myself: TODO this PR will need to change the run_tempest.yml playbook to consider the new tempest execution method in Ocata.
As of now, this card is still blocked on https://github.com/rcbops/u-suk-dev/issues/1680. The good news is that this is the last link in the dependency chain before we can unblock this.

tox.ini
```diff
@@ -12,7 +12,7 @@ whitelist_externals =
     sed
 deps =
     -rtest-requirements.txt
-    ansible{env:ANSIBLE_VERSION:==2.1.0}
+    ansible{env:ANSIBLE_VERSION:==2.2.0}
```
odyssey4me: You'll likely need to match this to the current version at the SHA you're using.

alextricity25:
Blocked by:
https://github.com/rcbops/rpc-openstack/pull/2298
https://rpc-openstack.atlassian.net/browse/RI-18

alextricity25:
UPDATE:
With the exception of one more pending change in the logstash repo, I'd say that this PR is just about ready to go. I've tested MaaS, tempest tests, rpc-support, and the osa deployment on my AIO and they all ran just fine. The only playbook that doesn't run is the setup-logging.yml playbook. This fix will need to be merged to get that working:
https://github.com/rcbops/rpc-role-logstash/pull/20

alextricity25:
@odyssey4me @evrardjp Could you guys take another look at this if you get the chance?

group_vars/all/rpc-o.yml
```diff
@@ -30,10 +30,21 @@ horizon_custom_uploads:
...
+# role then attempts to render these datasets, and will fail if it cannot resolve
+# a variable. Putting these variables in here ensures the repo_build
+# role can interpolate the variables correctly.
+elasticsearch_version: 2.4.1
+elasticsearch_reindex_version: 1.7.5
+
 # Tempest tests to run. A one-line, space-delimited string
-tempest_test_sets: "scenario defcore cinder_backup"
+tempest_test_whitelist:
+  - "tempest\\.scenario\\.test_(minimum|swift|server|dashboard)_basic(_ops)?"
```
odyssey4me: How does this differ from the defaults? It strikes me that a ceph deployment will fail with these tests thanks to the swift inclusion. Also, many of these tests have been extracted from the tempest repo and are now plugins, it seems that overriding these seems like a path to a bad day if we keep trying to do it as an override.

group_vars/all/rpc-o.yml
```diff
@@ -30,10 +30,21 @@ horizon_custom_uploads:
 repo_build_pip_extra_indexes:
   - "https://rpc-repo.rackspace.com/pools"
 
-# Tempest test execution options
-tempest_run_tempest_opts:
-  - "--serial"
-# Tempest testr options
-tempest_testr_opts: []
+# Elasticsearch versions
```
odyssey4me: Is there a reason these aren't just static values in the role itself?

releasenotes/notes/move-es-version-vars-17c22cfd558a68d5.yaml
```diff
@@ -0,0 +1,6 @@
+---
```
odyssey4me: Why would they need it? I'm not aware of any other role needing access to them.

releasenotes/notes/remove-elasticsearch-constraint-9dae50a1b4aeb7e6.yaml
```diff
@@ -0,0 +1,7 @@
+---
+fixes:
+  - |
+    The ``elasticsearch>=2.0.0,<3.0.0`` constraint has been removed from the
+    ElasticSearch role, and is now controlled via OpenStack upper-constraints
+    placed here:
+    https://github.com/openstack/requirements/blob/stable/ocata/upper-constraints.txt#L129
```
odyssey4me: This line will become ineffective over time due to changes. IT's probably better made a formal SHA-based link.

releasenotes/notes/rename-scenario-var-791bdeaf69a3eeab.yaml
```diff
@@ -0,0 +1,11 @@
+---
```
odyssey4me: Why is this included - it's not an operator-related note, but instead a developer note. This is a test only variable.

releasenotes/notes/use-bootstrap-user-variables-template-var-080d52eb63fb0f84.yaml
```diff
@@ -0,0 +1,6 @@
+---
```
odyssey4me: Again, this is a test only thing. I don't believe that it belongs in release notes which are targeted at Operators.

scripts/bootstrap-aio.yml
```diff
@@ -21,13 +21,17 @@
       set_fact:
         user_variables_overrides:
           apply_security_hardening: "{{ rpco_deploy_hardening }}"
+          # Tempest is turned off to prevent the tests from running by default
+          tempest_run: no
```
odyssey4me: Just add this to vars instead of using the set_fact task.

scripts/run_tempest.yml
```diff
@@ -13,23 +13,41 @@
...
   tags:
     - tempest_execute_tests
-
-
+  tasks:
+    - name: Run Tempest Tests
+      include_role:
+        name: os_tempest
+        tasks_from: tempest_run
+  vars:
```
odyssey4me: This is unnecessary as you're running the task file directly, unless I'm missing some reason for the conditional inclusion?

scripts/run_tempest.yml
```diff
@@ -13,23 +13,41 @@
...
 # Run the tempest playbook to install tempest
 - include: ../openstack-ansible/playbooks/os-tempest-install.yml
   tags:
     - tempest_install
+  vars:
+    tempest_run: no
 
-- name: Execute Tempest Tests
-  hosts: utility[0]
+# Run the tempest tests from the os_tempest role
```
odyssey4me: If you had tempest_run: yes above, it would just run - not sure why this is a second play?

scripts/run_tempest.yml
```diff
@@ -13,23 +13,41 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Get defcore tests and set tempest_test_whitelist fact
+- name: Get defcore tests and set tempest_test_whitelist fact
+  hosts: utility_all[0]
+  gather_facts: no
+  user: root
+  tasks:
+    - include: get_defcore_tests.yml
```
odyssey4me: Why's an included set of tasks being used here? Is there intent to re-use those tasks somewhere? Why not just included them in this play?

scripts/deploy.sh
```diff
@@ -171,7 +171,7 @@ if [[ "${DEPLOY_OA}" == "yes" ]]; then
 
   if [[ "${DEPLOY_TEMPEST}" == "yes" ]]; then
     # Deploy tempest
-    run_ansible os-tempest-install.yml
+    run_ansible ${BASE_DIR}/scripts/run_tempest.yml
```
odyssey4me: It looks like tempest gets deployed, then again. If you set ``tempest_install: no`` in the overrides above, then tempest will only be deployed is ``DEPLOY_TEMPEST == "yes"``.

scripts/bootstrap-aio.yml
```diff
@@ -38,6 +42,8 @@
   vars:
     bootstrap_host_apt_distribution_suffix_list: "{{ (lookup('env', 'RPCO_APT_ARTIFACTS_AVAILABLE') | bool) | ternary([], ['updates', 'backports']) }}"
     scenario: "{% if lookup('env', 'DEPLOY_MAGNUM') == 'yes' %}magnum{% else %}{{ lookup('env', 'DEPLOY_CEPH') |bool |ternary('ceph','swift') }}{%endif%}"
+    bootstrap_host_scenario: "{% if lookup('env', 'DEPLOY_MAGNUM') == 'yes' %}magnum{% else %}{{ lookup('env', 'DEPLOY_CEPH') |bool |ternary('ceph','swift') }}{%endif%}"
```
odyssey4me: This method is not going to scale very well. One more scenario and it's broken. I would recommend finding something simpler. In fact, if we go with https://github.com/rcbops/rpc-gating/pull/208 then we just take the scenario from the axes and stop trying to be clever here.

group_vars/all/rpc-o.yml
alextricity25: Good point. I was just looking at the defaults defined [here](https://github.com/openstack/openstack-ansible-os_tempest/blob/stable/ocata/defaults/main.yml#L59-L66), and it doesn't seem like we need to change anything to include all of the scenario tests that we need. I'll remove this override.

group_vars/all/rpc-o.yml
alextricity25: Putting these variables *only* in the role caused the repo process to fail when indexing the roles for `*_apt_pacakges` variables and their corresponding packages [here](https://github.com/rcbops/rpc-openstack/blob/master/rpcd/playbooks/roles/elasticsearch/vars/ubuntu-16.04.yml#L17) The repo process tried to interpolate these variables when there were being indexed, but failed since they were defined outside the scope of the repo roles. I had to place them her to compromise.

releasenotes/notes/move-es-version-vars-17c22cfd558a68d5.yaml
alextricity25: The repo_build role tries to interpolate these variables when indexing the elasticsearch role for apt packages.

releasenotes/notes/remove-elasticsearch-constraint-9dae50a1b4aeb7e6.yaml
alextricity25: Good point. Will do.

releasenotes/notes/rename-scenario-var-791bdeaf69a3eeab.yaml
alextricity25: Fair point. I'll remove this release note.

releasenotes/notes/use-bootstrap-user-variables-template-var-080d52eb63fb0f84.yaml
alextricity25: Fair point. I'll remove this release note.

scripts/bootstrap-aio.yml
alextricity25: I agree 100%! However, maybe this can be saved for another PR.

group_vars/all/rpc-o.yml
```diff
@@ -30,10 +30,21 @@ horizon_custom_uploads:
...
+elasticsearch_reindex_version: 1.7.5
+
 # Tempest tests to run. A one-line, space-delimited string
-tempest_test_sets: "scenario defcore cinder_backup"
+tempest_test_whitelist:
+  - "tempest\\.scenario\\.test_(minimum|swift|server|dashboard)_basic(_ops)?"
+# Do not run tempest as part of setup-openstack. Instead, tempest is run in deploy.sh
+# so we can control when to run it apart from setup-openstack.yml
+tempest_install: no
+tempest_run: no
```
alextricity25: These will need to move to group_vars/all/osa.yml

scripts/deploy.sh
alextricity25: The ``tempest_install`` variable is already set to ``no`` in the group vars. Is that enough, or am I missing something?

scripts/run_tempest.yml
alextricity25: There's no reason I put those task in a separate file. I can move them in here so we don't clutter up our ``scripts/`` dir.

scripts/run_tempest.yml
alextricity25: Good catch. I'll change that.

scripts/run_tempest.yml
alextricity25: You're absolutely right. I'll address that now.

scripts/run_tempest.yml
alextricity25: I remember now: The reason I had a separate play for this is so we can install tempest or execute tests using the defined tags ``tempest_install`` and ``tempest_execute_tests``. Our gating jobs rely on these tags to run the tempest installation and tempest test execution as two separate stages. This means that I needed a way to run *only* the tempest installation using the tag ``tempest_install``, and a way to run *only* the tempest tests execution using the tag ``tempest_execute_tests``.

releasenotes/notes/move-es-version-vars-17c22cfd558a68d5.yaml
odyssey4me: OK, but why are those package lists variablised? Why not just have them be static names?

scripts/deploy.sh
odyssey4me: Yes, the AIO bootstrap overrides it by default so you'd need to make sure that the RPCO AIO bootstrap replaces that override just like you're doing with ``tempest_run``.

scripts/bootstrap-aio.yml
odyssey4me: Agreed.

scripts/run_tempest.yml
odyssey4me: It seems to me like we should be pulling the tempest test execution out of rpc-gating and into RPC-O so that we can have different per-branch test execution strategies and make it easy to replicate the same tests by hand. That's a larger discussion though.

releasenotes/notes/move-es-version-vars-17c22cfd558a68d5.yaml
alextricity25: I was thinking the same thing. Maybe the role author wanted to make it clear that these package versions are being defined statically? Or maybe they wanted to have the ability to override the package version?

scripts/deploy.sh
alextricity25: Ah! You're right! I'll fix that up now.

alextricity25:
To re-iterate. Only the required jobs (**travis and xenial-deploy-swift jobs**) are expected to pass here.

alextricity25:
UPDATE:
Need a couple of review here before this can merge.

scripts/get_defcore_tests.yml
```diff
@@ -0,0 +1,29 @@
+---
```
odyssey4me: Is this file needed any more now that the tasks are merged into run_tempest.yml ?

releasenotes/notes/move-es-version-vars-17c22cfd558a68d5.yaml
```diff
@@ -0,0 +1,6 @@
...
+fixes:
+  - |
+    The ``elasticsearch_version`` and ``elasticsearch_reindex_version``
+    variables have been moved to ``user_rpco_variables_defaults.yml``
```
odyssey4me: I don't think that this is accurate any more? It's been moved into the group_vars.

releasenotes/notes/add-upper-constraints-to-es-c2eb0b51c83e30e2.yaml
```diff
@@ -0,0 +1,6 @@
+---
+features:
+  - |
+    The elasticsearch role now respects upper constraints when installing
+    pip packages. The upper constriants file is defined in
+    ``rpcd/playbooks/group_vars/all.yml``.
```
odyssey4me: Is it? AFAIK it uses the OSA upper constraints setting, which uses the upstream requirements repo until the repo server exists, when it switches to using the repo's compiled/artifacted constraints.

scripts/run_tempest.yml
```diff
@@ -13,23 +13,53 @@
...
-      changed_when: false
   tags:
     - tempest_execute_tests
-
-
+  tasks:
+    - name: Run Tempest Tests
+      include_role:
+        name: os_tempest
+        tasks_from: tempest_run
```
odyssey4me: This is being skipped because the override file takes precedence of the role defaults. The override file will always take higher precedence, so if you want this to actually execute tests it will need to be executed without the extra vars, or you'll need to add ``-e tempest_run=yes`` on to the command that executes this playbook.

scripts/get_defcore_tests.yml
alextricity25: Duh! No it's not needed. I completely overlooked this. I'll remove that file.

releasenotes/notes/move-es-version-vars-17c22cfd558a68d5.yaml
alextricity25: You're right! I'll change the wording here to reflect that.

scripts/run_tempest.yml
alextricity25: @odyssey4me It's being skipped? Should't these tasks get executed regardless of what `tempest_run` is since the tasks are being run directly?

releasenotes/notes/add-upper-constraints-to-es-c2eb0b51c83e30e2.yaml
alextricity25: I think this is an old release note that hasn't been updated for a while. I'll remove the bits about the upper constraints file, there is no need for that information to be here.

scripts/run_tempest.yml
odyssey4me: If you take a look at the jenkins log, it got skipped. Not a single task in the playbook actually executed... unless I'm missing something?

alextricity25: That's not what I'm seeing:
```
TASK [os_tempest : Execute tempest tests] **************************************
task path: /etc/ansible/roles/os_tempest/tasks/tempest_run.yml:16
Friday 09 June 2017  20:52:36 +0000 (0:00:00.550)       0:00:16.418 *********** 
ok: [ramxdsp-4-f88e_utility_container-52bce778] => {"changed": false, "cmd": ". /openstack/venvs/tempest-r15.0.0rc1/bin/activate\n tempest run --serial --whitelist-file /openstack/venvs/tempest-r15.0.0rc1/workspace/etc/tempest_whitelist.txt", "delta": "0:26:04.496768", "end": "2017-06-09 21:18:41.309174", "rc": 0, "start": "2017-06-09 20:52:36.812406", "stderr": "/openstack/venvs/tempest-r15.0.0rc1/lib/python2.7/site-packages/tempest/clients.py:45: DeprecationWarning: Using the 'client_parameters' argument is deprecated\n  client_parameters=self._prepare_configuration())", "stdout": "running=OS_STDOUT_CAPTURE=${OS_STDOUT_CAPTURE:-1} \\\nOS_STDERR_CAPTURE=${OS_STDERR_CAPTURE:-1} \\\nOS_TEST_TIMEOUT=${OS_TEST_TIMEOUT:-500} \\\n${PYTHON:-python} -m subunit.run discover -t /openstack/venvs/tempest-r15.0.0rc1/lib/python2.7/site-packages/tempest /openstack/venvs/tempest-r15.0.0rc1/lib/python2.7/site-packages/tempest/test_discover --list \nrunning=OS_STDOUT_CAPTURE=${OS_STDOUT_CAPTURE:-1} \\\nOS_STDERR_CAPTURE=${OS_STDERR_CAPTURE:-1} \\\nOS_TEST_TIMEOUT=${OS_TEST_TIMEOUT:-500} \\\n${PYTHON:-python} -m subunit.run discover -t /openstack/venvs/tempest-r15.0.0rc1/lib/python2.7/site-packages/tempest /openstack/venvs/tempest-r15.0.0rc1/lib/python2.7/site-packages/tempest/test_discover  --load-list /tmp/tmpE2oyno\n{0} tempest.api.compute.flavors.test_flavors.FlavorsV2TestJSON.test_list_flavors [0.179945s] ... ok\n{0} tempest.api.compute.flavors.test_flavors.FlavorsV2TestJSON.test_list_flavors_detailed_filter_by_min_disk [0.124427s] ... ok\n{0} tempest.api.compute.flavors.test_flavors.FlavorsV2TestJSON.test_list_flavors_detailed_filter_by_min_ram [0.117361s] ... ok\n{0} tempest.api.compute.flavors.test_flavors.FlavorsV2TestJSON.test_list_flavors_detailed_limit_results [0.042078s] ... ok\n{0} tempest.api.compute.flavors.test_flavors.FlavorsV2TestJSON.test_list_flavors_detailed_using_marker [0.138216s] ... ok\n{0} tempest.api.compute.flavors.test_flavors.FlavorsV2TestJSON.test_list_flavors_filter_by_min_disk [0.130239s] ... ok\n{0} tempest.api.compute.flavors.test_flavors.FlavorsV2TestJSON.test_list_flavors_filter_by_min_ram [0.146180s] ... ok\n{0} tempest.api.compute.flavors.test_flavors.FlavorsV2TestJSON.test_list_flavors_limit_results [0.045058s] ... ok\n{0} tempest.api.compute.flavors.test_flavors.FlavorsV2TestJSON.test_list_flavors_using_marker [0.124572s] ... ok\n{0} tempest.api.compute.flavors.test_flavors.FlavorsV2TestJSON.test_list_flavors_with_detail [0.129413s] ... ok\n{0} tempest.api.compute.images.test_images.ImagesTestJSON.test_delete_saving_image [10.360987s] ... ok\n{0} tempest.api.compute.images.test_images_oneserver.ImagesOneServerTestJSON.test_create_delete_image [12.826685s] ... ok\n{0} tempest.api.compute.images.test_images_oneserver.ImagesOneServerTestJSON.test_create_image_specify_multibyte_character_image_name [6.228896s] ... ok\n{0} tempest.api.compute.images.test_list_image_filters.ListImageFiltersTestJSON.test_list_images_filter_by_changes_since [0.135192s] ... ok\n{0} tempest.api.compute.images.test_list_image_filters.ListImageFiltersTestJSON.test_list_images_filter_by_name [0.119631s] ... ok\n{0} 
......
```

odyssey4me: My bad, I was looking at the wrong thing.

releasenotes/notes/move-es-version-vars-17c22cfd558a68d5.yaml
cloudnull: While I can see why we need this for now I think we should work with the folks creating the elk bits to get these vars removed. These should just use whatever the repo provides which in our case will be contained within our own mirror.

rpcd/playbooks/roles/elasticsearch/defaults/main.yml
```diff
@@ -42,7 +42,7 @@ elasticsearch_version: 2.4.1
 elasticsearch_reindex_version: 1.7.5
 
 elasticsearch_pip_packages:
-  - elasticsearch>=2.0.0,<3.0.0
+  - elasticsearch
   - elasticsearch-curator==4.0.4
```
cloudnull: Is it possible to remove this constraint?

scripts/run_tempest.yml
cloudnull: ++ this should be removed from "rpc-gating". Tests should be written and co-located with project, gating should be nothing more than an executor and artifact collection tool.

tox.ini
```diff
@@ -12,7 +12,7 @@ whitelist_externals =
     sed
 deps =
     -rtest-requirements.txt
-    ansible{env:ANSIBLE_VERSION:==2.1.0}
+    ansible{env:ANSIBLE_VERSION:==2.2.2}
```
cloudnull: Is there a reason why we're setting this here? can we not just use the included ansible-bootstrap within osa or the openstack-ansible-tests repo to ensure this consistency?

evrardjp:
Ok reviewing.

group_vars/all/rpc-o.yml
```diff
@@ -30,10 +30,13 @@ horizon_custom_uploads:
...
-tempest_test_sets: "scenario defcore cinder_backup"
+# Elasticsearch versions
+# NOTE (alextricity25):
+# The elasticsearch version variables needs to be available to the repo_build role
+# due to the way it builds out the distro packages dataset. Distro packages within
+# a role are indexed using their raw form(jinja is not rendered). The repo_build
+# role then attempts to render these datasets, and will fail if it cannot resolve
+# a variable. Putting these variables in here ensures the repo_build
+# role can interpolate the variables correctly.
+elasticsearch_version: 2.4.1
```
evrardjp: Could you put these vars in group_vars/all/elasticsearch.yml instead?

tox.ini
evrardjp: Agreed. Upping the version of ansible can cause a lot of pain.

alextricity25: I"m not sure why this is here to be honest. I should be able to get rid of it, right?

alextricity25: I'm going to focus on doing that on another PR.

PR #2236, opened 2017-04-28
Use CLI to create flavors

evrardjp:
There is no good version (read: with compatible requirements)
for together shade, pbr, os client tools, and another conflicting
package.

This removes the dependency to shade and works out of the box.
Idempotency is ensured for existing/non existing flavors, without
checking their values.

This can be rolled back when the requirements will be bumped (i.e.
when we are in an upper branch, meant for the usage of shade and
these modules)

Connected https://github.com/rcbops/rpc-openstack/issues/2227

### DIFF
--- a/rpcd/playbooks/roles/rpc_support/tasks/create_default_flavors.yml
+++ b/rpcd/playbooks/roles/rpc_support/tasks/create_default_flavors.yml
@@ -13,32 +13,21 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-- name: Install shade and its dependencies
-  pip:
-    name: "{{ item.name }}"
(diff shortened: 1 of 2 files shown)


evrardjp:
Cherry picked from master.

BjoernT:
I got it working for newton, took me to find out the right version. Since the support role gets eliminated in rpc-o I would close this bug. In openstack-ops I was pushing shade into a own venv so we are not interfering with any other pips. ``--isolated`` felt wrong to me when using pip install....

evrardjp:
If you allow yourself of not respecting the upper constraints (which IMO is fine for shade calls), it's indeed better to have your own venv. ``--isolated`` is indeed wrong :p

PR #2245, opened 2017-05-03
Use group_vars in RPC-O

evrardjp:
RPC-O previously relied on user-space configuration files to set variables used in playbooks across both the OpenStack-Ansible and RPC-O playbooks. These user-space configuration files provided global overrides only and could not be scoped to specific host groups. These defaults have now been set in group_vars in the code tree, reducing the need to overwrite user-space files, making them easier to scope to host groups, and reducing their precedence level to improve future development flexibility.

User-space configurable group_vars and host_vars can now optionally be used to customise deployments without requiring the user of extra-vars. The user-space group_vars and host_vars will take precedence over the in-tree group_vars. The user-space group_vars should be placed in /etc/openstack_deploy/host_vars and the user-space group_vars should be placed in /etc/openstack_deploy/group_vars.

Connected https://github.com/rcbops/u-suk-dev/issues/1620
Connected https://github.com/rcbops/u-suk-dev/issues/1659


evrardjp:
recheck_cit_xenial

alextricity25:
recheck_cit_ceph

evrardjp:
recheck_cit_ceph

evrardjp:
It looks ready!

alextricity25:
Reviewing this now!

rpcd/etc/openstack_deploy/user_rpco_variables_defaults.yml
```diff
@@ -13,139 +13,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
```
alextricity25: I would like to see a note in this file explaining where the new defaults are defined. It might be a little confusing for someone to go into ``/etc/openstack_deploy/user_rpco_variables_defaults.yml`` and not find anything.

rpcd/etc/openstack_deploy/user_osa_variables_defaults.yml
```diff
@@ -13,173 +13,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
```
alextricity25: I would like to see a note in this file explaining where the new defaults are defined. It might be a little confusing for someone to go into /etc/openstack_deploy/user_rpco_variables_defaults.yml and not find anything.

evrardjp:
recheck_cit_swift

releasenotes/notes/multi-layered-group-vars-cb60fcf454ca3bbe.yaml
```diff
@@ -0,0 +1,10 @@
+---
+other:
+  - |
+    RPC-Openstack had multiple levels of variable precedence, and we
+    heavily relied on user variables to complement/define
+    OpenStack-Ansible variables. This is not required anymore.
+    RPC-Openstack can now ship safe defaults for OpenStack-Ansible
+    without needing to set user variables. The user variables can
+    still be used by the deployer, but we now use the group_vars/
+    and host_vars/ folder for safe defaults in tree.
```
odyssey4me: I don't think that this note is explaining the value to the end-user (our support teams). How about something like this:

```
  - RPC-O previously relied on user-space configuration files to set variables used in playbooks across both the OpenStack-Ansible and RPC-O playbooks. These user-space configuration files provided global overrides only and could not be scoped to specific host groups. These defaults have now been set in group_vars in the code tree, reducing the need to overwrite user-space files, making them easier to scope to host groups, and reducing their precedence level to improve future development flexibility.
  - User-space configurable group_vars and host_vars can now optionally be used to customise deployments without requiring the user of extra-vars. The user-space group_vars and host_vars will take precedence over the in-tree group_vars. The user-space group_vars should be placed in /etc/openstack_deploy/host_vars and the user-space group_vars should be placed in /etc/openstack_deploy/group_vars.
```

alextricity25:
Is the intent to get rid of the user_*_overrides.yml file completely? If so, the bootstrap-aio.yml playbook still writes variables to this file. Does that need to change too?
Also, I recommend creating a doc issue for this so we know what RPCO docs need to change to reflect these changes.

evrardjp:
Fair point on the docs.

odyssey4me:
@alextricity25 Forcing the use of extra-vars is very bad practise, so yes, over time the intend is to be rid of that mechanism. However, it can't be done for everything right now.

I'm curious about which docs need changing though - I was under the impression that the user_*defaults.yml files were for development only and not for editing by anyone else. The files are still there, so nothing's changed except the content. The mechanism to override the defaults is still the same.

With these changes, though, it may be worth considering switching back to using the simpler set of files - just user_variables.yml and user_secrets.yml.

evrardjp:
Docs and reno added.

alextricity25:
@odyssey4me Gotcha. So the {rpco,osa}_variables_overrides.yml files are staying? For some reason It looked to me that they were going away(i.e. renamed to release.yml here: https://github.com/rcbops/rpc-openstack/pull/2245/files#diff-c4235d01f5e70134554b3bb48cfb50c5L13). Is the intention to get rid of the *_overrides files completely? Or still use them for deployers to override variables?

odyssey4me:
@alextricity25 The user_*.yml files have to stay for a while until we manage to remove the need for them - that will likely still take a few cycles.

alextricity25:
Does that mean we need to put them back in place here?

odyssey4me:
@alextricity25 They only need to be there if they have a function. ``rpcd/etc/openstack_deploy/user_rpco_variables_defaults.yml`` no longer has a function, so it can be removed whereas ``rpcd/etc/openstack_deploy/user_rpco_variables_overrides.yml`` has a user-facing function, and therefore stays.

The file ``rpcd/etc/openstack_deploy/user_osa_variables_defaults.yml`` has to stay because we have to override the ``openstack_release`` OSA group var. That'll have to stay for some time. The The file ``rpcd/etc/openstack_deploy/user_osa_variables_overrides.yml`` has a user-facing function and therefore stays.

Quite honestly, though, the overrides files could consolidate... I think having two of them is just confusing. It could return to just being ``user_variables.yml`` and leave it at that. This bring us closer to the same model that OSA is using, and therefore the documentation from OSA is usable. It also means that the OSA migration processes will work just fine. Having two secrets files also seems a bit silly to me, but consolidation of these files is out of scope for this patch.

group_vars/all/release.yml
```diff
@@ -13,10 +13,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-## Using this file
-#
-# This file is laid down only once and is used for overriding variables based on the file name.
-# This file will not be overwritten by upgrades.
+# The release tag to use for the repo and venvs
+# This can't be overriden because OSA is using group_vars.
+#openstack_release: "{{ rpc_release }}"
```
odyssey4me: Why is this comment here. It seems somewhat pointless.

group_vars/all/support.yml
```diff
@@ -1,5 +1,5 @@
 ---
```
odyssey4me: I don't really understand why this is being done. The file contains nothing. This also changes the user interface, so I suggest that it remains as it was.

group_vars/all/release.yml
```diff
@@ -13,10 +13,12 @@
...
+
+# The release tag to use
+# Note that this is set here so that the openstack_release override in osa_defaults
+# is able to use it.
+rpc_release: "{{ lookup('pipe', 'cd /opt/rpc-openstack && git describe --tags --abbrev=0') }}"
```
odyssey4me: The user_rpco_variables_overrides.yaml file needs to remain in place to avoid changing the user interface for overriding rpco vars.

group_vars/all/release.yml
evrardjp: This will be modified to:
TODO(evrardjp): Uncomment this when OSA has changed its group vars folder location.

group_vars/all/support.yml
evrardjp: We are providing the same level of override here, it's to show it's possible.

group_vars/all/release.yml
evrardjp: When rebased on artifacts branch, this will get replaced to a static value and will be the official user interface. Documentation has already been updated.

group_vars/all/support.yml
odyssey4me: Of course it's possible, but it serves no purpose other than to confuse people or encourage them to edit the code tree. I would recommend it's removed.

group_vars/all/release.yml
odyssey4me: You're assuming that will ever happen. I would suggest adding a TODO to the defaults override file to move it when it's possible to do so, then removing this bit.

group_vars/all/release.yml
odyssey4me: That's not the point. The point is that you've removed the empty default file which goes into user space for a new deployment.

evrardjp: It doesn't make sense to keep an empty user_variables_overrides. Shipping an empty file is pointless and cause of issues. We'd now have to do conditionals for upgrades and all that jazz.

group_vars/all/release.yml
evrardjp: Done.

group_vars/all/support.yml
evrardjp: I emailed @BjoernT to have his opinion.
It's indeed unnecessary to ship this file. 

I'd love see support interact on rpc-o, bring their variables and playbooks here.

group_vars/all/release.yml
odyssey4me: The overrides files are a user interface for implementing overrides and have always been empty (except for commented common overrides). We are not supposed to be editing them in any upgrades as they are supposed to be managed by the deployers.

rpcd/etc/openstack_deploy/user_osa_variables_defaults.yml
evrardjp: If the file doesn't exist, I think ppl will ask themselves questions, read at the git log and find out what happened.
Also they could ask around. Or check into user_osa_variables_defaults. Or read the release notes.

Plenty of sources of information, I don't see why we need to keep a file on top of it, just for devs.
I'd agree on users, but this file was never intended to touch by deployers.

group_vars/all/release.yml
evrardjp: Why shipping them then? Doesn't make sense to me.

evrardjp: These files were created for N release. They could change anytime as long as we explain. Also it doesn't mean doc change. Userland is still userland.

I can do it, but I will just do it to end this bikeshed.

BjoernT:
@evrardjp I somewhat share @odyssey4me opinion that this will cause people to edit the group vars directly rather then overriding them. The problem I have with this module is it's neither a ansible role nor a real standalone project and I would prefer not making this project a role more just a set of playbooks and scripts like it is, which pulls down other dependencies. Looking at what we currently have, that causes some issue too where people will edit the user_osa_variables_defaults.yml and not use the overrides. From that situation, hiding the defaults outside of /etc/openstack_deploy maybe beneficial. But I don't know if group vars are better here, how about merging the defaults into  /etc/openstack_deploy/rpco_defaults.yml and add a user_variables.yml copying the default but commented out. So people would be incentivized to use the normal user_variables.yml.

odyssey4me:
@BjoernT The benefit of using the vars plugin here (putting the default into group vars) is that both the OSA and RPC playbooks have access to the same changes and vars. It is now one all-inclusive set which is found regardless of where it's executed from. We had this before, but only because we were using extra vars which were global. This now allows vars to be scoped and we're able to tailor things more carefully.

Over the long term I'd definitely prefer to see a simplification down to three extra vars files:

1. Some sort of defaults file for stuff we have to set at the highest precedence (ie user_osa_variables_defaults.yml) but it could be named something more suitable and we could adjust the openstack-ansible tooling to recognise this name better. An even better option could be to implement a facility where the openstack-ansible tooling could recognise an extra vars file from somewhere outside user-space, then we could put it in-tree.
2. A global override file: user_variables.yml - this can have commented common overrides.
3. A global secrets file: user_secrets.yml

Having just these simplifies everything and makes it clear where things should be done.

evrardjp:
I 100% agree with @odyssey4me on last comment, for once. 🎆 

I strongly believe that we need to work all together, but we should have deployer's config in /etc/openstack_deploy, and our "safe product defaults" in rpc-openstack. 

I think the solution I brought on the table is far more elegant, and make it possible to avoid this whole frankenmess of files, override file naming, etc. ("Wait, was it user_variables_defaults, user_osa_variables_defaults, or user_variables? What was in that folder, I can't remember!").

Here, clear separation. 

I just hope support is gonna use this facility to its whole potential.

group_vars/all/support.yml
```diff
@@ -0,0 +1,16 @@
+---
```
odyssey4me: This file adds no value, being in-tree. I suggest that it be removed.

group_vars/all/maas.yml
```diff
@@ -0,0 +1,51 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+#maas_tenant_id:
+#maas_username:
+#maas_api_key:
```
odyssey4me: These commented vars and explanations for them should probably go into the user_rpco_default file.

group_vars/all/maas.yml
```diff
@@ -0,0 +1,51 @@
...
+
+# Set the LB device name in CORE
+#lb_name: 'unspecified'
```
odyssey4me: Thi commented var and explanations for it should probably go into the user_rpco_default file.

rpcd/etc/openstack_deploy/user_osa_variables_overrides.yml
```diff
@@ -18,5 +18,4 @@
 # This file is laid down only once and is used for overriding variables based on the file name.
 # This file will not be overwritten by upgrades.
 
-# this variable is here because vars files need to have at least ONE variable defined.
```
odyssey4me: This comment should likely stay to explain the variable.

rpcd/etc/openstack_deploy/user_rpco_variables_overrides.yml
```diff
@@ -18,5 +18,4 @@
 # This file is laid down only once and is used for overriding variables based on the file name.
 # This file will not be overwritten by upgrades.
 
-# this variable is here because vars files need to have at least ONE variable defined.
```
odyssey4me: This comment should likely stay to explain the variable.

evrardjp:
Ok I'll split this into two PR. We definitely have different opinions on what should and should not be done. What I did was functional. 

If we want to improve the system, we can.

BjoernT:
If we do the group vars I see no need to keep doing user_rpco_variables_overrides.yml, user_rpco_variables_defaults.yml, user_osa_variables_overrides.yml, user_osa_variables_defaults.yml and just create a user_variables.yml

evrardjp:
@BjoernT and I agree with you. To avoid the misunderstanding this will be, like I said in 2PRs.

One that's functionally doing its thing, one that's cleaning up the vars.

evrardjp:
@odyssey4me This could merge as is, could we stop the discussion and I'll do another PR with the cleanups to a different interfacing.

odyssey4me:
@BjoernT Happy with that conceptually, but I'd prefer to see that done in aanother PR as that will put this PR in the situation where it's trying to solve two problems - not one. The move of defaults to group vars is something we need for artifacting to avoid having to move a bunch of config into user space. The merging of user_* files is not something that is needed and has not been socialised. I'd like to keep the issues seperate.

BjoernT:
That's fine splitting this up as long as we get there. For me as end user it's all the same there is no separation. Just keep that in mind

BjoernT:
That's fine splitting this up as long as we get there. For me as end user it's all the same there is no separation. Just keep that in mind

alextricity25:
recheck_cit_ceph

PR #2247, opened 2017-05-03
Move logstash role out and add log aggregation

shannonmitchell:
This change is for issue/942 with the overall goal of filtering
and pushing customer logs from the local elk stack to a shared
global elk cluster.  This should give support easier access to
search customer logs for a quicker resolution with customer issues.
The existing logstash role was moved out to rcbops/rpc-role-logstash
with additions to push logs to a central elasticsearch cluster.
Enabling this should be a simple OSA overrides update and a re-run
of the logstash playbook.

Connects https://github.com/rcbops/u-suk-dev/issues/942


shannonmitchell:
Cancelling now as changes will be needed as we update the logstash role to a newer version, pull the defaults into and uncouple the inventory variables from the role.

PR #2250, opened 2017-05-05
Move maas_rpc_scripts_dir and maas_venv_enabled

alextricity25:
This commit moves the maas_rpc_scripts_dir variable definition to the
verify-maas.yml playbook. This variable is used no where else in the
rpc-maas role and has been taken out per
https://github.com/rcbops/rpc-maas/pull/239

This commit also removes the maas_venv_enabled variable. MaaS plugins
will now always run inside a python virtualenv.

Connects https://github.com/rcbops/u-suk-dev/issues/1601


evrardjp:
We need https://github.com/rcbops/rpc-openstack/pull/2245 in Newton branch, so that would be great to merge it before.

evrardjp:
Not merging, until the other one merged.

PR #2256, opened 2017-05-15
Add run_tempest.yml playbook

alextricity25:
This playbook installs tempest and executes the tempest tests
defined in group_vars/all/rpc_o.yml in the variable
``tempest_test_sets``.

The goal for doing this is to eventually move out tempest test
execution from the rpc-gating repository, into the rpc-openstack
repository series branches. This is so, from the point of view
of the gating jobs, the tempest tests are consumed the same
way regardless of openstack version.

Connects https://github.com/rcbops/u-suk-dev/issues/1680


alextricity25:
I'm missing the RUN_TEMPEST_OPTS and TESTR_OPTS here. Fixing that now.

PR #2269, opened 2017-05-18
Verifies Orchestration services

dankolbrs:
Adds verification Orchestration service post upgrade. Also
includes Orchestration in pre-upgrade check.

Connects https://rpc-openstack.atlassian.net/browse/UG-543

### DIFF
--- a/rpcd/playbooks/roles/rpc_post_upgrade/tasks/post-upgrade-utility.yml
+++ b/rpcd/playbooks/roles/rpc_post_upgrade/tasks/post-upgrade-utility.yml
@@ -70,3 +70,20 @@
     msg: "One or more of the cinder services are in the down state. Check the output above for more detail."
   when: "'down' in '{{ item.split(' ')[-1] }}'"
   with_items: "{{ cinder_output.stdout_lines | default([]) }}"
+
+- name: Grab relevant output of orchestration service list
+  shell: |
+    . ~/openrc
+    openstack orchestration service list -c Binary -c host -c status -f value
+  register: orchestration_output
+  failed_when: "orchestration_output.rc != 0"
+
+- name: Display output of orchestration service list
+  debug:
+    var: orchestration_output
+
+- name: Fail if any of the orchestration services are down
+  fail:
+    msg: "One or more of the orchestration services are down. Check the output above for more detail."
+  when: "'down' in '{{ item.split(' ')[-1] }}'"
+  with_items: "{{ orchestration_output.stdout_lines | default([]) }}"
--- a/rpcd/playbooks/roles/rpc_pre_upgrade/tasks/gather_stats.yml
+++ b/rpcd/playbooks/roles/rpc_pre_upgrade/tasks/gather_stats.yml
@@ -44,6 +44,13 @@
   register: neutron_agentlist_result
   delegate_to: "{{ groups['utility_all'][0] }}"
 
+- name: Gather orchestration service list
+  shell: "source ~/openrc; openstack orchestration service list --insecure"
+  args:
+    executable: /bin/bash
+  register: orchestration_servicelist_result
+  delegate_to: "{{ groups['utility_all'][0] }}"
+
 - name: Gather openstack endpoint list
   shell: "source ~/openrc; openstack --insecure endpoint list"
   args:
--- a/rpcd/playbooks/roles/rpc_pre_upgrade/templates/status.txt.j2
+++ b/rpcd/playbooks/roles/rpc_pre_upgrade/templates/status.txt.j2
@@ -25,6 +25,11 @@ cinder service-list output as at {{ ansible_date_time.date }}-{{ ansible_date_ti
 {{ line }}
 {% endfor %}
 
+orchestration service list output as at {{ ansible_date_time.date }}-{{ ansible_date_time.time }}
+
+{% for line in orchestration_servicelist_result.stdout_lines %}
+{{ line }}
+{% endfor %}
 
 neutron agent-list output as at {{ ansible_date_time.date }}-{{ ansible_date_time.time }}
 


dankolbrs:
recheck_cit_all - pip timeout fail

dankolbrs:
recheck_cit_all

dankolbrs:
recheck_cit_all

dankolbrs:
recheck_cit_upgrade - `TEMPEST FAIL scenario defcore cinder_backup (301 tests)`

dankolbrs:
recheck_cit_ceph

rpcd/playbooks/roles/rpc_post_upgrade/tasks/post-upgrade-utility.yml
```diff
@@ -70,3 +70,21 @@
...
+  failed_when: "orchestration_output.rc != 0"
+
+- name: Display output of orchestration service list
+  debug:
+    var: orchestration_output
+
+- name: Fail if any of the orchestration services are down
+  fail:
+    msg: "One or more of the orchestration services are down. Check the output above for more detail."
+  when: "'down' in '{{ item.split(' ')[-1] }}'"
```
evrardjp: I think there is no need for {{ .
```
when:
  - " 'down' in item.split(' ')[-1]"
```

should do the trick

rpcd/playbooks/roles/rpc_pre_upgrade/tasks/gather_stats.yml
```diff
@@ -44,6 +44,13 @@
   register: neutron_agentlist_result
   delegate_to: "{{ groups['utility_all'][0] }}"
 
+- name: Gather orchestration service list
+  shell: "source ~/openrc; openstack orchestration service list --insecure"
```
evrardjp: openstack should use the clouds.yaml so no need to source the openrc if I'm not mistaken.

You can then move to use the command module.

dankolbrs: Legit point, but this commit is only adding checking Heat the same way nova, neutron, and cinder is already being checked. For clarity, it duplicates the way the checks for the other services are done rather than re-implementing the test for Heat in a slightly different way, or rewriting existing checks.

rpcd/playbooks/roles/rpc_post_upgrade/tasks/post-upgrade-utility.yml
dankolbrs: Good catch. See response in gather_stats.yml.

git-harry: When testing for errors I think it's better to make the validation a little more exacting, e.g.

```yaml
- name: Grab relevant output of orchestration service list
  shell: |
    . ~/openrc
    openstack orchestration service list --insecure  -f json
  register: orchestration_output
  failed_when: "orchestration_output.rc != 0"

- name: Fail if any of the orchestration services are down
  fail:
    msg: "One or more of the orchestration services are down. Check the output above for more detail."
  when: item["status"] != "up"
  with_items: "{{ orchestration_output | from_json }}"
```

By using JSON the specific piece of data, `status`, can be referenced instead of having to rely on ordering or successful text parsing. By testing for inequality with `"up"` it protects against changes to status names as well as the unexpected.

rpcd/playbooks/roles/rpc_post_upgrade/tasks/post-upgrade-utility.yml
```diff
@@ -70,3 +70,21 @@
     msg: "One or more of the cinder services are in the down state. Check the output above for more detail."
   when: "'down' in '{{ item.split(' ')[-1] }}'"
   with_items: "{{ cinder_output.stdout_lines | default([]) }}"
+
+# Adds orchestration service checks per https://rpc-openstack.atlassian.net/browse/UG-543
+- name: Grab relevant output of orchestration service list
+  shell: |
+    . ~/openrc
+    openstack orchestration service list --insecure
```
git-harry: I think requesting a JSON response provides a better experience because we can reference the specific key instead of searching the text, e.g.:

```yaml
    openstack orchestration service list --insecure  -f json
```

rpcd/playbooks/roles/rpc_post_upgrade/tasks/post-upgrade-utility.yml
dankolbrs: Since this is the second review recommending performing this test differently, I may need to change my stance. Is there a value in running this test in a different way than the existing nova, cinder, and neutron services are being checked?  For simplicity, I matched the three other tests so they would be logically grouped reading through them, being able to see each one is using the same method for testing.  Parsing JSON does make perfect sense and is the way I would have gone with it if not for the existing service checks using string parsing.

rpcd/playbooks/roles/rpc_post_upgrade/tasks/post-upgrade-utility.yml
```diff
@@ -70,3 +70,21 @@
...
+
+- name: Display output of orchestration service list
+  debug:
+    var: orchestration_output
+
+- name: Fail if any of the orchestration services are down
+  fail:
+    msg: "One or more of the orchestration services are down. Check the output above for more detail."
+  when: "'down' in '{{ item.split(' ')[-1] }}'"
+  with_items: "{{ orchestration_output.stdout_lines | default([]) }}"
```
evrardjp: the idea is to not burp when stdout_lines is not defined (can't iterate over a None object), and therefore giving a list would work, while doing nothing.

dankolbrs: If it has no lines, that would mean that Orchestration is not configured/installed on the OpenStack installation. While that would be valid for an OpenStack installation, is that not valid for an RPC installation? Specifically the card was asking for down services, and I'm not sure whether **all** RPC installations use Orchestration.

rpcd/playbooks/roles/rpc_pre_upgrade/tasks/gather_stats.yml
```diff
@@ -44,6 +44,13 @@
...
+  args:
+    executable: /bin/bash
+  register: orchestration_servicelist_result
```
dankolbrs: I missed that. Updated in latest commit. Thanks!

rpcd/playbooks/roles/rpc_post_upgrade/tasks/post-upgrade-utility.yml
evrardjp: The difference is important. IMO if we need to ensure that ALL RPC installations have orchestration, we should remove the  |default.

rpcd/playbooks/roles/rpc_pre_upgrade/templates/status.txt.j2
```diff
@@ -25,6 +25,11 @@ cinder service-list output as at {{ ansible_date_time.date }}-{{ ansible_date_ti
 {{ line }}
 {% endfor %}
 
+orchestration service-list output as at {{ ansible_date_time.date }}-{{ ansible_date_time.time }}
```
git-harry: The command is incorrect:
s/orchestration service-list/openstack orchestration service list/

rpcd/playbooks/roles/rpc_post_upgrade/tasks/post-upgrade-utility.yml
git-harry: I think you make a fair point, any general bugs with this file are separate from the addition of this functionality. That said, the task you have added does not match the implementation of the others. For example:
```yaml
openstack compute service list -c Binary -c Host -c State -f value
```
This command limits the output to an ordered list of space-separated specific values. While I would prefer the use of JSON or YAML, I think this is preferable to the default table output. `--insecure` is not used here or by any of the other commands in the file. It looks like you've copied this task from the other file which is doing things differently.

rpcd/playbooks/roles/rpc_post_upgrade/tasks/post-upgrade-utility.yml
git-harry: I can't find any detail in the card regarding the specific requirements of this task but if this is not doing the right thing, I'm happy to see it addressed more generally as a separate piece of work.

rpcd/playbooks/roles/rpc_pre_upgrade/templates/status.txt.j2
dankolbrs: In `rpcd/playbooks/roles/rpc_pre_upgrade/templates/status.txt.j2` that it correct, as that is just outputting to a text file and not running the command. 

It was incorrect in `rpcd/playbooks/roles/rpc_pre_upgrade/tasks/gather_stats.yml` which is probably what you meant to note.

git-harry: I appreciate it's not running the command but I believe it is supposed to be showing the command run to get the output. `nova service-list` and `cinder service-list` are both are run by `rpcd/playbooks/roles/rpc_pre_upgrade/tasks/gather_stats.yml`. To my knowledge there is no client called orchestration and `orchestration service-list` is not run by any task.

dankolbrs:
recheck_cit_all

mattt416:
Dismissing @git-harry's reviewing since he's out this week.

dankolbrs:
@evrardjp How do you feel about the changes and discussion on this after your review?

dankolbrs:
recheck_cit_all

dankolbrs:
recheck_cit_ceph
recheck_cit_swift  -TEMPEST FAIL scenario defcore (297 tests)

dankolbrs:
recheck_cit_ceph

PR #2271, opened 2017-05-23
Introducing leapfrog script

evrardjp:
This is a first pass, to make sure an OSA leapfrog can happen
in the gates.

Connected https://github.com/rcbops/u-suk-dev/issues/1488

### DIFF
--- a/scripts/leapfrog/ubuntu14-leapfrog.sh
+++ b/scripts/leapfrog/ubuntu14-leapfrog.sh
@@ -1,0 +1,110 @@
+#!/usr/bin/env bash
+# Copyright 2017, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


scripts/leapfrog/ubuntu14-leapfrog.sh
```diff
@@ -0,0 +1,110 @@
...
+set -e -u -x
+set -o pipefail
+
+## Standard Vars --------------------------------------------------------------
+export OA_OVERRIDES='/etc/openstack_deploy/user_osa_variables_overrides.yml'
+export OA_DEFAULTS='/etc/openstack_deploy/user_osa_variables_defaults.yml'
+export RPCD_OVERRIDES='/etc/openstack_deploy/user_rpco_variables_overrides.yml'
+export RPCD_DEFAULTS='/etc/openstack_deploy/user_rpco_variables_defaults.yml'
+export RPCD_SECRETS='/etc/openstack_deploy/user_rpco_secrets.yml'
+export RPCO_DEFAULT_FOLDER="/opt/rpc-openstack"
```
odyssey4me: Is there a reason that you aren't using the functions export that all the other scripts use to set these variables?

evrardjp: There are overlapping ones we could redefine a little below, and non-overlapping ones.
I'd like to move to functions in a later commit, but this is the minimum viable script :p

Issue #2266, opened 2017-05-17
Liberty nova-spicehtml5proxy service continues after upgrade to Mitaka


dankolbrs:
> **THIS IS A PUBLIC FACING ISSUE.**
>
> Do not include private or confidential information in this issue.

## Is this a feature or a bug?

- [ ] Feature _(request for enhancement, new functionality)_
- [x] Bug _(something is broken)_

## Which version of rpc-openstack does this issue relate to?
- [ ] master
- [ ] newton-14.0
- [x] mitaka-13.1
- [x] liberty-12.2

Run `git log --pretty=oneline | head -n 1` and provide the output below between
the backtick lines:

```
d6699311e9b87b84a9ed2e928b1f196fd7387c8b Merge 268243fde331a19e98377269e4ba20a5ce8b796b into 5d25ebcbea2204bc066aa0595dfc31f995b6831a
```

## What happened?
Testing Nova services after upgrade from Liberty-12.2 to Mitaka-13.1 displays `nova-spicehtml5proxy` continues running Liberty version due to the `nova-spicehtml5proxy` init script not being updated or service being restarted.

```
root@ramup-128-7782-nova-console-container-4d8760cc:~# ps awwfux | grep nova
nova      1888  1.5  0.2 296564 43528 ?        Ss   19:15   0:01 /openstack/venvs/nova-12.2.8/bin/python /openstack/venvs/nova-12.2.8/bin/nova-spicehtml5proxy --log-file=/var/log/nova/nova-spicehtml5proxy.log
nova      1922  5.0  0.6 198692 102400 ?       Ss   19:15   0:05 /openstack/venvs/nova-13.3.19/bin/python /openstack/venvs/nova-13.3.19/bin/nova-consoleauth --log-file=/var/log/nova/nova-consoleauth.log
nova      1944  2.9  0.4 328952 68944 ?        Ss   19:15   0:03 /openstack/venvs/nova-13.3.19/bin/python /openstack/venvs/nova-13.3.19/bin/nova-novncproxy --log-file=/var/log/nova/nova-novncproxy.log
root@ramup-128-7782-nova-console-container-4d8760cc:~# ls -lah /etc/init/nova*
-rw-r--r-- 1 root root 1.2K May 17 19:14 /etc/init/nova-consoleauth.conf
-rw-r--r-- 1 root root 1.2K May 17 19:14 /etc/init/nova-novncproxy.conf
[...]

PR #2273, opened 2017-05-24
Adds nova_console container rebuild

dankolbrs:
On L->M upgrade, the switch from Spice console to novnc results
in Spice console containers to continue running without being
rebuilt or destroyed. This will result in a failure on a post
upgrade check which ensures all nova services are the same version.

This destroys and rebuilds all nova_console containers so that
only novnc containers will be running in Mitaka.

Issue https://github.com/rcbops/rpc-openstack/issues/2266

Connects https://rpc-openstack.atlassian.net/browse/UG-538

### DIFF
--- a/scripts/test-upgrade.sh
+++ b/scripts/test-upgrade.sh
@@ -86,6 +86,21 @@ openstack-ansible lxc-containers-destroy.yml --limit repo_all
 openstack-ansible setup-hosts.yml --limit repo_all
 
 # TASK #5
+# Bug: https://github.com/rcbops/rpc-openstack/issues/2266
+# Issue: In Liberty the console was Spice, which was moved to novnc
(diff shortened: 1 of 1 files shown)


mattt416:
@BjoernT you happy with doing something like this in your liberty->mitaka upgrade plan?

BjoernT:
@mattt416  We had not talked about this, only expressed interest in exploring the option to disconnect the serial consoles and attach new ones, with vnc protocol. The option to bounce instances just because of that, it not desired. That should be left to the customer discretion of the customer

mattt416:
recheck_cit_ceph
recheck_cit_upgrade

dankolbrs:
Added documentation update at https://github.com/rackerlabs/docs-rpc/pull/1031

PR #2274, opened 2017-05-24
Move logstash role out and add log aggregation

shannonmitchell:
This change is for issue/942 with the overall goal of filtering
and pushing customer logs from the local elk stack to a shared
global elk cluster. This should give support easier access to
search customer logs for a quicker resolution with customer issues.
The existing logstash role was moved out to rcbops/rpc-role-logstash
with additions to push logs to a central elasticsearch cluster.
Enabling this should be a simple OSA overrides update and a re-run
of the logstash playbook.

Connects rcbops/u-suk-dev#942


shannonmitchell:
recheck_cit_swift

alextricity25:
Nic @shannonmitchell. Can we go ahead and delete the logstash role from the repo? I think we should go ahead and do that in this PR if you can.

shannonmitchell:
Awesome.  Removed it from the local repo, but forgot to add the removal to the commit.  Working on it now.

PR #2276, opened 2017-05-25
Improve flexibility of leapfrog in master

evrardjp:
This introduces the loading of post leap and pre leap steps.

Connected https://github.com/rcbops/u-suk-dev/issues/1734


evrardjp:
This will be dropped, as we now have newton 14.1 branch, and leapfrog code should only be in newton 14.1 branch.

PR #2286, opened 2017-05-31
Configure rabbitMQ HA policy

BjoernT:
The OSA policy replicated all queues to all nodes
ultimately leading to performance issues.
A new HA policy limits the queue mirroring to two
nodes allowing for much better scalability

Related-To: rcbops/u-suk-dev#461

### DIFF
--- a/group_vars/all/osa.yml
+++ b/group_vars/all/osa.yml
@@ -58,6 +58,19 @@ nova_console_type: novnc
 # RabbitMQ overrides
 rabbitmq_ulimit: 65535
 
+# Set RabbitMQ message replication count to 2
+# The desired setting would be {{ (groups.rabbitmq|length /2 +1)|round(0,'floor') |int }}
+# but can't be used due to https://github.com/ansible/ansible/issues/9362
+rabbitmq_policies:
+  - name: "HA"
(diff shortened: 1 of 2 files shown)


group_vars/all/osa.yml
```diff
@@ -58,6 +58,15 @@ nova_console_type: novnc
 # RabbitMQ overrides
 rabbitmq_ulimit: 65535
 
+rabbitmq_policies:
+  - name: "HA"
+    pattern: '^(?!amq\.).*'
+    tags:
+      ha-mode: exactly
+      ha-params: 2
```
prometheanfire: Can this be changed to a variable based off what the quorum is.  2 in a 3 node cluster 3 in a 5 node, etc.  That'd match the rabbit docs of this.  Other than that, lgtm.

https://www.rabbitmq.com/ha.html

BjoernT: I see no need to have more than 2 copies even when using 5 rabbitmq nodes. The quorum has nothing to do with the number of message copies, there is no correlation.

BjoernT:
Damn it ansible: https://github.com/ansible/ansible/issues/9362

alextricity25:

PR #2298, opened 2017-06-06
Adapt AIO bootstrap when apt artifacts unavailable

odyssey4me:
When the apt artifacts are not available (as they have
not yet been built for this release), the AIO bootstrap
must not clobber the updates/backports repository
configuration. This patch ensures that it adapts
accordingly.

A little bit of rpc_release override clean up has also
been removed as it is no longer necessary now that we
are setting the version statically.

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -58,10 +58,6 @@ if [[ "${DEPLOY_AIO}" == "yes" ]]; then
   sed -i "s/aio1/$(hostname)/" /etc/openstack_deploy/openstack_user_config.yml
   sed -i "s/aio1/$(hostname)/" /etc/openstack_deploy/conf.d/*.yml
 
-  # TODO(odyssey4me):
-  # Remove this once the rpc_release is statically defined.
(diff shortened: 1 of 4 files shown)


odyssey4me:
closed/opened to unbreak travis

scripts/deploy.sh
```diff
@@ -58,10 +58,6 @@ if [[ "${DEPLOY_AIO}" == "yes" ]]; then
   sed -i "s/aio1/$(hostname)/" /etc/openstack_deploy/openstack_user_config.yml
   sed -i "s/aio1/$(hostname)/" /etc/openstack_deploy/conf.d/*.yml
 
-  # TODO(odyssey4me):
-  # Remove this once the rpc_release is statically defined.
-  echo "rpc_release: $(/opt/rpc-openstack/scripts/artifacts-building/derive-artifact-version.sh)" >> /etc/openstack_deploy/user_rpco_variables_overrides.yml
```
odyssey4me: Removing this has uncovered that the holland test relies on an override being set for this. This really should not be the case, so I've pushed https://github.com/rcbops/rpc-gating/pull/218 up to resolve that.

evrardjp:
We'll need to backport to N14.1 too

odyssey4me:
closed/re-opened because travis is dumb

alextricity25:
Backport is here: https://github.com/rcbops/rpc-openstack/pull/2307

PR #2300, opened 2017-06-06
Add ironic deployment instructions to the readme

prometheanfire:
We deploy ironic if told to.  If not, do not deploy.  We do this by
assigning the ironic containers to the os-infra-containers groups.  By
default they are assigned to their own group.  We set up the env.d file
in both deploy.sh and boostrap-aio.yml as just because one may be used
does not mean the other one is.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,8 @@ maas_api_key: YourAPIKey
        `openstack_user_config.yml.aio`
     2. There is a tool to generate the inventory for RAX datacenters, otherwise
        it will need to be coded by hand.
+  6. If deploying ironic set up the `ironic-infra_hosts` and
+    `ironic-compute_hosts` parameters in `/etc/openstack_deploy/conf.d/ironic.yml`
 3. Run the RPC deploy script: `cd /opt/rpc-openstack && ./scripts/deploy.sh`
   1. If building without the ELK stack, set `DEPLOY_ELK=no` before running
 


prometheanfire:
@odyssey4me It sounds like you are over-complicating things.  A conf.d file could be used but we'd need to define which nodes the ironic containers go on.  This will be the os-infra-hosts in nearly all cases.  Instead of having to redefine that we can just set that up to be the default for ironic nodes.  I'm not sure what purpose the ironic scenario would serve since we are just deploying the infra pieces.  Does our gating not use deploy.sh?  If so, it sounds like our gating doesn't test how we tell people to deploy.

@cloudnull working on that now, thanks

odyssey4me:
@prometheanfire People should not be editing the env.d files if they can possibly help it - so your approach is actually making changes which are the lest desired when it comes to how a deployer should do the configuration. As you'll see here, it's two entries: https://github.com/openstack/openstack-ansible/blob/master/etc/openstack_deploy/conf.d/ironic.yml.aio

If env.d files are customised, they're very hard to maintain. We are trying to phase out their general use from rpc-openstack as far as possible because they should only be used when a customised group/layout is required. The conf.d/openstack_user_config file is explicitly for host:group allocations which is exactly what you're wanting to do here.

My vote stands - this patch is entirely unnecessary unless you wish to add a test scenario for gating purposes. If that's the case, then simply add the scenario into the aio bootstrap as previously suggested.

prometheanfire:
@odyssey4me can I set up the conf.d to deploy the ironic containers on infra hosts without specifying the hosts or IPs?  We are looking to automate it similar to how things like keystone and glance are deployed (on os-infra-hosts) without manually redefining which hosts are os-infra-hosts (since ironic by default uses os-ironic-hosts).

odyssey4me:
@prometheanfire I don't understand the problem. If you're already defining which hosts go into the os-infra group, then you're clearly able to define any other conf.d entries for other services. The purpose of the tests is to test with as close a configuration as possible to how a deployer tests it, and the mechanism being used in this patch is not that.

evrardjp:
Yeah, let's reuse the existing env.d, and only configure conf.d.

prometheanfire:
@odyssey4me yes, which hosts go into which group is already defined, but ironic containers don't get deployed to those hosts, that's what I'm changing here.  env.d is the only location that I can change that.

As far as testing how a deployer would run it, if we are the deployer and we deploy it on os-infra hosts then we will be testing how a deployer deploys it.

odyssey4me:
@prometheanfire That is not the only way to change it. Adding a conf.d file with the group:host mapping is the correct way to do it. Using an env.d file to change the group mappings is not the correct way to do it.

prometheanfire:
@odyssey4me first, the change doesn't negate what's already in the env.d file, simply adds to it.  The containers were already part of the ironic containers group, now they will also be a part of the infra group.

second, what would that look like?  Can I use the conf.d files to override which container group the ironic containers belong to?  I'm under the impression that this is not the case.

odyssey4me:
@prometheanfire The point is that a deployer should not be modifying the groups, only the group members. If you add a host to the ironic-infra_hosts group, as shown in https://github.com/openstack/openstack-ansible/blob/master/etc/openstack_deploy/conf.d/ironic.yml.aio then it will add the control plane containers for that service onto the host specified.

Modifying the group structure introduces technical debt which needs to be managed through upgrades. Modifying group members does not - it is an expected configuration.

prometheanfire:
Yes, that's how I was doing it at the start.  What I was hoping to avoid is yet another step in deploying rpc-o.  If we are going to use the same hosts then just use the host group.

The alternative to this PR would be to add a section to the readme to have the deployer set up the ironic infra host group.

odyssey4me:
@prometheanfire Ah, I now see where the confusion is. The only step required to deploy ironic, actually, outside of the gating system is to implement the host:group mapping. The playbook execution is already included in setup-openstack. A deployer should never have to set the environment variable. The environment variable's purpose is for *gating*, not for deployer use.

evrardjp:
@odyssey4me I think if deploy.sh is our standard interface, it makes sense for any deployer to use bash env variables. 

In this case, I think if an AIO and DEPLOY_IRONIC is set, export the proper ironic conf.d. If not an AIO, the deployer should probably write his own, and we'd have nothing to do.

Is that what you meant @odyssey4me ? I'm not sure to understand both of you.

odyssey4me:
@evrardjp Yes, the env var only has a gating purpose in this case, and should therefore be used to implement a new scenario or additional conf.d file in the AIO bootstrap. The use of env vars for any deployer purposes needs to be phased out entirely, but that will require a restructuring of deploy.sh into something used for deployers and a different script used for gating. Once that's done the purpose will be a lot clearer.

prometheanfire:
Ya, so right now the plan is to just have have the deployer be required to setup the conf.d mappings unless doing an aio (in which case the conf.d file will be laid down pointing to aio1).  At this stage I'm not sure we need another scenario as we are not setting up ironic hosts, just the containers.

odyssey4me:
@prometheanfire Then what would ever activate this as a test? What is the purpose of modifying it when there's no test execution for it?

prometheanfire:
updated just the readme now :(

prometheanfire:
recheck

prometheanfire:
superseded by https://github.com/rackerlabs/docs-rpc/pull/1061

PR #2301, opened 2017-06-06
Ensure that the apt artifact repo is configured

odyssey4me:
When using the apt artifacts, we currently remove
the updates source from /etc/apt/sources.list but
do not ensure that the rpco.list file is configured
prior to executing the openstack_hosts role. This
causes the build to fail because that role needs
packages not available in the base source.

This patch implements the execution of the
pip_install role which happens to also configure
the additional repo needed for everything to be
right with the world again.


odyssey4me:
@evrardjp I don't understand the feedback. The changes have been made to rpcd/playbooks/configure-apt-sources.yml as you suggested. The playbook now uses the pip_install role to avoid duplicating everything that's in that role to setup the repositories.

PR #2303, opened 2017-06-06
Ignore whitespaces when checking repo configuration

mkam:
Removes any leading whitespaces in sources.list
before checking for the existing repo configuration.

This resolves issues when building using the
multi-node AIO.


odyssey4me:
Closed/re-opened because travis is dumb.

PR #2312, opened 2017-06-08
UG-629 Add quotes around tempest variables

mkam:
If tempest_run_tempest_opts or tempest_testr_opts are
empty lists, have spaces in values, or have mulitple
values, the exports fail without these quotes.


mkam:
recheck_cit_ceph

PR #2313, opened 2017-06-09
[LA-316] [LA-317] Leapfrog with rpc

evrardjp:
Previous commit included the flexible leapfrog features. This
leverages it to load our own scripts to leap RPC, and redeploy
RPC bits (currently ELK and MaaS).

### DIFF
--- a/scripts/leapfrog/post_leap.sh
+++ b/scripts/leapfrog/post_leap.sh
@@ -20,12 +20,15 @@ set -e -u -x
 set -o pipefail
 
 echo "POST LEAP STEPS"
-#if [[ ! -f "${UPGRADE_LEAP_MARKER_FOLDER}/deploy-rpc.complete" ]]; then
-#  pushd ${RPCO_DEFAULT_FOLDER}/rpcd/playbooks/
-#    #scripts/deploy.sh
-#    openstack-ansible site.yml
-#  popd
-#  log "deploy-rpc" "ok"
-#else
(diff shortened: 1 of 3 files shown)


evrardjp:
@hughsaunders @odyssey4me Your comments on previous commit should be addressed here.

odyssey4me:
@evrardjp OK, just bear in mind that commit history lives beyond other systems which have details, for example JIRA. It's best to comment in the code (it's easiest to find information in context) or to describe what/why in the commit message to provide context. I do feel like the stuff being done in these commits requires more context than I have, making it difficult to do anything other than a syntax/spelling/sanity check.

evrardjp:
@odyssey4me Fair point. I'll have to rebase, so I'll include commit message changes at the same time.

evrardjp:
recheck_cit_swift

evrardjp:
as you can see, the trusty-leapfrogupgrade-swift works \o/

evrardjp:
CIT/ceph
Old job, expected to fail

CIT/leapfrogupgrade 
Old job, expected to fail

CIT/swift
Old job, expected to fail

CIT/trusty-leapfrogupgrade-ceph
ceph doesn't deploy on N. This job will disappear in the future

CIT/xenial
Old job, expected to fail

CIT/xenial-leapfrogupgrade-swift
kilo doesn't deploy on xenial. This job will disappear in the future.

So everything is fine!

PR #2315, opened 2017-06-09
Remove the AIO override for cinder-backup

odyssey4me:
Enabling cinder-backup causes a problem with upgrades
because the upgrades are not properly orchestrated to
ensure that the services are restarted in the right
order. The cinder-backup service is disabled by
default for RPC-O anyway, so this patch removes the
override.

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -95,11 +95,6 @@ if [[ "${DEPLOY_AIO}" == "yes" ]]; then
     # set network speed for vms
     echo "net_max_speed: 1000" >>$RPCD_VARS
 
-    # TODO(odyssey4me):
-    # This is disabled until work can be done to resolve the upgrade issues.
-    # https://github.com/rcbops/rpc-openstack/issues/2258
-    sed -i "s/cinder_service_backup_program_enabled: .*/cinder_service_backup_program_enabled: false/" /etc/openstack_deploy/user_osa_aio_variables.yml
-
     # set the necessary bits for ceph
     if [[ "$DEPLOY_CEPH" == "yes" ]]; then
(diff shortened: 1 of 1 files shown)


scripts/deploy.sh
```diff
@@ -92,11 +92,14 @@ if [[ "${DEPLOY_AIO}" == "yes" ]]; then
...
     sed -i "s/maas_notification_plan: .*/maas_notification_plan: npTechnicalContactsEmail/" $RPCD_VARS
-    # the AIO needs this enabled to test the feature, but $RPCD_VARS defaults this to false
-    sed -i "s/cinder_service_backup_program_enabled: .*/cinder_service_backup_program_enabled: true/" /etc/openstack_deploy/user_osa_variables_defaults.yml
     # set network speed for vms
     echo "net_max_speed: 1000" >>$RPCD_VARS
 
+    # TODO(odyssey4me):
+    # This is disabled until work can be done to resolve the upgrade issues.
+    # https://github.com/rcbops/rpc-openstack/issues/2258
+    sed -i "s/cinder_service_backup_program_enabled: .*/cinder_service_backup_program_enabled: false/" /etc/openstack_deploy/user_osa_aio_variables.yml
```
cloudnull: IDK if this is needed. Our default setting in the overrides files disables this already. https://github.com/rcbops/rpc-openstack/blob/mitaka-13.1/rpcd/etc/openstack_deploy/user_osa_variables_defaults.yml#L150

mattt416:
recheck_cit_all

alextricity25:

cloudnull:

alextricity25:

PR #2316, opened 2017-06-09
Remove the AIO override for cinder-backup

odyssey4me:
Enabling cinder-backup causes a problem with upgrades
because the upgrades are not properly orchestrated to
ensure that the services are restarted in the right
order. The cinder-backup service is disabled by
default for RPC-O anyway, so this patch removes the
override.

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -90,10 +90,13 @@ if [[ "${DEPLOY_AIO}" == "yes" ]]; then
     sed -i "s/lb_name: .*/lb_name: '$(hostname)'/" $RPCD_VARS
     # set the notification_plan to the default for Rackspace Cloud Servers
     sed -i "s/maas_notification_plan: .*/maas_notification_plan: npTechnicalContactsEmail/" $RPCD_VARS
-    # the AIO needs this enabled to test the feature, but $RPCD_VARS defaults this to false
-    sed -i "s/cinder_service_backup_program_enabled: .*/cinder_service_backup_program_enabled: true/" /etc/openstack_deploy/user_variables.yml
     # set network speed for vms
     echo "net_max_speed: 1000" >>$RPCD_VARS
+
+    # TODO(odyssey4me):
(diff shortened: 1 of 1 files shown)


scripts/deploy.sh
```diff
@@ -90,11 +90,14 @@ if [[ "${DEPLOY_AIO}" == "yes" ]]; then
...
     sed -i "s/maas_notification_plan: .*/maas_notification_plan: npTechnicalContactsEmail/" $RPCD_VARS
-    # the AIO needs this enabled to test the feature, but $RPCD_VARS defaults this to false
-    sed -i "s/cinder_service_backup_program_enabled: .*/cinder_service_backup_program_enabled: true/" /etc/openstack_deploy/user_variables.yml
     # set network speed for vms
     echo "net_max_speed: 1000" >>$RPCD_VARS
 
+    # TODO(odyssey4me):
+    # This is disabled until work can be done to resolve the upgrade issues.
+    # https://github.com/rcbops/rpc-openstack/issues/2258
+    sed -i "s/cinder_service_backup_program_enabled: .*/cinder_service_backup_program_enabled: false/" /etc/openstack_deploy/user_variables.yml
```
cloudnull: IDK if this is needed. Our default setting in the overrides files disables this already. https://github.com/rcbops/rpc-openstack/blob/liberty-12.2/rpcd/etc/openstack_deploy/user_variables.yml#L142

odyssey4me: I might be missing something, but I don't see where we use that file?

alextricity25: This looks good to me. Liberty AIO deployments use the `user_variables.yml.j2` template provided by the bootstrap-host role. Multi-node deployments in liberty required that the `user_variables.yml` file in `rpcd/etc/openstack_deploy` be copied over to `/etc/openstack_deploy/` manually.
In a nut shell, the `user_variables.yml file` in `rpcd/openstack_deploy/` isn't used by AIO deployments in liberty-12.2, so we need to override it here.

PR #2296, opened 2017-06-05
Use standalone RPC-MaaS

cloudnull:
This change converts our deployment of RPC MaaS from the old role method
to the new playbook method. This change will provide us access to a
faster and more flexible monitoring solution and allow all customers to
benifit from Monitoring as a Service using a unified code base.

Signed-off-by: Kevin Carter <kevin.carter@rackspace.com>

PR #2317, opened 2017-06-09
Remove hard coded cinder backup enablement

cloudnull:
This is being removed because it has been determined by the gating team
to be broken. Sadly the gating team has hard coded an additional
override for this override, which disbales this feature, so this code
path is never exercised in or outside of the gate. This removal will
allow us use well defined, and known overrides within the project
giving developers a clear sense of what is being developed vs what
is being tested and will unblock work in master.

Related-Issue: https://github.com/rcbops/rpc-gating/pull/224
Un-Blocks: https://github.com/rcbops/rpc-openstack/pull/2296


cloudnull:
I missed that there was an inflight patch already https://github.com/rcbops/rpc-openstack/pull/2315 closing infavor of that.

PR #2330, opened 2017-06-12
Move GROUP_VARS and HOST_VARS_PATH

alextricity25:


evrardjp:
Agreed with @odyssey4me .

odyssey4me:
I'm busy working through the fix for all the artifact jobs [1], but it looks like you're doing the right things here, and yes it is simpler. After [1] merges it will need some rework though.

[1] https://github.com/rcbops/rpc-openstack/pull/2337

alextricity25:
This is ready for reviews.

odyssey4me:
We should port this back to newton-14.1 too.

evrardjp:
@alextricity25 when backporting to Newton, could you check that the leapfrog script is adapted too?

PR #2343, opened 2017-06-15
Remove support role from standard testing

odyssey4me:
As the role's purpose is to implement access and
convenience scripts for operational purposes, it
really does not fit well in development testing.

The role takes over 20 minutes to execute on an
AIO due to the way it is implemented [1]. It is
also completely untested [2].

It should rather be implemented as some sort of
late stage testing before release, rather than
waste time and resources for PR testing.

[1] https://github.com/rsoprivatecloud/openstack-ops/issues/33
[2] https://github.com/rsoprivatecloud/openstack-ops/issues/20


odyssey4me:
recheck_cit_all

alextricity25:

odyssey4me:
All I did was a rebase to bring in the artifact job fixes.

BjoernT:

BjoernT:
👍 OK by me. It should not be your problem or burden, hence it can be removed

odyssey4me:

mattt416:
@odyssey4me @BjoernT side note, with this change the rpc support role will no longer get deployed by _default_.  Is that a behaviour we want to change?

BjoernT:
I'm ok with that we just flip DEPLOY_SUPPORT_ROLE before installing. 
Since I will restructure the role to a stand alone module anyway we will integrate differently in the long term

odyssey4me:
@BjoernT When it's been restructured and there is testing for the role, please ping us so that we can re-enable it for testing to ensure that it's part of the release verification process.

BjoernT:
If this will be back ported to newton 14.1, please let me know so I can talk to the deployment engineers

odyssey4me:
@BjoernT We will very likely port this back to all branches in order to ensure that we're able to continue development without the distractions and time taken when this role runs. The ideal state would be that this role is revised very soon, then we wouldn't need this, but we unfortunately do not have the time and resources to rewrite the role on your behalf.

evrardjp:

odyssey4me:
This patch cannot land until the holland check is removed from RPC-O, because it removes the deployment of holland (it's done in the role).

evrardjp:

evrardjp:
now that the other merged.

PR #2347, opened 2017-06-16
UG-641 Remove neutron checksum workaround

odyssey4me:
The checksum addition using this method was very
unreliable, so in OSA it was added [1] in a more
robust way to the neutron role and removed from
run-playbooks [2].

RPC-O therefore inherits it and does not need
to do this any more. In fact it may interfere
with it so it's better not to use this hack.

[1] https://review.openstack.org/330405
[2] https://review.openstack.org/330389

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -225,22 +225,6 @@ if [[ "${DEPLOY_OA}" == "yes" ]]; then
   # setup the infrastructure
   run_ansible setup-infrastructure.yml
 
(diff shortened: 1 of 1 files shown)


odyssey4me:
https://github.com/rcbops/rpc-openstack/pull/2349 will need to merge to get travis passing

mattt416:

mattt416:

odyssey4me:
I could not get rid of the submodule change, so I've closed this and made a new PR: https://github.com/rcbops/rpc-openstack/pull/2366

PR #2348, opened 2017-06-16
UG-617 Limit the number of Ansible forks used to 10

odyssey4me:
The default MaxSessions setting for SSHD is 10.
Each Ansible fork makes use of a Session, so
this patch still uses the CPU number to set the
number of forks used but limits it to 10 forks
when the number of CPU's is larger.

Bug Reference:
https://bugs.launchpad.net/openstack-ansible/+bug/1479812

The 'FORKS' environment variable is still
catered for in case it is used, but this
changes the implementation of fork setting
from using the CLI parameter to the standard
Ansible environment variable. As such, there
is no need to document the environment variable
in the way that FORKS required it.


odyssey4me:
recheck_xenial_deploy_swift

odyssey4me:

odyssey4me:
All I did was a rebase on top of the artifact job fixes.

PR #2350, opened 2017-06-16
UG-534 Correctly set backup_dir for pre/post upgrades

mattt416:
In order for the post upgrade neutron router tests to inspect state
prior to upgrade, we had to add number of backup_dir-related vars to
the post upgrade role.  Additionally, we added those vars to
group_vars/all.yml so that the pre and post upgrade roles would share
those vars.  What we missed unfortunately is that those vars reference
additional vars, which are evaluated at run-time.  This means if
someone runs an upgrade at 11 PM at night, the backup_dir would be (for
example) /root/rpc13-upgrade-2017-06-16.  By the time the post upgrade
runs, it'll be past midnight, and now backup_dir will be
/root/rpc13-upgrade-2017-06-17.  While this won't cause the post
upgrade role to fail, it won't see the router data from pre upgrade and
not provide any meaningful data to the deployer.

In a nutshell, this commit simply adds an override for backup_dir to
/etc/openstack_deploy/user_upgrade_variables.yml (created temporarily by
test-upgrade.sh) and removes all the unnecessary overrides added to
group_vars/all.yml.  A doc change and note for support will follow.

### DIFF
--- a/rpcd/playbooks/group_vars/all.yml
+++ b/rpcd/playbooks/group_vars/all.yml
@@ -28,9 +28,3 @@ migrations_dir: /etc/openstack_deploy/migrations
 
 # Ceph Conf Directory Mode
 conf_directory_mode: 755
-
-date_stamp: "{{ ansible_date_time.date }}"
-time_stamp: "{{ ansible_date_time.time }}"
-datetime_stamp: "{{ date_stamp }}-{{ time_stamp }}"
-local_home: "{{ lookup('env', 'HOME') }}"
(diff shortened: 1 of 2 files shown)


dankolbrs:
Looks good, but I'm going to hold off for gating. 

Also, a wise person once recommended to me to change [the docs](https://github.com/rackerlabs/docs-rpc/blob/master/doc/rpc-upgrade-internal/rpc-upgrade-v12-v13-perform.rst) on editing of `test-upgrade.sh`.

alextricity25:

PR #2351, opened 2017-06-20
UG-633 [newton] Backport artifact job fixes

odyssey4me:
This PR backports the appropriate non-ocata patches from https://github.com/rcbops/rpc-openstack/pull/2337

PR #2352, opened 2017-06-20
LA-321 Add support for generating test resources

git-harry:
The script `generate_resources.py` is added to provide basic support for
adding resources, e.g. compute instances or block storage volumes, to
allow the impact of an upgrade, from an end-users perspective, to be
measured.

The script assumes the openstack-ansible playbook
`os-tempest-install.yml` has already been run on the deployment.

The playbook `generate-resources.yml` ensures the necessary Pip
dependencies are installed prior to running `generate_resources.py`
using a virtualenv.

An example of the playbook's use is:

openstack-ansible \
    -e server_count=10 \
    -e volume_count=25 \
    -e network_count=12 \
    generate-resources.yml

Issue: [LA-321](https://rpc-openstack.atlassian.net/browse/LA-321)

### DIFF
--- a/scripts/leapfrog/generate_resources.py
+++ b/scripts/leapfrog/generate_resources.py
@@ -1,0 +1,326 @@
+#! /usr/bin/env python
+from argparse import ArgumentParser
+from datetime import datetime, timedelta
+from ipaddress import IPv4Network
+from itertools import chain, izip, repeat, islice
+from json import dumps
+import os
+from time import sleep
+
+from keystoneauth1.identity import v3
+from keystoneauth1 import session as session_
+from cinderclient import client as cinder_client
+from glanceclient import client as glance_client
+from neutronclient.v2_0 import client as neutron_client
+from novaclient import client as nova_client
+
+CLIENT_VERSION = "2"
+
+
+def get_conf():
+    parser = ArgumentParser()
+    parser.add_argument("--servers", type=int, default=1)
+    parser.add_argument("--volumes", type=int, default=0)
+    parser.add_argument("--networks", type=int, default=1)
+    parser.add_argument("--infinite-quotas", action="store_true")
+    args = parser.parse_args()
+    if args.servers > args.networks:
+        raise ValueError(
+            "Each server must have at least one network and so the number of "
+            "networks must be greater than or equal to the number of servers."
+        )
+    if args.servers < 1:
+        raise ValueError(
+            "The minimum number of servers that can be created is 1."
+        )
+
+    conf = {
+        "auth": {},
+        "infinite_quotas": args.infinite_quotas,
+        "resources": {},
+    }
+    conf["auth"]["endpoint_type"] = os.environ["OS_ENDPOINT_TYPE"]
+    conf["auth"]["username"] = os.environ["OS_USERNAME"]
+    conf["auth"]["password"] = os.environ["OS_PASSWORD"]
+    conf["auth"]["project_name"] = os.environ["OS_PROJECT_NAME"]
+    conf["auth"]["auth_url"] = os.environ["OS_AUTH_URL"]
+    conf["auth"]["user_domain_id"] = os.environ["OS_USER_DOMAIN_NAME"]
+    conf["auth"]["project_domain_id"] = os.environ["OS_PROJECT_DOMAIN_NAME"]
+
+    conf["resources"]["servers"] = {
+        "count": args.servers,
+        "flavor": "m1.tiny",
+        "image": "cirros",
+    }
+    conf["resources"]["volumes"] = {
+        "count": args.volumes,
+        "size": 1,
+    }
+    conf["resources"]["networks"] = {
+        "count": args.networks,
+        "external_gateway_network": "public",
+    }
+    return conf
+
+
+def _get_session(conf):
+    auth = v3.Password(
+        auth_url=conf["auth_url"],
+        username=conf["username"],
+        password=conf["password"],
+        project_name=conf["project_name"],
+        user_domain_id=conf["user_domain_id"],
+        project_domain_id=conf["project_domain_id"],
+    )
+
+    return session_.Session(auth=auth, verify=True)
+
+
+def session_caching_wrapper():
+    cached = {}
+
+    def get_session():
+        if not cached.get("session"):
+            conf = get_conf()
+            cached["session"] = _get_session(conf["auth"])
+        return cached["session"]
+    return get_session
+
+
+get_session = session_caching_wrapper()
+
+
+def get_flavor_id_from_name(name):
+    nova = nova_client.Client(CLIENT_VERSION, session=get_session())
+    flavors = [
+        flavor for flavor in nova.flavors.list() if flavor.name == name
+    ]
+    if len(flavors) > 1:
+        raise Exception("Flavor name, '{}', is not unique.".format(name))
+    try:
+        return flavors[0].id
+    except IndexError:
+        raise Exception("No flavor with name '{}' exists.".format(name))
+
+
+def get_network_id_from_name(name):
+    neutron = neutron_client.Client(session=get_session())
+    networks_ = neutron.list_networks(name=name)
+    networks = networks_["networks"]
+    if len(networks) > 1:
+        raise Exception("Network name, '{}', is not unique.".format(name))
+    try:
+        return networks[0]["id"]
+    except IndexError:
+        raise Exception("No network with name '{}' exists.".format(name))
+
+
+def get_image_id_from_name(name):
+    glance = glance_client.Client(CLIENT_VERSION, session=get_session())
+    images = list(glance.images.list(name=name))
+    if len(images) > 1:
+        raise Exception("Image name, '{}', is not unique.".format(name))
+    try:
+        return images[0].id
+    except IndexError:
+        raise Exception("No image with name '{}' exists.".format(name))
+
+
+def verify_active(building, timeout=600):
+    nova = nova_client.Client(CLIENT_VERSION, session=get_session())
+    active = []
+    delay = 10
+    timeout_ = datetime.utcnow() + timedelta(seconds=timeout)
+    while building:
+        if datetime.utcnow() < timeout_:
+            sleep(delay)
+        else:
+            raise Exception(
+                "{} servers failed to build within {} seconds.".format(
+                    len(building),
+                    timeout,
+                )
+            )
+
+        still_building = []
+        for server in building:
+            server = nova.servers.get(server.id)
+            if server.status == "ACTIVE":
+                active.append(server)
+            elif server.status == "BUILD":
+                still_building.append(server)
+            else:
+                raise Exception(
+                    "The resource {} is in an invalid state.".format(
+                        server.id
+                    )
+                )
+        building = still_building
+    return active
+
+
+def make_networks(conf):
+    neutron = neutron_client.Client(session=get_session())
+    net = IPv4Network(u'10.0.0.0/8')
+    snets = net.subnets(prefixlen_diff=16)
+    ext_gateway_net = get_network_id_from_name(
+        conf["external_gateway_network"]
+    )
+    for number, cidr in enumerate(snets):
+        network = neutron.create_network(
+            {
+                "network": {
+                    "name": "test_network_{}".format(number),
+                },
+            }
+        )
+        subnet = neutron.create_subnet(
+            {
+                "subnet": {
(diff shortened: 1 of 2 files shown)


git-harry:
recheck_cit_all

git-harry:
recheck_cit_artifact_trusty

git-harry:
recheck_cit_artifact_trusty_apt

git-harry:
recheck_cit_artifact_trusty_git
recheck_cit_artifact_trusty_python

git-harry:
recheck_cit_artifact_xenial_apt
recheck_cit_artifact_xenial_container
recheck_cit_artifact_xenial_git
recheck_cit_artifact_xenial_python

git-harry:
Blocked by https://github.com/rcbops/rpc-openstack/pull/2351

scripts/leapfrog/generate_resources.py
```diff
@@ -0,0 +1,322 @@
...
+                        server.id
+                    )
+                )
+        building = still_building
+        counter += delay
+        if counter < timeout:
+            sleep(delay)
+        else:
+            raise Exception(
+                "All resources failed to build within {} seconds.".format(
```
evrardjp: nit: "Timeout reached and some resources didn't build in a timely fashion" vs "All failed"

scripts/leapfrog/generate_resources.py
```diff
@@ -0,0 +1,322 @@
...
+                )
+            )
+    return active
+
+
+def make_networks(conf):
+    neutron = neutron_client.Client(session=get_session())
+    net = IPv4Network(u'10.0.0.0/8')
+    snets = net.subnets(16)
+    ext_gateway = get_network_id_from_name(conf["external_gateway"])
```
evrardjp: ext_gateway is confusing, it sounds like a gateway, while it's a network.

scripts/leapfrog/generate_resources.py
```diff
@@ -0,0 +1,322 @@
...
+                    timeout,
+                )
+            )
+    return active
+
+
+def make_networks(conf):
+    neutron = neutron_client.Client(session=get_session())
+    net = IPv4Network(u'10.0.0.0/8')
+    snets = net.subnets(16)
```
evrardjp: That sounds static. Is that generating only 16 subnets, or creating /16 subnets, or adding a 16 bits to 8 for netmask? I'd be happy to see a small word of explanation on what it's doing and why.

mancdaz:
recheck_cit_artifact_trusty_apt
recheck_cit_artifact_xenial_apt
recheck_cit_artifact_xenial_container

scripts/leapfrog/generate_resources.py
```diff
@@ -0,0 +1,322 @@
...
+            {
+                "subnet": {
+                    "network_id": network["network"]["id"],
+                    "ip_version": 4,
+                    "cidr": cidr,
+                    "name": "test_subnet_{}".format(number),
+                },
+            }
+        )
+        router = neutron.create_router(
```
evrardjp: The routers don't need to be created linearily, see card comment:
"For the subnets, we should create a router every 2 subnets created, attaching the two networks together."

But I think this is an okay-ish assumption for simplification. Everything is simplified, and all the tenant networks are now linked to the outside world. So, 1:1 mapping of subnet/routers.

We could iterate there.

scripts/leapfrog/generate_resources.py
```diff
@@ -0,0 +1,322 @@
+#! /usr/bin/env python
+from argparse import ArgumentParser
```
git-harry: It's passed the check. What doesn't look right to you?

evrardjp: my bad then.

scripts/leapfrog/generate_resources.py
```diff
@@ -0,0 +1,322 @@
...
+
+
+def get_conf():
+    parser = ArgumentParser()
+    parser.add_argument("--servers", type=int, default=1)
+    parser.add_argument("--volumes", type=int, default=0)
+    parser.add_argument("--networks", type=int, default=1)
+    parser.add_argument("--infinite-quotas", action="store_true")
+    args = parser.parse_args()
+    if args.servers > args.networks:
```
git-harry: The request was to keep it simple, there are obviously multiple ways these things can be combined however it only needs to support one at this time, the pattern I chose was to create separate networks for each server. Is this going to cause an issue with the testing?

scripts/leapfrog/generate_resources.py
```diff
@@ -0,0 +1,322 @@
...
+from time import sleep
+
+from keystoneauth1.identity import v3
+from keystoneauth1 import session as session_
+from cinderclient import client as cinder_client
+from glanceclient import client as glance_client
+from neutronclient.v2_0 import client as neutron_client
+from novaclient import client as nova_client
+
+CLIENT_VERSION = "2"
```
git-harry: Given this is only set once, you shouldn't need to keep referring back to it. I set it here so I didn't need to specify the number in multiple places. DRY

scripts/leapfrog/generate_resources.py
```diff
@@ -0,0 +1,322 @@
...
+        "auth": {},
+        "infinite_quotas": args.infinite_quotas,
+        "resources": {},
+    }
+    conf["auth"]["endpoint_type"] = os.environ["OS_ENDPOINT_TYPE"]
+    conf["auth"]["username"] = os.environ["OS_USERNAME"]
+    conf["auth"]["password"] = os.environ["OS_PASSWORD"]
+    conf["auth"]["project_name"] = os.environ["OS_PROJECT_NAME"]
+    conf["auth"]["auth_url"] = os.environ["OS_AUTH_URL"]
+    conf["auth"]["user_domain_id"] = os.environ["OS_USER_DOMAIN_NAME"]
```
git-harry: These settings are placed in the openrc file, I didn't see a comment there qualifying their use. What would you hope a comment would say?

scripts/leapfrog/generate_resources.py
```diff
@@ -0,0 +1,322 @@
...
+
+    def get_session():
+        if not cached.get("session"):
+            conf = get_conf()
+            cached["session"] = _get_session(conf["auth"])
+        return cached["session"]
+    return get_session
+
+
+get_session = session_caching_wrapper()
```
git-harry: The idea of it was to save me having to pass a session object around.

scripts/leapfrog/generate_resources.py
git-harry: I can see how this is ambiguous, I'll change it.

scripts/leapfrog/generate_resources.py
git-harry: I don't think it makes sense to start documenting functionality from other modules.
https://docs.python.org/3/library/ipaddress.html#ipaddress.IPv4Network.subnets

scripts/leapfrog/generate_resources.py
git-harry: Along the lines I mentioned above, if the pattern your suggesting is because of then some specific testing requirement I can change it. If not, what is the reason it would be better?

scripts/leapfrog/playbooks/generate-resources.yml
```diff
@@ -0,0 +1,36 @@
...
+  hosts: utility_all[0]
+  tasks:
+    - name: Install virtualenv
+      pip:
+        name: "virtualenv"
+    - name: Prepare a virtualenv for generate_resources.py
+      pip:
+        name: "{{ item }}"
+        virtualenv: "{{ venv }}"
+        extra_args: "--isolated"
```
git-harry: keystoneauth1 wasn't the right version so I needed to isolate it.

scripts/leapfrog/playbooks/generate-resources.yml
```diff
@@ -0,0 +1,36 @@
...
+      environment:
+        OS_ENDPOINT_TYPE: "internalURL"
+        OS_USERNAME: "admin"
+        OS_PASSWORD: "{{ keystone_auth_admin_password }}"
+        OS_PROJECT_NAME: "admin"
+        OS_AUTH_URL: "{{ keystone_service_internalurl_v3 }}"
+        OS_USER_DOMAIN_NAME: "Default"
+        OS_PROJECT_DOMAIN_NAME: "Default"
+        PATH: "{{ venv }}/bin"
+    - name: Remove virtualenv
```
git-harry: Yup, I was simply tidying up after myself.

scripts/leapfrog/generate_resources.py
git-harry: Naming is hard, what about ext_gateway_network?

scripts/leapfrog/playbooks/generate-resources.yml
evrardjp: good.

scripts/leapfrog/playbooks/generate-resources.yml
evrardjp: I guess it doesn't really matter if we have idempotency or not, as the whole process is completely not idempotent :p

scripts/leapfrog/generate_resources.py
evrardjp: What I meant here is CLIENT_VERSION 2 is used for glance and nova, is there a risk this would be causing issues with different versions of a deploy (Kilo vs Liberty, ...)

scripts/leapfrog/generate_resources.py
evrardjp: Not an issue by itself, but it's restricting what we can do in the testing. I think if we are looking about numbers, I don't see a hard issue of removing the restriction args.servers > args.networks, and deal with it.

scripts/leapfrog/generate_resources.py
evrardjp: Nevermind. We are just reusing.

scripts/leapfrog/generate_resources.py
evrardjp: Why not in main() ?

scripts/leapfrog/generate_resources.py
evrardjp: Thanks

scripts/leapfrog/generate_resources.py
evrardjp: Well, you don't use kwargs, so it makes the "16" less readable.

scripts/leapfrog/generate_resources.py
git-harry: done

evrardjp: https://twitter.com/secretGeek/status/7269997868

evrardjp: ext_gateway_network sounds better.

scripts/leapfrog/generate_resources.py
evrardjp: On this one, I think it's okay. I imagine we are here testing the external connectivity instead of testing intra tenant networking, which is not very important, because we don't have any view of how openstack is used in our customer environments (multi tier apps?)

In any case, we can change that later.

scripts/leapfrog/generate_resources.py
git-harry: This script was designed/tested on kilo. Anything else is outside the requirements given to me.

evrardjp: Fair enough. I'm gonna create a card to improve this generation system for other releases.

scripts/leapfrog/generate_resources.py
git-harry: Well the script requires each server to have at least one separate associated network. I can obviously change that, in which case this condition is not necessary, but when I raised this at the beginning there were no hard requirements on how the different resources should be plugged together beyond being able to specify a variable number.

scripts/leapfrog/generate_resources.py
git-harry: I'm not sure what specifically you are querying here. `session_caching_wrapper()` returns a closure, this allows `get_session` to be called whenever a session is required and so saves having to pass a session object around.

scripts/leapfrog/generate_resources.py
git-harry: That is a fair point, kwargs are self-documenting, I'll change this.

mancdaz:
@hughsaunders would very much appreciate some eyes if you have the cycles

mancdaz:
2 approved reviews, merging

PR #2357, opened 2017-06-21
K8S-116 Remove magnum deployment

odyssey4me:
The approach for implementing containers is
going to be completely re-done. This patch
removes the current approach to prevent bit
rot.

The remaining Octavia bits are adjusted to
ensure that the SHA's are fixed at a point
in time. As the development work for that
is progressed, those SHA's may be adjusted.

@jnoller says we can do it. :)

Note that this does not prevent magnum being
deployed with RPC-O, but it does remove the
scenario/test and the hacks implemented.


ansible-role-requirements.yml
```diff
@@ -14,10 +14,6 @@
   scm: git
   src: https://github.com/rsoprivatecloud/openstack-ops
   version: f87406880f67584b5fe562d2711734bbd63c0fdb
-- name: os_octavia
```
hughsaunders: [As discussed in slack](https://rackspace.slack.com/archives/C0EUKPSFR/p1498049639362131) we should probably leave the octavia bits.

PR #2360, opened 2017-06-21
UG-641 Remove neutron agent iptables from deploy.sh

mattt416:
This commit updates deploy.sh by removing the iptables rules which get
dropped inside the neutron agents containers.  The os_neutron role in
stable/mitaka handles this when neutron_metadata_checksum_fix is set,
and this variable is automatically overriden by the AIO bootstrap.


mattt416:
There's no reason for this to exist in mitaka-13.1's deploy.sh, right @odyssey4me ?

mattt416:
Closing, just realised https://github.com/rcbops/rpc-openstack/pull/2347 already exists for this.

PR #2367, opened 2017-06-23
[WIP] UG-653 Move test_holland.yml from rpc-gating

mattt416:
This commit moves rpc-gating's test_holland.yml into rpc-openstack.
This change is part of a wider effort to reduce the logic and files
that exist within rpc-gating and making them widely available to the
repos beings tested.

Issue: [UG-653](https://rpc-openstack.atlassian.net/browse/UG-653)

### DIFF
--- a/scripts/test_holland.yml
+++ b/scripts/test_holland.yml
@@ -1,0 +1,21 @@
+---
+
+- hosts: galera
+  user: root
+  vars:
+    # This fallback for when rpc_release is not defined is to cater for older releases
+    # where rpc_release is not available as a variable, but is instead derived by the
+    # execution of a module. If we are no longer testing branches older than Newton,
+    # the fallback mechanism can be removed.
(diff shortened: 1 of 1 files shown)


mattt416:

dankolbrs:
@mattt416 is this still `WIP` ?

mattt416:
@dankolbrs @odyssey4me Yeah, I'm a bit unclear now in light of https://github.com/rcbops/rpc-openstack/pull/2343.  Should this test not live in rpc-openstack?  If https://github.com/rsoprivatecloud/openstack-ops is going to do their own gate testing then we can probably allow testing to be done there, but I have no idea if/when that is going to happen.

odyssey4me:
@mattt416 I'm inclined to remove the holland tests from rpc-gating and to register an issue in the ops repo to have them test there, then leave it to them to execute.

mattt416:
Closing this, it would make most sense for this test to go into the support role.  It seems pointless merging this, backporting, and then undoing all those changes once it's landed in the support role.

PR #2368, opened 2017-06-28
UG-645 Port artifact gathering to rpc-o script

dankolbrs:
Ports artifact gathering function to bash script and adds to
rpc-o.

Adds gating top level directory for scripts and playbooks only
used for gating.

Issue: [UG-645](https://rpc-openstack.atlassian.net/browse/UG-645)

### DIFF
--- a/scripts/gather-artifacts.sh
+++ b/scripts/gather-artifacts.sh
@@ -1,0 +1,47 @@
+#!/usr/bin/env bash
+#
+# Copyright 2014-2017, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# -------------- Import ---------------------#
+export BASE_DIR=${BASE_DIR:-"/opt/rpc-openstack"}
+source ${BASE_DIR}/scripts/functions.sh
+# ------------ End import -------------------#
+
+export WORKSPACE=${WORKSPACE:-"$HOME"}
+
+echo "Archiving logs and configs..."
+d="${WORKSPACE}/logs"
+mkdir -p $d
+
+# logs and config from host
+mkdir -p $d/$HOSTNAME/log
+cp -rp /openstack/log/$HOSTNAME-* $d/$HOSTNAME/log ||:
+cp -rp /etc/ $d/$HOSTNAME/etc
+cp -rp /var/log/ $d/$HOSTNAME/var_log
+
+# logs and config from the containers
+while read c; do
+  mkdir -p $d/$c/log
+  cp -rp /openstack/log/$c/* $d/$c/log 2>/dev/null ||:
+  cp -rp /var/lib/lxc/$c/rootfs/etc $d/$c 2>/dev/null ||:
+  cp -rp /var/lib/lxc/$c/delta0/etc $d/$c 2>/dev/null ||:
+done < <(lxc-ls)
+
+# compress to reduce storage space requirements
+ARTIFACT_SIZE=$(du -sh $d | cut -f1)
+echo "Compressing $ARTIFACT_SIZE of artifact files..."
+tar cjf "$d".tar.bz2 $d
+echo "Compression complete."
+rm -rf ${d}


hughsaunders:
This is 645.

dankolbrs:
Updated title and commit message for card

dankolbrs:
@odyssey4me , updated commit separating artefact gathering and upload

scripts/archive-artifacts.sh
```diff
@@ -0,0 +1,50 @@
...
+# -------------- Import ---------------------#
+export BASE_DIR=${BASE_DIR:-"/opt/rpc-openstack"}
+source ${BASE_DIR}/scripts/functions.sh
+# ------------ End import -------------------#
+
+#---------- Required Vars -------------------#
+export BUILD_TAG=${BUILD_TAG:-""}
+export CONTAINER=${CONTAINER:-""}
+export SRC=${SRC:-""}
+#--------------- End Vars -------------------#
```
odyssey4me: I'm not sure I understand the point of these vars. It places an unnecessary complexity on the repo owner to have to do these things.

My suggestion would be to simply make the directory something like ${WORKSPACE}/logs - put everything in there, then tarball/compress into logs.tar.bz2

The gating system then takes that file, renames it or whatever need to be done and handles the upload. The only requirement for the repo owner is to have a logs.tar.bz2 in the workspace (or some well known directory which *all* repositories use).

dankolbrs: Good call, I missed those in moving the variables, which showed a larger issue with `BUILD_TAG` in the upload script.

Latest commit removes and uses the `${WORKSPACE}/log` directory

scripts/archive-artifacts.sh
```diff
@@ -0,0 +1,46 @@
...
+source ${BASE_DIR}/scripts/functions.sh
+# ------------ End import -------------------#
+
+export WORKSPACE=${WORKSPACE:-"$HOME"}
+
+echo "Archiving logs and configs..."
+d="${WORKSPACE}/logs"
+mkdir -p $d
+
+# logs and config from the single use slave
```
odyssey4me: Can you change this comment to something like:

# logs and config from the host

It's more generic.

scripts/archive-artifacts.sh
```diff
@@ -0,0 +1,46 @@
...
+mkdir -p $d/$HOSTNAME/log
+cp -rp /openstack/log/$HOSTNAME-* $d/$HOSTNAME/log ||:
+cp -rp /etc/ $d/$HOSTNAME/etc
+cp -rp /var/log/ $d/$HOSTNAME/var_log
+
+# logs and config from the containers
+while read c; do
```
odyssey4me: This one caught me by surprise, the population of ``c`` at the end of the loop is confusing.

I'd suggest rather using a for loop like so:

```
for c in $(lxc ls -1); do
...
done
```

hughsaunders: I prefer while read, it avoids expanding the command within the substitution into a single line which is subject to length limitation. The redirect after done sets the loop's standard input, c is populated at the beginning of each iteration as with the for example.

scripts/archive-artifacts.sh
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env bash
```
hughsaunders: Maybe gather or prepare artefacts would be a better name as this script doesn't actually archive them, just prepares them in order for Jenkins to archive them.

scripts/archive-artifacts.sh
odyssey4me: Interesting - fair enough.

scripts/archive-artifacts.sh
dankolbrs: done

scripts/archive-artifacts.sh
dankolbrs: I prefer the `for c in $(lxc ls -1); do` as well, but didn't feel the need to change the way the loop was done in the port.

scripts/archive-artifacts.sh
dankolbrs: done

dankolbrs:
@hughsaunders , should this be run on every deploy? The log gathering is probably more valid after running tempest or other tests or upgrading, rather than immediately after deploy. I'm not sure that at this time it would be worth the additional time to gather and compress log files on every rpc-openstack build when it's currently only done by Jenkins after upgrades

alextricity25:
I agree with @dankolbrs. It doesn't make much sense to run this after each deploy before running tempest.

odyssey4me:
@dankolbrs @alextricity25 Agreed. We might need to get the tempest execution moved into rpc-o (instead of the gating executing it as a separate stage) before it makes sense to use this. That would allow the test to run and then artifacts gathered. Either that or we do some sort of interim stage thing while we transition.

hughsaunders:
My bad, I thought we had a conditional to run tempest from deploy.sh. As we don't, calling this script will have to wait until we have a single entry point gate script ([UG-645](https://rpc-openstack.atlassian.net/browse/UG-647)).

scripts/gather-artifacts.sh
```diff
@@ -0,0 +1,47 @@
...
+# logs and config from the containers
+while read c; do
+  mkdir -p $d/$c/log
+  cp -rp /openstack/log/$c/* $d/$c/log 2>/dev/null ||:
+  cp -rp /var/lib/lxc/$c/rootfs/etc $d/$c 2>/dev/null ||:
+  cp -rp /var/lib/lxc/$c/delta0/etc $d/$c 2>/dev/null ||:
+done < <(lxc-ls)
+
+# compress to reduce storage space requirements
+ARTIFACT_SIZE=$(du -sh $d | cut -f1)
```
mattt416: Should we skip creating an archive if `ARTIFACT_SIZE` exceeds a certain amount?

odyssey4me: Why don't we collect the whole folder, instead of forcing the collection of a single archive? That way we can restrict individual file sizes and the files can be viewed directly in the artifacts link for the job instead of having to decompress. We can encourage the compression of each individual file, much like OpenStack-Infra does.

odyssey4me: Well, never mind - this is a current state conversion. My question is more aimed at the general facility rather than this patch in particular.

@mattt416 I do not think that file size restrictions apply here, although they can skip adding a file to the archive if it's too large - but that's down to the product owner to do in their repo, so they'll need to understand why it's good for them to do it.

mattt416: Yeah, I think my suggestion is not suitable for this iteration since we're just porting functionality over to rpc-openstack.  I do think this warrants a discussion afterwards though.

PR #2373, opened 2017-06-30
UG-641 Don't create untracked files in os_neutron

mattt416:
In [1], we backported a change from stable/mitaka's os_neutron and
apply via the patcher role. This is working fine except it is causing
an issue when we upgrade from liberty-12.2->mitaka-13.1. The issue is
that the patch file creates new, untracked files and these don't get
cleaned up when the gate job runs a git reset --hard. When the
rpc-openstack checkout is switched to mitaka-13.1 these few untracked
files are left behind, meaning ansible tries to use an imcomplete role
in rpcd/playbooks/roles instead of /etc/ansible/roles.

A simple solution would be to update the gate job to remove these
untracked file before/after running a git reset, however this would
also need to be done by support and require a change to their
maintenance plan.

To avoid that, this commit simply removes the new file from the
os_neutron patch, and stores it under rpcd/patches/os_neutron instead.
The patched os_neutron playbook then simply references this file in
rpcd/patches/os_neutron instead of under
openstack-ansible/playbooks/roles/os_neutron/files.

Note that we also move the os_neutron patch file previously in
rpcd/patches into rpcd/patches/os_neutron, to keep things a little more
tidy.

[1] https://github.com/rcbops/rpc-openstack/pull/2370

Issue: [UG-641](https://rpc-openstack.atlassian.net/browse/UG-641)


mattt416:
I'm currently testing a liberty->mitaka upgrade in https://rpc.jenkins.cit.rackspace.net/job/mattt-RPC-AIO_mitaka-trusty-majorupgrade-swift-periodic/13/, please hold off on merging this until we know the upgrade is passing.

hughsaunders:
Would it be easier to add git clean -df to the checkout?

mattt416:
https://rpc.jenkins.cit.rackspace.net/job/mattt-RPC-AIO_mitaka-trusty-majorupgrade-swift-periodic/13/ failed, with what I /think/ is an unrelated issue.  I logged into the VM and was able to run tempest scenario tests without issue.  I've spun up [another test](https://rpc.jenkins.cit.rackspace.net/job/mattt-RPC-AIO_mitaka-trusty-majorupgrade-swift-periodic/16/console) to be sure but I think we're probably OK to merge this.

mattt416:
The MaaS failure I keep running into is legitimate, but unrelated to this PR.  We're actually seeing this failing in the periodic jobs so I'll create a bug for this so we can address.

I think we're OK to merge this one.

PR #2374, opened 2017-06-30
[WIP] UG-646 Handle config within RPCO

hughsaunders:

build that is required.

This duplicates gating functionality so must not be merged until it's
counterpart in rpc-gating is merged:
  https://github.com/rcbops/rpc-gating/pull/266


hughsaunders:
Closing this PR to prevent unproductive gate jobs on push.

PR #2187, opened 2017-04-21
Remove the nova environment file

cloudnull:
This file was only used to get a round a bug. That bug was fixed a long
while ago so this env.d file does nothing.

PR that fixed the bug:
  https://review.openstack.org/#/c/411348/

Signed-off-by: Kevin Carter <kevin.carter@rackspace.com>

Issue: [UG-651](https://rpc-openstack.atlassian.net/browse/UG-651)

PR #2375, opened 2017-06-30
[AR-192] Move env variable overrides to groupvars

alextricity25:
This commit moves as many property variables possible from env.d
overrides to group_vars.

Issue: [AR-192](https://rpc-openstack.atlassian.net/browse/AR-192)


group_vars/cinder_api_container.yml
```diff
@@ -13,8 +13,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-container_skel:
-  repo_container:
-    properties:
-      service_name: repo
-      lxc_container_variant: "repo_server-{{ rpc_release }}"
+lxc_container_variant: "cinder-{{ rpc_release }}"
```
odyssey4me: Any reason why you're not using the 'cinder_all' group here? The variant is the same for all cinder containers and this var will only apply to containers anyway.

scripts/bootstrap-aio.yml
```diff
@@ -132,26 +132,12 @@
...
           mode: "0440"
-        - name: "env.d/cinder.yml"
         - name: "env.d/elasticsearch.yml"
           condition: "{{ rpco_deploy_elk | bool }}"
-        - name: "env.d/glance.yml"
-        - name: "env.d/heat.yml"
-        - name: "env.d/horizon.yml"
-        - name: "env.d/ironic.yml"
-        - name: "env.d/keystone.yml"
         - name: "env.d/kibana.yml"
```
odyssey4me: I think that we can move the variant setting for the ELK roles from the env.d files into group_vars too, for consistency.

rpcd/etc/openstack_deploy/env.d/nova.yml
```diff
@@ -1,45 +0,0 @@
...
-      lxc_container_variant: "nova-{{ rpc_release }}"
-  nova_api_os_compute_container:
-    properties:
-      service_name: nova
-      lxc_container_variant: "nova-{{ rpc_release }}"
-  nova_cert_container:
-    properties:
-      service_name: nova
-      lxc_container_variant: "nova-{{ rpc_release }}"
-  nova_compute_container:
```
odyssey4me: This should technically stay, but I see that https://github.com/rcbops/rpc-openstack/pull/2187 is trying to remove it.

alextricity25:
Thanks for the feedback, Jesse. I'm working on addressing your comments now.

group_vars/cinder_api_container.yml
alextricity25: This value was copied over from what was previously in the `rpcd/etc/openstack_deploy/env.d/cinder.yml` file. Should they be using the `*_all` groups instead?

scripts/bootstrap-aio.yml
alextricity25: Done.

group_vars/cinder_api_container.yml
odyssey4me: Yup, it achieves the same goal and cuts a lot of duplication.

alextricity25: Done.

group_vars/elasticsearch_all.yml
```diff
@@ -0,0 +1,2 @@
+---
```
odyssey4me: Add license boilerplate.

group_vars/kibana_all.yml
```diff
@@ -0,0 +1,2 @@
+---
```
odyssey4me: Add license boilerplate.

group_vars/logstash_all.yml
```diff
@@ -0,0 +1,2 @@
+---
```
odyssey4me: Add license boilerplate.

rpcd/etc/openstack_deploy/env.d/nova.yml
odyssey4me: I would suggest reverting the nova env.d file to https://github.com/rcbops/rpc-openstack/blob/db55a85c6cb7a58e0e7d996a524eac81aefb1c54/rpcd/etc/openstack_deploy/env.d/nova.yml in this commit, but then removing it in another commit as per https://github.com/rcbops/rpc-openstack/pull/2187 so that we deal with both issues in a single PR but have a record of both issues being solved independently.

group_vars/kibana_all.yml
alextricity25: Done.

group_vars/logstash_all.yml
alextricity25: Done.

rpcd/etc/openstack_deploy/env.d/nova.yml
alextricity25: Sure thing.

alextricity25: Done.

PR #2376, opened 2017-07-03
[LA-339] Make pre_redeploy script friendlier

evrardjp:
When running outside gates, the pre_redeploy script didn't
ask for a verification of the variable migration.

This commit ensures to ask a user verification of the variable
migration, but also notices the user shouldn't interrupt the
process to verify them. As a result, the process should be
more user-friendly.

### DIFF
--- a/scripts/leapfrog/pre_redeploy.sh
+++ b/scripts/leapfrog/pre_redeploy.sh
@@ -15,10 +15,24 @@
 #
 # (c) 2017, Jean-Philippe Evrard <jean-philippe.evrard@rackspace.co.uk>
 
+## Notice:
+# Please do not run this script separately. This script is meant to run
+# Inside the OSA Leapfrog.
+
 ## Shell Opts ----------------------------------------------------------------
 set -e -u -x
 set -o pipefail
 
+# These variables should have been defined in the main script, with a
+# conditional value depending on if this run is inside our gates or not
+# For safety, we ensure that these variables are defined, and set
+# to the safest value if not already defined.
+# The following must be TRUE to skip questions
+export IS_GATING="${IS_GATING:-FALSE}"
+# The following must be "--for-testing-take-new-vars-only" to skip questions
+export AUTOMATIC_VAR_MIGRATE_FLAG="${AUTOMATIC_VAR_MIGRATE_FLAG:-}"
+
+warning "Please DO NOT interrupt this process."
 notice "Pre redeploy steps"
 pushd ${LEAPFROG_DIR}
     if [[ ! -f "${UPGRADE_LEAP_MARKER_FOLDER}/variable-migration.complete" ]]; then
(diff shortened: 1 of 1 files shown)


evrardjp:
Please note this is WIP.

evrardjp:
Could someone have a look why trusty/xenial-swift doesn't work? It's not linked to my commit.

scripts/leapfrog/pre_redeploy.sh
```diff
@@ -73,6 +87,11 @@ pushd ${LEAPFROG_DIR}
                 log "user_secrets_migration" "skipped"
             fi
         popd
+        if [[ "${IS_GATING}" != "TRUE" ]]; then
+            notice "Please verify your migrated secrets in /etc/openstack_deploy"
```
git-harry: Where do `notice` and `warning` come from?

scripts/leapfrog/pre_redeploy.sh
```diff
@@ -73,6 +87,11 @@ pushd ${LEAPFROG_DIR}
...
+            warning "DO NOT CTRL-C this process to verify your secrets."
```
git-harry: Why not? What happens if I do?

scripts/leapfrog/pre_redeploy.sh
evrardjp: This script is sourced in the main process, that defines this.

scripts/leapfrog/pre_redeploy.sh
evrardjp: You're on your own then.
The idea is to never interrupt the leapfrog.

It should be fine theoretically, because the process will skip this part (this warning happened at the end) + the parts before.
Interrupting and restarting will continue to the next steps, and not show this warning anymore.

However, some files were already migrated, and some containers were not, so we are sitting between two chairs, and it can be weird.

evrardjp: Oh I see, the log line should be before these lines. Thanks for the notice!

evrardjp:
Updated.

mancdaz:
trusty/xenial-swift failed on an instance reboot test. Assuming it's a transient issue and trying again:

mancdaz:

evrardjp:
Thanks for babysitting @mancdaz :p

evrardjp:
Ready to go

scripts/leapfrog/pre_redeploy.sh
```diff
@@ -74,6 +88,14 @@ pushd ${LEAPFROG_DIR}
             fi
         popd
         log "variable-migration" "ok"
+        # This message only needs to appears the first time the user variable migration was successful
+        # This way, if a user mistakenly Ctrl-C, the variable migration process will be skipped on
+        # the next run
+        if [[ "${IS_GATING}" != "TRUE" ]]; then
+            notice "Please verify your migrated secrets in /etc/openstack_deploy"
+            warning "DO NOT CTRL-C this process to verify your secrets."
```
git-harry: I don't think this is very helpful, it implies there will be a negative impact from terminating the script however given the use of `.complete` files there don't appear to be any. If someone does it by mistake, this message creates FUD.

evrardjp: If someone does it, it will imply a need for workarounding, or weird expectations, as expressed before.

It's better if we do it that way, IMO, for user friendliness.

Issue #2377, opened 2017-07-03
maas_use_api removed from RPC-O but still referenced


hughsaunders:
> **THIS IS A PUBLIC FACING ISSUE.**
>
> Do not include private or confidential information in this issue.

## Is this a feature or a bug?

- [ ] Feature _(request for enhancement, new functionality)_
- [x] Bug _(something is broken)_

## Which version of rpc-openstack does this issue relate to?
- [x] master
- [ ] newton-14.0
- [ ] mitaka-13.1
- [ ] liberty-12.2

Run `git log --pretty=oneline | head -n 1` and provide the output below between
the backtick lines:

```
4509a36 Merge pull request #2358 from mattt416/ug/651-master
```

## What happened?

Gate job failed because maas_use_api is referenced but not defined:
```
PLAY [Verify MaaS API setup] ***************************************************

TASK [Verify Checks & Alarms are registered] ***********************************
Friday 30 June 2017  15:57:01 +0000 (0:00:34.474)       0:00:36.337 *********** 
fatal: [ramxdsp-16-2a6e]: FAILED! => {"failed": true, "msg": "The conditional check 'maas_use_api | bool' failed. The error was: error while evaluating conditional (maas_use_api | bool): 'maas_use_api' is undefined\n\nThe error appears to have been in '/opt/rpc-openstack/rpcd/playbooks/verify-maas.yml': line 21, column 7, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n  tasks:\n    - name: \"Verify Checks & Alarms are registered\"\n      ^ here\n"}
```
https://rpc.jenkins.cit.rackspace.net/blue/organizations/jenkins/RPC-AIO_master-xenial-deploy-swift-periodic/detail/RPC-AIO_master-xenial-deploy-swift-periodic/16/pipeline

Verify-maas.yml uses a variable `maas_use_api` which is defined in the rpc-maas role defaults, and accessed via a vars_file include.  [PR 2296](https://github.com/rcbops/rpc-openstack/pull/2296) removed the role and the [vars_file include](https://github.com/rcbops/rpc-openstack/pull/2296/files#diff-3ed445628fb90724fdcaf0845937b356L76) so now that variable is undefined and the veryify-maas playbook fails. 

## What were you doing when it happened?

Happened during a periodic gate job. 

## What did you expect to happen instead of what actually happened?

I expect all required variables have suitable defaults.

[...]

PR #2378, opened 2017-07-03
Add default value for maas_use_api

hughsaunders:
Add default value for maas_use_api as not all scenarios define it.
This playbook will probably removed in a future refactor, but for
now we need to unblock the gate.

Connects: rcbops/rpc-openstack#2377.

### DIFF
--- a/rpcd/playbooks/verify-maas.yml
+++ b/rpcd/playbooks/verify-maas.yml
@@ -52,3 +52,6 @@
       delay: "5"
       when:
         - maas_use_api | bool
+
+  vars:
+    maas_use_api: true


rpcd/playbooks/verify-maas.yml
```diff
@@ -52,3 +52,6 @@
       delay: "5"
       when:
         - maas_use_api | bool
+
+  vars:
+    maas_use_api: true
```
evrardjp: this force a non-overridable maas_use_api: True, overridable only through a CLI var.
I'd put that into a group var instead.

hughsaunders: Hmm, I checked and it was [true previously](https://github.com/rcbops/rpc-openstack/pull/2166/files#diff-1d0825392161b9f338fa4db5ddd23713L616). However we do need to be able to override as we use false for PR jobs.

evrardjp: I'd rather leave it as False by default, and override it to True on some cases in our gating "contract" script.

evrardjp: Also, I think we should change the when: to be default(False) .

cloudnull:
recheck_cit_artifact-xenial-Git

hughsaunders:
failed to download unattended upgrades, recheck_cit_artifact-xenial-Git

hughsaunders:
failed to download unattended upgrades, recheck_cit_artifact_xenial_Git

PR #2379, opened 2017-07-03
Remove all of the maas API tests

cloudnull:
This change removes all of the maas API testing because the
functionality is being moved to the rpc-maas repo.

Related: https://github.com/rcbops/rpc-maas/pull/290

Issue: [UG-651](https://rpc-openstack.atlassian.net/browse/UG-651)


scripts/deploy-rpc-playbooks.sh
```diff
@@ -36,13 +36,15 @@ if [[ "${DEPLOY_ELK}" == "yes" ]]; then
   run_ansible setup-logging.yml
 fi
 
+# Download the latest release of rpc-maas
+run_ansible maas-get.yml
```
odyssey4me: nit: I'm not sure why this should be outside the ``DEPLOY_MAAS`` conditional below, but it doesn't hurt.

cloudnull: this was done because the rpc-gating system disables sets DEPLOY_MAAS to galse even though it deploys maas.

PR #2381, opened 2017-07-03
UG-646 Handle config within RPCO

hughsaunders:

build that is required.

Depends:
-  https://github.com/rcbops/rpc-gating/pull/266
-  rcbops/rpc-gating#296

### DIFF
--- a/scripts/bootstrap-aio.yml
+++ b/scripts/bootstrap-aio.yml
@@ -17,27 +17,142 @@
   hosts: localhost
   user: root
   tasks:
-    - name: Set ceph user_variables override
+    - name: Set user_variables_overrides fact
       set_fact:
         user_variables_overrides:
+          ## Three vars below migrated from existing boostrap-aio task.
           apply_security_hardening: "{{ rpco_deploy_hardening }}"
           # Tempest is turned off to prevent the tests from running by default
           tempest_run: no
           tempest_install: no
-      when: "{{ not rpco_deploy_ceph | bool }}"
-    - name: Set ceph user_variables override
+          # This version is proposed to openstack-ansible-os_tempest
+          # master, but won't be backported all the way, so overridden here.
+          cirros_version: 0.3.5
+          # needs to be overridden to ensure that SHAs match cirros_version
+          # escaped jinja markup so that its not templated during bootstrap.
+          tempest_images:
+            - url: "http://download.cirros-cloud.net/0.3.5/cirros-0.3.5-x86_64-disk.img"
+              sha256: "e137062a4dfbb4c225971b67781bc52183d14517170e16a3841d16f962ae7470"
+            - url: "http://download.cirros-cloud.net/0.3.5/cirros-0.3.5-x86_64-uec.tar.gz"
+              sha256: "32ca0867f4099f33537625a2fcb1bc5fea1621c62833cdc58eaef2f93e10f154"
+
+
+    # migrated from rpc-gating/playbooks/vars/all.yml
+    - name: Set gating overrides fact
       set_fact:
-        user_variables_overrides:
-          apply_security_hardening: "{{ rpco_deploy_hardening }}"
-          glance_default_store: rbd
-          nova_libvirt_images_rbd_pool: vms
-          # Tempest is turned off to prevent the tests from running by default
-          tempest_run: no
-          tempest_install: no
-      when: "{{ rpco_deploy_ceph | bool }}"
+        gating_user_variables_overrides:
+          maas_notification_plan: npTechnicalContactsEmail
+          nova_virt_type: qemu
+          maas_auth_method: "token"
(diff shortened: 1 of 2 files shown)


hughsaunders:
Had a successful test build, with modified rpc-gating branch: https://rpc.jenkins.cit.rackspace.net/blue/organizations/jenkins/RPC-AIO_master-xenial-deploy-swift-pr/detail/RPC-AIO_master-xenial-deploy-swift-pr/172/pipeline/257

hughsaunders:
recheck_cit_xenail_deploy_swift

hughsaunders:

hughsaunders:
recheck_cit_xenial_deploy_swift previous job aborted

scripts/bootstrap-aio.yml
```diff
@@ -17,27 +17,140 @@
...
+              reseller_admin_role: ResellerAdmin
+            compute-feature-enabled:
+              personality: false
+              attach_encrypted_volume: false
+            volume-feature-enabled:
+              snapshot: true
+          # This is being increased from the default of 85 as the default value may be
+          # too low for the liberty->mitaka upgrade job where more space is used by
+          # additional packages, venvs, logs, etc.
+          percent_used_critical_threshold: 95
```
mattt416: Suppose we could skip this task when `lookup ('env', 'JOB_NAME') != ""`?

scripts/bootstrap-aio.yml
```diff
@@ -17,27 +17,140 @@
...
-          # Tempest is turned off to prevent the tests from running by default
-          tempest_run: no
-          tempest_install: no
-      when: "{{ rpco_deploy_ceph | bool }}"
+        gating_user_variables_overrides:
+          maas_notification_plan: npTechnicalContactsEmail
+          nova_virt_type: qemu
+          maas_auth_method: "token"
+          maas_auth_token: "{{lookup('env', 'MAAS_AUTH_TOKEN')}}"
+          maas_api_url: "{{lookup('env', 'MAAS_API_URL')}}"
```
mattt416: As per IRC, MAAS_AUTH_TOKEN and MAAS_API_URL are set when common.prepareConfigs() runs, which won't happen here.

mkam:
Do we still need the user_zzz_gating_variables.yml related tasks still if we're combining everything into user_variables_overrides?

scripts/bootstrap-aio.yml
hughsaunders: addressed in https://github.com/rcbops/rpc-gating/pull/296

scripts/bootstrap-aio.yml
hughsaunders: Done

hughsaunders:
@mkam good point, removed those.

hughsaunders:

hughsaunders:
This shouldn't be merged as it requires a change to RPC-Gating to pass the MaaS token and URL to the bootstrap process. Without that periodics will break, PRs may work (hence the successfull checks) because they don't do any MaaS verifications that require a connected agent or access to the MaaS api.

scripts/bootstrap-aio.yml
```diff
@@ -17,27 +17,142 @@
...
         user_variables_overrides:
+          ## Three vars below migrated from existing boostrap-aio task.
           apply_security_hardening: "{{ rpco_deploy_hardening }}"
           # Tempest is turned off to prevent the tests from running by default
           tempest_run: no
           tempest_install: no
-      when: "{{ not rpco_deploy_ceph | bool }}"
-    - name: Set ceph user_variables override
+          # This version is proposed to openstack-ansible-os_tempest
+          # master, but won't be backported all the way, so overridden here.
```
odyssey4me: Why won't it? The proposal has never been made. If you make it, it might.

scripts/bootstrap-aio.yml
```diff
@@ -17,27 +17,142 @@
...
+          percent_used_critical_threshold: 95
+      when:
+        - lookup ('env', 'JOB_NAME') != ""
+    - name: Add gating user vars to user_variables_overrides
+      set_fact:
+        user_variables_overrides: "{{ user_variables_overrides|combine(gating_user_variables_overrides)}}"
+      when:
+        - lookup ('env', 'JOB_NAME') != ""
+
+    # migrated from rpc-gating/rpc_jobs/rpc_aio.yml
```
odyssey4me: There is no ceph job for master or either of the newton branches, so this doesn't really need to be done here.

scripts/bootstrap-aio.yml
hughsaunders: It won't make it back to kilo, so we have to override it here.

scripts/bootstrap-aio.yml
hughsaunders: I'll leave the ceph vars in untill we rip ceph out fully.

hughsaunders:

PR #2382, opened 2017-07-04
LA-271 Remove liberasurecode-dev requirement

evrardjp:
If the leapfrog venv is published we don't need to install
liberasurecode-dev on the deploy node, and therefore no
requirements on the repos used for the deploy node.

This bumps openstack-ansible-ops sha to include the
patch that removes the installation of liberasurecode-dev on
the deploy node (while keeping it on the venv build node)


mancdaz:

mancdaz:
without manually testing the change in the other repo, I can at least confirm that this sha bump includes the change that is referenced.

evrardjp:
Ready to go

PR #2385, opened 2017-07-07
[LA-350] Ensure ansible_host doesn't exist in vars

evrardjp:
If the new RPC-Maas is deployed on K,L,M, ansible_host will be
defined to ansible_ssh_host and will break the leapfrog.

Issue: [LA-350](https://rpc-openstack.atlassian.net/browse/LA-350)

### DIFF
--- a/scripts/leapfrog/pre_leap.sh
+++ b/scripts/leapfrog/pre_leap.sh
@@ -1,0 +1,27 @@
+#!/usr/bin/env bash
+# Copyright 2017, Rackspace US, Inc.
+#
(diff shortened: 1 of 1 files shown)


cloudnull:
recent changes here pull in the default variable file for a given release from the rpcm project [ https://github.com/cloudnull/rpc-openstack/blob/ba33df91d6f302ca206947f80760d7584597da3c/rpcd/playbooks/maas-get.yml ] While this change will work it may not be needed so I'm holing off merging this until a test run is completed.

PR #2390, opened 2017-07-11
AR-204 Remove maas-get from bootstrap-ansible

odyssey4me:
As maas-get implements some configuration in
/etc/openstack_deploy which breaks the apt
artifact building, and maas-get is effectively
run in three different places, it is removed
from the Ansible bootstrap.

Issue: [AR-204](https://rpc-openstack.atlassian.net/browse/AR-204)

### DIFF
--- a/scripts/deploy-rpc-playbooks.sh
+++ b/scripts/deploy-rpc-playbooks.sh
@@ -39,6 +39,12 @@ if [[ "${DEPLOY_ELK}" == "yes" ]]; then
 fi
 
 # Download the latest release of rpc-maas
+# TODO(odyssey4me):
+# Remove this once rpc-gating no longer tries
+# to run rpc-maas as its own thing and instead
+# just uses deploy.sh end-to-end. This line
+# should not be necessary as setup-maas.yml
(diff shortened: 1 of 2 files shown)


scripts/deploy-rpc-playbooks.sh
```diff
@@ -39,6 +39,12 @@ if [[ "${DEPLOY_ELK}" == "yes" ]]; then
 fi
 
 # Download the latest release of rpc-maas
+# TODO(odyssey4me):
+# Remove this once rpc-gating no longer tries
+# to run rpc-maas as its own thing and instead
+# just uses deploy.sh end-to-end. This line
+# should not be necessary as setup-maas.yml
+# below includes maas-get.yml
 run_ansible maas-get.yml
```
evrardjp: Could you also add that maas-get is needed even if DEPLOY_MAAS is not "yes"?
For example, it will be used for the influx db integration.

evrardjp: After discussion, my comment here could be done in another PR.

PR #2388, opened 2017-07-10
[LA-314] Bring the ability to use influxdb

evrardjp:
RPC-Maas is able to push metrics to influx, we should expose
this facility to our users, which includes our gates.

Issue: [LA-314](https://rpc-openstack.atlassian.net/browse/LA-314)

PR #2345, opened 2017-06-16
Use upstream RPC-MaaS (mitaka)

cloudnull:
This change converts our deployment of RPC MaaS from the old role method
to the new playbook method. This change will provide us access to a
faster and more flexible monitoring solution and allow all customers to
benifit from Monitoring as a Service using a unified code base.

The maas group_vars have been removed and the values of the old file
have been placed into the `user_osa_variables_defaults.yml` file. This
has been done to allow the rpc-maas playbooks to inherit from the user
variables in cases where the group vars not not available.

Signed-off-by: Kevin Carter <kevin.carter@rackspace.com>
(cherry picked from commit dbd0253e619710b926230e573c9f273559102bf1)

PR #2395, opened 2017-07-12
Bring the ability to use influxdb (#2388)

evrardjp:
RPC-Maas is able to push metrics to influx, we should expose
this facility to our users, which includes our gates.

On top of it, we should bring the facility to deploy influx, if
need be.

RPC-Maas is able to push metrics to influx, we should expose
this facility to our users, which includes our gates.

Issue: [LA-314](https://rpc-openstack.atlassian.net/browse/LA-314)
Manual cherry-pick from eae30c4f74ad471cccfd6f17217dc6eeb0aa347b

Issue: [mitaka-13](https://rpc-openstack.atlassian.net/browse/mitaka-13)


evrardjp:
This needs new maas to be merged into this branch.

evrardjp:
``+ openstack-ansible --forks 4 maas-get.yml
Variable files: "-e @/etc/openstack_deploy/user_conf_files.yml -e @/etc/openstack_deploy/user_osa_aio_variables.yml -e @/etc/openstack_deploy/user_osa_secrets.yml -e @/etc/openstack_deploy/user_osa_variables_defaults.yml -e @/etc/openstack_deploy/user_osa_variables_overrides.yml -e @/etc/openstack_deploy/user_rpco_secrets.yml -e @/etc/openstack_deploy/user_rpco_variables_defaults.yml -e @/etc/openstack_deploy/user_rpco_variables_overrides.yml -e @/etc/openstack_deploy/user_zzz_gating_variables.yml "
ERROR: the playbook: maas-get.yml could not be found``

I will fix that.

evrardjp:
depends on https://github.com/rcbops/rpc-openstack/pull/2345 that will not get fixed until a team decides to bypass the majorupgrade jobs, or, alternatively, fix them.

cloudnull:
the major upgrade jobs are no longer required because the code path at this time is unmaintained and a blocker for progress. Please review #2345 when at all possible so we can get this done.

cloudnull:
Updated to resolve merge conflicts

scripts/functions.sh
```diff
@@ -1,4 +1,52 @@
...
+
+## Vars ----------------------------------------------------------------------
+
+# Gating
+export BUILD_TAG=${BUILD_TAG:-}
+export INFLUX_IP=${INFLUX_IP:-}
+export INFLUX_PORT=${INFLUX_IP:-"8086"}
+
+# Other
+export ADMIN_PASSWORD=${ADMIN_PASSWORD:-"secrete"}
```
hughsaunders: I don't like this, should be user secrets? Also seems to only be used for kibana so should labeled as such. However I realise this is a backport, so will let it go.

evrardjp:
@cloudnull could you vote/merge?

scripts/functions.sh
odyssey4me: Yeah, this is a bad habit holdover from bygone days.

PR #2397, opened 2017-07-12
Bring the ability to use influxdb (#2388)

evrardjp:
RPC-Maas is able to push metrics to influx, we should expose
this facility to our users, which includes our gates.

On top of it, we should bring the facility to deploy influx, if
need be.

RPC-Maas is able to push metrics to influx, we should expose
this facility to our users, which includes our gates.

Note: To include this in a more standardized way, I had to
create scripts/(functions|deploy-rpc-playbooks).sh, and use these
from deploy.sh. Further work can be done to simplify deploy.sh
script to look more like the one on upper branches.

Issue: [LA-314](https://rpc-openstack.atlassian.net/browse/LA-314)
Manual cherry pick from 45046ee09887b4eb3e91e10de9707c9634011b23

### DIFF
--- a/scripts/deploy-rpc-playbooks.sh
+++ b/scripts/deploy-rpc-playbooks.sh
@@ -1,0 +1,88 @@
+#!/usr/bin/env bash
+#
+# Copyright 2014-2017, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


evrardjp:
run_ansible is undefined.

evrardjp:
This also need new maas to be backported first.

evrardjp:
Maas get not found.

PR #2400, opened 2017-07-12
LA-308 Leapfrog Documentation

hughsaunders:
Document the Leapfrog Upgrade Process for users and future developers.

Issue: [LA-308](https://rpc-openstack.atlassian.net/browse/LA-308)


scripts/leapfrog/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+#### Upgrade
+This step has a section for each major version between the source (Kilo) and
+target (Newton) versions. Each section includes:
+* Variable & Secrets Migration (OSA)
+* Fact Cleanup
+* Hostname Compatibility checks/modifications.
+
+This step also includes host upgrades.
+
```
mancdaz: Is this where DB migrations happen? I don't see it mentioned anywhere

Also, what does 'This step includes host upgrades' mean?

hughsaunders: Good point the migrations heading was missed. Will add that and clarify host updates.

hughsaunders: Decided to remove the host upgrade line, as its not significant in that context or easy to explain briefly.

scripts/leapfrog/README.md
```diff
@@ -0,0 +1,136 @@
...
+It is crucial that the rpco repo and openstack-ansible submodule are both clean
+and at the appropriate SHAs for newton-14.1 at this stage. The [github
+newton-14.1 branch
+page](https://github.com/rcbops/rpc-openstack/tree/newton-14.1) shows both SHAs,
+the RPCO SHA is the latest commit, and the OSA SHA is shown as an annotation
+after @ on the openstack-ansible submodule.
+
+The next step is to execute the leapfrog upgrade script
+
```
evrardjp: I think it would be good to say that the database should be cleaned up of unnecessary things. (We do have something for it, right?)

scripts/leapfrog/README.md
```diff
@@ -0,0 +1,136 @@
...
+This step executes pre-flight checks, and prompts the user for confirmation. It
+also ensures that the databases are backed up.
+
+#### Upgrade
+This step has a section for each major version between the source (Kilo) and
+target (Newton) versions. Each section includes:
+* Variable & Secrets Migration (OSA)
+* Fact Cleanup
+* Hostname Compatibility checks/modifications.
+
```
evrardjp: * Inventory upgrades

scripts/leapfrog/README.md
```diff
@@ -0,0 +1,136 @@
...
+#### Migrations
+The step runs the database migrations for each major upgrade in turn.
+For a RPC Kilo --> Newton upgrade this involves running the liberty,
+mitaka and newton migrations in sequence.
+
+#### Re-Deploy
+This step first runs the pre_re deploy script which handles RPC variable and
+secret migrations. Next the target version of OSA is deployed.
+
```
evrardjp: Maybe we should say in this section that the redeploy voluntarily leaves a neutron agent container running, and rebalances it to the Newton neutron agent containers when ready.

scripts/leapfrog/README.md
```diff
@@ -0,0 +1,136 @@
...
+operations
+guide](https://docs.openstack.org/openstack-ansible/latest/admin/maintenance-tasks/galera.html).
+
+#### Confirmation Prompts and the Ctrl-c warning
+The confirmation prompts instruct the user not to interrupt the leapfrog process
+via ctrl-c. While an uninterrupted upgrade is the smoothest, the consequences
+of interruption are not as dire as implied. The process can be resumed by
+re-running the top level script, which will skip the steps that have already
+been completed by checking for the existence of marker files.
+
```
evrardjp: Keep in mind, when resuming, that the detection system could say that you are running Newton, even if leapfrog the process didn't fully run.
That's normal, because some of the files used for detection could have been migrated (depending on when the Ctrl-C happened), while the rest haven't been migrated.

Please continue the process as if you were already on newton, to ensure everything ran, if something happened after the upgrade steps.

scripts/leapfrog/README.md
```diff
@@ -0,0 +1,136 @@
...
+* [RPCO](https://github.com/rcbops/rpc-openstack): Rackspace Private Cloud powered by OpenStack
+* [OSA](https://github.com/openstack/openstack-ansible):  OpenStack Ansible
+* [OSA-OPS](https://github.com/openstack/openstack-ansible-ops):  OpenStack Operations
+* [Kilo](https://github.com/rcbops/rpc-openstack/tree/kilo): The RPCO release of OpenStack Kilo
+* [Newton-14.1](https://github.com/rcbops/rpc-openstack/tree/newton-14.1) The second RPCO release of OpenStack Newton. This version includes Leapfrog upgrade tools.
+
+## Executing a leapfrog upgrade
+
+The start point is a kilo deployment. It is useful to check the kilo deployment
```
odyssey4me: 'useful'? I would think it's essential.

scripts/leapfrog/README.md
```diff
@@ -0,0 +1,136 @@
...
+is healthy before starting the leapfrog process, for example by running tempest.
+
+```
+root@kilo:/opt/rpc-openstack/rpcd/playbooks# openstack-ansible ../../scripts/run_tempest.yml -t tempest_execute_tests -vv
```
odyssey4me: I'm not sure we should be prescribing this as a health check. I think it's best to remove it. The support team have a much broader and more complete health check.

scripts/leapfrog/README.md
odyssey4me: We do not need to go into the details of database maintenance. All we need to do is say that it should be done. We should definitely add it as a reminder above where the health check reminder is.

scripts/leapfrog/README.md
evrardjp: Agreed

scripts/leapfrog/README.md
evrardjp: Ok with that too, but we should insist on the fact that a health check should be done.

scripts/leapfrog/README.md
```diff
@@ -0,0 +1,136 @@
...
+<img align="right" src="docs/images/leapfrog_structure_diagram.png"/>
+The RPCO leapfrog scripts are a thin wrapper around OSA-OPS leapfrog tools. 
+
+For details please refer to the scripts themselves. Paths are omitted for
+brevity, scripts may not be in the root of the relevant repo.
+
+
+#### Pre Leap
+This step removes modifications to RPCO Kilo that aren't compatible with OSA
+Newton. Currently this only contains an Ansible 1.9 compatibility workaround.
```
odyssey4me: OSA Newton? Do you mean RPCO Newton?

scripts/leapfrog/README.md
```diff
@@ -0,0 +1,136 @@
...
+
+#### Prep
+This step executes pre-flight checks, and prompts the user for confirmation. It
+also ensures that the databases are backed up.
```
odyssey4me: Where are the database backed up to?

scripts/leapfrog/README.md
mancdaz: with some additional detail on why it does that (to keep data plane access as long as possible during the upgrade).
Maybe each section at a high level should have an indicator of which bits are 'still up' and which bits are 'hard down'. Not how long, because we don't have the data yet, just which components are impacted during which phase.

scripts/leapfrog/README.md
```diff
@@ -0,0 +1,136 @@
...
+* Fact Cleanup
+* Hostname Compatibility checks/modifications.
+
+#### Migrations
+The step runs the database migrations for each major upgrade in turn.
+For a RPC Kilo --> Newton upgrade this involves running the liberty,
+mitaka and newton migrations in sequence.
+
+#### Re-Deploy
+This step first runs the pre_re deploy script which handles RPC variable and
```
odyssey4me: pre_redeploy ?

scripts/leapfrog/README.md
```diff
@@ -0,0 +1,136 @@
...
+secret migrations. Next the target version of OSA is deployed.
+
+#### Post Leap
+This step deploys the RPC additions on top of the already deployed OSA.
```
odyssey4me: ... already deployed OpenStack services.

scripts/leapfrog/README.md
hughsaunders: Done

scripts/leapfrog/README.md
hughsaunders: Removed

scripts/leapfrog/README.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Rollback and Interruption
+The Leapfrog upgrade is a one way process, once it has begun there is no
+automated rollback. If a leapfrog upgrade is interrupted, it can be resumed.
```
odyssey4me: there is no way to roll back, automated or otherwise. While it may be possible to roll back the configuration and databases and the control plane, it is not possible to roll back the package changes on the hosts.

odyssey4me: Add the above (or something like it) to explain why there is no roll back, then make a new paragraph for the interruption description.

scripts/leapfrog/README.md
```diff
@@ -0,0 +1,136 @@
...
+will be used to skip completed tasks on subsequent runs.
+
+If a step fails information about that step will be printed along with all the
+remaining steps. The operator must fix the failure before re-running the
+leapfrog or completing the remaining steps manually.
+
+## Confirmation Prompts
+The leapfrog process does not run unattended, it requires the operator to
+confirm at various points. This is to ensure that an irreversible leapfrog is
+not initiated accidentally.
```
odyssey4me: Are these all at the start, or do they happen at specific stages. It'd be good to set the expectation here otherwise the reader thinks that they literally have to watch the screen constantly.

scripts/leapfrog/README.md
hughsaunders: Done

scripts/leapfrog/README.md
hughsaunders: Done

scripts/leapfrog/README.md
```diff
@@ -0,0 +1,136 @@
...
+
+
+## Problems
+
+#### Repo Build
+The redeploy phase involves many git clones from github, this may cause rate
+limits, which in turn can cause timeouts and the repo build process to fail. If
+this happens, wait for rate limits to reset before retrying.
```
odyssey4me: Uh, what? We should not be doing a repo build at all. We should be using the artifact staging which is all http downloads only.

scripts/leapfrog/README.md
hughsaunders: Done

scripts/leapfrog/README.md
```diff
@@ -0,0 +1,136 @@
...
+
+#### Galera
+Occasionally the galera cluster may be in a non-functional state after the
+leapfrog. If this is the case, follow the [Galera Maintenance section of the OSA
+operations
+guide](https://docs.openstack.org/openstack-ansible/latest/admin/maintenance-tasks/galera.html).
```
odyssey4me: Do we mean to point at latest here? That's Pike-targeted information. Perhaps we should rather point at the newton-targeted information?

scripts/leapfrog/README.md
evrardjp: At the start, and before re-deploying.

scripts/leapfrog/README.md
hughsaunders: Done.

scripts/leapfrog/README.md
hughsaunders: Done

scripts/leapfrog/README.md
hughsaunders: done

scripts/leapfrog/README.md
hughsaunders: Done

scripts/leapfrog/README.md
hughsaunders: Done.

scripts/leapfrog/README.md
hughsaunders: As discussed earlier, using mirror.rackspace can lead to failures downloading artefacts and a fallback to building packages.

scripts/leapfrog/README.md
hughsaunders: Done

scripts/leapfrog/README.md
evrardjp: Agreed. https://rpc-openstack.atlassian.net/browse/LA-361

scripts/leapfrog/README.md
evrardjp: Sounds good.

hughsaunders: Done.

scripts/leapfrog/README.md
```diff
@@ -0,0 +1,152 @@
...
+* [RPCO](https://github.com/rcbops/rpc-openstack): Rackspace Private Cloud powered by OpenStack
+* [OSA](https://github.com/openstack/openstack-ansible):  OpenStack Ansible
+* [OSA-OPS](https://github.com/openstack/openstack-ansible-ops):  OpenStack Operations
+* [Kilo](https://github.com/rcbops/rpc-openstack/tree/kilo): The RPCO release of OpenStack Kilo
+* [Newton-14.1](https://github.com/rcbops/rpc-openstack/tree/newton-14.1) The second RPCO release of OpenStack Newton. This version includes Leapfrog upgrade tools.
+
+## Executing a leapfrog upgrade
+
+The start point is a kilo deployment. It is essential to check the kilo deployment
+is healthy before starting the leapfrog process. Many tools are availble for this,
```
hughsaunders: Done

scripts/leapfrog/README.md
```diff
@@ -0,0 +1,152 @@
...
+```
+
+It is crucial that the rpco repo and openstack-ansible submodule are both clean
+and at the appropriate SHAs for newton-14.1 at this stage. The [github
+newton-14.1 branch
+page](https://github.com/rcbops/rpc-openstack/tree/newton-14.1) shows both SHAs,
+the RPCO SHA is the latest commit, and the OSA SHA is shown as an annotation
+after @ on the openstack-ansible submodule.
+
+Database housekeeping should be performed at this stage.
```
odyssey4me: I would suggest that this be done before the checkout. Perhaps there should be a heading called something like 'pre-leapfrog tasks' and the health check and housekeeping should be listed there.

hughsaunders: Done

scripts/leapfrog/README.md
```diff
@@ -0,0 +1,152 @@
...
+
+
+#### Pre Leap
+This step removes modifications to RPCO Kilo that aren't compatible with RPCO
+Newton. Currently this only contains an Ansible 1.9 compatibility workaround.
+
+#### Prep
+This step executes pre-flight checks, and prompts the user for confirmation. It
+also ensures that the databases are backed up. Backups are stored in /var/backup
+on the first galera node.
```
odyssey4me: node is ambiguous - is that the host or the container? I'm guessing that's the location in the container, which has a bind mount to the host. I think it may be better to refer to the host's location and specifically state that it's available there.

hughsaunders: Done

scripts/leapfrog/README.md
odyssey4me: Ah, so that's not the repo build process - that's the leapfrog venv build process. ie the venvs used to facilitate the db migrations. The two are distinctly different.

scripts/leapfrog/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+#### Terms
+
+* [RPCO](https://github.com/rcbops/rpc-openstack): Rackspace Private Cloud powered by OpenStack
+* [OSA](https://github.com/openstack/openstack-ansible):  OpenStack Ansible
+* [OSA-OPS](https://github.com/openstack/openstack-ansible-ops):  OpenStack Operations
+* [Kilo](https://github.com/rcbops/rpc-openstack/tree/kilo): The RPCO release of OpenStack Kilo
+* [Newton-14.1](https://github.com/rcbops/rpc-openstack/tree/newton-14.1) The second RPCO release of OpenStack Newton. This version includes Leapfrog upgrade tools.
+
+## Preparing for a lepfrog upgrade
```
odyssey4me: lepfrog (spelling error)

scripts/leapfrog/README.md
```diff
@@ -0,0 +1,155 @@
...
+The start point is a kilo deployment. It is essential to check the kilo
+deployment is healthy before starting the leapfrog process. Many tools are
+available for this, for example tempest.
```
odyssey4me: The sentencing here is a little clumsy and doesn't read well. I'd suggest keeping it very simple and stating:

* The state of the source environment must be healthy.
* Database  housekeeping should be done to prevent unnecessary database migrations.

scripts/leapfrog/README.md
hughsaunders: done

scripts/leapfrog/README.md
hughsaunders: Done. I can not haz engrish

scripts/leapfrog/README.md
hughsaunders: Changed heading from repo build to Clone failures.

weezer:
LGTM

scripts/leapfrog/README.md
```diff
@@ -0,0 +1,154 @@
+# Leapfrog Documentation
+
+#### Overview
+A Leapfrog upgrade is a major upgrade that skips at least one release. Currently
+RPCO supports leapfrog upgrades from kilo to newton-14.1.
+
+#### Terms
+
```
mancdaz: at least in the preview, this heading isn't rendering properly

scripts/leapfrog/README.md
```diff
@@ -0,0 +1,154 @@
...
+
+* [RPCO](https://github.com/rcbops/rpc-openstack): Rackspace Private Cloud powered by OpenStack
+* [OSA](https://github.com/openstack/openstack-ansible):  OpenStack Ansible
+* [OSA-OPS](https://github.com/openstack/openstack-ansible-ops):  OpenStack Operations
+* [Kilo](https://github.com/rcbops/rpc-openstack/tree/kilo): The RPCO release of OpenStack Kilo
+* [Newton-14.1](https://github.com/rcbops/rpc-openstack/tree/newton-14.1) The second RPCO release of OpenStack Newton. This version includes Leapfrog upgrade tools.
+
+## Pre Leapfrog Tasks
+
```
mancdaz: at least in the preview, this heading isn't rendering properly

scripts/leapfrog/README.md
hughsaunders: I've noticed that, and I'm not sure whats wrong with it :( its the same as the other headings. It also renders fine in my local preview. I think the  best thing to do is merge, see if it renders fine in the merged version. If not, we'll fix it.

PR #2401, opened 2017-07-12
Restart Neutron Agents

evrardjp:
Neutron agents seem to behave correctly after leapfrog, but
they are not. This implements a workaround, which restarts
the neutron-agent containers post the OSA leapfrog, in the
meantime we find and resolve the root cause of the issue.


evrardjp:
A card has been created to solve this issue on long term: https://rpc-openstack.atlassian.net/browse/LA-358

mattt416:

mattt416:

evrardjp:
We got the two reviews, let's go forward!

PR #2409, opened 2017-07-14
[Backport - newton-14.1] Move GROUP_VARS and HOST_VARS_PATH

alextricity25:

In addition, some logic has been added so that the variables are set to their
original value before running the OSA bootstrap playbook. This is so there
are no extra settings when the bootstrap playbook runs.

(cherry picked from commit ce2b72aabc1ab95ebbc9be499973200b633c0841)


hughsaunders:
closing in favour of https://github.com/rcbops/rpc-openstack/pull/2412 as the branch name is causing problems.

PR #2332, opened 2017-06-12
[RT-15] Break out of remaining ELK roles

d34dh0r53:
As part of the effort to move ELK upstream this commit removes ELK
roles in the RPC-O tree in favor of the new rcbops repos.  This commit
also upgrades ELK versions to the 5.x release.

Issue: [RT-15](https://rpc-openstack.atlassian.net/browse/RT-15)

Issue: [RE-173](https://rpc-openstack.atlassian.net/browse/RE-173)

Issue: [K86-116](https://rpc-openstack.atlassian.net/browse/K86-116)

Issue: [RE-185](https://rpc-openstack.atlassian.net/browse/RE-185)

Issue: [RE-187](https://rpc-openstack.atlassian.net/browse/RE-187)

PR #2410, opened 2017-07-14
RT-15 Update ELK versions

d34dh0r53:
This commit updates the artifacts building ELK versions to the latest require for the ELK role move out in #2332.

Issue: [RT-15](https://rpc-openstack.atlassian.net/browse/RT-15)

### DIFF
--- a/scripts/artifacts-building/apt/aptly-vars.yml
+++ b/scripts/artifacts-building/apt/aptly-vars.yml
@@ -399,7 +399,7 @@ aptly_mirrors:
     update_flags:
       - "-keyring='trustedkeys.gpg'"
     do_update: "{{ aptly_mirror_do_updates }}"
-- name: elastic-logstash-5.5-ALL
+  - name: elastic-logstash-5.5-ALL
     archive_url: http://artifacts.elastic.co/packages/5.x/apt
     distribution: stable
     component1:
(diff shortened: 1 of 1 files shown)


d34dh0r53:
recheck_cit_artifact_xenial_apt

scripts/artifacts-building/apt/aptly-vars.yml
```diff
@@ -400,6 +399,66 @@ aptly_mirrors:
     update_flags:
       - "-keyring='trustedkeys.gpg'"
     do_update: "{{ aptly_mirror_do_updates }}"
+- name: elastic-logstash-5.5-ALL
```
alextricity25: I'm not sure if it matters, but the indentation here doesn't seem to be consistent with the above entries. `-name` should be indented two more spaces like it is above. i.e.:
```
    - name: <name>
      archive_url: <url>
      ..
```

d34dh0r53: Good catch, I'll fix that.

d34dh0r53:
recheck_cit_artifact_xenial_apt

d34dh0r53:

mancdaz:
Hi @d34dh0r53 you probably want to squash these commits at the same time as adding the issue number to the commit message/title

odyssey4me:
I'll be doing an artifact rebuild soon as the SHA for master will need bumping after the setuptools issue is resolved.

odyssey4me:
The SHA bump proposal is here: https://github.com/rcbops/rpc-openstack/pull/2425

PR #2411, opened 2017-07-16
[LA-365] Force yaprt to build setuptools 24

evrardjp:
There is no constraint at all for setuptools in the rpc builds.
So we can install any version of these software, and yaprt
is building latest wheels.

We need to not use latest wheels for setuptools, therefore
I list it as a pip_package. It will therefore be listed in the
deps, and will be used in the yaprt process. No new setuptools
will be used, as the dependency will be met.

On top of it, it was free for all for elasticsearch, so I took
the liberty (pun intended) to strengthen the elasticsearch and
elastisearch-curator constraints, with a value coming from
Liberty-12.2 branch.

Issue: [LA-365](https://rpc-openstack.atlassian.net/browse/LA-365)

### DIFF
--- a/rpcd/playbooks/roles/beaver/defaults/main.yml
+++ b/rpcd/playbooks/roles/beaver/defaults/main.yml
@@ -14,8 +14,20 @@
 # limitations under the License.
 
 # the pip packages to install
+# Only beaver is required, the rest are there to constrain the version of deps.
 beaver_pip_packages:
   - "beaver==33.3.0"
+  - "boto==2.48.0"
+  - "conf-d==0.0.4"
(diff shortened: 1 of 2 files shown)


evrardjp:

evrardjp:
The job for trusty leapfrog upgrade swift from a Kilo PR apparently never worked with Kibana selenium testing. The reference used was not existing.
I've modified the job temporarily, hoping the PR would work with my changes without affecting coverage. I will revert tomorrow and make a proper fix in tree.

evrardjp:

weezer:
LGTM

PR #2412, opened 2017-07-17
Move GROUP_VARS and HOST_VARS_PATH

hughsaunders:

In addition, some logic has been added so that the variables are set to their
original value before running the OSA bootstrap playbook. This is so there
are no extra settings when the bootstrap playbook runs.

(cherry picked from commit ce2b72aabc1ab95ebbc9be499973200b633c0841)

### DIFF
--- a/scripts/artifacts-building/containers/build-process.sh
+++ b/scripts/artifacts-building/containers/build-process.sh
@@ -60,10 +60,6 @@ source scripts/bootstrap-ansible.sh
 # that no containers are created in the inventory.
 rm -f /etc/openstack_deploy/conf.d/*
 
-# Now use GROUP_VARS of OSA and RPC
-sed -i "s|GROUP_VARS_PATH=.*|GROUP_VARS_PATH=\"\${GROUP_VARS_PATH:-${BASE_DIR}/openstack-ansible/playbooks/inventory/group_vars/:${BASE_DIR}/group_vars/:/etc/openstack_deploy/group_vars/}\"|" /usr/local/bin/openstack-ansible.rc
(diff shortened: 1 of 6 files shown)


alextricity25:
recheck_cit_all

alextricity25:
recheck_cit_all

scripts/leapfrog/pre_redeploy.sh
```diff
@@ -102,8 +102,6 @@ pushd ${LEAPFROG_DIR}
     if [[ ! -f "${UPGRADE_LEAP_MARKER_FOLDER}/rebootstrap-ansible-for-rpc.complete" ]]; then
         pushd ${RPCO_DEFAULT_FOLDER}
             ansible-galaxy install --force -r ansible-role-requirements.yml
-            sed -i "s|GROUP_VARS_PATH=.*|GROUP_VARS_PATH=\"\${GROUP_VARS_PATH:-${RPCO_DEFAULT_FOLDER}/openstack-ansible/playbooks/inventory/group_vars/:${RPCO_DEFAULT_FOLDER}/group_vars/:/etc/openstack_deploy/group_vars/}\"|" /usr/local/bin/openstack-ansible.rc
```
evrardjp: We are not rebootstrapping ansible, so this will break the leapfrog.

alextricity25: Gotcha. I'll keep the lines in here then.

evrardjp:
Looks fine when it will pass gates. Need further investigation on my side.

alextricity25:

odyssey4me:

PR #2422, opened 2017-07-19
RI-38 - Create ansible role to configure and deploy ironic

prometheanfire:
Ironic requires some user action and values before it can finish
being deployed.  This new role will ask for these values then write
them to the correct overrides file.

### DIFF
--- a/rpcd/playbooks/roles/configure_ironic/CONTRIBUTING.md
+++ b/rpcd/playbooks/roles/configure_ironic/CONTRIBUTING.md
@@ -1,0 +1,1 @@
+../../../../CONTRIBUTING.md
--- a/rpcd/playbooks/roles/configure_ironic/README.rst
+++ b/rpcd/playbooks/roles/configure_ironic/README.rst
@@ -1,0 +1,20 @@
+Configure Ironic
+========
+
+Configure deployment specific Ironic settings.
+
+Example Playbook
+----------------
+
+Including an example of how to use your role (for instance, with variables passed in as parameters) is always nice for users too:
+
+.. code-block:: yaml
+
+    - hosts: servers
+      roles:
+         - { role: "filebeat" }
+
+License
+-------
+
+Apache 2.0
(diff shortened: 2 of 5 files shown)


prometheanfire:
The pause prompt doesn't output what you type, known bug.

https://github.com/ansible/ansible/issues/14160

prometheanfire:
Well, that got shot down quick, I was told to do it this way by @cloudnull 

As far as making this a role goes, this is just the first of the values that would be there.

odyssey4me:
@prometheanfire I personally feel that if you feel inclined to edit user config, then that means that the product should be adjusted to be smarter. We should never edit configs for our customer, but we can make the product behave smarter through other means - whether that means doing stuff upstream, then consuming it,

odyssey4me:
Oops - sorry, wrong button.

prometheanfire:
Ya, I mostly agree with you, but this role is for the deployer / end user to run.  It'd be mentioned in the docs (that don't exist yet).

This particular override is needed (and can't be set by us) because it relies on the deployer / end user to create the network in neutron first.

prometheanfire:
@odyssey4me also, I don't think this is automation, at least in the since that it's not fully automatic, it does require user input.

    TASK [configure_ironic : Get cleaning_network_uuid] ****************************
    [configure_ironic : Get cleaning_network_uuid]
    What is the ironic cleaning network uuid from neutron?:
    ok: [localhost]

the input isn't displayed as you type though (mentioned earlier in this discussion)

odyssey4me:
@prometheanfire This should be a documented procedure. That is all.

1. Run this command to create your cleaning network.
2. Add the returned value to your user variables.

If they're administering an Ironic environment they will have to be doing a lot of these sorts of things - asking them to run a playbook so that they can enter the values of things is just adding more tech debt for us to maintain with very, very little in return value.

And again - if you do anything that's a simple task for convenience, a role is just over the top. The reason I reacted so quickly is because I saw a new PR adding another role into the RPC-O codebase. We are trying to remove roles, not add them.

evrardjp:
This is a doc change, no need for a playbook, not even a role!

However, if you want to integrate this for gates, it should probably need just one tasks inside deploy.sh, but wrapped around our checks to see if you are in gating..

prometheanfire:
Gating for ironic will be based on how rpc-maas does it I hope (as in I'm targeting the simplicity of it).

Beyond that, we talked about this 'in person' and decided to convert this fully to docs instead of any type of automation.

PR #2428, opened 2017-07-24
UG-646 Handle config within RPCO [Backport: newton-14.0]

hughsaunders:

build that is required.

This duplicates gating functionality so must not be merged until it's
counterpart in rpc-gating is merged:
  https://github.com/rcbops/rpc-gating/pull/266

(cherry picked from commit 4435dda995de74894017ba48eeced19a3ebe4715)
(cherry picked from commit 72de4174739131c07aea98dedaae320e0b091023)

### DIFF
--- a/ansible-role-requirements.yml
+++ b/ansible-role-requirements.yml
@@ -14,3 +14,10 @@
   scm: git
   src: https://github.com/rsoprivatecloud/openstack-ops
   version: 1383d0db302c6fc400a50fae4f85e77bf1633581
+  # TODO(odyssey4me):
+  # Remove this override once https://review.openstack.org/486754/ merges
+  # and is included in the RPC-O SHA bump.
+- name: plugins
+  scm: git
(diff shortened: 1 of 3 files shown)


hughsaunders:
recheck cit_trusty_deploy_swift failed on test server basic ops

hughsaunders:
recheck_cit_trusty_deploy_swift failed on test server basic ops

ansible-role-requirements.yml
```diff
@@ -14,3 +14,7 @@
   scm: git
   src: https://github.com/rsoprivatecloud/openstack-ops
   version: 1383d0db302c6fc400a50fae4f85e77bf1633581
```
mattt416: Can we add a note about this needing to come out, as exists here:

https://github.com/rcbops/rpc-openstack/blob/newton-14.1/ansible-role-requirements.yml#L21-L23

hughsaunders: Done.

PR #2430, opened 2017-07-25
RE-97 Add support to configure multi-node AIO

mkam:
Adds a configure_multi_node_aio script and playbook
that runs if CONFIGURE_MNAIO is set to "yes". This
prepares the configuration that was previously done in
rpc-gating by reusing bootstrap-aio.

Issue: [RE-97](https://rpc-openstack.atlassian.net/browse/RE-97)

### DIFF
--- a/scripts/bootstrap-aio.yml
+++ b/scripts/bootstrap-aio.yml
@@ -151,12 +151,15 @@
     rpco_deploy_hardening: "{{ lookup('env', 'DEPLOY_HARDENING') }}"
     TRIGGER: "{{ lookup('env', 'TRIGGER')}}"
     SCENARIO: "{{ lookup('env', 'SCENARIO')}}"
-    TARGET: "{{ lookup('env', 'TARGET')}}"
+    TARGET: "{{ lookup('env', 'TARGET') | default('aio', true) }}"
 
 
 - name: Execute the OSA AIO bootstrap
   include: "{{ lookup('env', 'OA_DIR') }}/tests/bootstrap-aio.yml"
+  when:
+    - TARGET == "aio"
   vars:
+    TARGET: "{{ lookup('env', 'TARGET') | default('aio', true) }}"
     bootstrap_host_apt_distribution_suffix_list: "{{ (lookup('env', 'RPCO_APT_ARTIFACTS_AVAILABLE') | bool) | ternary([], ['updates', 'backports']) }}"
     bootstrap_host_scenario: "{{ (lookup('env', 'DEPLOY_CEPH') | bool) | ternary('ceph','swift') }}"
     bootstrap_user_variables_template: "user_variables.aio.yml.j2"
--- a/scripts/configure-multi-node-aio.sh
+++ b/scripts/configure-multi-node-aio.sh
@@ -1,0 +1,48 @@
+#!/usr/bin/env bash
+#
+# Copyright 2017, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+## Shell Opts ----------------------------------------------------------------
+set -e -u -x
+
+## Functions -----------------------------------------------------------------
+
+export BASE_DIR=${BASE_DIR:-"/opt/rpc-openstack"}
+source ${BASE_DIR}/scripts/functions.sh
+
+## Main ----------------------------------------------------------------------
+
+# Check the openstack-ansible submodule status
+check_submodule_status
+
+# Run multi-node AIO config setup playbook
+export TARGET=${TARGET:-"mnaio"}
+openstack-ansible -vvv ${BASE_DIR}/scripts/configure-multi-node-aio.yml \
+                  -i "localhost," -c local
+
+if ! apt_artifacts_available; then
+  # Remove the AIO configuration relating to the use
+  # of apt artifacts. This needs to be done because
+  # the apt artifacts do not exist yet.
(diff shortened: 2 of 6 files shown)


scripts/bootstrap-aio.yml
```diff
@@ -142,21 +142,27 @@
...
-    TARGET: "{{ lookup('env', 'TARGET')}}"
+    target: "{{ TARGET | default(lookup('env', 'TARGET')) }}"
 
 
 - name: Execute the OSA AIO bootstrap
-  include: "{{ lookup('env', 'OA_DIR') }}/tests/bootstrap-aio.yml"
+  hosts: localhost
+  user: root
+  tasks:
+   - include: "{{ lookup('env', 'OA_DIR') }}/tests/bootstrap-aio.yml"
```
hughsaunders: I don't think this will work as bootstrap-aio.yml is a playbook rather than a tasks file.

scripts/bootstrap-aio.yml
```diff
@@ -142,21 +142,27 @@
 
     - name: Add target user vars to user_variables_overrides
       set_fact:
-        user_variables_overrides: "{{ user_variables_overrides|combine(target_user_variables_overrides[TARGET])}}"
+        user_variables_overrides: "{{ user_variables_overrides|combine(target_user_variables_overrides[target])}}"
       when:
-        - TARGET is defined
-        - TARGET in target_user_variables_overrides
+        - target is defined
```
hughsaunders: Why target and TARGET? (L154 includes both). If you do decide to go with lower case, SCENARIO and TRIGGER should also be lowercased for consitency.

scripts/bootstrap-aio.yml
mkam: Hrm, okay, I'll need to find another way to run it only when the target is aio then. I did it this way so that I could use the conditionals.

scripts/bootstrap-aio.yml
mkam: So this was a whole thing where I was trying to get the playbook to take TARGET as either a variable you pass with vars or through the environment. Due to uncreative naming choices, I made the new value used target (since if I had `TARGET: "{{ TARGET | default(...`, that throws recursive errors). I'll change it to be more consistent and have clearer names.

scripts/bootstrap-aio.yml
mkam: Ah, so I actually can use `when` on playbook includes because that was added in ansible 2.0. It'll be trickier for the branches with older versions of ansible, but I can address that in the future, I suppose.

mkam:
Artifact jobs failed because TARGET isn't set in those jobs, so the AIO isn't being bootstrapped anymore. I'm thinking I should change this to only skip when the target is mnaio (instead of not aio) since a bootstrap-aio.yml playbook that doesn't bootstrap by default seems counterintuitive. 

@hughsaunders Thoughts?

scripts/bootstrap-aio.yml
```diff
@@ -151,12 +151,15 @@
     rpco_deploy_hardening: "{{ lookup('env', 'DEPLOY_HARDENING') }}"
     TRIGGER: "{{ lookup('env', 'TRIGGER')}}"
     SCENARIO: "{{ lookup('env', 'SCENARIO')}}"
-    TARGET: "{{ lookup('env', 'TARGET')}}"
+    TARGET: "{{ target_override | default(lookup('env', 'TARGET')) }}"
```
hughsaunders: `lookup('env', 'TARGET')|default('aio')` then `set env.TARGET=mnaio` in multi_node_aio.yml, similar to https://github.com/rcbops/rpc-gating/blob/master/rpc_jobs/rpc_aio.yml#L290 ? Then in jobs that don't set target, aio will be used by default (eg the artifact jobs).

hughsaunders:
@mkam I agree, the aio playbook should bootstrap by default, and not require the `TARGET` variable. You mentioned problems with older versions of ansible, it doesn't turn out to be so much of a problem as the older rpco branches don't contain bootstrap-aio.yml. I'm currently attempting a shell version for mitaka: https://github.com/rcbops/rpc-openstack/pull/2433/files

scripts/bootstrap-aio.yml
mkam: I have it in the job, but I also added exporting TARGET in the configure-multi-node-aio.sh so that the script can easily be run outside of the gate.

scripts/configure-multi-node-aio.sh
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env bash
```
odyssey4me: I'm not sure that I understand the value of having this as a different file. The difference between this and https://github.com/rcbops/rpc-openstack/blob/master/scripts/bootstrap-aio.sh is that script partitions the largest secondary disk as far as I can see. Other than that, not much else.

scripts/configure-multi-node-aio.yml
```diff
@@ -0,0 +1,34 @@
+---
```
odyssey4me: There's a lot being done in https://github.com/rcbops/rpc-openstack/blob/master/scripts/bootstrap-aio.yml which is not being done here, and also a lot being done by the bootstrap is unnecessary (configuring loopback disks and such). Perhaps we could figure out a way to re-use the same thing and enable/disable actions based on the environment being deployed? Either that or we should differentiate this one properly.

It seems that the MNAIO tooling does a ton of config related things - why're we using the AIO bootsrap?

I would ideally prefer to use the same config prep mechanism for both test environments. It makes it easier to context switch.

mkam: I reused bootstrap-aio.yml because previously @hughsaunders added the ability to set the variable TARGET to set configurations for mnaio (https://github.com/rcbops/rpc-openstack/blob/master/scripts/bootstrap-aio.yml#L131-L141). I'm using also TARGET to skip the "Execute the OSA AIO bootstrap". I could add the final tasks to create user_osa_variables_overrides.yml to bootstrap-aio.yml and have them run conditionally as well.

Though I guess the real issue is that boostrap-aio is doing more than just bootstrapping AIOs. We could split the configuration parts and the actual bootstrap part into separate playbooks and only use the config playbook for MNAIO.

scripts/configure-multi-node-aio.sh
mkam: The only difference is it calls the configure-multi-node-aio.yml playbook and sets the TARGET variable to mnaio if it hasn't been already. We'll set TARGET in the job, but I thought it would be easier for people to use outside of the gate if the script also set it.

PR #2432, opened 2017-07-25
RI-44: Add post deployment Ironic instructions

prometheanfire:
This documentation goes over how to create images for the deployed ironic
environment along with registering those images for CentOS 7, Ubuntu Xenial and
Ubuntu Trusty.  It also goes over how to register nodes to Ironic.


prometheanfire:
hitting the button so we can move forward

PR #2436, opened 2017-07-28
UG-646 Handle config within RPCO [BP: Liberty]

hughsaunders:
Backport of mitaka version, which was a re-implementation of the
newton backport.

(cherry picked from commit 10338487fb03ec7dcc8d4ee9a93e760fcb953874)

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -79,6 +79,15 @@ if [[ "${DEPLOY_AIO}" == "yes" ]]; then
         ../scripts/update-yaml.py "/${filename}" "${filename}";
       done
     popd
+
+    add_config all
+    if [[ "${TARGET:-}" == "aio" ]]; then
+      add_config aio
+    fi
+    if [[ "${TRIGGER:-}" == "pr" ]]; then
+      add_config pr
+    fi
+
     # ensure that the elasticsearch JVM heap size is limited
     sed -i 's/# elasticsearch_heap_size_mb/elasticsearch_heap_size_mb/' $RPCD_VARS
     # set the kibana admin password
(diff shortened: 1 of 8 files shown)


scripts/functions.sh
```diff
@@ -36,3 +36,14 @@ export RPCD_DIR="${BASE_DIR}/rpcd"
 function run_ansible {
   openstack-ansible ${ANSIBLE_PARAMETERS} --forks ${FORKS} $@
 }
+
+function add_config {
+  src="$1"
+  src_path="${BASE_DIR}/scripts/config/${src}.yml"
+  if [[ -e "${src_path}" ]]; then
+      echo "Adding vars from $src_path to $RPCD_VARS: $(cat $src_path)"
+      cat ${src_path} >> $RPCD_VARS
```
evrardjp: Is there a risk to duplicate some vars, or mess up with user configured files?
If this is only used for gating, that shouldn't be a problem.
However, it looks like ``add_config ceph`` is not only used in gating.
So all of the variables should probably be copied to a separate file (if there is at some point an idea to use this for non gating).

scripts/deploy.sh
```diff
@@ -103,17 +112,16 @@ if [[ "${DEPLOY_AIO}" == "yes" ]]; then
...
-      echo "osd_directory: true" | tee -a $RPCD_VARS
-      echo "osd_directories:" | tee -a $RPCD_VARS
-      echo "  - /var/lib/ceph/osd/mydir1" | tee -a $RPCD_VARS
       sed -i "s/glance_default_store:.*/glance_default_store: rbd/" /etc/openstack_deploy/user_variables.yml
-      echo "nova_libvirt_images_rbd_pool: vms" | tee -a /etc/openstack_deploy/user_variables.yml
-      echo "cinder_ceph_client_uuid:"  | tee -a /etc/openstack_deploy/user_secrets.yml
+      add_config ceph
+    else
+      if [[ "$DEPLOY_SWIFT" == "yes" ]]; then
+        echo "glance_default_store: swift" | tee -a /etc/openstack_deploy/user_osa_variables_defaults.yml
```
evrardjp: This looks like it can cause duplicate vars to appear if deploy.sh is running twice.
Probably worth grepping and adding if not found.

odyssey4me: If it does do duplicate values, that issue will be short lived and has minimal effect. I say 'short lived' because the duplicate executions will be removed from gating once this merges across all branches, and I say 'minimal effect' because the values will be the same and therefore it does no harm.

evrardjp: well you get a large warning that appears on the screen, which is not helping for confidence. But i's an old branch, and it's not a problem, so I understand why it's not very important.

PR #2442, opened 2017-08-01
RE-133 Remove previously configured ansible

odyssey4me:
On the long running apt artifact builder we are
switching Ansible versions for each series, but
leaving behind parts of the installation which
cause the bootstrap to fail.
This is easily seen when doing the ansible
bootstrap using ocata, then trying to bootstrap
using newton. It fails if the role fetch method
is set to use git-clone.

This patch just ensures that we remove
everything related to the ansible execution so
that the new boostrap can do its thing without
worrying about previous installs.

Issue: [RE-133](https://rpc-openstack.atlassian.net/browse/RE-133)

### DIFF
--- a/scripts/artifacts-building/apt/build-apt-artifacts.sh
+++ b/scripts/artifacts-building/apt/build-apt-artifacts.sh
@@ -63,7 +63,16 @@ fi
 
 # The derive-artifact-version.sh script expects the git clone to
 # be at /opt/rpc-openstack, so we link the current folder there.
-ln -sfn ${PWD} /opt/rpc-openstack
+if [[ "${PWD}" != "/opt/rpc-openstack" ]]; then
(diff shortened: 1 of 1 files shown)


odyssey4me:
recheck_cit_all

odyssey4me:
@hughsaunders Yeah, we could do something like that. Let's raise a card for that but get this merged and move on for now.

odyssey4me:
@hughsaunders Card registered: https://rpc-openstack.atlassian.net/browse/RE-161

odyssey4me:
The CIT/artifact-* jobs no longer exist and are replaced by the CIT/artifact test. That test cannot pass due to a problem with the container builds for the ELK stack in their current state.

Also, the CIT/xenial-deploy-swift job will fail due to https://rpc-openstack.atlassian.net/browse/RE-155 which I'm working on resolving.

PR #2446, opened 2017-08-02
[backport] RE-133 Remove previously configured ansible

odyssey4me:
On the long running apt artifact builder we are
switching Ansible versions for each series, but
leaving behind parts of the installation which
cause the bootstrap to fail.
This is easily seen when doing the ansible
bootstrap using ocata, then trying to bootstrap
using newton. It fails if the role fetch method
is set to use git-clone.

This patch just ensures that we remove
everything related to the ansible execution so
that the new boostrap can do its thing without
worrying about previous installs.

(cherry picked from commit 71ff2d6dd90d890b1d452dd03635ad9e0456057b)

### DIFF
--- a/scripts/artifacts-building/apt/build-apt-artifacts.sh
+++ b/scripts/artifacts-building/apt/build-apt-artifacts.sh
@@ -63,7 +63,16 @@ fi
 
 # The derive-artifact-version.sh script expects the git clone to
 # be at /opt/rpc-openstack, so we link the current folder there.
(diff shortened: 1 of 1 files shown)


odyssey4me:
The lint test cannot pass until https://github.com/rcbops/rpc-openstack/pull/2447 merges.

odyssey4me:
This passed the artifact build for apt, which is the only section which is affected by this patch. I would suggest that this just be merged due to this being a blocker for any other PR's against this branch.

PR #2447, opened 2017-08-02
RE-171 Update newton-14.1 refs to newton/tags

odyssey4me:
The release notes are now configured to use the branch
name and the documentation now refers to release tags
instead of the branch name.

Issue: [RE-171](https://rpc-openstack.atlassian.net/browse/RE-171)


evrardjp:
ok to refer to tags, it just won't work for multinode...

odyssey4me:
This is a docs-only patch and there's a bug [1] in the upgrade tests. This should just be merged.

[1] https://rpc-openstack.atlassian.net/browse/RE-172

PR #2450, opened 2017-08-02
RE-155 Add rpco apt source to lxc base cache

odyssey4me:
Prior to https://github.com/rcbops/rpc-openstack/commit/70b86f6a1759f92ef789e3d7028bb3e75ad6e722 the rpco.list file was copied into the container cache
and worked fine because the deploy node and lxc host were the same.
With that patch, this no longer happens and the cache prep dies.

In order for the lxc cache prep process to make use
of the extra rpco apt source we need to ensure that
the repo source file is copied into the cache prior
to doing any package actions. This patch ensures
that the repo config is there.

The OSA SHA is also updated to ensure that the
unbound variable bug is fixed when installing pip.
ref: https://review.openstack.org/489191

Issue: [RE-155](https://rpc-openstack.atlassian.net/browse/RE-155)

Issue: [RE-174](https://rpc-openstack.atlassian.net/browse/RE-174)

Issue: [RE-173](https://rpc-openstack.atlassian.net/browse/RE-173)

### DIFF
--- a/ansible-role-requirements.yml
+++ b/ansible-role-requirements.yml
@@ -30,3 +30,9 @@
   scm: git
   src: https://git.openstack.org/openstack/openstack-ansible-lxc_hosts
   version: fdd159a0238cede501e91a9850c47b4692d334c9
+# TODO(odyssey4me)
+# Remove this once this SHA is included in our OSA SHA
+- name: rabbitmq_server
+  scm: git
+  src: https://git.openstack.org/openstack/openstack-ansible-rabbitmq_server
+  version: e944366f98f358d600f79062465a31ee11cecf4b
--- a/ansible-role-requirements.yml
+++ b/ansible-role-requirements.yml
@@ -24,3 +24,9 @@
   scm: git
   src: https://github.com/rcbops/rpc-role-logstash.git
   version: 3f5e7a6044629e76acb1933af19f09a89e55917b
+# TODO(odyssey4me)
+# Remove this once this SHA is included in our OSA SHA
+- name: lxc_hosts
+  scm: git
+  src: https://git.openstack.org/openstack/openstack-ansible-lxc_hosts
+  version: fdd159a0238cede501e91a9850c47b4692d334c9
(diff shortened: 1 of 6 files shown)


odyssey4me:
Travis tests will fail until https://github.com/rcbops/rpc-openstack/pull/2449 merges.

cjloader:
Version '2.4.1' for 'elasticsearch' was not found -- CIT/xenial-deploy-swift

odyssey4me:
Alright, I've rebuild the apt artifacts for master based on this branch, so the ELK stack will hopefully deploy now.

odyssey4me:
New error:

```
TASK [kibana : Deploy Kibana configuration files] ******************************
Thursday 03 August 2017 10:44:11 +0000 (0:00:00.345) 0:03:13.893 ******* 
fatal: [ramxdsp-275-9505_kibana_container-cd9f9202]: FAILED! =>
{"changed": true, "failed": true, "msg": "Destination directory /opt/kibana/config does not exist"}
```

I'm building an instance to see what the issue is - I suspect that a newer version of kibana is being installed which doesn't use the same folder structure. I'll try to verify that, then figure out a patch to get it working.

odyssey4me:
Confirmed that kibana 5.1.1 is being deployed, so we'll have to work around this.

odyssey4me:
The artifact jobs are currently expected to fail. This PR does require the xenial-deploy-swift job to pass though.

odyssey4me:
recheck_cit_artifact

odyssey4me:
recheck_cit_artifact

odyssey4me:
The selenium tests are failing. @mkam do you know much about these tests - can you help to diagnose the failure?

odyssey4me:
I have validated that the ELK stack everything is working as expected by logging into kibana and seeing events, graphs, etc. So it appears that this may be an issue with the selenium tests themselves.

mkam:
@odyssey4me I haven't seen that particular issue before, but selenium can be finicky, especially when there's a chain of actions (like clicking dropdowns and waiting for them to load and then verifying that the element is there). A retry might be enough to get a passing run, but we could also try to put in retries to the tests too.

mkam:

odyssey4me:
@mkam OK, we'll have to figure it out because without this patch master is broken. So either we need to update the selenium test, or we need to remove it. Obviously the latter is less desirable.

PR #2456, opened 2017-08-03
LA-393 Fix unbound variable DEPLOY_SUPPORT_ROLE

evrardjp:
Even when outside the gates, the DEPLOY_SUPPORT_ROLE variable
needs to be set.

Issue: [LA-393](https://rpc-openstack.atlassian.net/browse/LA-393)


evrardjp:
This slipped through cracks.

odyssey4me:
This needs to be patched to master too, does it not?

evrardjp:
No, it slipped through the cracks. Master has it.

PR #2457, opened 2017-08-03
LA-394 Ensure checkout is done if folder exists

evrardjp:
If the openstack-ansible-ops repo exists, the checkout won't
be done. Re-running the leapfrog after modifying the ops repo
branch env variable won't be taken into account.

This should fix it.

Issue: [LA-394](https://rpc-openstack.atlassian.net/browse/LA-394)

### DIFF
--- a/scripts/leapfrog/ubuntu14-leapfrog.sh
+++ b/scripts/leapfrog/ubuntu14-leapfrog.sh
@@ -97,12 +97,14 @@ pushd ${LEAPFROG_DIR}
 
     # Get the OSA LEAPFROG
     if [[ ! -d "openstack-ansible-ops" ]]; then
-        git clone ${OA_OPS_REPO}
-        pushd openstack-ansible-ops
-            git checkout ${OA_OPS_REPO_BRANCH}
-        popd
+        git clone ${OA_OPS_REPO} openstack-ansible-ops
         log "clone" "ok"
     fi
+    pushd openstack-ansible-ops
+        git fetch --all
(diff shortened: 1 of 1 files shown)


odyssey4me:
recheck_cit_artifact

scripts/leapfrog/ubuntu14-leapfrog.sh
```diff
@@ -98,11 +98,11 @@ pushd ${LEAPFROG_DIR}
     # Get the OSA LEAPFROG
     if [[ ! -d "openstack-ansible-ops" ]]; then
         git clone ${OA_OPS_REPO}
-        pushd openstack-ansible-ops
-            git checkout ${OA_OPS_REPO_BRANCH}
-        popd
         log "clone" "ok"
     fi
+    pushd openstack-ansible-ops
```
odyssey4me: You should add a git fetch --all here to ensure that you get the most up to date refs, otherwise the checkout may fail if it was done prior to the SHA you want to use existing.

scripts/leapfrog/ubuntu14-leapfrog.sh
```diff
@@ -98,11 +98,11 @@ pushd ${LEAPFROG_DIR}
...
+        git checkout ${OA_OPS_REPO_BRANCH}
```
odyssey4me: Maybe add a log entry here too?

evrardjp:
@odyssey4me Done.

PR #2459, opened 2017-08-03
RE-178 Use rpc-maas 1.0.0 tag

hughsaunders:
This is in preparation for the r14.1.1 rpco release and the 2017.8 meta
release. It is important that we use a tagged version of rpc-maas so
that the 2017.8 meta release can be accurately recreated.

Issue: [RE-178](https://rpc-openstack.atlassian.net/browse/RE-178)


odyssey4me:
There should be an equivalent patch to master too, otherwise the same mistake may be made. It's plausible that this change may also affect how tests are run in the rpc-maas repo too.

mancdaz:
recheck_cit_artifact
recheck_cit_deploy_swift

mancdaz:

PR #2467, opened 2017-08-04
RE-184 Ensure that user_defaults are copied

odyssey4me:
Currently the user_defaults file is only
implemented in the AIO bootstrap, but it
should be implemented for *all* deployments.

This patch switches the implementation into
the deploy script to cover all deployments.

(cherry picked from commit 5f66deea1d73e6bb279a32b5a6e0d2a3560b67b3)

Issue: [RE-184](https://rpc-openstack.atlassian.net/browse/RE-184)


odyssey4me:
recheck_cit_all

PR #2472, opened 2017-08-08
LA-401 Safeguard against tentative of ceph leap

evrardjp:
Leapfrog don't handle ceph right now.
We should prevent the leap until the ceph cases are handled.

Issue: [LA-401](https://rpc-openstack.atlassian.net/browse/LA-401)


mancdaz:
recheck_cit_artifact

evrardjp:
@mancdaz gates are broken, no need for recheck.

evrardjp:
We've got enough votes and manual testing, we should go ahead.

PR #2474, opened 2017-08-11

mattt416:


mattt416:

mattt416:

mattt416:
Apparently there's an upstream change being backported now that will unblock this review.

mattt416:

odyssey4me:
recheck_cit_artifact
recheci_cit_xenial_deploy_swift

odyssey4me:

mattt416:
@prometheanfire just a simple rebase to pull in some changes that unbroke the gate

prometheanfire:
yep, still wishing I could do an interdiff between patchsets on github, but lgtm

PR #2475, opened 2017-08-15
RE-187 Resolve master build/test failures

odyssey4me:
This is a PR to resolve all the issues for building artifacts and test deploying RPC-O master.

Issue: [RE-187](https://rpc-openstack.atlassian.net/browse/RE-187)

Issue: [RE-189](https://rpc-openstack.atlassian.net/browse/RE-189)

Issue: [RE-184](https://rpc-openstack.atlassian.net/browse/RE-184)

### DIFF
--- a/group_vars/all/apt.yml
+++ b/group_vars/all/apt.yml
@@ -60,9 +60,10 @@ rabbitmq_erlang_repo: "{{ rpco_apt_repo }}"
 rabbitmq_gpg_keys: "{{ rpco_apt_gpg_keys }}"
 
 # ceph_client wiring
-ceph_apt_repos:
-  uca: "{{ rpco_apt_repo }}"
-ceph_pkg_source: "uca"
+# We use 'distro' to avoid any apt repo being setup by the ceph roles.
+# The repository needed (RPC-O apt artifacts repo) will already be on
+# the host.
+ceph_pkg_source: "distro"
 
 # galera_client role wiring
 galera_client_repo: "{{ rpco_mirror_apt_deb_line }}"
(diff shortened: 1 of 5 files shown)


group_vars/all/apt.yml
```diff
@@ -60,9 +60,10 @@ rabbitmq_erlang_repo: "{{ rpco_apt_repo }}"
 rabbitmq_gpg_keys: "{{ rpco_apt_gpg_keys }}"
 
 # ceph_client wiring
-ceph_apt_repos:
-  uca: "{{ rpco_apt_repo }}"
-ceph_pkg_source: "uca"
+# We use 'distro' to avoid any apt repo being setup by the ceph roles.
+# The repository needed (RPC-O apt artifacts repo) will already be on
+# the host.
+ceph_pkg_source: "distro"
```
mattt416: Are we still required to set anything ceph-related variables here?  Surely that should be done by the ceph deployment itself.

odyssey4me: Regardless of the ceph cluster, there will always be the client parts and the client integration. The packages used for that will need to be in the apt artifacts, so yes this is still required.

mattt416: Do the client/server components need to be in sync?  What happens if someone today deploys a cluster w/ kraken, for example.  Will those packages be available as distro packages?

odyssey4me: @mattt416 That's not something we can answer at this time. Once there is a storage product to test against, that question can be figured out and a solution put together. For now if we don't do this we end up with a rogue repo configured.

Note that from a product standpoint there is no support for Ceph for Newton+ until there is a ceph-based storage product developed, so this is a bit of a no-op which is only here to help the artifacts get configured right when executing tests.

PR #2471, opened 2017-08-08
RE-187 Resolve newton build/test failures

odyssey4me:
This PR pulls together all the fixes necessary to allow newton to build and deploy again, and also includes an update to the release version and SHA's in preparation for cutting the 14.2 stabilisation branch.

Issue: [RE-113](https://rpc-openstack.atlassian.net/browse/RE-113)

Issue: [RE-185](https://rpc-openstack.atlassian.net/browse/RE-185)

Issue: [RE-187](https://rpc-openstack.atlassian.net/browse/RE-187)

Issue: [RE-152](https://rpc-openstack.atlassian.net/browse/RE-152)

Issue: [RE-189](https://rpc-openstack.atlassian.net/browse/RE-189)

Issue: [RE-184](https://rpc-openstack.atlassian.net/browse/RE-184)

Issue: [RE-155](https://rpc-openstack.atlassian.net/browse/RE-155)

Issue: [RE-202](https://rpc-openstack.atlassian.net/browse/RE-202)

Issue: [RE-204](https://rpc-openstack.atlassian.net/browse/RE-204)

Issue: [RE-206](https://rpc-openstack.atlassian.net/browse/RE-206)

Issue: [RE-213](https://rpc-openstack.atlassian.net/browse/RE-213)

Issue: [RE-214](https://rpc-openstack.atlassian.net/browse/RE-214)

Issue: [RE-215](https://rpc-openstack.atlassian.net/browse/RE-215)

Issue: [RE-218](https://rpc-openstack.atlassian.net/browse/RE-218)

PR #2478, opened 2017-08-18

mattt416:

(cherry picked from commit 663e2eb7fc360ac4522ea97cd08961fe6c15614e)
(cherry picked from commit 8b4336a73c7d5a6474775e635e11b9b226bf15a2)

### DIFF
--- a/scripts/migrate-yaml.py
+++ b/scripts/migrate-yaml.py
@@ -37,6 +37,10 @@ def parse_args():
     parser.add_argument('--for-testing-take-new-vars-only', dest='testing',
                         action="store_true",
                         help="don't take old overrides over new defaults")
+    parser.add_argument('--for-testing-keep-old-overrides-only',
+                        dest='testing_keep_overrides',
+                        action="store_true",
+                        help="take old overrides over new defaults")
     return parser.parse_args()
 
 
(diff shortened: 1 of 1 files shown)


odyssey4me:
recheck_cit_artifact

odyssey4me:

mattt416:
Blocked by https://github.com/rcbops/rpc-openstack/pull/2484

mattt416:
@odyssey4me https://rpc.jenkins.cit.rackspace.net/job/RPC-AIO_newton-trusty-leapfrogupgrade-swift-pr/96/ failed on the kibana test, however it seems that https://rpc.jenkins.cit.rackspace.net/job/RPC-AIO_newton-trusty-leapfrogupgrade-swift-pr/99/ passed that test so either something has merged recently or that test is just unstable.

mattt416:
recheck_cit_artifact

odyssey4me:
artifact job success is blocked by https://github.com/rcbops/rpc-openstack/pull/2471
recheck_cit_trusty_leapfrogupgrade_swift - kibana tests fail

odyssey4me:
@mattt416 In the deploy jobs I see the kibana tests fail around 50% of the time for the same PR. I think it's just unstable.

odyssey4me:
To pass the artifacts job, this will require a rebase.

mattt416:
recheck_cit_trusty_leapfrogupgrade_swift - rpc-maas clone failure

mattt416:

mattt416:
recheck_cit_trusty_leapfrogupgrade_swift - tempest failure

PR #2479, opened 2017-08-18

mattt416:

(cherry picked from commit 663e2eb7fc360ac4522ea97cd08961fe6c15614e)
(cherry picked from commit 8b4336a73c7d5a6474775e635e11b9b226bf15a2)

### DIFF
--- a/scripts/migrate-yaml.py
+++ b/scripts/migrate-yaml.py
@@ -37,6 +37,10 @@ def parse_args():
     parser.add_argument('--for-testing-take-new-vars-only', dest='testing',
                         action="store_true",
                         help="don't take old overrides over new defaults")
+    parser.add_argument('--for-testing-keep-old-overrides-only',
+                        dest='testing_keep_overrides',
+                        action="store_true",
+                        help="take old overrides over new defaults")
     return parser.parse_args()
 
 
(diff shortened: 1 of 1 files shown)


mattt416:

mattt416:
Blocked by https://github.com/rcbops/rpc-openstack/pull/2484.

mattt416:

evrardjp:
It seems a bad idea. Now we are hiding some code paths into this toolkit.

If some gate work NEEDS a variable to be set, shouldn't this be a scenario, with different overrides, explicitly defined instead? In that case, no need for this change.

I think that deploy.sh is the interface we are using in general, and maybe we should go ahead and use deploy.sh for upgrades too.

mattt416:
@evrardjp that's good feedback, but unfortunately this new functionality has already been added and this change is just making the flags mutually exclusive.  We can certainly look at making deploy.sh handle upgrades but that would be a separate piece of work.

mattt416:
@evrardjp I just deleted my previous comment as it was incorrect, but basically this functionality has already merged in mitaka (where we actually use this functionality).  This PR here (as well as https://github.com/rcbops/rpc-openstack/pull/2478) just brings the script in line with the version that exists in the mitaka branch.  I certainly think we can revisit how we do upgrades (and making deploy.sh the single "entry-point" for handling upgrades, but I think that's a much larger piece of work that needs consideration.

PR #2480, opened 2017-08-23
Convert the pre/post upgrade roles into playbooks

cloudnull:
The pre/post upgrade roles used over extended host groups and task
matching for group membership. The original playbooks ran actions across
most of the cluster cluster during an upgrade scenario which caused
transient errors and extended run-times. This change moves converts the
roles into playbooks to facilitate better orchestration which will
improve deployment times and allow operators to better understand
what is being executed and where.

This change is being done as part of [Phase 1](https://github.com/rcbops/architecture-review/blob/master/blueprints/rpco-modernization.md#phase-1) of the RPCO modernization
effort.

Issue: [phase-1](https://rpc-openstack.atlassian.net/browse/phase-1)


cloudnull:
@odyssey4me +1 We should add tests for these playbooks. Is there no longer an upgrade specific job? of not can we remove these playbooks all together?

odyssey4me:
@cloudnull There is, it's just not applied to this series (or any other beyond Mitaka). Until RPC-O considers 'upgrades' to be 'supported' no-one would be looking at the job results, so its just a waste of resources and adds more pressure to the few who are paying attention to the jobs.

cloudnull:
@odyssey4me I added an option to the functions.sh lib `DEPLOY_UPGRADE_TOOLS` which defaults to no and can be set to yes to enable testing that code path. However I'm not sure where I should implement the actual activation of that var. Is that something we're still doing in RPC-Gating?

odyssey4me:
@cloudnull for now, yes - that said, perhaps we can try to carve a new path in master here

The job already sets env vars for the trigger, scenario, target, action and series in order for the in-repo scripts to react to. See https://github.com/rcbops/rpc-gating/blob/master/rpc_jobs/rpc_aio.yml#L279-L283

Unfortunately the ability is not yet merged into newton, which is the last place it needs to be before we can materially change the rpc-gating to make it stop doing everything. The newton merge is waiting for the stabilisation of the branch - primarily for https://github.com/rcbops/rpc-openstack/pull/2471

claco:
Yes I did it. I want to make sure these changes and that RFC get input from the folks who will be consuming it.

cloudnull:
@odyssey4me lets work on what that new path will be in a subsequent PR. I think we can do a lot using the OSA tests repo and the framework we've already created in the integrated OSA repo.

PR #2485, opened 2017-08-23
RE-219 Bump percona-xtrabackup

mattt416:
Liberty/mitaka builds are failing to build due to galera not deploying
as a result of an implicit bump of mariadb (10.0.31 to 10.0.32) which
now requires a newer version of XtraBackup than what we deploy.  The
galera cluster fails to build as 2/3 nodes cannot perform a state
transfer from the primary node. The logged error is:

```
WSREP_SST: [ERROR] FATAL: The innobackupex version is 1.5.1. \
Needs xtrabackup-2.3.5 or higher to perform SST (20170823 15:11:28.126)
```

This commit simply updates user_osa_variables_defaults.yml so we deploy
a more current version of percona-xtrabackup.  The specified version is
2.3.5 exactly (the minimum requirement) to keep the amount of change as
minimal as possible.

(cherry picked from commit 452bfea1dc7a6cbe613b1f9cdf6e09a78407aeae)

Conflicts:

Issue: [RE-219](https://rpc-openstack.atlassian.net/browse/RE-219)

### DIFF
--- a/rpcd/etc/openstack_deploy/user_variables.yml
+++ b/rpcd/etc/openstack_deploy/user_variables.yml
@@ -169,3 +169,8 @@ openstack_kernel_options:
   - { key: 'net.ipv6.neigh.default.gc_interval', value: 60 }
   - { key: 'net.ipv6.neigh.default.gc_stale_time', value: 120 }
   - { key: 'fs.aio-max-nr', value: 131072 }
+
+galera_package_url: "https://www.percona.com/downloads/XtraBackup/Percona-XtraBackup-2.3.5/binary/debian/trusty/x86_64/percona-xtrabackup_2.3.5-1.trusty_amd64.deb"
+galera_package_sha256: "7b407be253b922f8bb772d87877486c0ce8eef8af1304abec702dba962b497dd"
+galera_client_package_url: "{{ galera_package_url }}"
+galera_client_package_sha256: "{{ galera_package_sha256 }}"


mattt416:
The rpc-support role is failing with this error:

```
failed: [raltdcp-43-5ab0_galera_container-a980eb76] => (item=bash-completion,bridge-utils,dstat,htop,iotop,iptables-persistent,lsof,openssh-server,sudo,tcpdump,vim) => {"attempts": 3, "failed": true, "item": "bash-completion,bridge-utils,dstat,htop,iotop,iptables-persistent,lsof,openssh-server,sudo,tcpdump,vim"}
stderr: E: Unmet dependencies. Try 'apt-get -f install' with no packages (or specify a solution).

stdout: Reading package lists...
Building dependency tree...
Reading state information...
You might want to run 'apt-get -f install' to correct these:
The following packages have unmet dependencies:
 percona-xtrabackup : Depends: libev4 (>= 1:4.04) but it is not going to be installed
                      Conflicts: percona-xtrabackup-22 but 2.2.13-1.trusty is to be installed
 percona-xtrabackup-22 : Conflicts: percona-xtrabackup but 2.3.5-1.trusty is to be installed
 tcpdump : Depends: libpcap0.8 (>= 1.5.1) but it is not going to be installed

msg: Task failed as maximum retries was encountered
```

I'll need to spin up an instance to troubleshoot further.

mattt416:

evrardjp:
Because 10.0.32 has CVE fixes, it's good that we go forward instead of pinning 10.0.31.

@mattt416 Have you checked if there is a risk for consuming roles (using the xtrabackup CLI) of the bump to 2.3? https://www.percona.com/doc/percona-xtrabackup/LATEST/release-notes/2.3/2.3.2.html#new-features . Maybe the support role or their maintenance plans?

For me, if it passes tests, it means its good. I'm waiting for your reply 'though.

mattt416:
@evrardjp I've not had a chance to test anything further than what was tested here.  I can manually spin up a deploy to verify holland works as expected (assuming that uses xtrabackup), but am not sure what else needs to be checked here.

mattt416:
Kicked off a [build](https://rpc.jenkins.cit.rackspace.net/job/RPC-AIO_liberty-trusty-deploy-swift-pr/46/) to test Holland stage.

evrardjp:
Just popping an email to support that this could affect their maintenance plans on those branch would be enough.

evrardjp:
I think we are good to remove the WIP.

mattt416:
https://rpc.jenkins.cit.rackspace.net/blue/organizations/jenkins/RPC-AIO_liberty-trusty-deploy-swift-pr/detail/RPC-AIO_liberty-trusty-deploy-swift-pr/46/pipeline/194 ran successfully.  @BjoernT, any concerns here before we merge this?

evrardjp:
We'll have concerns without it, so I'd say let's merge and contact support to prevent further concerns...

evrardjp:
@mattt416 FYI I've just sent an email to @JCallicoat and @BjoernT .

PR #2486, opened 2017-08-24
RE-187 Port newton changes to Ocata

odyssey4me:
This ports forward the changes made in https://github.com/rcbops/rpc-openstack/pull/2471 to the master branch to ensure that there are no regressions in Ocata work.

Issue: [RE-218](https://rpc-openstack.atlassian.net/browse/RE-218)

### DIFF
--- a/rpcd/etc/openstack_deploy/user_rpco_variables_defaults.yml
+++ b/rpcd/etc/openstack_deploy/user_rpco_variables_defaults.yml
@@ -42,9 +42,46 @@ maas_target_alias: public0_v4
 maas_monitor_cinder_backup: "{{ cinder_service_backup_program_enabled | default(false) }}"
 
 # MaaS pathing and versions
-maas_release: master
+maas_release: "1.0.0"
 maas_venv: "/openstack/venvs/maas-{{ maas_release }}"
 maas_venv_bin: "{{ maas_venv }}/bin"
 
 # Allow all of the host machiens running maas agents to restart once at the end of a maas playbook.
 maas_restart_independent: false
+
+# Apt artifact repo configuration
+# The rest of the apt artifact repo configuration is held in
+# group_vars/all/apt.yml but due to the way that the rpc-maas
+# implementation is done we cannot use group_vars to override
+# the defaults, so we have to use extra-vars.
+# TODO(odyssey4me):
+# Refactor the rpc-maas implementation to make use of roles
+# to implement the base repo/software so that these settings
+# can be implemented in group_vars instead of extra-vars.
+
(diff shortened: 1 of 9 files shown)


rpcd/etc/openstack_deploy/user_rpco_variables_defaults.yml
```diff
@@ -42,9 +42,46 @@ maas_target_alias: public0_v4
...
 # MaaS pathing and versions
-maas_release: master
+maas_release: "1.0.0"
 maas_venv: "/openstack/venvs/maas-{{ maas_release }}"
 maas_venv_bin: "{{ maas_venv }}/bin"
 
 # Allow all of the host machiens running maas agents to restart once at the end of a maas playbook.
 maas_restart_independent: false
+
+# Apt artifact repo configuration
```
evrardjp: We should probably mark that the content of this file, starting from this line to the end of the file, will be deleted when no artifacts are available. Less black magic.

rpcd/etc/openstack_deploy/user_rpco_variables_defaults.yml
```diff
@@ -42,9 +42,46 @@ maas_target_alias: public0_v4
 maas_monitor_cinder_backup: "{{ cinder_service_backup_program_enabled | default(false) }}"
 
 # MaaS pathing and versions
-maas_release: master
+maas_release: "1.0.0"
```
evrardjp: do we need to pin in master? it's not a release or anywhere near releasing, right?

odyssey4me: Sure, but it could require a release candidate at any time - so I was thinking that instead of only then pinning (delaying an RC) we should pin and sha bump just like any other development branch we have. ie This branch should follow the same release/update process.

rpcd/etc/openstack_deploy/user_rpco_variables_defaults.yml
odyssey4me: That only takes place for the AIO, not for any other deployment. Personally I don't think we should be exposing that to a file used in production deployments - it may make people think they can/could do that for production if it suits them.

evrardjp: Oh yeah. Good.

rpcd/etc/openstack_deploy/user_rpco_variables_defaults.yml
evrardjp: good for me. better safe than sorry.

PR #2487, opened 2017-08-24
DNM: how long does gating take?

prometheanfire:


evrardjp:
Alternatively, you can click on details of the build, and check the history of the job/history of the periodics.

cloudnull:
@prometheanfire good idea.

PR #2493, opened 2017-08-30
Remove the inline injections for better upstream support

cloudnull:
This change removes the injections being done with `sed` so that we're
not breaking use of our upstream tools by injecting our own code into
the default rc file. Prior to this commit any use of the upstream
bootstrap script would break RPC-O almost entirely. Using the known
interfaces, we can sync {group,host}_vars into the
"/etc/openstack_ansible" directory which retain upstream tool
compatibility.

### DIFF
--- a/scripts/artifacts-building/remove-container-aio-config.sh
+++ b/scripts/artifacts-building/remove-container-aio-config.sh
@@ -18,7 +18,8 @@
 # Remove the env.d configurations that set the build to use
 # container artifacts. We don't want this because container
 # artifacts do not currently exist when this script is used.
-sed -i.bak '/lxc_container_variant: /d' /opt/rpc-openstack/group_vars/*.yml
+sed -i.bak '/lxc_container_variant: /d' /opt/rpc-openstack/group_vars/*.yml \
+                                        /opt/rpc-openstack/group_vars/all/*.yml
 
 # Remove the RPC-O default configurations that are necessary
 # for deployment, but cause the build to break due to the fact
--- a/scripts/bootstrap-aio.sh
+++ b/scripts/bootstrap-aio.sh
@@ -60,16 +60,6 @@ if ! container_artifacts_available; then
   ./scripts/artifacts-building/remove-container-aio-config.sh
 fi
 
-# Run AIO bootstrap playbook
-# Setting GROUP_VARS and HOST_VARS to their original
-# values here so that the OSA bootstrap playbooks
-# can run with the correct variables.
-export GROUP_VARS_PATH="/etc/openstack_deploy/group_vars/"
-export HOST_VARS_PATH="/etc/openstack_deploy/host_vars/"
 openstack-ansible -vvv ${BASE_DIR}/scripts/bootstrap-aio.yml \
                   -i "localhost," -c local \
                   -e "${BOOTSTRAP_OPTS}"
-# Unset GROUP_VARS_PATH and HOST_VARS_PATH so that the
-# defaults are taken in openstack-ansible.rc
-unset GROUP_VARS_PATH
-unset HOST_VARS_PATH
--- a/scripts/bootstrap-ansible.sh
+++ b/scripts/bootstrap-ansible.sh
@@ -91,8 +91,19 @@ pushd ${OA_DIR}
     exit 99
   fi
 
-  # Now use GROUP_VARS of OSA and RPC
-  sed -i "s|GROUP_VARS_PATH=.*|GROUP_VARS_PATH=\"\${GROUP_VARS_PATH:-${OA_DIR}/playbooks/inventory/group_vars/:${BASE_DIR}/group_vars/:/etc/openstack_deploy/group_vars/}\"|" /usr/local/bin/openstack-ansible.rc
-  sed -i "s|HOST_VARS_PATH=.*|HOST_VARS_PATH=\"\${HOST_VARS_PATH:-${OA_DIR}/playbooks/inventory/host_vars/:${BASE_DIR}/host_vars/:/etc/openstack_deploy/host_vars/}\"|" /usr/local/bin/openstack-ansible.rc
+  # NOTE(cloudnull): Sync {group,host}_vars from rpc-o into the proper OSA
+  #                  directory which supports the default override interface.
+  #                  to ensure our group and host vars are always in sync the
+  #                  rsync command is used with the --delete flag and the vars
+  #                  are set to read only. If a deployer wishes to override our
+  #                  defaults they can do so using the user_.* files.
+  for dir_name in group_vars host_vars; do
+    if [[ -d "${BASE_DIR}/${dir_name}" ]];then
+      rsync -av --exclude '*.bak' --delete "${BASE_DIR}/${dir_name}" /etc/openstack_deploy/
+      if [[ -d "/etc/openstack_deploy/${dir_name}" ]]; then
+        chmod ugo+rX "/etc/openstack_deploy/${dir_name}"
+      fi
+    fi
+  done
 
 popd
--- a/releasenotes/notes/ro-group-host-vars-1e72a733eebb9cfa.yaml
+++ b/releasenotes/notes/ro-group-host-vars-1e72a733eebb9cfa.yaml
@@ -1,0 +1,8 @@
+---
+other:
+  - `{group,host}_vars` are now set in the `/etc/openstack_deploy` directory
+    and are being set to READ ONLY. This ensures we're able to control our
+    defaults in a sane way without having to modify the OSA interfaces to
+    suit our needs. Should a deployer need want to modify one of our default
+    settings they can follow our existing override pattern by setting their
+    desired option within the `user_.*` files.
--- a/scripts/artifacts-building/containers/build-process.sh
+++ b/scripts/artifacts-building/containers/build-process.sh
@@ -50,9 +50,15 @@ if [[ "${PWD}" != "/opt/rpc-openstack" ]]; then
 fi
 
 # Bootstrap Ansible
+cd /opt/rpc-openstack
+
+# Remove the AIO configuration relating to the use
+# of container artifacts. This needs to be done
+# because the container artifacts do not exist yet.
+./scripts/artifacts-building/remove-container-aio-config.sh
+
 # This script is sourced to ensure that the common
 # functions and vars are available.
-cd /opt/rpc-openstack
 source scripts/bootstrap-ansible.sh
 
 # Bootstrap the AIO configuration
@@ -74,11 +80,6 @@ fi
 if [[ "$(echo ${REPLACE_ARTIFACTS} | tr [a-z] [A-Z])" == "YES" ]]; then
   export PUSH_TO_MIRROR="YES"
 fi
-
-# Remove the AIO configuration relating to the use
-# of container artifacts. This needs to be done
-# because the container artifacts do not exist yet.
-./scripts/artifacts-building/remove-container-aio-config.sh
 
 # Set override vars for the artifact build
 cd scripts/artifacts-building/
@@ -195,4 +196,3 @@ if [[ "$(echo ${PUSH_TO_MIRROR} | tr [a-z] [A-Z])" == "YES" ]]; then
 else
   echo "Skipping upload to rpc-repo as the PUSH_TO_MIRROR env var is not set to 'YES'."
 fi
-
--- a/scripts/artifacts-building/python/build-python-artifacts.sh
+++ b/scripts/artifacts-building/python/build-python-artifacts.sh
@@ -47,9 +47,15 @@ if [[ "${PWD}" != "/opt/rpc-openstack" ]]; then
 fi
 
 # Bootstrap Ansible
+cd /opt/rpc-openstack
+
+# Remove the AIO configuration relating to the use
+# of container artifacts. This needs to be done
+# because the container artifacts do not exist yet.
+./scripts/artifacts-building/remove-container-aio-config.sh
+
 # This script is sourced to ensure that the common
 # functions and vars are available.
-cd /opt/rpc-openstack
 source scripts/bootstrap-ansible.sh
 
 # Bootstrap the AIO configuration
@@ -57,11 +63,6 @@ source scripts/bootstrap-ansible.sh
 
 # Copy the current user-space defaults file
 cp "/opt/rpc-openstack/rpcd/etc/openstack_deploy/user_osa_variables_defaults.yml" /etc/openstack_deploy/
-
-# Remove the AIO configuration relating to the use
-# of container artifacts. This needs to be done
-# because the container artifacts do not exist yet.
-./scripts/artifacts-building/remove-container-aio-config.sh
 
 # Set override vars for the artifact build
 echo "repo_build_wheel_selective: no" >> /etc/openstack_deploy/user_osa_variables_overrides.yml


cloudnull:
It was created to be consumed, not to be whole sale modified. As a downstream consumer we use the interface not modify it.

scripts/bootstrap-aio.sh
```diff
@@ -63,16 +63,6 @@ if ! container_artifacts_available; then
   ./scripts/artifacts-building/remove-container-aio-config.sh
 fi
 
-# Run AIO bootstrap playbook
```
odyssey4me: These need to stay - it actually an upstream bug which hasn't been dealt with yet IIRC.

scripts/bootstrap-aio.sh
```diff
@@ -51,7 +51,7 @@ if apt_artifacts_available; then
   export RPCO_APT_ARTIFACTS_AVAILABLE="yes"
 else
   export RPCO_APT_ARTIFACTS_AVAILABLE="no"
-  rm -f ${BASE_DIR}/group_vars/all/apt.yml
```
odyssey4me: This intentionally dirties the git tree - only for an AIO, and only when apt artifacts aren't available. It would only ever happen in the gate, so better to leave as-is.

scripts/bootstrap-ansible.sh
```diff
@@ -75,8 +75,13 @@ pushd ${OA_DIR}
   # path for Ansible to search.
   sed -i "s|/etc/ansible/roles:|/etc/ansible/roles:${RPCD_DIR}/playbooks/roles:|" /usr/local/bin/openstack-ansible.rc
 
-  # Now use GROUP_VARS of OSA and RPC
-  sed -i "s|GROUP_VARS_PATH=.*|GROUP_VARS_PATH=\"\${GROUP_VARS_PATH:-${OA_DIR}/playbooks/inventory/group_vars/:${BASE_DIR}/group_vars/:/etc/openstack_deploy/group_vars/}\"|" /usr/local/bin/openstack-ansible.rc
-  sed -i "s|HOST_VARS_PATH=.*|HOST_VARS_PATH=\"\${HOST_VARS_PATH:-${OA_DIR}/playbooks/inventory/host_vars/:${BASE_DIR}/host_vars/:/etc/openstack_deploy/host_vars/}\"|" /usr/local/bin/openstack-ansible.rc
+  # NOTE(cloudnull): Sync {group,host}_vars from rpc-o into the proper OSA
```
odyssey4me: Doing this will mean that our support group won't be able to change these files or add any other overrides to these group vars. If they're OK with that, then I guess this is OK. I would prefer it if the files were deliberately set to be read-only just to make it a little clearer.

odyssey4me:
It still needs changes, but after discussion and understanding the objective I can go along with this. I do think that user-space config is going to cause trouble and confusion, but this is the interface we have access to until something better is implemented.

scripts/bootstrap-aio.sh
cloudnull: these are the default settings in the RC file. Do they still need to be here if we're not modifying that pathing?

scripts/bootstrap-aio.sh
cloudnull: done.

scripts/bootstrap-ansible.sh
cloudnull: ++ I'll make them 0400.

scripts/bootstrap-aio.sh
cloudnull: if the group_vars are in the "/etc/openstack_deploy" do we still need this ?

odyssey4me: Yes, because we want to prevent them from getting there in gate tests where the apt artifacts do not yet exist for a release version. This is important for PR testing where the release version is changing.

scripts/bootstrap-aio.sh
odyssey4me: Yes, sorry - I got a little confused. This should be able to go.

scripts/bootstrap-ansible.sh
cloudnull: done.

cloudnull:

scripts/bootstrap-aio.sh
```diff
@@ -51,7 +51,8 @@ if apt_artifacts_available; then
...
+  rm -f /etc/openstack_deploy/group_vars/all/apt.yml
```
odyssey4me: This line is now no longer necessary.

scripts/bootstrap-ansible.sh
```diff
@@ -64,8 +64,19 @@ pushd ${OA_DIR}
     exit 99
   fi
 
-  # Now use GROUP_VARS of OSA and RPC
-  sed -i "s|GROUP_VARS_PATH=.*|GROUP_VARS_PATH=\"\${GROUP_VARS_PATH:-${OA_DIR}/playbooks/inventory/group_vars/:${BASE_DIR}/group_vars/:/etc/openstack_deploy/group_vars/}\"|" /usr/local/bin/openstack-ansible.rc
-  sed -i "s|HOST_VARS_PATH=.*|HOST_VARS_PATH=\"\${HOST_VARS_PATH:-${OA_DIR}/playbooks/inventory/host_vars/:${BASE_DIR}/host_vars/:/etc/openstack_deploy/host_vars/}\"|" /usr/local/bin/openstack-ansible.rc
+  # RPC-O has roles in its own git tree, so we need to add it to the
+  # path for Ansible to search.
```
odyssey4me: There are no roles in the rpc-openstack git tree any more.

scripts/bootstrap-ansible.sh
```diff
@@ -64,8 +64,19 @@ pushd ${OA_DIR}
...
+  sed -i "s|/etc/ansible/roles:|/etc/ansible/roles:${RPCD_DIR}/playbooks/roles:|" /usr/local/bin/openstack-ansible.rc
+
+  # NOTE(cloudnull): Sync {group,host}_vars from rpc-o into the proper OSA
+  #                  directory which supports the default override interface.
+  if [[ -d "${BASE_DIR}/group_vars" ]];then
+    rsync -av ${BASE_DIR}/group_vars /etc/openstack_deploy/
+    var_perms "/etc/openstack_deploy/group_vars"
```
odyssey4me: Instead of nesting a function here (considering the objective is to decrease the layers) you could also just do: ``chmod ugo+rX <path>``

scripts/bootstrap-ansible.sh
```diff
@@ -64,8 +64,19 @@ pushd ${OA_DIR}
...
-  sed -i "s|GROUP_VARS_PATH=.*|GROUP_VARS_PATH=\"\${GROUP_VARS_PATH:-${OA_DIR}/playbooks/inventory/group_vars/:${BASE_DIR}/group_vars/:/etc/openstack_deploy/group_vars/}\"|" /usr/local/bin/openstack-ansible.rc
-  sed -i "s|HOST_VARS_PATH=.*|HOST_VARS_PATH=\"\${HOST_VARS_PATH:-${OA_DIR}/playbooks/inventory/host_vars/:${BASE_DIR}/host_vars/:/etc/openstack_deploy/host_vars/}\"|" /usr/local/bin/openstack-ansible.rc
+  # RPC-O has roles in its own git tree, so we need to add it to the
+  # path for Ansible to search.
+  sed -i "s|/etc/ansible/roles:|/etc/ansible/roles:${RPCD_DIR}/playbooks/roles:|" /usr/local/bin/openstack-ansible.rc
+
+  # NOTE(cloudnull): Sync {group,host}_vars from rpc-o into the proper OSA
+  #                  directory which supports the default override interface.
+  if [[ -d "${BASE_DIR}/group_vars" ]];then
+    rsync -av ${BASE_DIR}/group_vars /etc/openstack_deploy/
```
odyssey4me: I would suggest including the use of ``--delete`` to remove any files removed in-tree too, then add a release note to note to deployers that any use of ``/etc/openstack_deploy/group_vars`` will be overwritten and that overrides should only be done in the user_* files.

scripts/bootstrap-ansible.sh
```diff
@@ -64,8 +64,19 @@ pushd ${OA_DIR}
...
+    var_perms "/etc/openstack_deploy/group_vars"
+  fi
+  if [[ -d "${BASE_DIR}/host_vars" ]];then
+    rsync -av ${BASE_DIR}/host_vars /etc/openstack_deploy/
```
odyssey4me: As above.

scripts/bootstrap-ansible.sh
```diff
@@ -64,8 +64,19 @@ pushd ${OA_DIR}
...
+    var_perms "/etc/openstack_deploy/host_vars"
```
odyssey4me: As above.

scripts/bootstrap-ansible.sh
cloudnull: I'm not sure we should have the yaml files be executable. I mean it shouldn't cause any harm but there's no reason to set the execution bit.

scripts/bootstrap-aio.sh
cloudnull: done.

scripts/bootstrap-ansible.sh
cloudnull: done.

scripts/bootstrap-ansible.sh
cloudnull: done.

scripts/bootstrap-ansible.sh
odyssey4me: Capital X would only set the executable bit on directories, not files.

scripts/bootstrap-ansible.sh
```diff
@@ -91,8 +91,17 @@ pushd ${OA_DIR}
...
+  # NOTE(cloudnull): Sync {group,host}_vars from rpc-o into the proper OSA
+  #                  directory which supports the default override interface.
+  #                  to ensure our group and host vars are always in sync the
+  #                  rsync command is used with the --delete flag and the vars
+  #                  are set to read only. If a deployer wishes to override our
+  #                  defaults they can do so using the user_.* files.
+  for dir_name in group_vars host_vars; do
+    if [[ -d "${BASE_DIR}/${dir_name}" ]];then
+      rsync -av --delete ${BASE_DIR}/${dir_name} /etc/openstack_deploy/
+      var_perms "/etc/openstack_deploy/${dir_name}"
```
odyssey4me: Instead of nesting a function here (considering the objective is to decrease the layers) you could also just do: chmod ugo+rX <path>

This would set the files to 0444 and the directories to 0755.

odyssey4me:
```
find: ‘/etc/openstack_deploy//etc/openstack_deploy/group_vars’: No such file or directory
```

scripts/bootstrap-ansible.sh
cloudnull: done.

scripts/bootstrap-aio.sh
odyssey4me: Why was this removed again? It needs to be there to cater for tests where the container artifacts for a release do not exist when building an AIO. That's why the tests failed.

cloudnull:

odyssey4me:
@cloudnull The reason for the failure is that the line ``rm -f ${BASE_DIR}/group_vars/all/apt.yml`` was removed from the AIO bootstrap, and there were no container artifacts. Without that line the build fails when there are no artifacts... and artifacts can't be built either.

scripts/bootstrap-aio.sh
cloudnull: done.

odyssey4me:
recheck_cit_xenial_deploy_swift - failed kibana tests

scripts/artifacts-building/remove-container-aio-config.sh
```diff
@@ -18,7 +18,8 @@
 # Remove the env.d configurations that set the build to use
 # container artifacts. We don't want this because container
 # artifacts do not currently exist when this script is used.
-sed -i.bak '/lxc_container_variant: /d' /opt/rpc-openstack/group_vars/*.yml
+sed -i.bak '/lxc_container_variant: /d' /opt/rpc-openstack/group_vars/*.yml \
+                                        /opt/rpc-openstack/group_vars/all/*.yml
```
odyssey4me: This extra line is not necessary as the only file in /all that contains that var is the one that gets deleted in the line below.

scripts/artifacts-building/remove-container-aio-config.sh
```diff
@@ -18,7 +18,8 @@
 # Remove the env.d configurations that set the build to use
 # container artifacts. We don't want this because container
 # artifacts do not currently exist when this script is used.
-sed -i.bak '/lxc_container_variant: /d' /opt/rpc-openstack/group_vars/*.yml
+sed -i.bak '/lxc_container_variant: /d' /opt/rpc-openstack/group_vars/*.yml \
```
odyssey4me: I suspect that the plugin may be loading *any* files from ``/etc/openstack_deploy/group_vars``, including the .bak files which are being copied over:

```
++ rsync -av --delete /opt/rpc-openstack/group_vars /etc/openstack_deploy/
sending incremental file list
group_vars/
group_vars/cinder_all.yml
group_vars/cinder_all.yml.bak
group_vars/glance_all.yml
group_vars/glance_all.yml.bak
group_vars/heat_all.yml
group_vars/heat_all.yml.bak
group_vars/horizon_all.yml
group_vars/horizon_all.yml.bak
group_vars/ironic_all.yml
group_vars/ironic_all.yml.bak
group_vars/keystone_all.yml
group_vars/keystone_all.yml.bak
group_vars/kibana_all.yml
group_vars/kibana_all.yml.bak
group_vars/log_hosts.yml
group_vars/log_hosts.yml.bak
group_vars/logstash_all.yml
group_vars/logstash_all.yml.bak
group_vars/memcached_container.yml
group_vars/memcached_container.yml.bak
group_vars/neutron_all.yml
group_vars/neutron_all.yml.bak
group_vars/nova_all.yml
group_vars/nova_all.yml.bak
group_vars/repo_all.yml
group_vars/repo_all.yml.bak
group_vars/rsyslog_all.yml
group_vars/rsyslog_all.yml.bak
group_vars/swift_all.yml
group_vars/swift_all.yml.bak
group_vars/utility_all.yml
group_vars/utility_all.yml.bak
group_vars/all/
group_vars/all/apt.yml
group_vars/all/apt.yml.bak
group_vars/all/ceph.yml
group_vars/all/ceph.yml.bak
group_vars/all/elasticsearch.yml
group_vars/all/elasticsearch.yml.bak
group_vars/all/lxc.yml.bak
group_vars/all/osa.yml
group_vars/all/osa.yml.bak
group_vars/all/release.yml
group_vars/all/release.yml.bak
group_vars/all/rpc-o.yml
group_vars/all/rpc-o.yml.bak
```

Maybe we should try ditching the backup when doing the inline editing. It's not necessary and was simply a force of habit.

scripts/artifacts-building/remove-container-aio-config.sh
odyssey4me: nit: This is still an unnecessary change.

odyssey4me:

odyssey4me:
It looks like this is failing on this known issue: https://rpc-openstack.atlassian.net/browse/RO-2781

odyssey4me:
recheck_xenial_swift_deploy

odyssey4me:
The ``xenial-deploy-swift`` test is legacy and no longer exists, so its results can be ignored.

prometheanfire:

@cloudnull has this been superseded by your megapatch?

prometheanfire:

cloudnull:
Abandoned in favor of INTAKERPC-132

PR #2494, opened 2017-08-30
Remove the galaxy legacy role removal

cloudnull:
This command is no needed and has not been needed for sometime. With the
total removal of all of our nested roles in rpc-o this code is no longer
required and should be removed.


odyssey4me:
recheck_cit_artifact

PR #2495, opened 2017-08-30
This removes redundant scripts and update linters

cloudnull:
This change beings using the tests repo from osa to better leverage well
curated and functional linters so that we're not having to mantain code
that is more/less a duplication to what is already free and upstream. This
update follows the patterns we've already used in RPC-MaaS.


doc/source/conf.py
```diff
@@ -0,0 +1,338 @@
...
+# This file is execfile()d with the current directory set to its
+# containing dir.
+#
+# Note that not all possible configuration values are present in this
+# autogenerated file.
+#
+# All configuration values have a default; values that are commented out
+# serve to show the default.
+
+import openstackdocstheme
```
odyssey4me: non-official project are not allowed to use this theme

doc/makefile
```diff
@@ -0,0 +1,196 @@
+# Makefile for Sphinx documentation
```
odyssey4me: Why're we adding doc stuff which is unused? It's a bit crufty.

setup.cfg
```diff
@@ -0,0 +1,24 @@
+[metadata]
```
odyssey4me: setup.cfg/py I think are only used by the doc build process, so they're not necessary

doc/source/conf.py
cloudnull: Is that documented somewhere?

doc/makefile
cloudnull: we have no docs atm however we could/should soon.

doc/source/conf.py
cloudnull: done.

doc/makefile
odyssey4me: Instead of adding this cruft now, why not add it then. It just becomes more lines of code to manage and another set of tests to deal with.

doc/source/conf.py
odyssey4me: You can ask @major - the monitorstack repo had to remove it.

cloudnull: done.

cloudnull:

major:
Looks like an invalid SHA of openstack-ansible-ops is being used somewhere:

> failed: [localhost] (item={'scm': u'git', 'src': u'https://github.com/rsoprivatecloud/openstack-ops', 'version': u'9cafb5d87463ba927a9315bd85dcc23e05683089', 'name': u'rcbops_openstack-ops'}) => {"attempts": 2, "cmd": "/usr/bin/git checkout --force 9cafb5d87463ba927a9315bd85dcc23e05683089", "failed": true, "item": {"name": "rcbops_openstack-ops", "scm": "git", "src": "https://github.com/rsoprivatecloud/openstack-ops", "version": "9cafb5d87463ba927a9315bd85dcc23e05683089"}, "msg": "Failed to checkout 9cafb5d87463ba927a9315bd85dcc23e05683089", "rc": 128, "stderr": "fatal: reference is not a tree: 9cafb5d87463ba927a9315bd85dcc23e05683089\n", "stdout": "", "stdout_lines": []}

major:
recheck_cit_issue_link

cloudnull:
@evrardjp I agree it'd be simpler to just use what upstream is producing. That said, we need an incremental steps to remove some of our lesser used and outdated bits and this is just one of those steps.

cloudnull:
recheck_cit_issue_link

prometheanfire:
recheck_cit_issue_link

prometheanfire:
@cloudnull has this been superseded by your newest megapatch?

cloudnull:
Abandoned in favor of INTAKERPC-132

PR #2497, opened 2017-08-31
RO-2429 Remove neutron_api_workers override

major:
This patch removes the `neutron_api_workers` override that set the
number of API workers to 64. OSA already has code that calculates
an optimal number of neutron api workers by taking the number of
CPU cores and dividing by 2. If that value is incorrect, we should
change the logic in OSA.

Issue: [RO-2429](https://rpc-openstack.atlassian.net/browse/RO-2429)


major:
This one will fail linter checks until #2495 gets in.

odyssey4me:

odyssey4me:
recheck_cit_artifact

evrardjp:
Good. 👍  after rebase.

major:
Combined in #2516.

PR #2499, opened 2017-08-31
RO-2431 Remove rpc_response_timeout overrides

major:
In the Havana/Grizzly releases, RPC thread response timeout overrides
were required because of an issue with iptables when lots of
simultaneous builds were occurring. This was fixed in the Havana
release and these overrides are no longer needed.

The OSA default of 60 seconds is strongly recommended by the
oslo.messaging project. This patch removes the override so that the
safe default can be used.

Upstream Bug: https://bugs.launchpad.net/nova/+bug/1199433
Upstream Fix: https://review.openstack.org/#/c/40688/

Issue: [RO-2431](https://rpc-openstack.atlassian.net/browse/RO-2431)

### DIFF
--- a/group_vars/all/osa.yml
+++ b/group_vars/all/osa.yml
@@ -14,24 +14,20 @@
 # limitations under the License.
 
 # SQLAlchemy/Olso Thread Pool Settings
-rpc_response_timeout: 180
 rpc_thread_pool_size: 180
 db_max_pool_size: 120
 db_pool_timeout: 60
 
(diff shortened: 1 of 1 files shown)


major:
This will fail linters until #2495 merges.

odyssey4me:

odyssey4me:
recheck_cit_artifact

evrardjp:
Rebase needed but 👍

major:
Combined in #2516.

PR #2503, opened 2017-09-01
RO-2434 Remove pool_timeout overrides

major:
RPC-O overrides the pool_timeout to 60 seconds, wheras oslo.db,
SQLAlchemy, and OSA all recommend 30 seconds as a maximum. There
were some bugs around building lots of instances and hitting db pool
timeouts back in the Icehouse/Kilo timeframes, but these bugs have
been fixes for quite some time.

This patch removes the override and uses the OSA default of 30 seconds.
If we're really expecting to wait up to 60 seconds for a successful
database connection, then we're making some serious MariaDB/OpenStack
issues in a deployment.

Issue: [RO-2434](https://rpc-openstack.atlassian.net/browse/RO-2434)

### DIFF
--- a/group_vars/all/osa.yml
+++ b/group_vars/all/osa.yml
@@ -17,19 +17,16 @@
 rpc_response_timeout: 180
 rpc_thread_pool_size: 180
 db_max_pool_size: 120
-db_pool_timeout: 60
 
 cinder_rpc_executor_thread_pool_size: "{{ rpc_thread_pool_size }}"
 cinder_rpc_response_timeout: "{{ rpc_response_timeout }}"
 
 keystone_database_max_pool_size: "{{ db_max_pool_size }}"
-keystone_database_pool_timeout: "{{ db_pool_timeout }}"
 
 neutron_api_workers: 64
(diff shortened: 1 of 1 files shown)


major:

odyssey4me:
Odd - lint failed due to elasticsearch role missing. Might need to rebase or close/open to kick them all off again.

major:
recheck_cit_artifact

evrardjp:
Good but needs conflicts merging.

major:
Combined in #2516.

BjoernT:
How did we verify that we can boot 400-500 instances at once as our customers provision currently. None of the OPS folks have been consulted and a change of our default parameter is simply not desired unless we can prove that above requirements still work ?

BjoernT:
@devx As said, don't make any changes on default parameters without consulting Jordan or myself. The parameters have been defined as default on purpose so that we don't have to locally override all parameters all the time. Also this cuts down on post installation issues quite a bit, since the configuration was usable for a large set of customer environments. I will follow up offline how that parameter change was verified

PR #2504, opened 2017-09-01
RO-2435 Remove max_overflow overrides

major:
The max_overflow override sets the maximum number of simultaneous
connected database threads. SQLAlchemy recommends setting this to 10
and oslo.db accepts this default. OSA uses the default for most
services but it bumps the max to 20 for very busy services, such as
nova and neutron.

This patch removes the override since it is no longer necessary to
override the max_overflow value. Bugs existed in Icehouse/Kilo that
caused excessive database threads when lots of instances were being
built, but those bugs have been fixed for years.

Issue: [RO-2435](https://rpc-openstack.atlassian.net/browse/RO-2435)

### DIFF
--- a/group_vars/all/osa.yml
+++ b/group_vars/all/osa.yml
@@ -17,7 +17,6 @@
 rpc_conn_pool_size: 180
 rpc_response_timeout: 180
 rpc_thread_pool_size: 180
-db_max_overflow: 60
 db_max_pool_size: 120
 db_pool_timeout: 60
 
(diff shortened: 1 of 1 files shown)


odyssey4me:

odyssey4me:
recheck_cit_artifact

BjoernT:
How did we verify that we can boot 400-500 instances at once as our customers provision currently. None of the OPS folks have been consulted and a change of our default parameter is simply not desired unless we can prove that above requirements still work ?

PR #2505, opened 2017-09-01
RO-2436 Remove rpc_conn_pool_size overrides

major:
This patch removes the rpc_conn_pool_size overrides from RPC-O. These
overrides were necessary back in Icehouse/Kilo, but the bugs that
required these increases have long been fixed.

Upstream fix: https://review.openstack.org/#/c/132202
Upstream bug: https://bugs.launchpad.net/neutron/+bug/1372049

Issue: [RO-2436](https://rpc-openstack.atlassian.net/browse/RO-2436)

### DIFF
--- a/group_vars/all/osa.yml
+++ b/group_vars/all/osa.yml
@@ -14,7 +14,6 @@
 # limitations under the License.
 
 # SQLAlchemy/Olso Thread Pool Settings
-rpc_conn_pool_size: 180
 rpc_response_timeout: 180
 rpc_thread_pool_size: 180
 db_max_overflow: 60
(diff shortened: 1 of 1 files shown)


evrardjp:
Looks good, but what about the other pool sizes?

major:
@evrardjp See the other PR's. ;)

odyssey4me:

odyssey4me:
@major I have the feeling these are all going to merge conflict, so it might be better to have them all in one PR. You can do multiple commits, but a single PR.

BjoernT:
@odyssey4me @major  How did we verify that we can boot 400-500 instances at once as our customers provision currently. None of the OPS folks have been consulted and a change of our default parameter is simply not desired unless we can prove that above requirements still work ?

PR #2506, opened 2017-09-01
RO-2437 Remove CPU/RAM allocation ratio overrides

major:
RPC-O is setting overrides that are identical to OSA's defaults.
This patch removes those overrides to avoid problems later.

Issue: [RO-2437](https://rpc-openstack.atlassian.net/browse/RO-2437)


major:

odyssey4me:
recheck_cit_artifact

PR #2509, opened 2017-09-01
RO-2433 Fix openstack-ops SHA

major:
The SHA used for openstack-ops has disappeared, likely due to a
forced push. This patch uses the head of openstack-ops within
rpc-openstack.

Issue: [RO-2433](https://rpc-openstack.atlassian.net/browse/RO-2433)


major:
:red_circle: The incorrect SHA is blocking rpc-openstack's master branch.

mattt416:
Was just about to add my review (+2 from me), but two things worth noting:

1. @BjoernT you should probably be aware of this.  Any idea what happened?
2. There was chatter about removing this from role requirements, and maybe bringing it in as we do w/ rpc-maas.  That's for further discussion, but just an FYI.

PR #2513, opened 2017-09-01
RO-2433 Remove untested role from rpco

cloudnull:
The openstack-ops role is untested and under active development. Until
this role can be stabalized, is no longer being force pushed against
and has some testing it should not be part of the rpco product
offering. If this role is being used by folks in the support
organization it can cloned and installed as needed outside the confinds
of the RPCO prodcut offering.


mattt416:
Is the plan to remove this all the way down?  We'll probably want to adjust https://github.com/rcbops/rpc-gating/blob/master/rpc_jobs/rpc_aio.yml#L95 accordingly before this merges otherwise we'll get periodic failures as a result.

cloudnull:
@mattt416 I'd say yes we would want to remove this all of the way down. If we're not using this and its not being developed by engineering/has no testing then it should be completely excised from the code base. We can re-introduce this role and its bits once we have a stable-ish role to consume.

major:
@BjoernT or @npawelek -- Do either of you know who maintains or owns the openstack-ops repository? We've had some issues with disappearing SHAs that have blocked the RPC-O gates for extended periods.

I'm inclined to take this role out until we figure out a good way to test commits in that repo.

BjoernT:
Yes take it out. Maintaining is on me.

odyssey4me:
The rpc-gating PR to remove all traces of the support role and holland testing is here: https://github.com/rcbops/rpc-gating/pull/405

odyssey4me:
For the record, a card has been registered to make this new testing gap clear: https://rpc-openstack.atlassian.net/browse/RO-2440

major:

The new changes to gating should help this change pass.

odyssey4me:
recheck_cit_artifact

PR #2516, opened 2017-09-05
Remove unneeded overrides

major:
This is a combined PR of several patches that clean up unneeded overrides.

Issue: [RO-2434](https://rpc-openstack.atlassian.net/browse/RO-2434)

### DIFF
--- a/group_vars/all/osa.yml
+++ b/group_vars/all/osa.yml
@@ -25,7 +25,6 @@ cinder_rpc_response_timeout: "{{ rpc_response_timeout }}"
 keystone_database_max_pool_size: "{{ db_max_pool_size }}"
 keystone_database_pool_timeout: "{{ db_pool_timeout }}"
 
-neutron_api_workers: 64
 neutron_rpc_response_timeout: "{{ rpc_response_timeout }}"
 neutron_rpc_thread_pool_size: "{{ rpc_thread_pool_size }}"
 neutron_db_max_pool_size: "{{ db_max_pool_size }}"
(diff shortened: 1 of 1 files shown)


cloudnull:
Merging the combined patch as this has all been thoroughly tested elsewhere.

BjoernT:
@odyssey4me @major How did we verify that we can boot 400-500 instances at once as our customers provision currently. None of the OPS folks have been consulted and a change of our default parameter is simply not desired unless we can prove that above requirements still work ?

evrardjp:
@BjoernT I agree on the notification part. We should have a mechanism to warn you. But what you're saying is probably an edge case for some specific (maybe very important) customers.

What I mean by that is that we should probably have tests for these specific customers, with overrides for them. Overlay those specifics on top of the global defaults for all the customers.

I'll not be the judge of whether or not the 400-500 instances at once test is good enough for testing those variables. I guess there are "many ways to skin a cat" (if that's the proper expression).

Should a session be organised on the collaboration/notification bit?

BjoernT:
@evrardjp 
To rephrase my expectation, RPC-O is targeted for production workload for control planes with more than 12 cores (HT enabled) and 64GB RAM, hence we choose these default parameters because they have proven to scale from to these workloads to large multi hundred nodes clouds.
My opinion is that the default parameter should reflect this, they need to target the intended end user and not the other way around.
If there is a need to lower the threads to account fro AIO or other dev related activities, that's fine and treat it like the openstack_user_config.yml where we have a separate set of configuration used. 

That is why I request the revert of this commit.

PR #2518, opened 2017-09-05
[Newton] RO-2439 Optimize AIO with one galera/rabbit node

major:
In OSA, we thoroughly test building three Galera nodes and three
RabbitMQ nodes, including upgrading both. The OSA AIO builds only
use one Galera/RabbitMQ node each to save time/resources since the
role tests are quite thorough.

It would make sense to use one Galera/RabbitMQ node each in RPC-O
AIO builds to save time and resources. This patch removes the
configuration override and allows the OSA defaults to be used.

(cherry picked from commit c5bd64b0f6a6aede85955085fa08173ed86384fd)

Issue: [RO-2439](https://rpc-openstack.atlassian.net/browse/RO-2439)

### DIFF
--- a/scripts/bootstrap-aio.yml
+++ b/scripts/bootstrap-aio.yml
@@ -71,13 +71,6 @@
         - name: swift.yml.aio
         - name: ceph.yml.aio
           path: "{{ lookup('env', 'RPCD_DIR') ~ '/etc/openstack_deploy/conf.d' }}"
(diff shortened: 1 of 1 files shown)


major:

Horizon LXC container never came up.

mattt416:

major:

major:
Leapfrog upgrade keeps timing out. :(

PR #2532, opened 2017-09-14
RLM-207 Update rpc-maas SHA to use ping plugin

git-harry:
The override for `maas_influxdata_repo` is incorrect, rpc-maas is
looking for `maas_influxdata_repo.url` whereas the override provides
`maas_influxdata_repo.repo`. When this is corrected the telegraf package
is not found. These overrides are removed until telegraf is stored as an
artefact.

Instance monitoring for leapfrog upgrades makes use of the Telegraf ping
plugin. `maas_release` is updated to a SHA that support this
functionality.

Issue: [RLM-207](https://rpc-openstack.atlassian.net/browse/RLM-207)


git-harry:
recheck_cit_artifact

PR #2536, opened 2017-09-18
RO-2428 Bump OSA SHA to head of Newton

major:
This patch bumps the OSA SHA to the head of the Newton branch.

    9b8f3ffd Update SHA for os_tempest
    dff42ad3 Switch to use ansible-hardening
    c351cdb8 Suppress curl warning w/shell module
    6d10e575 Fix LXC container start order

This one is important since it has the `openstack-ansible-security` > `ansible-hardening` changeover.

PR #2538, opened 2017-09-19
RO-2428 Bump RabbitMQ role SHA

major:
This patch provides some temporary relief from RO-2428 until there is
an upstream release of OpenStack-Ansible's Newton branch that
includes the fix.

This also brings in the fix for RO-2438.

Issue: [RO-2428](https://rpc-openstack.atlassian.net/browse/RO-2428)

### DIFF
--- a/rpcd/etc/openstack_deploy/user_osa_variables_defaults.yml
+++ b/rpcd/etc/openstack_deploy/user_osa_variables_defaults.yml
@@ -39,3 +39,8 @@ octavia_git_repo: 'https://git.openstack.org/openstack/octavia'
 octavia_git_project_group: 'octavia_all'
 octavia_git_install_fragments: "venvwithindex=True&ignorerequirements=True"
 
+# RabbitMQ
+rabbitmq_install_method: "external_repo"
+rabbitmq_repo: "{{ rpco_apt_repo }}"
+rabbitmq_erlang_repo: "{{ rpco_apt_repo }}"
+rabbitmq_gpg_keys: "{{ rpco_apt_gpg_keys }}"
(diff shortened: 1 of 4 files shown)


odyssey4me:
On its own, bringing in the SHA won't make a difference. This PR will also require an adjustment to the artifact repo configuration which tells aptly to mirror the erlang repo (seeing as it's a new repo) and to include it in the snapshot.

major:
Thanks for that, @odyssey4me. I'll try to get this fixed up.

rpcd/etc/openstack_deploy/user_osa_variables_defaults.yml
```diff
@@ -39,3 +39,8 @@ octavia_git_repo: 'https://git.openstack.org/openstack/octavia'
 octavia_git_project_group: 'octavia_all'
 octavia_git_install_fragments: "venvwithindex=True&ignorerequirements=True"
 
+# RabbitMQ
```
odyssey4me: Wrong file - this stuff now goes into https://github.com/rcbops/rpc-openstack/blob/newton/group_vars/all/apt.yml with the rest of the apt repo configs.

scripts/artifacts-building/apt/aptly-vars.yml
```diff
@@ -281,6 +282,25 @@ aptly_mirrors:
     update_flags:
       - "-keyring='trustedkeys.gpg'"
     do_update: "{{ aptly_mirror_do_updates }}"
+  # Erlang (used by RabbitMQ):
+  # Got the key from:
+  # wget https://packages.erlang-solutions.com/ubuntu/erlang_solutions.asc
+  # gpg --list-packets erlang_solutions.asc
+  - name: erlang-xenial
+    archive_url: https://packages.erlang-solutions.com/ubuntu
+    distribution: xenial
```
odyssey4me: Need to add one for trusty too, as Newton is both trusty and xenial.

major:
#2536 takes care of this one now.

PR #2544, opened 2017-09-21
Make pre upgrade tasks kilo compatible

BjoernT:
In a leapfrog case, the pre upgrade tasks are
executed against a Kilo environment.
The tasks are now compatible with the older
python-openstack and neutron client.

This is a back port of 274d9d98d0e8903c5a1c48bdf955eb5f689d684f

Issue: [RLM-64](https://rpc-openstack.atlassian.net/browse/RLM-64)


cloudnull:
recheck_cit_artifact

cloudnull:

claco:
Passed required. Backport from master (which is why I also asked @major @prometheanfire).

PR #2548, opened 2017-09-26
RLM-207 Correct influxdb/grafana repo overrides

odyssey4me:
The data structure used in rpc-maas is different to
all the OSA repositories, so the standard data in
'rpco_apt_repo' does not work.

In this patch we correct the overrides to ensure
that the right data structure is provided to make
use of the apt artifacts when executing the rpc-maas
playbooks.

(cherry picked from commit 88ffdeacfe2c54878f8f8481fd48a1d97acd9031)

Issue: [RLM-207](https://rpc-openstack.atlassian.net/browse/RLM-207)

### DIFF
--- a/rpcd/etc/openstack_deploy/user_rpco_variables_defaults.yml
+++ b/rpcd/etc/openstack_deploy/user_rpco_variables_defaults.yml
@@ -74,13 +74,19 @@ maas_keys:
   state: "present"
 
 # Influxdata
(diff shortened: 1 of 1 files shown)


odyssey4me:
recheck_cit_xenial_deploy_swift - tempest failures

odyssey4me:
recheck_cit_xenial_deploy_swift - previous recheck didn't happen

PR #2559, opened 2017-09-29
RO-2469: Bumping sha to top of rcp-ops newton - cherry pick from newton

xgerman:
JIRA link: https://rpc-openstack.atlassian.net/browse/RO-2469 -- the current sha will overwrite the arp-utils so Floating IPs won't work on a new install of rpc-o. This sha is taking top of the openstack-ops package per  @BjoernT  suggestion

Issue: [RO-2469](https://rpc-openstack.atlassian.net/browse/RO-2469)

PR #2560, opened 2017-09-29
 RE-2451 Clone openstack-ansible in ansible-lint

mattt416:
In #2483, we drop the use of submodules but did not update tox.ini to
 handle the clone of openstack-ansible.  The reason this passed the lint
 check on #2483 is because it first checked out master, did a submodule
 update, then checked out that commit and proceeded to run tests.  At
 that point, the submodules had already been cloned and therefore
 ansible-lint worked as expected.

 This commit simply updates ansible-lint to clone openstack-ansible,
 check out the correct commit, and then proceed as it did before.

 Note that we also update how we handle submodules in .travis.yml.  This
 isn't an issue in master for now (since we dropped submodules), but it
 currently affects other branches since we checkout a differing commit
 and then never update submodules.

(cherry picked from commit b8a749ab99962c40a1b129921c30a8d695aa76a0)

Issue: [RE-2451](https://rpc-openstack.atlassian.net/browse/RE-2451)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,9 +11,10 @@ before_script:
 # the shallow clone that travis creates, then clones
 # the full repository
 install:
-  - git clone --recursive https://github.com/rcbops/rpc-openstack.git rcbops/rpc-openstack
+  - git clone https://github.com/rcbops/rpc-openstack.git rcbops/rpc-openstack
   - cd rcbops/rpc-openstack
   - "if [[ $TRAVIS_PULL_REQUEST != 'false' ]]; then git fetch origin +refs/pull/$TRAVIS_PULL_REQUEST/merge && git checkout -qf FETCH_HEAD; else git checkout -qf $TRAVIS_COMMIT; fi"
+  - git submodule update --init
 
 script: tox
 


mattt416:
@mancdaz @odyssey4me @hughsaunders @hughsaunders @evrardjp I believe this will need to be backported into newton-rc in order for https://github.com/rcbops/rpc-openstack/pull/2559 to pass the travis test.  I don't know if we want to merge that one with a broken travis check or if we want to merge this, then rebase https://github.com/rcbops/rpc-openstack/pull/2559.  Can one of you have a look at this on Monday before we create a release?

Thanks!

evrardjp:
I am very confused. In travis we start clean everytime.
I don't see why all of this is needed.

For newton, the previous ``git clone --recursive https://github.com/rcbops/rpc-openstack.git rcbops/rpc-openstack`` should be enough to have all the code, and then we can apply the proper patch. What is lacking in my understanding?

For branches above, shouldn't travis already take care of checking out the proper reference?

odyssey4me:
@evrardjp As ``master`` is the default branch, the ``git clone --recursive`` clones the branch but has no submodule to init/update. This patch makes it clone, then checkout to newton, then init/update the submodule.

evrardjp:
@odyssey4me still don't understand how master is the default branch.
Travis should check out your version, that's it.

odyssey4me:
@evrardjp In the sequence of commands given to travis to do, it appears to me that we're telling travis how to do the checkout. I don't know much about how travis works, but given that set of commands this change makes sense.

evrardjp:
@odyssey4me Don't get me wrong, I agree with you: the code doesn't seem wrong.

I just wonder WHY the code happens in the first place. This becomes very obscure as no comment is provided, and IMO, travis should be smart. 

On top of that, it worked before on that branch, so why do we need to change it? -> That's where I need to find an explanation. The explanation of the backport and what happened in master branch doesn't hold because .travis.yml is in tree and should therefore be properly checked out already. No link there.

evrardjp:
We should use ``git fetch --unshallow`` for the reno, and I'll create a card for it.

PR #2564, opened 2017-10-02
RE-197 Add update_dependencies hook

hughsaunders:
This script keeps the OSA sha tracking OSA stable/newton.

Issue: [RE-197](https://rpc-openstack.atlassian.net/browse/RE-197)

### DIFF
--- a/gating/update_dependencies/run
+++ b/gating/update_dependencies/run
@@ -1,0 +1,70 @@
+#!/bin/bash -xe
+
+# Update OSA SHA to head of stable/ocata
+
+# These var must be set per branch of RPC-Openstack
+osa_branch=stable/ocata
+rpco_branch=${BRANCH:-master} # BRANCH injected by Jenkins.
+
+# Env vars injected by Jenkins:
+WORKSPACE="${WORKSPACE:-$PWD}"
+
+# Note for local testing: this script expects to be executed from the root
+# of an rpco clone, checkout at master.
+
+# Get current head of stable/ocata
+git clone "https://github.com/openstack/openstack-ansible" ${WORKSPACE}/oa
+pushd ${WORKSPACE}/oa
+  git checkout $osa_branch
+  osa_sha="$(git log -n 1 --format=%H)"
+popd
+
+# Insert current SHA into functions.sh
+sed -i '/OSA_RELEASE:-/ c\
+export OSA_RELEASE=${OSA_RELEASE:-"'${osa_sha}'"}' scripts/functions.sh
+
+
+# Update rpc_release. The mainline branch should always be one version
+# ahead of the rc branch, as the next rc branch will be cut from mainline
+# post release.
+
+# This update script will only ever run against a mainline branch, as
+# dependencies should not change once an RC branch has been cut.
+
+# Get RC branch version
+rc_branch="${rpco_branch}-rc"
+git fetch origin
+git show origin/${rc_branch} || {
+  echo "RC branch ${rc_branch} not found, skipping rpc_release version bump.
+If there is no RC branch then the mainline branch is considered unreleased
+and therefore the rpc_release value is left alone. It is still important
+for the dependencies to be updated regularly though, so that part continues
+to be done."
+  exit 0
+}
+
+# can't use derive-artifact-version.sh as that hardcodes the rpc repo path
+extract_rpc_release(){
+  awk '/rpc_release/{print $2}' | tr -d '"'
+}
+
+rc_branch_version="$(git show origin/${rc_branch}:group_vars/all/release.yml \
+                    | extract_rpc_release)"
+echo "rpc_release version from rc-branch (${rc_branch}): ${rc_branch_version}"
+
+current_branch_version="$(extract_rpc_release < group_vars/all/release.yml)"
+echo "rpc_release version from current branch (${rpco_branch}): ${current_branch_version}"
+
+# Extract the required version info
+major_version=$( echo ${rc_branch_version} | cut -d. -f1 )
+minor_version=$( echo ${rc_branch_version} | cut -d. -f2 )
+patch_version=$( echo ${rc_branch_version} | cut -d. -f3 )
+
+# increment the minor version
+minor_version=$(( minor_version + 1 ))
+
+incremented_version="${major_version}.${minor_version}.${patch_version}"
+echo "Incremented rpc_release version: ${incremented_version}"
+
+sed -i "s/${current_branch_version}/${incremented_version}/" \
+  group_vars/all/release.yml


hughsaunders:

odyssey4me:
@git-harry The purpose here is just to implement a hook to demonstrate it. It can be extended from there. If hook implementations across multiple repositories turn out to be very similar we can consider implementing a facility to make it easier, but for now that's unnecessary. The purpose of this patch is to get something simple in which the team can iterate.

hughsaunders:
I think this particular script demonstrates the need for repos to have their own mechanisms for updating deps, who would have thought we'd need to sed a shell script to bump a version?

hughsaunders:
Example Job: https://rpc.jenkins.cit.rackspace.net/job/Dep-Update_rpc-openstack-newton/18/console ✅ 
Created Jira Issue: https://rpc-openstack.atlassian.net/browse/RE-656
Created PR: https://github.com/rcbops/rpc-openstack/pull/2568

hughsaunders:
@evrardjp documentation: https://rpc-openstack.atlassian.net/wiki/spaces/RE/pages/19005457/RE+for+Projects

evrardjp:
@hughsaunders I know this documentation, they may have forgotten it. Because it's in this review, I think we are good.

evrardjp:
Oh no.

hughsaunders:
@evrardjp you make a good point, about the branch conflict. For master this should be stable/ocata.  I'll change that.

gating/update_dependencies/run
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash -xe
+
+# Update OSA SHA to head of stable/ocata
+
+# Get current head of stable/ocata
+git clone "https://github.com/openstack/openstack-ansible" ${WORKSPACE}/oa
+pushd ${WORKSPACE}/oa
+  git checkout stable/ocata
+  CURRENT_SHA="$(git log -n 1 --format=%H)"
+popd
```
odyssey4me: This is also going to need to update the value of rpc_release if the current value matches the latest tag on this branch.

evrardjp:
The ideal would be to trigger functions from the scripts/functions, but I think editing it and using it at the same time looks like a bad idea.

odyssey4me:
@evrardjp I don't understand what you mean by the previous comment. Can you please clarify?

evrardjp:
@odyssey4me Here we are basically adding one more source of truth for the shas: The branch name is stored, in hard, here.

The ideal would be, IMO, to have one source of all those, and have convenience tool that bumps all that. I'd say that vars/functions already have functions for the parsing of RPC_VERSION (https://github.com/rcbops/rpc-openstack/blob/master/scripts/functions.sh#L60) , and it has vars. So I'd say, our script should probably source the scripts/functions , and apply what's needed.

However, I don't think it's smart to source scripts/functions, EDIT it, and use it at the same time.

Also I don't think it's our choice, whatever the "hat" we have (good citizen hat isn't a good citizen if we are abstracting work and RPC-O team don't realise we've added a new interface).

So I guess the sooner we give responsibility to RPC-O team the better, and this commit can therefore go ahead (maybe we need to create a card to tidy up this process for clarity).

gating/update_dependencies/run
odyssey4me: Something like this:

```
RPC_RELEASE="$(/opt/rpc-openstack/scripts/artifacts-building/derive-artifact-version.sh)"
LATEST_RPC_RELEASE="$(git describe --abbrev=0 --tags)"

if [[ ${RPC_RELEASE} == ${LATEST_RPC_RELEASE} ]]; then
      echo "Updating the release version..."

      # Extract the required version info
      major_version=$( echo ${currentversion} | cut -d. -f1 )
      minor_version=$( echo ${currentversion} | cut -d. -f2 )
      patch_version=$( echo ${currentversion} | cut -d. -f3 )

      # increment the minor version
      minor_version=$(( minor_version + 1 ))

      sed -i "s/${RPC_RELEASE}/${major_version}.${minor_version}.${patch_version}/" group_vars/all/release.yml
fi
```

evrardjp: The code seem ok to me.

odyssey4me:
@evrardjp Ah, I see what you mean. Yes, I do think there are perhaps smarter ways of implementing this hook, but the sooner we can put the implementation into the hands of the product owners, the better.

evrardjp:
@odyssey4me We are definitely in sync then.

gating/update_dependencies/run
git-harry: This doesn't sound like a dependency update, this sounds like it should be part of the release hook.

odyssey4me: I just realised that the latest tag would be two versions behind the development branch:

```
newton -> rpc_release = 14.2.0
newton-rc -> rpc_release = 14.1.0
latest tag -> rpc_release = 14.0.0
```

This means that the previously suggested script won't work. We'd have to read the rpc_release version from the rc branch. :/

@git-harry You're probably right, except that this is an in-repo change which requires a PR. It's a quirk of this particular repository so one could argue that it better fits being done in this hook. However I would not object to some sort of post-release job proposing some sort of change to the repo in some kind of post release hook.

gating/update_dependencies/run
```diff
@@ -0,0 +1,49 @@
...
+
+# Update rpc_release. The mainline branch should always be one version
+# ahead of the rc branch, as the next rc branch will be cut from mainline
+# post release.
+
+# This update script will only ever run against a mainine branch, as
+# dependencies should not change once an RC branch has been cut.
+
+# Get RC branch version
+rc_branch="${BRANCH}-rc"
```
odyssey4me: Where is ``BRANCH`` defined?

gating/update_dependencies/run
```diff
@@ -0,0 +1,53 @@
...
+git fetch origin
+git show origin/${rc_branch} || {
+  echo "rc-branch ${rc_branch} not found, aborting."
+  exit 1
```
odyssey4me: Perhaps instead of exiting, we just assume that we do not need to edit the release version because there has never been a release? Otherwise the job will fail and the SHA update will never be done.

hughsaunders: Done

gating/update_dependencies/run
hughsaunders: Added comments and defaults for env vars

hughsaunders:
Updated to also bump rpc_release version number. 

Test job: https://rpc.jenkins.cit.rackspace.net/job/Dep-Update_rpc-openstack-newton/24/ ✅ 
Created Jira Issue: https://rpc-openstack.atlassian.net/browse/RE-662
Created PR: https://github.com/rcbops/rpc-openstack/pull/2570

gating/update_dependencies/run
```diff
@@ -0,0 +1,66 @@
...
+# Insert current SHA into functions.sh
+sed -i '/OSA_RELEASE:-/ c\
+export OSA_RELEASE=${OSA_RELEASE:-"'${osa_sha}'"}' scripts/functions.sh
+
+
+# Update rpc_release. The mainline branch should always be one version
+# ahead of the rc branch, as the next rc branch will be cut from mainline
+# post release.
+
+# This update script will only ever run against a mainine branch, as
```
odyssey4me: Spelling error ``mainline``

gating/update_dependencies/run
```diff
@@ -0,0 +1,66 @@
...
+# dependencies should not change once an RC branch has been cut.
+
+# Get RC branch version
+rc_branch="${rpco_branch}-rc"
+git fetch origin
+git show origin/${rc_branch} || {
+  echo "Rc-branch ${rc_branch} not found, skipping rpc_release version bump."
```
odyssey4me: nit:
``RC branch`` instead of ``Rc-branch``

gating/update_dependencies/run
hughsaunders: Done

gating/update_dependencies/run
hughsaunders: Done

gating/update_dependencies/run
```diff
@@ -0,0 +1,66 @@
...
+
+# This update script will only ever run against a mainline branch, as
+# dependencies should not change once an RC branch has been cut.
+
+# Get RC branch version
+rc_branch="${rpco_branch}-rc"
+git fetch origin
+git show origin/${rc_branch} || {
+  echo "RR branch ${rc_branch} not found, skipping rpc_release version bump."
+  exit 0
```
odyssey4me: It looks like my other comment got missed:

It'd be best if we also document why the release is not updated if there is no -rc branch. Something long the lines of:

```
If there is no RC branch then the mainline branch is considered unreleased and therefore the ``rpc_release`` value is left alone. It is still important for the dependencies to be updated
regularly though, so that part continues to be done.
```

gating/update_dependencies/run
```diff
@@ -0,0 +1,66 @@
...
+# post release.
+
+# This update script will only ever run against a mainline branch, as
+# dependencies should not change once an RC branch has been cut.
+
+# Get RC branch version
+rc_branch="${rpco_branch}-rc"
+git fetch origin
+git show origin/${rc_branch} || {
+  echo "RR branch ${rc_branch} not found, skipping rpc_release version bump."
```
odyssey4me: s/RR/RC/

hughsaunders: done

gating/update_dependencies/run
hughsaunders: done

gating/update_dependencies/run
```diff
@@ -0,0 +1,70 @@
...
+
+# Get RC branch version
+rc_branch="${rpco_branch}-rc"
+git fetch origin
+git show origin/${rc_branch} || {
+  echo "RC branch ${rc_branch} not found, skipping rpc_release version bump.
+If there is no RC branch then the mainline branch is considered unreleased
+and therefore the rpc_release value is left alone. It is still important
+for the dependencies to be updated regularly though, so that part continues
+to be done."
```
odyssey4me: LOL, I meant that to be a comment in between 27-32 somewhere. :/ But I guess this works too.

hughsaunders: 😂

gating/update_dependencies/run
```diff
@@ -0,0 +1,70 @@
...
+# Update OSA SHA to head of stable/ocata
+
+# These var must be set per branch of RPC-Openstack
+osa_branch=stable/ocata
+rpco_branch=${BRANCH:-master} # BRANCH injected by Jenkins.
+
+# Env vars injected by Jenkins:
+WORKSPACE="${WORKSPACE:-$PWD}"
+
+# Note for local testing: this script expects to be excecuted from the root
```
odyssey4me: s/excecuted/executed/

hughsaunders: Done

PR #2567, opened 2017-10-04
Make artifacting togglable

prometheanfire:
Adds a new variable 'DEPLOY_ARTIFACTING', defaulting to 'yes', which
is used to specify if artifacting is to be used in the deployment.

### DIFF
--- a/scripts/functions.sh
+++ b/scripts/functions.sh
@@ -39,6 +39,7 @@ export DEPLOY_SWIFT=${DEPLOY_SWIFT:-"yes"}
 export DEPLOY_HARDENING=${DEPLOY_HARDENING:-"yes"}
 export DEPLOY_RPC=${DEPLOY_RPC:-"yes"}
 export DEPLOY_ARA=${DEPLOY_ARA:-"no"}
+export DEPLOY_ARTIFACTING=${DEPLOY_ARTIFACTING:-"yes"}
 export DEPLOY_SUPPORT_ROLE=${DEPLOY_SUPPORT_ROLE:-"no"}
 export BOOTSTRAP_OPTS=${BOOTSTRAP_OPTS:-""}
 export UNAUTHENTICATED_APT=${UNAUTHENTICATED_APT:-no}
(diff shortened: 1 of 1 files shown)


prometheanfire:
recheck cit_xenial_deploy_swift

prometheanfire:

prometheanfire:

odyssey4me:
This only works for an AIO because the functions are only called for gate tests in AIO's. If you want this to be useful for MNAIO tests then the MNAIO config prep process will need to make use of the same functions, just as the AIO config prep does. If this is needed for physical labs, then the prep process for those labs will need to action the same user_/group_var file removal being done in the AIO config prep process. TL/DR if this is intended for use outside of AIO's then it is incomplete.

prometheanfire:
This doesn't circumvent anything, it defaults to 'true'.  As far as the MNAIO tests, it sounds like they don't represent reality.

Also, isn't deploy.sh supposed to be used to deploy?

prometheanfire:

prometheanfire:

prometheanfire:
sure, I just disagree with your reasoning.  As far as I know we still state that deployment should use deploy.sh still.

odyssey4me:
@prometheanfire ``deploy.sh`` is used, sure, but this is implementing a way to circumvent the use of artifacts when using ``deploy.sh`` and my suggestion is that it's counter-productive to do so. 

My first item of feedback, however, is that this patch does not work for anything beyond an AIO. The patch does not make it clear what the purpose of this toggle is other than to disable the use of artifacts... and if it's general purpose, it is incomplete.

odyssey4me:
@major The place to do it is in the preparation of the configuration, as is already done for the AIO. Disabling the use of artifacts is already done in the AIO config preparation when they are not available. If this is needed for MNAIO's too then it would have to be baked into the MNAIO config prep. If it's needed for production environments then it would require documentation to describe how to do it.

prometheanfire:
is the mnaio prep done in rpc-o?

prometheanfire:

prometheanfire:

odyssey4me:
It looks like this is failing on this known issue: https://rpc-openstack.atlassian.net/browse/RO-2781

prometheanfire:

antonym:

prometheanfire:
Updating on my previous comment, this change is not for our gating (methods that are not used by deployers).  The target of this PR is deployer used scripts (aka deploy.sh).

cloudnull:
change is simple enough and doesn't change the default behavior. LGTM

PR #2573, opened 2017-10-05
Bump OSA SHA to include os_ironic bump

major:
This patch bumps the OpenStack-Ansible SHA to include an upstream
SHA bump for os_ironic:

    b0579865 Update SHA for os_ironic


major:
Can't install the right packages:

```
"The following packages have unmet dependencies:", " build-essential : Depends: g++ (>= 4:5.2) but it is not going to be installed"
```

odyssey4me:
recheck_cit_artifact

PR #2576, opened 2017-10-05
RLM-219 Reduce memory usage in AIOs

antonym:
This adds a few optimizations for Galera and reduces the number of threads and workers in AIOs to help alleviate memory pressure and allow failing Tempest tests to pass.

Will backport this to Newton branch as well.

Issue: [RLM-219](https://rpc-openstack.atlassian.net/browse/RLM-219)

### DIFF
--- a/scripts/bootstrap-aio.yml
+++ b/scripts/bootstrap-aio.yml
@@ -1,5 +1,5 @@
 ---
-# Copyright 2015, Rackspace US, Inc.
+# Copyright 2017, Rackspace US, Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


claco:
@odyssey4me So what's your suggestion then? That we use the upstream file instead for rpc-o gating, or that we just target this patch at Newton directly, or both?

For the moment, I assume the goal is to do what we need to do in Newton for upgrades for RPC-O.

For later, I'm viewing this from the lens that Pike is OSA directly, with some RPC things installed on top (like RPC-MaaS), but not in the way we have them tied together now, and not as two separate products (RPC-O & OSA). @cloudnull RPC modernization RFC.

antonym:
@odyssey4me I pulled these from upstream master here which seems to have been a bit more optimized for less memory usage:

https://github.com/openstack/openstack-ansible/blob/master/tests/roles/bootstrap-host/templates/user_variables.aio.yml.j2#L23-L137

I can trim them down the variables that don't exist for Ocata and Newton if that helps but we wouldn't benefit from them unless we were running the latest OSA.

odyssey4me:
@antonym Most of those vars only pertain to the use of uwsgi to run the services, which only applies to Pike onwards. You would have to work through them all. Still - the point is that these are all actually implemented in OSA already. If you find any from master that apply to the older branches, I'd recommend submitting a review to add it upstream.

antonym:
@odyssey4me Most of the settings are reducing upstreams worker from 2 to 1 and making some tweaks to Galera settings.  If that won't break any upstream gating and that's acceptable, I can push the review upstream instead.  I wasn't sure if there's a dependency upstream to test at least two workers in the gate or not.  It would probably delay us getting gating passing until we'd bumped to the latest OSA that included those for newton.

Should we do both upstream and our repo and then pull out the settings once we're cut over to the upstream settings?

antonym:
Closing this out since really we're needing the changes for Newton and don't really need this for the Ocata+ since it's already upstream.

PR #2581, opened 2017-10-06
RE-674 Downgrade pkgs to apt artifact versions [WIP]

odyssey4me:
When using images which have unattended upgrades
enabled, or have been built using packages which
are more recent than the available apt artifacts
the builds fail in strange ways.

This patch implements a function which will try
to downgrade any packages installed on the
deployment host and not available in the apt
artifacts to the versions which are.

Issue: [RE-674](https://rpc-openstack.atlassian.net/browse/RE-674)


odyssey4me:
As a workaround for now, https://github.com/rcbops/rpc-gating/pull/458 has been merged. A better long term solution is being implemented in https://github.com/rcbops/rpc-gating/pull/460 so this is not necessary.

PR #2595, opened 2017-10-13
RO-2739 Bump OSA newton SHA to include dns_domain fix

johnsom:
This patch increments the OSA SHA to the current newton head which
includes a neutron role fix for the dns_domain setting:
    7267d21 Update os_neutron role SHA to include dns_domain


odyssey4me:
recheck_trusty_swift_deploy

PR #2599, opened 2017-10-16
RLM-54 Exclude ES re-deploy from post leap stage

Frank-ZhangXin:
Our goal is to make all ES logs are kept and barely changed in leap.
So we are supposed to separate Elasticsearch upgrade from whole leap
ugprade for now, because it needs pre-requisite operations both on ES
container and logstash container. For ease sake, we make ES ugprade into
post stage of leapfrog upgrade. And next to-do is integrating these step
back to leapfrog post ugprade script again.

Issue: [RE-621](https://rpc-openstack.atlassian.net/browse/RE-621)


Frank-ZhangXin:
Base is wrong, closed.

PR #2601, opened 2017-10-17
RE-360 Include Reno notes in release notes

git-harry:
This changes updates the release notes that will be generated by the
rpc-gating generate_release_notes hook to include those managed by Reno.

The result of this change is that the release notes now returned will be
any Reno notes followed by those produced by rpc-differ. If there are no
Reno notes for the release that will be explicitly stated in the release
notes.

The code was restructured so that the code for generating the different
types of notes could be kept separate. A wrapper script is used to
create a single set of notes from the separate files.

(cherry picked from commit 773413d94a6df65576c27fe8f68733e43f4c4b8a)

Issue: [RE-360](https://rpc-openstack.atlassian.net/browse/RE-360)

### DIFF
--- a/gating/generate_release_notes/release_notes_dockerfile
+++ b/gating/generate_release_notes/release_notes_dockerfile
@@ -9,9 +9,11 @@ RUN echo "jenkins ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers
 RUN apt-get install -y pandoc
 
 # use released versions of the differs when they are available.
-RUN pip install rpc_differ==0.3.0 reno==2.5.0
+RUN pip install rpc_differ==0.3.0 reno==2.5.1
 RUN pip install git+https://github.com/major/osa_differ@0.3.1
 
 RUN rpc-differ --debug --update master master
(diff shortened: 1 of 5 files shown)


gating/generate_release_notes/release_notes_dockerfile
```diff
@@ -9,9 +9,11 @@ RUN echo "jenkins ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers
 RUN apt-get install -y pandoc
 
 # use released versions of the differs when they are available.
-RUN pip install rpc_differ==0.3.0 reno==2.5.0
+RUN pip install rpc_differ==0.3.0 reno==2.5.1
 RUN pip install git+https://github.com/major/osa_differ@0.3.1
```
hughsaunders: Should we bump this to 0.3.2?

git-harry: That's a separate piece of work - https://rpc-openstack.atlassian.net/browse/RE-622

PR #2602, opened 2017-10-17
RE-360 Include Reno notes in release notes

git-harry:
This changes updates the release notes that will be generated by the
rpc-gating generate_release_notes hook to include those managed by Reno.

The result of this change is that the release notes now returned will be
any Reno notes followed by those produced by rpc-differ. If there are no
Reno notes for the release that will be explicitly stated in the release
notes.

The code was restructured so that the code for generating the different
types of notes could be kept separate. A wrapper script is used to
create a single set of notes from the separate files.

(cherry picked from commit 773413d94a6df65576c27fe8f68733e43f4c4b8a)

Issue: [RE-360](https://rpc-openstack.atlassian.net/browse/RE-360)


git-harry:
recheck_cit_trusty_swift_deploy

git-harry:
recheck_cit_trusty_swift_deploy

PR #2605, opened 2017-10-18
RE-622 Update osa-differ to fix rpc-differ links

git-harry:
Updating osa-differ ensures that the report, produced by rpc-differ,
includes valid links when a GitHub URL, used by the tool, ends with
`".git"`.

(cherry picked from commit f67e1160c037a22519d7b60d6537d0e3a530d463)

Issue: [RE-622](https://rpc-openstack.atlassian.net/browse/RE-622)

### DIFF
--- a/gating/generate_release_notes/release_notes_dockerfile
+++ b/gating/generate_release_notes/release_notes_dockerfile
@@ -8,9 +8,8 @@ RUN echo "jenkins ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers
 
 RUN apt-get install -y pandoc
 
-# use released versions of the differs when they are available.
 RUN pip install rpc_differ==0.3.0 reno==2.5.1
-RUN pip install git+https://github.com/major/osa_differ@0.3.1
+RUN pip install git+https://github.com/major/osa_differ@0.3.2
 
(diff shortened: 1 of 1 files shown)


gating/generate_release_notes/release_notes_dockerfile
```diff
@@ -8,9 +8,8 @@ RUN echo "jenkins ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers
 
 RUN apt-get install -y pandoc
 
-# use released versions of the differs when they are available.
 RUN pip install rpc_differ==0.3.0 reno==2.5.1
-RUN pip install git+https://github.com/major/osa_differ@0.3.1
+RUN pip install git+https://github.com/major/osa_differ@0.3.2
```
major: Could we use pypi for this rather than GitHub? https://pypi.python.org/pypi/osa_differ

git-harry: Maybe but that's not something I think should be addressed here. This patch is working code that is a back port of something already merged in both master and newton.

Is there a material downside with leaving it how it is? @hughsaunders may be able to explain why it was done that way. If this needs changing to install from PyPi it should be done as a separate commit.

PR #2606, opened 2017-10-18
Use the git-clone fetch mode instead of galaxy

cloudnull:
This change modifies the fetch mode so that we're keeping the git
history within roles when doing a deployment. This makes it possible to
know which role we're using from the perspective of a checkout and
provides operators the ability to locally fetch refs without having to
re-bootstrap ansible.


odyssey4me:
If this merges, don't forget to also do this in master.

PR #2608, opened 2017-10-19
Ensure our repo cache is enabled

cloudnull:
The repo cache in OSA creates a mesh within a deployment so that all
package installations come from our repo-server infrastructure. This
allows us to limit our download traffic to our apt mirrors and ensures
fast package installation from a local proxy. This change will improve
the overall deployment experience with or without our package mirrors.

### DIFF
--- a/group_vars/all/apt.yml
+++ b/group_vars/all/apt.yml
@@ -39,12 +39,6 @@ rpco_gpg_key_id: 52AA252F #SET IN STATIC (to force key verification per release)
 lxc_container_cache_files_from_host:
   - "/etc/apt/sources.list.d/{{ rpco_mirror_apt_filename }}.list"
 
-# We won't be using the repo package cache, as we'll have a full mirror.
-# TODO (odyssey4me):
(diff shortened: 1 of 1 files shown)


BjoernT:
@odyssey4me is it ok to use it for r14.2.0, is the mirror now complete ?

odyssey4me:
@BjoernT I don't think I understand what you're asking.

The apt repo used by RPC has been complete since r14.1.0. This patch changes the default to activate the repo cache in the repo container, but still uses the same apt repo.

If you're asking whether it's safe to use an override to enable the cache in any newton RPC-O releases, then the answer is yes.

Does that answer your question?

BjoernT:
Yes that was the question because I noticed that this was not in place after leapfrog

odyssey4me:
@BjoernT What is ``this`` which was not in place after the leapfrog?

BjoernT:
The availability of apt cacher NG. The vip is down and apt seems not configured to use the cache in the first place on the host. The vip is called RPC_POOL_REPO_CACHE.

PR #2617, opened 2017-10-24
Make artifacting togglable (#2567)

prometheanfire:
Adds a new variable 'DEPLOY_ARTIFACTING', defaulting to 'yes', which
is used to specify if artifacting is to be used in the deployment.

(cherry picked from commit df90d89f282471a4270eb836d6bb6248256b405a)


prometheanfire:
RE-694 is nice, but doesn't allow artifacting to be disabled

PR #2622, opened 2017-10-26
RO-2938

cloudnull:
The index-system items need to all have a trailing slash otherwise the
lxc-container-create, lxc-host, and the lxc-create command will not
parse the URL correctly. This change adds the trailing slash to all
itmes as they're being added to the index list.

Issue: [RO-2938](https://rpc-openstack.atlassian.net/browse/RO-2938)


odyssey4me:
I've ported this back to newton: https://github.com/rcbops/rpc-openstack/pull/2624

PR #2624, opened 2017-10-30
RO-2938 (#2622)

odyssey4me:
The index-system items need to all have a trailing slash otherwise the
lxc-container-create, lxc-host, and the lxc-create command will not
parse the URL correctly. This change adds the trailing slash to all
itmes as they're being added to the index list.

(cherry picked from commit 5d6ac040bc0e3d21b6c41f174da2676ad42637fa)

Issue: [RO-2938](https://rpc-openstack.atlassian.net/browse/RO-2938)

### DIFF
--- a/scripts/artifacts-building/containers/artifact-build-chroot.yml
+++ b/scripts/artifacts-building/containers/artifact-build-chroot.yml
@@ -155,7 +155,7 @@
     - name: Create release index entries
       lineinfile:
         dest: "{{ built_container_artifact_metadata_file }}"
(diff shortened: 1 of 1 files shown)


antonym:
recheck_cit_artifact

odyssey4me:
Setting this to do not merge given the release is to be implemented tomorrow and there won't be time to figure out any issues caused by new artifacts.

odyssey4me:
Oh actually, the release comes from the newton-rc branch so this should not affect the release at all.

PR #2625, opened 2017-10-30
RE-694 Replace artifacts if it is safe to do so

odyssey4me:
Currently the periodic job which builds artifacts is not
set to replace the artifacts every time it runs. Instead
of setting that on, therefore always replacing them, we
use a function to verify that the rpc_release version
set in the mainline branch differs from the value set in
the release candidate branch.

This protection is implemented to prevent artifacts from
being replaced by mistake once a new release candidate
branch has been cut, but the mainline branch has not had
the rpc_release value changed to the next version yet.

Using this method still allows the periodic job to be
executed manually with REPLACE_ARTIFACTS="YES", forcing
them to be replaced.

We only implement this check when PUSH_TO_MIRROR is already
set to "YES" when the script executes. This is only the
case in the periodic jobs which are the ones designed to
be run in sequence to to upload their changes.

Issue: [RE-694](https://rpc-openstack.atlassian.net/browse/RE-694)

### DIFF
--- a/scripts/artifacts-building/apt/build-apt-artifacts.sh
+++ b/scripts/artifacts-building/apt/build-apt-artifacts.sh
@@ -74,9 +74,36 @@ fi
 # from previously installed/configured items.
 rm -rf /etc/ansible /etc/openstack_deploy /usr/local/bin/ansible* /usr/local/bin/openstack-ansible*
 
-# Install Ansible
-./scripts/bootstrap-ansible.sh
+# Bootstrap Ansible
+# This script is sourced to ensure that the common
+# functions and vars are available.
+source scripts/bootstrap-ansible.sh
 cp scripts/artifacts-building/apt/lookup/* /etc/ansible/roles/plugins/lookup/
+
+# Figure out when it is safe to automatically replace artifacts
+if [[ "$(echo ${PUSH_TO_MIRROR} | tr [a-z] [A-Z])" == "YES" ]]; then
+
+  if apt_artifacts_available; then
+    # If there are artifacts for this release already, and it is not
+    # safe to replace them, then set PUSH_TO_MIRROR to NO to prevent
+    # them from being overwritten.
+    if ! safe_to_replace_artifacts; then
+      export PUSH_TO_MIRROR="NO"
+
+    # If there are artifacts for this release already, and it is safe
+    # to replace them, then set REPLACE_ARTIFACTS to YES to ensure
+    # that they do get replaced.
+    else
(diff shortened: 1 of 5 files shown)


scripts/functions.sh
```diff
@@ -187,3 +187,31 @@ function configure_apt_sources {
   curl --silent --fail ${HOST_RCBOPS_REPO}/apt-mirror/rcbops-release-signing-key.asc | apt-key add -
 
 }
+
+function safe_to_replace_artifacts {
+
+  # The mainline branch should always be one version ahead of the rc branch,
+  # as the next rc branch will be cut from mainline. This function checks
+  # the current branch against the rc branch and returns 1 if it is the same
```
evrardjp: First phrase is useful. 

"This function checks the current branch against the rc branch and returns 1 if it is the same" is basically the code, so I doubt it's useful.

I'd replace this part of the comment with:
"A return of 0 means it's safe to replace artifacts. This should happen when there is no rc branch, or the versions in the two branches differ, to avoid overlaps."

odyssey4me:
recheck_cit_artifact

The python/container builds both failed when trying to set git up:

```
+ git config --global user.email rpc-jenkins-svc@github.com
error: could not lock config file /var/lib/jenkins/.gitconfig: File exists
```

scripts/artifacts-building/apt/build-apt-artifacts.sh
```diff
@@ -74,10 +74,37 @@ fi
...
 
-# Install Ansible
-./scripts/bootstrap-ansible.sh
+# Bootstrap Ansible
+# This script is sourced to ensure that the common
+# functions and vars are available.
+source scripts/bootstrap-ansible.sh
 cp scripts/artifacts-building/apt/lookup/* /etc/ansible/roles/plugins/lookup/
 
+# Figure out when it is safe to automatically replace artifacts
```
hughsaunders: Why not put this whole repeated block inside safe_to_replace_artifacts? Or some other function?

evrardjp: That would make it very abstracted IMO, while this is easily readable.

odyssey4me: @hughsaunders The reason is that line 81 is different in each set of artifact scripts.

hughsaunders: Ahh, apologies, didn't spot that.

PR #2626, opened 2017-10-30
RLM-218 Ensure /etc/rpc-release is created

antonym:
Since rpc-support was pulled out of rpc-o, /etc/rpc-release
no longer gets set or updated.  When running a leapfrog, the
rpc-release file is not updated and shows an older version.

This file should live inside rpc-o and removed from the
rpc-support role.

DISTRIB_CODENAME has also been dropped unless we want to
maintain codenames, and if so we could add that back in.

Original template is here:

https://github.com/rsoprivatecloud/openstack-ops/blob/master/templates/rpc-release.j2

Will backport to Newton as well.

Issue: [RLM-218](https://rpc-openstack.atlassian.net/browse/RLM-218)

### DIFF
--- a/rpcd/playbooks/rpc-release.yml
+++ b/rpcd/playbooks/rpc-release.yml
@@ -1,0 +1,29 @@
+---
+# Copyright 2017, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+- name: Install rpc-release file
+  hosts: hosts
+  gather_facts: true
+  tasks:
+    - name: Install rpc-release file
+      include_role:
+        name: openstack_hosts
+        tasks_from: openstack_release.yml
+      vars:
+        openstack_distrib_id: "RPC"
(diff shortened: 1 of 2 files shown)


antonym:
@odyssey4me We currently drop both openstack-release and rpc-release in /etc.  During a leapfrog though, only the openstack-release is updated and rpc-release is not since the rpc-support role is no longer present.

We could do a few things:

Ensure rpc-release is updated
Drop rpc-release 
Symlink rpc-release to openstack-release

I'm not sure what scripts support/ops have been written to key off either file so I was looking to at least ensure it's updated so that it doesn't reflect incorrect version information after a leapfrogupgrade.

The release files are redundant, so it might be a good time to pick the one we want and go with it.

odyssey4me:
@antonym The point is that we can just re-use the same role to implement both files, just give it different var values. ie OSA runs and drops openstack-release, then RPC runs with a different playbook to drop rpc-release.

My intent when I added that facility was to allow this method instead of carrying another template and set of tasks - slimming down the implementation to nothing more than a set of vars and a playbook to start with... but then hopefully to only drop one of the files and not bother with having the playbook any more.

The goes towards the goal of cutting tech debt and having rpc-openstack as a repo be as little as possible on top of OSA.

antonym:
recheck_cit_artifact

antonym:
@odyssey4me Using include_role should work post with Ansible >2.2 as it lets you run specific tasks.  

Newton is running 2.1.x so any suggestions for running that?  Should I just include the location of the task in /etc/ansible/roles for the backport to newton?

odyssey4me:
@antonym Oh yes, I see what you mean. This is the trouble with having both files, and I don't really understand why we'd want both files. Why not just have rpc-release and not openstack-release, and therefore only have the vars set in group_vars for rpc-openstack, doing away with the rest of it. The only thing we might want is to clean up the old release file, and I'd even suggest adding that to OSA rather than here. @cloudnull Your thoughts?

PR #2629, opened 2017-11-01
RE-694 Correct RC branch name

odyssey4me:
The current verification is happening against
the wrong RC branch, causing artifacts to always
refresh, which was not the intent. It should only
refresh when the mainline and RC branch differ
in the value of rpc_release.

Issue: [RE-694](https://rpc-openstack.atlassian.net/browse/RE-694)


odyssey4me:
Added 'do not merge' label because the current artifact refresh is in progress and the merge of this might prevent the upload, resulting in a mixed set. Once the job is done we can merge it.

PR #2630, opened 2017-11-01
RO-2864 Remove a-r-r specs included in OSA

odyssey4me:
These role SHA's are now included in OSA, so we should
not override them.

Issue: [RO-2864](https://rpc-openstack.atlassian.net/browse/RO-2864)

### DIFF
--- a/ansible-role-requirements.yml
+++ b/ansible-role-requirements.yml
@@ -10,10 +10,6 @@
   scm: git
   src: https://github.com/ceph/ansible-ceph-osd.git
   version: 434a13f4d3ee3f60c410052d2670d5dfea034bf8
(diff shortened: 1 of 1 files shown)


antonym:
recheck_cit_artifact

odyssey4me:
Oh hum, it appears that there's something broken in there... the git artifact prep is failing with:

```
{u'msg': u'Looking up repositories'}}, 'msg': u'Looking up repositories', '_ansible_verbose_always': True}]}: 'galera_client_major_version' is undefined"}
```

odyssey4me:
I've updated the patch to remove the role specs that can be removed, and to update those that can't until RO-3006 is addressed.

PR #2634, opened 2017-11-01
K8S-527 Remove old implementation of Octavia

xgerman:
Octavia is now it's own package in rpc-octavia and can be installed on
top of RPC-O newton. So remove all remnants.

Issue: [K8S-527](https://rpc-openstack.atlassian.net/browse/K8S-527)


odyssey4me:
The ``release`` test appears to have an issue if the PR is not made using a branch on the RPC-O repo, so I think it can be safely ignored, especially given there being no release related content here.

xgerman:
Yep, I wanted to cherry-pick that change...

PR #2637, opened 2017-11-02
RE-1082 Hotfix issues with nova deployment

odyssey4me:
This overrides the OSA integrated repo role SHA in order
to include two critical fixes:

- Ensure nova is started after everything else
  This affects Trusty deployment (including leapfrogs).

- Force the installation of qemu packages
  If this is not done, libvirt fails to restart properly.

Issue: [RE-1082](https://rpc-openstack.atlassian.net/browse/RE-1082)


odyssey4me:
recheck_cit_xenial_swift_deploy - job artifact upload failed, causing the job to fail

PR #2644, opened 2017-11-03
RLM-327 Removal of rpc-support role

BjoernT:
The rpc-support role is removed from the RPC-O distribution
and now individually maintained as standalone module
openstack-ops at https://github.com/rsoprivatecloud/openstack-ops


antonym:
Want to remove references from the README.md as well?

antonym:
recheck_cit_issue_link

PR #2661, opened 2017-11-21
RLM-301 Add release note for es upgrade

Frank-ZhangXin:
Explain new env var in release note.


claco:
Why is it tied to RLM-9 in the title?

PR #2670, opened 2017-11-27
RE-1178 Remove initial rpc-differ and update osa-differ

hughsaunders:
The initial rpc-differ which is run in the dockerfile
is not really necessary as the later execution will
pull down everything it needs.

OSA differ 0.3.3 correctly pulls refs for fresh clones so
the pre-clone is no longer required.

Issue: [RE-1178](https://rpc-openstack.atlassian.net/browse/RE-1178)

### DIFF
--- a/gating/generate_release_notes/release_notes_dockerfile
+++ b/gating/generate_release_notes/release_notes_dockerfile
@@ -9,11 +9,7 @@ RUN echo "jenkins ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers
 RUN apt-get install -y pandoc
 
 RUN pip install rpc_differ==0.3.0 reno==2.5.1
-RUN pip install git+https://github.com/major/osa_differ@0.3.2
-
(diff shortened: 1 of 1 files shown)


gating/generate_release_notes/release_notes_dockerfile
```diff
@@ -9,7 +9,7 @@ RUN echo "jenkins ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers
 RUN apt-get install -y pandoc
 
 RUN pip install rpc_differ==0.3.0 reno==2.5.1
-RUN pip install git+https://github.com/major/osa_differ@0.3.2
+RUN pip install git+https://github.com/major/osa_differ@0.3.3
 
 RUN rpc-differ --debug --update master master
```
git-harry: Shouldn't this line have been removed?

odyssey4me: Done.

PR #2679, opened 2017-11-29
RE-1174 Add workaround for nova-novncproxy fail

odyssey4me:
When upgrading the nova playbook will try to reload
nova-novncproxy and fail. This patch adds a release
note to describe this as a known issue and to provide
a workaround.

Issue: [RE-1174](https://rpc-openstack.atlassian.net/browse/RE-1174)


odyssey4me:
recheck_cit_xenial_swift_deploy - error when uploading job artifacts

major:
Swift deploy is still running, but this PR is just for a reno. Merging...

PR #2680, opened 2017-11-29
RO-3117 Fix issues with downloading aptly data from rpc-repo

odyssey4me:
The Ansible documentation specifies that the rsync_opts
should be an array, not a string. While this has worked
for us so far with Ansible 2.1.x it could break at any
time and definitely causes failure with Ansible 2.3.x
so we change this up to be implemented the correct way.

When synchronising the aptly database from rpc-repo
rsync changes the owner:group to root when it is only
modifying the attributes of the file instead of the
contents - at least I think that's the condition.

After spending several hours trying to figure out
how to do it better in the rsync options, and failing,
this patch opts to use a simpler change which just
forces the owner:group after the sync is completed.

Some line spaces are added into the playbook to improve
readability.

Issue: [RO-3117](https://rpc-openstack.atlassian.net/browse/RO-3117)

### DIFF
--- a/scripts/artifacts-building/apt/aptly-all.yml
+++ b/scripts/artifacts-building/apt/aptly-all.yml
@@ -32,7 +32,11 @@
         dest: "{{ artifacts_aptrepos_dest_folder | dirname }}"
         mode: pull
         delete: yes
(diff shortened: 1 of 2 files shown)


major:
Waiting on the linting tests to finish. This patch itself is unchanged and was passing tests properly a moment ago.

PR #2692, opened 2017-11-30
[RO-3145] ELK Linting Fixes

d34dh0r53:
Update Logstash, Filebeat and Kibana SHAs to incorporate linting fixes.

Issue: [RO-3145](https://rpc-openstack.atlassian.net/browse/RO-3145)

### DIFF
--- a/ansible-role-requirements.yml
+++ b/ansible-role-requirements.yml
@@ -16,15 +16,15 @@
 - name: logstash
   scm: git
   src: https://github.com/rcbops/rpc-role-logstash.git
-  version: 5ee5229e9b80e1193b1dd4ba56ef16edf244f237
+  version: f42bd2ccdc9c4608e2a357a9f088e3873a6ecad4
 - name: filebeat
   scm: git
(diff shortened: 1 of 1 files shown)


git-harry:
Why is ansible-lint being run on these roles when they are not code in this repo? This seems like the wrong way to fix this issue.

prometheanfire:
It's still our code, while it'd be better to have the linting on that repo as well, I think that we should care about all the code we use and more importantly all the code we manage.

odyssey4me:
The dep update scripts be updated to also include updating these deps, and jobs should be implemented on those roles to test them on PR's and periodically so that we pick up these issues proactively instead of reactively.

major:
recheck_cit_dep_update

odyssey4me:
It would seem that this needs a rebase to bring in the dep_update fix.

odyssey4me:
Oh, I guess not - bad browser.

cloudnull:

PR #2693, opened 2017-11-30
RO-3149 Update rpc_release file location

odyssey4me:
The dep_update script checks the value of the rpc_release
in the rc branch and the current branch, then does things
from there. Currently the location of the file is wrong,
so the script is failing.

Issue: [RO-3149](https://rpc-openstack.atlassian.net/browse/RO-3149)


odyssey4me:
Note: The ELK role lint failures are fixed in https://github.com/rcbops/rpc-openstack/pull/2692

PR #2697, opened 2017-11-30
[RO-3150] Fix partial artifacts installation

cloudnull:
This change moves the pip install role to the configure-python-sources
task file. This is done so we're only ever executing it to setup our
hosts when our sources are present. This change also uses two facts when
running with artifacts. First the artifacts must be online and availanle.
Second the artifact source must be "enabled". This gives deployers the
ability to enable or disable artifacts as needed.

Issue: [RO-3150](https://rpc-openstack.atlassian.net/browse/RO-3150)

### DIFF
--- a/playbooks/configure-apt-sources.yml
+++ b/playbooks/configure-apt-sources.yml
@@ -19,6 +19,22 @@
   user: root
   gather_facts: true
   pre_tasks:
+    - name: Ensure local facts directory exists
+      file:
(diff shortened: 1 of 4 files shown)


cloudnull:
Rebased to pull in the linter fixes

cloudnull:
Merging as it was approved by Major and does unblock our master gates.

PR #2711, opened 2017-12-05
Set a package pin for the RPC-OpenStack repo

cloudnull:
While the RPC-OpenStack repo "should" be the only repository used within
an artifacted deployment there are times when "other" repositories may
be added to a deployment post build which could have adverse affects on
our artifacted builds when it comes to day to operations. This change
adds a package pin instructing apt to prefer packages from our repos.
This change implements a pin using the highest priorty possible without
also instructing the APT to downgrade packages by default. The pin will
only be implemented when apt package artifacts are enabled and found to
be available.

Issue: [RO-3263](https://rpc-openstack.atlassian.net/browse/RO-3263)

### DIFF
--- a/playbooks/configure-apt-sources.yml
+++ b/playbooks/configure-apt-sources.yml
@@ -165,11 +165,29 @@
         - apt_artifact_found | bool
         - apt_artifact_enabled | bool
 
+    - name: Remove apt preference file if artifacts are disbaled
+      file:
+        dest: "/etc/apt/preferences.d/rpc_openstack_pinned_packages.pref"
+        state: absent
+      when:
+        - not apt_artifact_enabled | bool
+
     - name: Update apt-cache
       apt:
         update_cache: yes
       when:
         - (apt_artifact_found | bool) and (apt_sources_configure | changed or apt_sources_configure_rpco | changed)
+        - apt_artifact_enabled | bool
+
+  roles:
+    - role: "apt_package_pinning"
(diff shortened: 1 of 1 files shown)


playbooks/configure-apt-sources.yml
```diff
@@ -171,7 +171,16 @@
...
-
+  roles:
+    - role: "apt_package_pinning"
+      apt_package_pinning_file_name: "rpc_openstack_pinned_packages.pref"
+      apt_pinned_packages:
+        - package: "*"
+          release: "{{ rpc_release }}-{{ ansible_distribution_release }}"
+          priority: "998"
+        - package: "*"
+          release: "{{ ansible_distribution_release }}"
```
odyssey4me: Why is this a higher priority? I'm a little confused about the intention here? I thought that the artifacts should be highest priority?

We could also set this to priority 1001 which would cause forcible downgrades if something else is installed. This might help us with the day 1 deployment issues?

cloudnull: Yup I've these backwards and I'm not sure if I need the second entry at all.

As for the priority, I considered doing 1001 but the general consensus I got from folks is dowgraded packages in any form is a bad idea and I tend to agree.

odyssey4me:
I've implemented something similar in https://github.com/rcbops/rpc-openstack/pull/2740 but I think it's a bit more thorough. I've done the same mechanism for newton in https://github.com/rcbops/rpc-openstack/pull/2739.

cloudnull:
Abandoning in favor of #2740

PR #2712, opened 2017-12-05
Add properties to the images we provide

cloudnull:
This change adds property flags to the image we provide which will
ensure all of the images for distros that may run customer workloads
are operating at the highest level of performance possible.

### DIFF
--- a/playbooks/openstack-image-setup.yml
+++ b/playbooks/openstack-image-setup.yml
@@ -42,6 +42,7 @@
         name: "{{ item.name }}"
         filename: "/var/backup/os_image_{{ item.name }}"
         disk_format: "{{ item.format }}"
+        properties: "{{ item.properties | default(omit) }}"
       with_items: "{{ openstack_images }}"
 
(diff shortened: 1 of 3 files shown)


BjoernT:
The benchmarks  I did with customer workload has shown no benefit for virtio-scsi over virtio-blk. I had better latency with virtio-blk over virtio-scsi. Also multiqueueing should only be turned on when file handles are increased for qemu, someting we usually forget....

cloudnull:
@BjoernT I can update this to use the virtio-blk driver by default. Is there a file descriptor limit we should set on compute hosts by default ?

BjoernT:
From a RH errata I found that you should at least configure 2048 file handles but from running ceph we know that at least 32k are needed and I would like to use the same initial limit

cloudnull:
done.

prometheanfire:
recheck_cit_artifact

cloudnull:
the artifact build process is busted. We're working elsewhere to fix this.

PR #2716, opened 2017-12-10
Fix hw_scsi_model option

cloudnull:
This change sets the hw_scsi_model to "virtio-scsi". The use of
"virtio-blk" is assumed however for performance reasons, as noted in
horizon and by redhat we need to set this to "virtio-scsi" [0] to get
the most out of the images we provide by default.

[0] - https://redhatstackblog.redhat.com/2017/01/18/9-tips-to-properly-configure-your-openstack-instance/

Issue: [RO-3080](https://rpc-openstack.atlassian.net/browse/RO-3080)


antonym:
recheck_cit_artifact

PR #2720, opened 2017-12-12
Add tasks to revert artifacts if they're disabled

cloudnull:
Presently if artifacts are enabled there's no way to automate their
removal. This change adds tasks to backup files that we modify for
artifacts and to restore original files should a deployer choose to
disable the artifacts during a run.

Issue: [RLM-1120](https://rpc-openstack.atlassian.net/browse/RLM-1120)

Issue: [RO-3080](https://rpc-openstack.atlassian.net/browse/RO-3080)

Issue: [RE-1287](https://rpc-openstack.atlassian.net/browse/RE-1287)

### DIFF
--- a/playbooks/configure-apt-sources.yml
+++ b/playbooks/configure-apt-sources.yml
@@ -119,49 +119,104 @@
         - not apt_artifact_enabled | bool
 
   tasks:
-    - name: Sync artifact files (all)
-      copy:
-        src: "{{ item }}"
-        dest: "/etc/openstack_deploy/group_vars/all/{{ item | basename }}"
-      with_items:
-        - files/apt.yml
-      run_once: true
-      delegate_to: localhost
+    - name: Deploy apt artifacts
+      block:
+        - name: Backup the original sources file
+          copy:
+            src: /etc/apt/sources.list
+            dest: /etc/apt/sources.list.original
+            remote_src: yes
+            force: no
+
+        - name: Sync artifact group files (all)
+          copy:
+            src: "{{ item }}"
+            dest: "/etc/openstack_deploy/group_vars/all/{{ item | basename }}"
+          with_items: "{{ artifact_apt_group_files }}"
+          run_once: true
+          delegate_to: localhost
+
+        - name: Sync artifact user files (all)
+          copy:
+            src: "{{ item }}"
+            dest: "/etc/openstack_deploy/{{ item | basename }}"
+            mode: "0444"
+          with_items: "{{ artifact_apt_user_files }}"
+          run_once: true
+          delegate_to: localhost
+
+        - name: Determine the existing Ubuntu repo configuration
+          shell: 'sed "s/^[ \t]*//" /etc/apt/sources.list | grep -oP "^deb \K(\[?.*\]?.*ubuntu\S*\/?)(?= {{ ansible_distribution_release }} main)"'
+          register: _ubuntu_repo
+          changed_when: false
+          delegate_to: "{{ physical_host | default(omit) }}"
+          when:
+            - host_ubuntu_repo is not defined
+
+        - name: Set host_ubuntu_repo fact
+          set_fact:
+            host_ubuntu_repo: "{{ _ubuntu_repo.stdout_lines[0] }}"
+          when:
+            - host_ubuntu_repo is not defined
+            - _ubuntu_repo.stdout_lines is defined
+
+        - name: Replace the apt sources file with our content
+          copy:
+            content: |
+              # Base repository
+              deb {{ host_ubuntu_repo }} {{ ansible_distribution_release }} main universe
+            dest: "/etc/apt/sources.list"
+            backup: yes
+          register: apt_sources_configure
+          when:
+            - apt_artifact_mode == "strict"
+
+        - name: Create the rpco apt sources file
+          copy:
+            content: |
+              # RPC-OpenStack repository
+              deb {{ rpco_mirror_base_url }}/apt-mirror/integrated/ {{ rpc_release }}-{{ ansible_distribution_release }} main
+            dest: "/etc/apt/sources.list.d/rpco.list"
+            backup: yes
+          register: apt_sources_configure_rpco
+
+        - name: Add rpco keys
+          apt_key:
+            url: "{{ rpco_mirror_base_url }}/apt-mirror/rcbops-release-signing-key.asc"
+            state: "present"
+          register: add_keys
+          until: add_keys | success
+          retries: 5
+          delay: 2
+
       when:
         - apt_artifact_found | bool
         - apt_artifact_enabled | bool
 
-    - name: Sync artifact files (all)
-      copy:
-        src: "{{ item }}"
-        dest: "/etc/openstack_deploy/{{ item | basename }}"
-        mode: "0444"
-      with_items:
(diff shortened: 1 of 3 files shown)


major:
This looks pretty good, but there's a conflict with one of the more important files from this commit. ;)

playbooks/configure-container-sources.yml
```diff
@@ -139,6 +139,25 @@
   vars:
     ansible_python_interpreter: "/usr/bin/python"
     container_search_string: ".*{{ ansible_distribution_release }};.*{{ rpc_release }};"
+    container_search_string: ".*{{ ansible_distribution_release }};.*{{ rpc_release }};"
```
odyssey4me: Why the duplicate var definition?

playbooks/configure-apt-sources.yml
```diff
@@ -119,48 +119,129 @@
...
+        - name: Add rpco keys
+          apt_key:
+            url: "{{ rpco_mirror_base_url }}/apt-mirror/rcbops-release-signing-key.asc"
+            state: "present"
+          register: add_keys
+          until: add_keys | success
+          retries: 5
+          delay: 2
+
+        - name: Remove extra sources
```
odyssey4me: Line 160 replaces the entire content of the file, so this is unnecessary (and a regression).

playbooks/configure-apt-sources.yml
```diff
@@ -119,48 +119,129 @@
...
-        - files/apt.yml
-      run_once: true
-      delegate_to: localhost
-      when:
-        - apt_artifact_found | bool
-        - apt_artifact_enabled | bool
+    - name: Deploy apt artifacts
+      block:
+        - name: Backup the original sources list
+          command: "cp /etc/apt/sources.list /etc/apt/sources.list.original"
```
odyssey4me: Is there a reason for the use of command rather than the source which used copy?

```
    - name: Backup the original sources file
      copy:
        src: /etc/apt/sources.list
        dest: /etc/apt/sources.list.original
        remote_src: yes
        force: no
```

playbooks/configure-apt-sources.yml
```diff
@@ -119,48 +119,129 @@
...
+          with_items: "{{ artifact_apt_user_files }}"
+          run_once: true
+          delegate_to: localhost
+
+        - name: Check for original sources file
+          stat:
+            path: "/etc/apt/sources.list.original"
+          register: ogs
+
+        - name: Remove the modified sources file
```
odyssey4me: Why remove the file, when the next task replaces whatever's there anyway?

playbooks/configure-apt-sources.yml
```diff
@@ -119,48 +119,129 @@
...
+          file:
+            dest: "/etc/apt/sources.list"
+            state: absent
+          register: remove_sources
+          when:
+            - ogs.stat.exists | bool
+
+        - name: Restore the original sources list and remove rpco.list
```
odyssey4me: Why move the file when we can just copy and leave the original in-place?

playbooks/configure-apt-sources.yml
```diff
@@ -119,48 +119,129 @@
...
+
+        - name: Remove artifact user vars (all)
+          file:
+            dest: "/etc/openstack_deploy/group_vars/all/{{ item | basename }}"
+            state: absent
+          with_items: "{{ artifact_apt_user_files }}"
+          run_once: true
+          delegate_to: localhost
+
+        - name: Check for original sources file
```
odyssey4me: In the current environment, the backup is always created if it's not there, negating the need to have this task and always ensuring that the original is backed up. Why not leave that in place and do away with the extra tasks?

playbooks/configure-apt-sources.yml
```diff
@@ -119,48 +119,129 @@
...
-      changed_when: false
-      delegate_to: "{{ physical_host | default(omit) }}"
+        - name: Replace the apt sources file with our content
+          copy:
+            content: |
+              # Base repository
+              deb {{ host_ubuntu_repo }} {{ ansible_distribution_release }} main universe
+            dest: "/etc/apt/sources.list"
+            backup: yes
+          register: apt_sources_configure
```
odyssey4me: This should not happen in loose mode.

playbooks/configure-apt-sources.yml
```diff
@@ -119,48 +119,129 @@
...
+        - name: Remove the modified sources file
+          file:
+            dest: "/etc/apt/sources.list"
+            state: absent
+          register: remove_sources
+          when:
+            - ogs.stat.exists | bool
+
+        - name: Restore the original sources list and remove rpco.list
+          command: "mv /etc/apt/sources.list.original /etc/apt/sources.list"
```
odyssey4me: Currently the file is restored when switching from strict to loose mode. This patch only reverts when disabling artifacts, so this is breaking that functionality.

Also, currently the backup can be restored from the remote host or the deploy host. That functionality is not provided in this patch.

playbooks/configure-python-sources.yml
```diff
@@ -118,6 +133,14 @@
         - py_artifact_found | bool
         - py_artifact_enabled | bool
 
+    - name: Backup the original repo-build playbook
```
odyssey4me: Instead of doing this copy/restore mechanism, why not just reset the git repo to the original state at the appropriate SHA?

playbooks/configure-container-sources.yml
cloudnull: done.

playbooks/configure-apt-sources.yml
cloudnull: done.

playbooks/configure-apt-sources.yml
cloudnull: yes I only want to create the original file once bu I see setting the force flag should do what I need.

playbooks/configure-apt-sources.yml
cloudnull: done.

playbooks/configure-apt-sources.yml
cloudnull: done.

playbooks/configure-apt-sources.yml
cloudnull: done.

playbooks/configure-apt-sources.yml
cloudnull: done.

playbooks/configure-apt-sources.yml
cloudnull: done.

playbooks/configure-python-sources.yml
cloudnull: There may be other modifications at play and I'd rather not deal with unwinding them. At some point I hope to find a better way to accomplish that playbook without running any of those tasks.

odyssey4me: The plan has always been to push the staging functionality upstream, as described in https://github.com/openstack/openstack-ansible-specs/blob/master/specs/queens/deployment-stages.rst and https://github.com/openstack/openstack-ansible-specs/blob/master/specs/queens/python-build-install-simplification.rst so its time here is limited.

prometheanfire:
this still valid or can it be closed?

PR #2721, opened 2017-12-12
RE-1199 Cater for alpha/beta release versions

odyssey4me:
The dependency update script also handles the automated update of
rpc_release when it is appropriate to do so. Currently it only
does minor version increments. There is no way to cater to alpha
or beta releases.

Also, the dependency update script is currently not working correctly.
With the implementation of c4132cd the dependency update script was
adjusted to update playbooks/vars/rpc-release.yml but it did not remove
the adjustments to etc/openstack_deploy/group_vars/all/release.yml. The
resulting PR's from the changes are therefore malformed.

This patch implements changes to the script in order to cater for
alpha/beta releases according to the rpc_release naming scheme in
RE-1199. It will auto-increment both pre-release and minor versions
depending on the current value of rpc_release. This gives us more
flexibility to cater to releasing internally before doing customer
facing releases.

Issue: [RE-1199](https://rpc-openstack.atlassian.net/browse/RE-1199)

### DIFF
--- a/gating/update_dependencies/get-rpc_release.py
+++ b/gating/update_dependencies/get-rpc_release.py
@@ -1,0 +1,43 @@
+#!/usr/bin/env python
+# Copyright 2014-2017, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+import sys
+import yaml
+
+# Read the release file path from an environment variable
+if len(sys.argv) > 1:
+    release_file = sys.argv[1]
+else:
+    release_file = os.environ['RELEASE_FILE']
+
+# Read the RPC release series name from an environment variable
+if len(sys.argv) > 2:
+    release_series = sys.argv[2]
+else:
+    release_series = os.environ['RPC_PRODUCT_RELEASE']
+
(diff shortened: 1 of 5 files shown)


odyssey4me:
The release test is broken due to issues unrelated to this patch. Please see https://rpc-openstack.atlassian.net/browse/RO-3244

playbooks/openstack-flavor-setup.yml
```diff
@@ -40,8 +40,6 @@
         ephemeral: "{{ item.ephemeral }}"
       with_items: "{{ openstack_vm_flavors }}"
 
-  environment: "{{ deployment_environment_variables | default({}) }}"
```
cloudnull: this seems out of place.

odyssey4me: Yup, see the specific commit for it.

cloudnull: ah makes sense. :+1:

odyssey4me:
Validated to be working with:
https://rpc.jenkins.cit.rackspace.net/job/Dep-Update_rpc-openstack-master-PM/15/
https://rpc.jenkins.cit.rackspace.net/job/Dep-Update_rpc-openstack-master-PM/16/

gating/update_dependencies/get-rpc_release.py
```diff
@@ -0,0 +1,43 @@
...
+    release_file = sys.argv[1]
+else:
+    release_file = os.environ['RELEASE_FILE']
+
+# Read the RPC release series name from an environment variable
+if len(sys.argv) > 2:
+    release_series = sys.argv[2]
+else:
+    release_series = os.environ['RPC_PRODUCT_RELEASE']
+
```
mancdaz: would using argparse or similar not make this less brittle?

odyssey4me: Perhaps, but that can be evolved from here. Given the timelines I went with quick and simple.

odyssey4me:
I'll go ahead with force merging given that this is part of the solution to unblocking the gate for rpc-o master.

PR #2729, opened 2017-12-14
RO-3267 Only copy yml files for openstack_deploy

odyssey4me:
Instead of copying *all* the files, resulting in
/etc/openstack_deploy being filled with a bunch of
example files and CI-related config files, we only
copy over the *.yml files as they're required and
useful.

Issue: [RO-3267](https://rpc-openstack.atlassian.net/browse/RO-3267)

### DIFF
--- a/playbooks/configure-release.yml
+++ b/playbooks/configure-release.yml
@@ -120,16 +120,25 @@
         force: "{{ osa_force_clone | default(false) }}"
 
     - name: Copy basic files into place
-      command: "cp -Rv /opt/openstack-ansible/etc/openstack_deploy /etc/openstack_deploy"
+      shell: |
+        rsync -av \
+              --include='*.yml' --include='*/' --exclude='*' \
+              "/opt/openstack-ansible/etc/openstack_deploy/" \
+              /etc/openstack_deploy/
       args:
+        executable: /bin/bash
         creates: /etc/openstack_deploy
+      tags:
+        - skip_ansible_lint
 
     - name: Sync configuration for RPC-OpenStack files
       shell: |
         rsync -av \
-              --exclude '*.bak' \
+              --include='*.yml' --include='*/' --exclude='*' \
               "{{ playbook_dir }}/../etc/openstack_deploy/" \
               /etc/openstack_deploy/
+      args:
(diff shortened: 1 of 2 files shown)


antonym:
recheck_cit_artifact

odyssey4me:
recheck_cit_xenial_swift_deploy That's odd, it seems that the test node went offline during the test.

odyssey4me:
recheck_cit_release

odyssey4me:

odyssey4me:
@antonym @cloudnull This is ready for review now, thanks!

playbooks/configure-release.yml
```diff
@@ -120,16 +120,25 @@
         force: "{{ osa_force_clone | default(false) }}"
 
     - name: Copy basic files into place
```
cloudnull: This task, the "Sync configuration for RPC-OpenStack files" task seems like it's doing the same thing twice. The initial intent of these tasks was to create the base directory if needed using the example files from OSA and then sync our files as needed. If we're just going to plain sync we should create the base dir and rsync just the one time.

odyssey4me: Well spotted, that's been corrected to copy from the correct source.

playbooks/configure-release.yml
```diff
@@ -120,16 +120,25 @@
...
-      command: "cp -Rv /opt/openstack-ansible/etc/openstack_deploy /etc/openstack_deploy"
+      shell: |
+        rsync -av \
+              --include='*.yml' --include='*/' --exclude='*' \
+              "/opt/openstack-ansible/etc/openstack_deploy/" \
```
prometheanfire: I guess this is fine (hard coded path), since we are being prescriptive with rpc-o

scripts/deploy-rpco.sh
```diff
@@ -27,9 +27,9 @@ source "${SCRIPT_PATH}/functions.sh"
 
 ## Main ----------------------------------------------------------------------
 
-# Generate the scretes required for the deployment.
+# Generate the secrets required for the deployment.
```
prometheanfire: the secrets need to be secrete

PR #2731, opened 2017-12-14
RO-3268 Use apt artifacts for bootstrap-ansible

odyssey4me:
When install.sh runs it uses the bootstrap-ansible script from
OSA to install packages which are needed on the deployment host.
If the apt artifacts are not used for this, the host may end up
with packages which are newer than those available in the apt
artifacts and this will result in a failure later in the build.

This patch ensures that the availability of apt packages is
checked, and if they are available it uses them. There is also
an environment variable to override this behaviour as may be
necessary (for example, this is necessary when preparing the apt
artifacts).

The 'Remove extra sources' task in configure-apt-sources.yml is
removed as the task 'Replace the apt sources file with our content'
lays down a replaced file and the remove task therefore is always
a no-op.

Issue: [RO-3268](https://rpc-openstack.atlassian.net/browse/RO-3268)

### DIFF
--- a/scripts/functions.sh
+++ b/scripts/functions.sh
@@ -44,10 +44,11 @@ if [ -z ${OSA_RELEASE+x} ]; then
   fi
 fi
 
-# Other
+# Vars used for bootstrapping artifact configurations
 export BASE_DIR=${BASE_DIR:-"/opt/rpc-openstack"}
+export HOST_SOURCES_REWRITE=${HOST_SOURCES_REWRITE:-"yes"}
+export HOST_UBUNTU_REPO=${HOST_UBUNTU_REPO:-"http://mirror.rackspace.com/ubuntu"}
 export HOST_RCBOPS_REPO=${HOST_RCBOPS_REPO:-"http://rpc-repo.rackspace.com"}
-export RPC_RELEASE="$(${BASE_DIR}/scripts/get-rpc_release.py)"
 
 # Read the OS information
 for rc_file in openstack-release os-release lsb-release redhat-release; do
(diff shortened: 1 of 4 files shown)


scripts/functions.sh
```diff
@@ -55,3 +57,35 @@ for rc_file in openstack-release os-release lsb-release redhat-release; do
...
+    return 1
+  fi
+
+}
+
+function configure_apt_sources {
+
+  # Replace the existing apt sources with the artifacted sources.
+
+  sed -i '/^deb-src /d' /etc/apt/sources.list
```
cloudnull: in order to be able rewind artifacts the old file should be backed up first. Something simple like 

``` shell
if [[ ! -f "/etc/apt/sources.list.original" ]]; then
  cp /etc/apt/sources.list /etc/apt/sources.list.original
fi
```

scripts/functions.sh
```diff
@@ -55,3 +57,35 @@ for rc_file in openstack-release os-release lsb-release redhat-release; do
...
+  sed -i '/-security /d' /etc/apt/sources.list
+  sed -i '/-updates /d' /etc/apt/sources.list
+
+  # Add the RPC-O apt repo source
+  echo "deb ${HOST_RCBOPS_REPO}/apt-mirror/integrated/ ${RPC_RELEASE}-${DISTRIB_CODENAME} main" \
+    > /etc/apt/sources.list.d/rpco.list
+
+  # Install the RPC-O apt repo key
+  curl --silent --fail ${HOST_RCBOPS_REPO}/apt-mirror/rcbops-release-signing-key.asc | apt-key add -
+
```
cloudnull: upstream's `bootstrap-ansible.sh` installs the following packages:

``` text
git-core curl gcc netcat python-minimal python-dev python3 python3-dev \
libssl-dev libffi-dev python-apt python3-apt python-virtualenv
```

maybe we can install these things first using the rpco repository without 
forcing a rewrite of the default sources list outside of the playbooks.
This way a deployer can still chose to disable artifacts without casting 
environment variables.

``` shell
apt-get update
DEBIAN_FRONTEND=noninteractive apt-get -t "${RPC_RELEASE}-${DISTRIB_CODENAME}" \
                                       -y install \
                                       git-core curl gcc netcat python-minimal \
                                       python-dev python3 python3-dev libssl-dev \
                                       libffi-dev python-apt python3-apt \
                                       python-virtualenv
```

cloudnull:
The biggest issue here is that we will not able to source the function.sh on a clean system if `python-yaml` is not install at a minimum. I [created a different approach](https://github.com/rcbops/rpc-openstack/pull/2732) for this problem which should help with the chicken and egg issue, however this will never work quite right so long as we're not able to PXE from the rpc-repo mirrors.

scripts/functions.sh
odyssey4me: Sure, that may work for this - but then there are other actions which may happen prior to the sources being rewritten which put the deployment at risk. And now all developers have to be aware of this when reviewing patches. It makes it less definitive. As such, I don't think this is how it should be done.

odyssey4me:
recheck_cit_all

odyssey4me:
recheck_issue_link

odyssey4me:
recheck_issue_link

odyssey4me:
recheck_cit_all

odyssey4me:
I'm working on a way to extract the rpc_release without having to install python-yaml.

odyssey4me:
Given that https://github.com/rcbops/rpc-openstack/pull/2732 has merged, this is being abandoned.

PR #2732, opened 2017-12-14
[RO-3268] Bootstrap rpco without breaking artifacts

cloudnull:
Because APT artifacts can not be used at kick time it's possible that
the basic installation of OSA can bring an environment ahead of our
current artifacted release. This change makes it possible for APT
artifacts to be present when bootstrapping the environment. When a
deployer runs a basic installation of RPC-OpenStack they have the choice
to enable or disable artifacts however if they chose to enable them all
package installations should come from our artifacted repo assuming
they're available.

This change ensures only the minimum requirements are installed on the system
prior to setting up the release and configuring our artifacted solutions.
To keep everything as idempotent as possible and to ensure we're using
playbooks to configure the deployment, the initial install will create an
*rpc-ansible* virtual environment in "/opt" using a cached version of ansible
from [ rpc-repo.rackspace.com ]. This simple ansible installation runs only
the required playbooks to get the deployment online. Once OSA has been
bootstrapped all future interactions can be run through the
`openstack-ansible` cli tool. 

Non-Idempotent commands in scripts were have been moved into
playbooks where possible. Modules have replaced most of the
functions and if no suitable module was available the command module
was used but in the most idempotent way possible by leveraging local
facts.

Issue: [RO-3268](https://rpc-openstack.atlassian.net/browse/RO-3268)

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -17,59 +17,69 @@
 set -euv
 set -o pipefail
 
+## Run first ------------------------------------------------------------------
+# NOTE(cloudnull): Install the minimum packages required before anything can be
+#                  executed. While these should have been included in the base
+#                  kick, if they were not, they will need to be installed prior
+#                  to doing anything else.
+apt-get update
+DEBIAN_FRONTEND=noninteractive apt-get install -y \
+                                       python \
+                                       python-yaml \
+                                       python-virtualenv
+
 ## Vars ----------------------------------------------------------------------
 export SCRIPT_PATH="$(readlink -f $(dirname ${0}))"
 
 ## Functions -----------------------------------------------------------------
 source "${SCRIPT_PATH}/functions.sh"
 
+function install_ansible_source {
+  DEBIAN_FRONTEND=noninteractive apt-get -y install \
+                                            gcc libssl-dev libffi-dev \
+                                            python-apt python3-apt \
+                                            python-dev python3-dev \
+                                            python-minimal python-virtualenv
+
+  /opt/rpc-ansible/bin/pip install "ansible==${RPC_ANSIBLE_VERSION}"
+}
+
+function install_ansible_wheel {
+  /opt/rpc-ansible/bin/pip install --trusted-host "${HOST_RCBOPS_DOMAIN}" \
+                                   --find-links "${RPC_LINKS}" \
+                                   "${RPC_ANSIBLE}"
+}
+
 ## Main ----------------------------------------------------------------------
-
-# If /opt/openstack-ansible exists, delete it if it is not a git clone
+# If /opt/openstack-ansible exists, delete it if it is not a git checkout
 if [[ -d "/opt/openstack-ansible" ]] && [[ ! -d "/opt/openstack-ansible/.git" ]]; then
-  rm -rf /opt/openstack-ansible
+  mv /opt/openstack-ansible /opt/openstack-ansible.original
 fi
 
-# Git clone the openstack-ansible repository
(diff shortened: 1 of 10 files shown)


cloudnull:
@prometheanfire this would not help the release distro version issues. starting from a .3 release with updates installed when the artifacts only support .2 would still result in a broken build. This change simply ensures we're not actively installing things that are known to break artifacts prior to the repos getting into place. We're now ONLY installing the three packages [python, python-yaml, python-virtualenv] which are the minimal requirements for anything else to run within RPC-OpenStack. While the argument could be made that even those three packages could break the build, they're necessary when starting from a fresh, minimal, system. If we want 100% chain of custody for package installs we're going to need to have the ability to preseed from rpc-repo, but we're not there yet.

odyssey4me:
I'm not sure what to say. Instead of a simple, low tech debt approach as done in https://github.com/rcbops/rpc-openstack/pull/2731 you would rather implement this. Personally, I think this is madness.

scripts/install.sh
```diff
@@ -17,59 +17,69 @@
...
+## Run first ------------------------------------------------------------------
+# NOTE(cloudnull): Install the minimum packages required before anything can be
+#                  executed. While these should have been included in the base
+#                  kick, if they were not, they will need to be installed prior
+#                  to doing anything else.
+apt-get update
+DEBIAN_FRONTEND=noninteractive apt-get install -y \
+                                       python \
+                                       python-yaml \
+                                       python-virtualenv
```
odyssey4me: This is still installing packages which do not match the packages that were tested with the release, so it does not actually solve the problem at all. The amount of packages installed are reduced, sure, but it is still using packages that are not from the artifacted repo.

cloudnull:
@odyssey4me it's not madness at all and address all of the concerns we had with idempotency and package chain of custody. The other patch is implementing techdebt by forcing users to consume artifacts and rewrite sources in a way that is irreversible. This solution identifies exactly what we need and only installs those components.

scripts/install.sh
cloudnull: Untill we can install an OS using RPC-Repo this will always be true.

odyssey4me:
@cloudnull Suit yourself, I think this is madness. This does not help us implement a properly repeatable build and is a massive amount of tech debt to carry in comparison. This is over 300 lines added, plus a new venv, plus a wholly different way of doing the setup. The point of this rewrite was to reduce the wrappers and flatten the build. This is just adding layers, adding more possibilities for variation in our builds and making it harder to debug what's going on.

If this merges, I will work with it only because I have to. But, this is not an approach that I support.

odyssey4me:
@cloudnull Sure, as mentioned before - how this all goes ahead is not my call. All I'm looking for is resolution to the issue at hand, which is that the initial bootstrap of configuration currently does not cater for and does not respect the use of artifacts. This is causing downstream breakage in the artifact builds. So as long as this implements the use of apt artifacts immediately after executing install.sh then that helps. If it doesn't, then please let me know what must be executed afterwards so that I can adjust the artifact build process accordingly. Right now, this issue is a blocker for getting the builds to produce the correct artifacts as they're being polluted with non-artifacted apt packages when building.

So, if this is the way the team wants to go, that's fine. I just need a solution implemented to unblock the downstream work.

prometheanfire:
Now that this is reducing what we install to a KNOWN set of packages, can we just force install the same versions from the artifacted repo when that's set up?  AKA, go through known remediation steps to get artifacting working?

cloudnull:
Rebased to pull in the latest changes from master.

PR #2739, opened 2017-12-19
RLM-316 Implement strict/loose artifact options

odyssey4me:
When deploying to a new environment it is often the case
that the base OS installed has packages which are newer
than those available in the apt artifacts.

To ease the initial installation and provide the ability
to transition to using the apt artifacts exclusively, an
option is made available to do a 'loose' mode for the apt
artifacts. This mode leaves the existing apt sources in
place, ensuring that we don't end up with apt returning
that packages are installed which are not available in
the sources configured as it does when in 'strict' mode.

The facility to switch between modes is also implemented
so that an initial implementation trying to use strict
mode can switch to loose mode and continue the deployment.

Note that various forms of pinning to the repository were
tried to try and prefer the RPC-O repository, but allow
other sources to be used. This only works when the pins
are equal.

The default behaviour for 'loose' mode will be to make
use of the latest available package from the already
installed sources and the RPC-O source. This will mean
that a deployment today will not be guaranteed to provide
the same result tomorrow.

A little bit of tidying up to the AIO bootstrap has been
included in order to make it easier to read.

Jobs to test this as a periodic are implemented in
https://github.com/rcbops/rpc-gating/pull/575

### DIFF
--- a/gating/pre_merge_test/run_deploy.sh
+++ b/gating/pre_merge_test/run_deploy.sh
@@ -41,6 +41,17 @@ elif [[ "${RE_JOB_SCENARIO}" == "ironic" ]]; then
   export DEPLOY_IRONIC="yes"
 fi
 
+if [[ ${RE_JOB_IMAGE} =~ loose$ ]]; then
+
+  # Set the apt artifact mode
+  export RPCO_APT_ARTIFACTS_MODE="loose"
+
+  # Upgrade to the absolute latest
+  # available packages.
+  apt-get update
+  DEBIAN_FRONTEND=noninteractive apt-get -y upgrade
+fi
+
 # Run the deployment script
 cd ${BASE_DIR}
 source ${BASE_DIR}/scripts/deploy.sh
--- a/rpcd/playbooks/configure-apt-sources.yml
+++ b/rpcd/playbooks/configure-apt-sources.yml
@@ -33,28 +33,49 @@
         - host_ubuntu_repo is not defined
         - _ubuntu_repo.stdout_lines is defined
 
-    - name: Replace the apt sources file with our content
+    - name: Backup the original sources file
+      copy:
+        src: /etc/apt/sources.list
+        dest: /etc/apt/sources.list.original
+        remote_src: yes
+        force: no
+
+    - name: Replace the apt sources file with our content (artifact 'strict' mode)
       template:
         src: configure-apt-sources.j2
         dest: "/etc/apt/sources.list"
         backup: yes
       register: apt_sources_configure
+      when:
+        - rpco_apt_artifacts_mode == "strict"
       tags:
         - always
+
+    # Set rpco_apt_sources_restore_source to 'local' to have the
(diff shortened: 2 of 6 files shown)


gating/pre_merge_test/run_deploy.sh
```diff
@@ -41,6 +41,11 @@ elif [[ "${RE_JOB_SCENARIO}" == "ironic" ]]; then
   export DEPLOY_IRONIC="yes"
 fi
 
+# Set the apt artifact mode
+if [[ ${RE_JOB_IMAGE} =~ loose$ ]]; then
+  export RPCO_APT_ARTIFACTS_MODE="loose"
```
odyssey4me: Add an upgrade of all packages here so that we have the absolute latest packages installed.

rpcd/playbooks/configure-apt-sources.yml
```diff
@@ -33,15 +33,43 @@
...
+          {{ (rpco_apt_sources_restore_source | default('remote') == 'remote') | ternary('yes', 'no') }}
+        force: yes
+      when:
+        - rpco_apt_artifacts_mode != "strict"
+
+    - name: Set the preferred repo to use to be the apt artifact repo
+      copy:
+        content: >-
+          APT::Default-Release "{{ rpc_release }}-{{ ansible_distribution_release }}";
+        dest: /etc/apt/apt.conf.d/99rpco-pin
```
odyssey4me: register this and apt-get update after if it changes

rpcd/playbooks/configure-apt-sources.yml
```diff
@@ -33,15 +33,43 @@
...
+        src: /etc/apt/sources.list.original
+        dest: /etc/apt/sources.list
+        remote_src: >-
+          # Set rpco_apt_sources_restore_source to 'local' to have the
+          # original apt sources file from the deploy host be pushed out
+          # to all hosts and containers.
+          {{ (rpco_apt_sources_restore_source | default('remote') == 'remote') | ternary('yes', 'no') }}
+        force: yes
+      when:
+        - rpco_apt_artifacts_mode != "strict"
```
odyssey4me: register this and apt-get update after if it changes

odyssey4me:
Trying the 'loose' mode out in https://rpc.jenkins.cit.rackspace.net/job/PM_rpc-openstack-newton-xenial_loose-swift-deploy/2/

odyssey4me:
Giving it another go: https://rpc.jenkins.cit.rackspace.net/job/PM_rpc-openstack-newton-xenial_loose-swift-deploy/3/

rpcd/playbooks/configure-apt-sources.yml
odyssey4me: Done.

antonym:
Gave this a test by cherry picking on top of 14.6.0 and it appeared to work in both modes.  Going to give one more try on 14.5.0 but overall LGTM.

PR #2740, opened 2017-12-19
RLM-316 Implement strict/loose artifact options

odyssey4me:
When deploying to a new environment it is often the case
that the base OS installed has packages which are newer
than those available in the apt artifacts.

To ease the initial installation and provide the ability
to transition to using the apt artifacts exclusively, an
option is made available to do a 'loose' mode for the apt
artifacts. This mode leaves the existing apt sources in
place, ensuring that we don't end up with apt returning
that packages are installed which are not available in
the sources configured as it does when in 'strict' mode.

The facility to switch between modes is also implemented
so that an initial implementation trying to use strict
mode can switch to loose mode and continue the deployment.

Note that various forms of pinning to the repository were
tried to try and prefer the RPC-O repository, but allow
other sources to be used. This only works when the pins
are equal.

The default behaviour for 'loose' mode will be to make
use of the latest available package from the already
installed sources and the RPC-O source. This will mean
that a deployment today will not be guaranteed to provide
the same result tomorrow.

Issue: [RLM-917](https://rpc-openstack.atlassian.net/browse/RLM-917)

### DIFF
--- a/playbooks/configure-apt-sources.yml
+++ b/playbooks/configure-apt-sources.yml
@@ -28,7 +28,7 @@
         mode:  "0755"
         recurse: no
 
-    - name: initialize local facts
+    - name: Initialize local facts
       ini_file:
         dest: "/etc/ansible/facts.d/rpc_openstack.fact"
         section: "rpc_artifacts"
@@ -46,12 +46,19 @@
       set_fact:
         rpc_openstack: "{{ ansible_local['rpc_openstack']['rpc_artifacts'] }}"
 
-    - name: Set the rpc-artifact variables
+    - name: Set apt_artifact_enabled (based on previous settings)
       set_fact:
         apt_artifact_enabled: "{{ rpc_openstack['apt_artifact_enabled'] }}"
       when:
         - rpc_openstack['apt_artifact_enabled'] is defined
         - apt_artifact_enabled is undefined
+
+    - name: Set apt_artifact_mode (based on previous settings)
+      set_fact:
+        apt_artifact_mode: "{{ rpc_openstack['apt_artifact_mode'] }}"
+      when:
+        - rpc_openstack['apt_artifact_mode'] is defined
+        - apt_artifact_mode is undefined
 
     - name: Check for artifacts
       uri:
@@ -60,15 +67,21 @@
       failed_when: false
       register: check_artifacts
 
-    - name: Set artifacts found
+    - name: Set apt_artifact_found
       set_fact:
         apt_artifact_found: "{{ check_artifacts.status == 200 }}"
 
-    - name: Set artifacts enabled
+    - name: Set apt_artifact_enabled (based on whether they exist)
       set_fact:
         apt_artifact_enabled: "{{ apt_artifact_found | bool }}"
       when:
         - apt_artifact_enabled is undefined
+
+    - name: Set apt_artifact_mode default value
+      set_fact:
+        apt_artifact_mode: "strict"
+      when:
+        - apt_artifact_mode is undefined
 
     - name: Check if artifacts are enabled but not found
       fail:
@@ -79,6 +92,17 @@
       when:
         - apt_artifact_enabled | bool
         - not apt_artifact_found | bool
+      tags:
+        - always
+
+    - name: Check if artifact mode is set incorrectly
+      fail:
+        msg: |
+          The apt artifact mode is set to the invalid value of '{{ apt_artifact_mode }}'.
+          The deployment has halted. Valid values are 'strict' or 'loose'.
+      when:
+        - apt_artifact_enabled | bool
+        - apt_artifact_mode not in ['strict', 'loose']
       tags:
         - always
 
@@ -138,7 +162,14 @@
         - apt_artifact_found | bool
         - apt_artifact_enabled | bool
 
-    - name: Replace the apt sources file with our content
+    - name: Backup the original sources file
+      copy:
+        src: /etc/apt/sources.list
+        dest: /etc/apt/sources.list.original
+        remote_src: yes
+        force: no
+
+    - name: Replace the apt sources file with our content (artifact 'strict' mode)
       copy:
         content: |
           # Base repository
@@ -149,6 +180,22 @@
       when:
         - apt_artifact_found | bool
         - apt_artifact_enabled | bool
+        - apt_artifact_mode == "strict"
+
+    # Set rpco_apt_sources_restore_source to 'local' to have the
+    # original apt sources file from the deploy host be pushed out
+    # to all hosts and containers.
+    - name: Restore the original sources file (artifact 'loose' mode)
+      copy:
+        src: /etc/apt/sources.list.original
+        dest: /etc/apt/sources.list
+        remote_src: "{{ rpco_apt_sources_restore_source | default('remote') == 'remote' }}"
+        force: yes
+      register: apt_sources_restore
+      when:
+        - apt_artifact_found | bool
+        - apt_artifact_enabled | bool
+        - apt_artifact_mode != "strict"
 
     - name: Create the rpco apt sources file
       copy:
@@ -174,26 +221,13 @@
         - apt_artifact_found | bool
         - apt_artifact_enabled | bool
 
-    - name: Remove extra sources
-      lineinfile:
-        dest: /etc/apt/sources.list
-        state: absent
-        regexp: "{{ item }}"
-      with_items:
-        - "^deb-src"
-        - "-backports"
-        - "-security"
-        - "-updates"
-      when:
-        - apt_artifact_found | bool
-        - apt_artifact_enabled | bool
-
     - name: Update apt-cache
       apt:
         update_cache: yes
       when:
-        - (apt_artifact_found | bool) and (apt_sources_configure | changed or apt_sources_configure_rpco | changed)
         - apt_artifact_enabled | bool
+        - apt_artifact_found | bool
+        - apt_sources_configure | changed or apt_sources_configure_rpco | changed or apt_sources_restore | changed
 
   post_tasks:
     - name: Set artifact local fact
@@ -203,8 +237,12 @@
         option: "{{ item.option }}"
         value: "{{ item.value }}"
       with_items:
-        - { option: "apt_artifact_enabled", value: "{{ apt_artifact_enabled }}" }
-        - { option: "apt_artifact_found", value: "{{ apt_artifact_found }}" }
+        - option: "apt_artifact_enabled"
+          value: "{{ apt_artifact_enabled }}"
+        - option: "apt_artifact_found"
+          value: "{{ apt_artifact_found }}"
+        - option: "apt_artifact_mode"
+          value: "{{ apt_artifact_mode }}"
 
   vars:
     ansible_python_interpreter: "/usr/bin/python"
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -47,11 +47,15 @@ if [ "${DEPLOY_AIO}" != false ]; then
   # RO-3316 has been resolved.
   openstack-ansible -i 'localhost,' \
                     -e 'apt_target_group=localhost' \
+                    -e "apt_artifact_mode='${RPCO_APT_ARTIFACTS_MODE}'" \
                     -e 'container_artifact_enabled=false' \
                     "${SCRIPT_PATH}/../playbooks/site-artifacts.yml"
 
   # Install OpenStack-Ansible
   openstack-ansible "${SCRIPT_PATH}/../playbooks/openstack-ansible-install.yml"
+
+  # Set the AIO config bootstrap options to take apt artifacts into consideration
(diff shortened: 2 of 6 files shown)


playbooks/configure-apt-sources.yml
```diff
@@ -126,6 +146,20 @@
       when:
         - apt_artifact_found | bool
         - apt_artifact_enabled | bool
+        - apt_artifact_mode == "strict"
+
+    # Set rpco_apt_sources_restore_source to 'local' to have the
+    # original apt sources file from the deploy host be pushed out
+    # to all hosts and containers.
+    - name: Restore the original sources file (artifact 'loose' mode)
```
cloudnull: I don't think this task should be part of this PR. The restore actions we're proposing in #2720 are aiming to address this issue. From the POV of this PR it make sense to simply make the modifications as it is today when enabled, unless we want to merge these two PRs.

playbooks/configure-apt-sources.yml
```diff
@@ -151,16 +185,12 @@
...
-      lineinfile:
-        dest: /etc/apt/sources.list
-        state: absent
-        regexp: "{{ item }}"
-      with_items:
-        - "^deb-src"
-        - "-backports"
-        - "-security"
-        - "-updates"
+    - name: Set the preferred repo to use to be the apt artifact repo
```
cloudnull: This pin makes sense however I believe it should only be run when artifacts are enabled and  "apt_artifact_mode" is "loose". Otherwise we can assume the strict mode is in place and the apt sources have all been wiped out.

playbooks/configure-apt-sources.yml
```diff
@@ -151,16 +185,12 @@
         - apt_artifact_found | bool
         - apt_artifact_enabled | bool
 
-    - name: Remove extra sources
-      lineinfile:
```
cloudnull: Are we no longer needing to remove these lines from the default sources, even in strict mode?

playbooks/configure-apt-sources.yml
```diff
@@ -115,7 +128,14 @@
         - apt_artifact_found | bool
         - apt_artifact_enabled | bool
 
-    - name: Replace the apt sources file with our content
+    - name: Backup the original sources file
```
cloudnull: I don't think this task should be part of this PR. The restore actions we're proposing in #2720 are aiming to address this issue. From the POV of this PR it make sense to simply make the modifications as it is today when enabled, unless we want to merge these two PRs.

playbooks/configure-apt-sources.yml
```diff
@@ -151,16 +185,12 @@
...
-        regexp: "{{ item }}"
-      with_items:
-        - "^deb-src"
-        - "-backports"
-        - "-security"
-        - "-updates"
+    - name: Set the preferred repo to use to be the apt artifact repo
+      copy:
+        content: >-
+          APT::Default-Release "{{ rpc_release }}-{{ ansible_distribution_release }}";
```
cloudnull: What is the reason for using apt config instead of an actual apt pin? Doesn't https://github.com/rcbops/rpc-openstack/pull/2720 accomplish much of the same thing and doesn't apt-package-pin role with a high priority do what we need?

scripts/deploy.sh
```diff
@@ -66,13 +66,28 @@ if [ "${DEPLOY_AIO}" != false ]; then
...
+    apt_artifact_enabled=$(awk '/apt_artifact_enabled/ {print $3}' /etc/ansible/facts.d/rpc_openstack.fact)
+    if [[ "${apt_artifact_enabled}" == "True" ]]; then
+      if [[ "${RPCO_APT_ARTIFACTS_MODE}" == "strict" ]]; then
+        # Prevent the AIO bootstrap from re-implementing
+        # the updates, backports and UCA sources.
+        export BOOTSTRAP_OPTS='{ "bootstrap_host_apt_distribution_suffix_list": [], "uca_enable": "False" }'
+      else
+        # Prevent the AIO bootstrap from re-implementing
+        # the UCA sources.
+        export BOOTSTRAP_OPTS='uca_enable="False"'
```
cloudnull: Does this export also need to be wrapped in braces?

playbooks/configure-apt-sources.yml
```diff
@@ -46,30 +46,43 @@
...
         apt_artifact_enabled: "{{ apt_artifact_found | bool }}"
       when:
         - apt_artifact_enabled is undefined
 
+    - name: Set apt_artifact_mode default value
+      set_fact:
+        apt_artifact_mode: "strict"
+      when:
+        - apt_artifact_mode is undefined
+
```
cloudnull: We should add a failure condition when the "apt_artifact_mode" is not set to "strict" or "loose". From an operator POV, the system should provide feedback to combat against typos especially when with an option that may impact the entire deployment.

odyssey4me: It's not really necessary. All the conditionals test against the default of 'strict'. If the value is anything else, it considers it loose.

playbooks/configure-apt-sources.yml
odyssey4me: In the line above, the entire file is being replaced. This task does nothing.

scripts/deploy.sh
odyssey4me: No, the braces above are only due to the expression of the list value. This is purely string value so it's not necessary.

playbooks/configure-apt-sources.yml
odyssey4me: It sets the priority of the apt artifact repo to 990, which is above the default of 500 for the other repositories. It's effectively the same as the pin, but this is the recommended way in Ubuntu/Debian docs to define one repo as the explicitly preferred source.

playbooks/configure-apt-sources.yml
odyssey4me: That PR would be a good evolution over and above what's being done here because that takes the flipping further. This is only addressing the apt package flip between strict and loose mode, it does not handle removing the setting of artifacts all together.

playbooks/configure-apt-sources.yml
odyssey4me: As above.

playbooks/configure-apt-sources.yml
odyssey4me: The host OS may also have other repositories configured on them for other tooling. This ensures that our repository is still preferred above *any* other repository on the host.

playbooks/configure-apt-sources.yml
```diff
@@ -126,6 +146,20 @@
...
+    # original apt sources file from the deploy host be pushed out
+    # to all hosts and containers.
+    - name: Restore the original sources file (artifact 'loose' mode)
+      copy:
+        src: /etc/apt/sources.list.original
+        dest: /etc/apt/sources.list
+        remote_src: "{{ rpco_apt_sources_restore_source | default('remote') == 'remote' }}"
+        force: yes
+      register: apt_sources_restore
+      when:
```
cloudnull: This should do nothing if artifacting are disabled.

odyssey4me: Good catch - updating.

playbooks/configure-apt-sources.yml
odyssey4me: As it turns out, no form of pinning I could work out resulted in a working deployment when the RPC-O repository has older packages than the updates/security repository. I've tried pinning at various levels - 501 or more. In every case this results in a broken build because it want to use the older packages. The solution therefore has to be to have the RPC-O repo at the same priority as the rest of the repositories on the system.

This will give unpredictable results, but it helps day one and does assist a transition process which can be done later to 'strict' mode.

scripts/deploy.sh
```diff
@@ -63,12 +63,27 @@ if [ "${DEPLOY_AIO}" != false ]; then
...
     # Install OpenStack-Ansible
     openstack-ansible "${SCRIPT_PATH}/../playbooks/openstack-ansible-install.yml"
 
+    # Set the AIO config bootstrap options
+    apt_artifact_enabled=$(awk '/apt_artifact_enabled/ {print $3}' /etc/ansible/facts.d/rpc_openstack.fact)
+    if [[ "${apt_artifact_enabled}" == "True" ]]; then
+      if [[ "${RPCO_APT_ARTIFACTS_MODE}" == "strict" ]]; then
+        # Prevent the AIO bootstrap from re-implementing
+        # the updates, backports and UCA sources.
+        export BOOTSTRAP_OPTS='{ "bootstrap_host_apt_distribution_suffix_list": [], "uca_enable": "False" }'
```
odyssey4me: Understandably a mix of json and string options isn't working, and the gate test uses a flavor with a large secondary disk. This will likely require an upstream patch - working on it now.

odyssey4me:

playbooks/configure-apt-sources.yml
cloudnull: I'd rather this error on the side of caution. I can imagine a scenario where someone wants strict mode but typos it in some way, like using a capital letter somehow, resulting in a deployment in loose mode unexpectedly.

odyssey4me:
@cloudnull Please read the PR message, or the commit message, or the slack messages or the JIRA messages where this has been explained and noted. Pinning has been tried and does not work - if one source is pinned higher than the others, it will always try to use that source and this causes a build fail if the packages installed are not from that source.

Bear in mind that the RPC-O repo includes UCA, so it will still be selected for many packages as they will be newer - but yes, the purpose of this patch (as explained in the commit message) is to provide a way to transition and to cater for the unique situation on day one. Once day one is done, it is very likely that strict mode can be used at the next RPC-O upgrade.

playbooks/configure-apt-sources.yml
odyssey4me: Alright, once I'm back from holiday I can do that. Or we can merge this and a patch can go on top of it to evolve it.

odyssey4me:
@cloudnull Given that the purpose of this release is an 'alpha' release to test the waters, I think it is worth the try to reduce the vector of change within the deployment and to ease switching from day one to day two. If the team's position is to summarily dismiss it without even trying it, so be it.

playbooks/configure-apt-sources.yml
odyssey4me: Done.

odyssey4me:
@cloudnull Given the recent discussions, please re-review. This is passing all required tests.

odyssey4me:
I've updated the gate execution to handle 'loose' mode testing as per the changes in https://github.com/rcbops/rpc-gating/pull/586

odyssey4me:
recheck_cit_all

scripts/deploy.sh
```diff
@@ -50,13 +50,32 @@ if [ "${DEPLOY_AIO}" != false ]; then
...
+  # Set the AIO config bootstrap options
+  apt_artifact_enabled=$(awk '/apt_artifact_enabled/ {print $3}' /etc/ansible/facts.d/rpc_openstack.fact)
+  if [[ "${apt_artifact_enabled}" == "True" ]]; then
+    if [[ "${RPCO_APT_ARTIFACTS_MODE}" == "strict" ]]; then
+      # Prevent the AIO bootstrap from re-implementing
+      # the updates, backports and UCA sources. We have
+      # to use parameters here as BOOTSTRAP_OPTS only
+      # accepts string value parameters and the suffix
+      # parameter is a list.
+      echo -e '---\n{ "bootstrap_host_apt_distribution_suffix_list": [], "uca_enable": False }' > /opt/bootstrap-opts.yml
```
cloudnull: Given this is something we're always going to do for the "strict" gate we should just carry this var file and point at it instead of creating the file within the script.

cloudnull:
One small change, otherwise looks good to me.

scripts/deploy.sh
odyssey4me: I'll see whether this could be re-implemented through the use of local facts or something, given that the AIO bootstrap does not use any of the /etc/openstack_deploy files (nor should it). I'll need to fire up a test env to verify so I'll pick it up tomorrow.

scripts/functions.sh
```diff
@@ -55,6 +55,8 @@ done
 export BASE_DIR=${BASE_DIR:-"/opt/rpc-openstack"}
 export HOST_RCBOPS_DOMAIN="rpc-repo.rackspace.com"
 export HOST_RCBOPS_REPO=${HOST_RCBOPS_REPO:-"http://${HOST_RCBOPS_DOMAIN}"}
+export RPCO_APT_ARTIFACT_ENABLED=${RPCO_APT_ARTIFACT_ENABLED:-"yes"}
```
cloudnull: For the sake of consistency these variables should be RPC_ not RPCO_

scripts/functions.sh
```diff
@@ -55,6 +55,8 @@ done
...
+export RPCO_APT_ARTIFACT_MODE=${RPCO_APT_ARTIFACT_MODE:-"strict"}
```
cloudnull: Same as above.

odyssey4me: Done.

scripts/functions.sh
odyssey4me: Done.

PR #2745, opened 2017-12-20
RO-3303 Use python artifacts in AIO tests

odyssey4me:
With the artifact build problems resolved, and the various
issues relating to python artifact usage also resolved, we
can now revert to using the artifacts by default for AIO
builds.

Issue: [RO-3303](https://rpc-openstack.atlassian.net/browse/RO-3303)


odyssey4me:

PR #2747, opened 2017-12-20
[RO-3152] Add dep-update script for role requirements

cloudnull:
This change implements a python script to automatically update the role
requirements file for a given product release.

Issue: [RO-3152](https://rpc-openstack.atlassian.net/browse/RO-3152)

### DIFF
--- a/gating/update_dependencies/role-requirements-update.py
+++ b/gating/update_dependencies/role-requirements-update.py
@@ -1,0 +1,107 @@
+#!/usr/bin/env python
+# Copyright 2014-2017, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


gating/update_dependencies/run
```diff
@@ -67,6 +67,12 @@ for the dependencies to be updated regularly though, so that part continues
 to be done."
 fi
 
+
+# Execute role requirements file update
+export ROLE_REQUIREMENTS_FILE="${GATING_PATH}/../../ansible-role-${RPC_PRODUCT_RELEASE}-requirements.yml"
```
prometheanfire: I feel like the ordering here could be better, define the file then point to it, but it's a VERY minor nit.

gating/update_dependencies/role-requirements-update.py
```diff
@@ -0,0 +1,107 @@
...
+with TempDirMake() as mkd:
+    for item in release_file_content:
+        if item['scm'] != 'git':
+            pass
+
+        branch = item.get('branch', 'master')
+        src = item['src']
+
+        subprocess.call(
+            ["git", "clone", src],
```
prometheanfire: should this be shallow?  It can also clone the branch/sha you want down in the same command instead of doing it in a separate command below.

cloudnull: I chose not to use shallow as I didn't want to run into issues where the a specific checkout may not be present.

cloudnull: as for cloning with the branch, that will only work for tags and branches however if we use a SHA it will break. For simplicity I chose to clone and checkout which will work regardless.

PR #2752, opened 2018-01-02
[RO-3272] Cleanup legacy user variable files

cloudnull:
This change removes legacy user variable files that may have been
created on the system over the life of the deployment. This change will
remove all of the legacy user variable files on disk which contained
"defaults" options already set within OSA and RPC-O group variables or
contained within the roles. All of the overrides implemented by RPC-O
will remain in the "override" files which will ensure anything
implemented by the deployer remains intact.

** This change does not validate options in variable files. All options
   in the override variable files are the responsibility of the
   deployer.

** This change is looking for and removing legacy variable files going
   as far back as liberty as it's very probable that these erroneous
   files will still exist, especially if the deployment has been leap
   upgraded.

[RO-3272](https://rpc-openstack.atlassian.net/browse/RO-3272)

Issue: [RO-3272](https://rpc-openstack.atlassian.net/browse/RO-3272)

### DIFF
--- a/playbooks/configure-release.yml
+++ b/playbooks/configure-release.yml
@@ -166,6 +166,25 @@
         - skip_ansible_lint
 
   post_tasks:
+    - name: Remove legacy user variable files
+      file:
+        dest: "/etc/openstack_deploy/{{ item }}"
+        state: absent
+      with_items:
+        - "user_extras_variables.yml"
+        - "user_liberty_vars.yml"
+        - "user_master_vars.yml"
+        - "user_newton_vars.yml"
+        - "user_osa_variables_defaults.yml"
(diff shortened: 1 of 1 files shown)


BjoernT:
@cloudnull Is there a step to migrate existing configurations during upgrades into the new files? Otherwise I find it dangerous to remove "legacy" files not knowing where changes have been made. I would assume that accidental changes happened inside default files and user_variables has been made guaranteed.

odyssey4me:
@BjoernT As this has already merged, I would suggest continuing the conversation in https://rpc-openstack.atlassian.net/browse/RO-3442 instead of here as it's more visible.

PR #2758, opened 2018-01-09
remove dead code (scripts/release.py)

prometheanfire:


odyssey4me:
recheck_cit_release

prometheanfire:
recheck_cit_release
recheck_cit_artifact

cloudnull:
release job failures are known issues when using an internal branch instead of a fork because the failures are unrelated to this change :+1: .

prometheanfire:
ah, right, forgot about that

PR #2759, opened 2018-01-09
remove dead code (scripts/release.py)

prometheanfire:


odyssey4me:
Tox is failing due to a failed test. Looks like that will need adjustment of tox config, maybe gating config and another script removed.

prometheanfire:

tox.ini
```diff
@@ -50,9 +50,6 @@ commands =
         ansible-galaxy install -r {homedir}/openstack-ansible-role-requirements.yml; \
         {toxinidir}/scripts/linting-ansible.sh"
 
-[testenv:release-script]
-commands = python -m unittest discover -s scripts -p test_release.py -v
```
odyssey4me: Given that test_release.py is no longer used, it should be removed

prometheanfire: addressed

PR #2760, opened 2018-01-09
RLM-316 Update gating artifact option setting

odyssey4me:
In [1] the names for the jobs were normalised to ensure more
consistency.

This updates the script to ensure that the new values are
properly interpreted.

The package upgrade consistently uses 'upgrade' instead
of 'dist-upgrade' as using the latter requires a reboot
to make all core OS packages upgraded active. A reboot
cannot be facilitated through this script and upgrading
the extra packages is unnecessary given the frequent image
updates anyway.

[1] https://github.com/rcbops/rpc-gating/pull/586

### DIFF
--- a/gating/pre_merge_test/run_deploy.sh
+++ b/gating/pre_merge_test/run_deploy.sh
@@ -33,12 +33,6 @@ source ${BASE_DIR}/scripts/functions.sh
 
 ## Main ----------------------------------------------------------------------
 
-if [[ "${RE_JOB_ACTION}" == "deploy_no_artifacts" ]]; then
-  export DEPLOY_ARTIFACTING="no"
-  apt-get update
-  DEBIAN_FRONTEND=noninteractive apt-get -y dist-upgrade
-fi
-
 # Set the appropriate scenario variables
 if [[ "${RE_JOB_SCENARIO}" == "ceph" ]]; then
   export DEPLOY_CEPH="yes"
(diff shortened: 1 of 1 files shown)


gating/pre_merge_test/run_deploy.sh
```diff
@@ -47,8 +41,17 @@ elif [[ "${RE_JOB_SCENARIO}" == "ironic" ]]; then
...
 
-if [[ ${RE_JOB_IMAGE} =~ loose$ ]]; then
+if [[ ${RE_JOB_IMAGE} =~ no_artifacts$ ]]; then
+  # Set the env var to disable artifact usage
+  export DEPLOY_ARTIFACTING="no"
+
+  # Upgrade to the absolute latest
+  # available packages.
+  apt-get update
+  DEBIAN_FRONTEND=noninteractive apt-get -y dist-upgrade
```
antonym: Line 61 uses DEBIAN_FRONTEND=noninteractive apt-get -y upgrade instead, I assume they should be the same, so can we align them?

odyssey4me: Using dist-upgrade requires a reboot to make everything active, so I don't really understand the point... but in this case I just moved the setting as it was.

Which would you like it to be to make it consistent?

antonym: Probably upgrade, sounds like dist-upgrade can remove packages and may resolve dependencies which I’m not sure we need here.

odyssey4me: Done.

PR #2761, opened 2018-01-09
Move maas variables into user vars

cloudnull:
The group vars for maas will not function as initially expected. This
change moves the group vars into user vars so that the openstack-ansible
wrapper automatically sources the file loading it as the highest
precedence variable source.

Issue: [RO-3440](https://rpc-openstack.atlassian.net/browse/RO-3440)


playbooks/configure-apt-sources.yml
```diff
@@ -101,7 +101,18 @@
         dest: "/etc/openstack_deploy/group_vars/all/{{ item | basename }}"
       with_items:
         - files/apt.yml
-        - files/maas.yml
+      run_once: true
```
odyssey4me: Why change this to run_once? As these are fixed defaults which should not be edited, they should *always* be copied, otherwise how do we handle minor upgrades after release which include changes to these files?

playbooks/configure-apt-sources.yml
```diff
@@ -101,7 +101,18 @@
...
+      when:
+        - apt_artifact_found | bool
+        - apt_artifact_enabled | bool
+
+    - name: Sync artifact files (all)
+      copy:
+        src: "{{ item }}"
+        dest: "/etc/openstack_deploy/{{ item | basename }}"
+      with_items:
+        - files/user_rpcm_variables.yml
```
odyssey4me: How do we handle updating this file for the next release? The former releases clearly used the word 'defaults' and made the file read-only, giving a clear indication that upgrades of RPC-O would overwrite this file. This is putting the file into the 'you can edit it as you please' realm, so the question is how do we make changes when we upgrade?

playbooks/configure-apt-sources.yml
cloudnull: `run_once` is used so the task, which is delegated to the localhost, the deployment host, is only executed one time instead of iterating through all "hosts" and running on the delegated target.

playbooks/configure-apt-sources.yml
cloudnull: To ensure we own this file we could simply make it read-only and indicate via a comment, at the top of the file, that any addition will be removed upon upgrade. All that said, the old defaults are all deleted so there should be little risk, if any, from the current setup.

cloudnull: done. see https://github.com/rcbops/rpc-openstack/pull/2805

PR #2762, opened 2018-01-09
Do not pass --system-site-packages

prometheanfire:
Passing this flag allows the system to leak into the venv, this can cause
some problems particularly for compiled modules (openssl)
ERROR! Unexpected Exception: 'module' object has no attribute 'SSL_ST_INIT'


prometheanfire:
it's your code, so review please :D

PR #2764, opened 2018-01-10
[Newton] RLM-1078 Adds initial MNAIO gating capabilities

antonym:
Checks if image contains mnaio and if so, runs the
run_deploy_mnaio.sh.

This will create the MNAIO before pushing rpc-o to
infra1 and will kick off the deploy.sh from infra1.

Issue: [RLM-1078](https://rpc-openstack.atlassian.net/browse/RLM-1078)

### DIFF
--- a/gating/pre_merge_test/run_deploy_mnaio.sh
+++ b/gating/pre_merge_test/run_deploy_mnaio.sh
@@ -1,0 +1,164 @@
+#!/usr/bin/env bash
+
+# Copyright 2017, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+## Shell Opts ----------------------------------------------------------------
+
+set -evu
+
+echo "Building a Multi Node AIO (MNAIO)"
+echo "+-------------------- MNAIO ENV VARS --------------------+"
+env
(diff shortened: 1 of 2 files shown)


gating/pre_merge_test/run_deploy_mnaio.sh
```diff
@@ -0,0 +1,162 @@
...
+echo "+-------------------- MNAIO ENV VARS --------------------+"
+
+## Vars ----------------------------------------------------------------------
+
+# These vars are set by the CI environment, but are given defaults
+# here to cater for situations where someone is executing the test
+# outside of the CI environment.
+export RE_JOB_NAME="${RE_JOB_NAME:-}"
+export RE_JOB_IMAGE="${RE_JOB_IMAGE:-}"
+export RE_JOB_BRANCH="${RE_JOB_BRANCH:-master}"
```
mattt416: Any point defaulting this to `newton` incase someone forgets to set it?

gating/pre_merge_test/run_deploy_mnaio.sh
```diff
@@ -0,0 +1,164 @@
...
+  export OSA_BRANCH=${OSA_COMMIT}
+popd
+
+# apply various modifications for mnaio
+pushd /opt/openstack-ansible-ops/multi-node-aio
+  # By default the MNAIO deploys metering services, so we override
+  # osa_enable_meter to prevent those services from being deployed.
+  sed -i 's/osa_enable_meter: true/osa_enable_meter: false/' playbooks/group_vars/all.yml
+  # bump up cryptography version to avoid exception detailed in RLM-1104
+  pip install cryptography==1.5 --upgrade
```
mattt416: The only issue here is that there's no guarantee that pip is installed at this stage.  I just ran it on a fresh (manual) deploy and it failed because no pip.

antonym:
recheck_cit_trusty_swift_deploy

gating/pre_merge_test/run_deploy_mnaio.sh
mattt416: As discussed, this can go in until rcbops/rpc-gating#591 merges, at which point it shouldn't be necessary and can be reverted.

PR #2772, opened 2018-01-12
RLM-1104 Bump cryptography version to let MNAIO start

antonym:
Bump cryptography version to let MNAIO start and not exception.

Issue: [RLM-1104](https://rpc-openstack.atlassian.net/browse/RLM-1104)

### DIFF
--- a/gating/pre_merge_test/run_deploy_mnaio.sh
+++ b/gating/pre_merge_test/run_deploy_mnaio.sh
@@ -93,6 +93,8 @@ pushd /opt/openstack-ansible-ops/multi-node-aio
   # By default the MNAIO deploys metering services, so we override
   # osa_enable_meter to prevent those services from being deployed.
   sed -i 's/osa_enable_meter: true/osa_enable_meter: false/' playbooks/group_vars/all.yml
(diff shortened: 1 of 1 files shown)


antonym:

gating/pre_merge_test/run_deploy_mnaio.sh
```diff
@@ -93,6 +93,8 @@ pushd /opt/openstack-ansible-ops/multi-node-aio
   # By default the MNAIO deploys metering services, so we override
   # osa_enable_meter to prevent those services from being deployed.
   sed -i 's/osa_enable_meter: true/osa_enable_meter: false/' playbooks/group_vars/all.yml
+  # bump up cryptography version to avoid exception detailed in RLM-1104
+  pip install cryptography==1.5 --upgrade
```
mattt416: The only issue here is that there's no guarantee that pip is installed at this stage. I just ran it on a fresh (manual) deploy and it failed because no pip.

antonym: Yeah, I was worried about that too but pip is installed via the setup_jenkins_slave so in the gates it would work.  We should definitely revisit this but at least we can start baking this out a bit.

PR #2777, opened 2018-01-15
[Newton] RLM-1104 Install cryptography after mnaio bootstrap

antonym:
In order to work around issues from the gating env,
we have to upgrade cryptography but it needs to be
done after running the bootstrap from build.sh so
dependencies like gcc are already in place.

Once installed, we should be able to run build.sh as
normal.

Issue: [RLM-1104](https://rpc-openstack.atlassian.net/browse/RLM-1104)


antonym:
recheck_cit_container_flake8_tox-test

PR #2778, opened 2018-01-15
RLM-1120 Use dist-upgrade for loose/no artifacts

antonym:
Ensures disto is fully updating including kernel
to surface any dependency issues.

Issue: [RLM-1120](https://rpc-openstack.atlassian.net/browse/RLM-1120)

PR #2779, opened 2018-01-15
[Newton] RLM-1120 Use dist-upgrade for loose/no artifacts

antonym:
Ensures disto is fully updating including kernel
to surface any dependency issues.

Cherry-picked from https://github.com/rcbops/rpc-openstack/pull/2778

Issue: [RLM-1120](https://rpc-openstack.atlassian.net/browse/RLM-1120)

### DIFF
--- a/gating/pre_merge_test/run_deploy.sh
+++ b/gating/pre_merge_test/run_deploy.sh
@@ -48,7 +48,7 @@ if [[ ${RE_JOB_IMAGE} =~ no_artifacts$ ]]; then
   # Upgrade to the absolute latest
   # available packages.
   apt-get update
-  DEBIAN_FRONTEND=noninteractive apt-get -y upgrade
+  DEBIAN_FRONTEND=noninteractive apt-get -y dist-upgrade
 
 elif [[ ${RE_JOB_IMAGE} =~ loose_artifacts$ ]]; then
   # Set the apt artifact mode
@@ -57,7 +57,7 @@ elif [[ ${RE_JOB_IMAGE} =~ loose_artifacts$ ]]; then
   # Upgrade to the absolute latest
   # available packages.
   apt-get update
-  DEBIAN_FRONTEND=noninteractive apt-get -y upgrade
+  DEBIAN_FRONTEND=noninteractive apt-get -y dist-upgrade
 fi
 
 # Run the deployment script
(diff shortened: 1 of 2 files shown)


gating/pre_merge_test/run_deploy.sh
```diff
@@ -48,7 +48,7 @@ if [[ ${RE_JOB_IMAGE} =~ no_artifacts$ ]]; then
   # Upgrade to the absolute latest
   # available packages.
   apt-get update
-  DEBIAN_FRONTEND=noninteractive apt-get -y upgrade
+  DEBIAN_FRONTEND=noninteractive apt-get -y dist-upgrade
```
odyssey4me: Doing this upgrade may upgrade the kernel packages, but won't make them active. It requires a reboot to be active as far as I know. Is this what you want?

gating/pre_merge_test/run_deploy_mnaio.sh
```diff
@@ -109,11 +109,11 @@ env | grep RE_ | sed 's/^/export /' > /opt/rpc-openstack/RE_ENV
 # check if we're using artifacts or not
 if [[ ${RE_JOB_IMAGE} =~ no_artifacts$ ]]; then
   echo "export RPC_APT_ARTIFACT_ENABLED=no" >> /opt/rpc-openstack/RE_ENV
-  ${MNAIO_SSH} "apt-get -qq update; DEBIAN_FRONTEND=noninteractive apt-get -y upgrade"
+  ${MNAIO_SSH} "apt-get -qq update; DEBIAN_FRONTEND=noninteractive apt-get -y dist-upgrade"
```
odyssey4me: Given that this is just a host for the VM's, is this even necessary?

gating/pre_merge_test/run_deploy.sh
antonym: Yeah, @cloudnull and I chatted and given there are some package dependencies on the kernel, this should still simulate the dependency tree without the reboot.

gating/pre_merge_test/run_deploy_mnaio.sh
antonym: It's actually running the command on the infra1 host which will be doing the deployment.

antonym: Technically this will probably noop given the VMs are freshly provisioned.

PR #2780, opened 2018-01-16
RE-1267 Ensure host has python/python-yaml

odyssey4me:
In order to get the value for rpc_release, we need to use
a python script, so we need to make sure that python and
the python yaml library are installed. If not, the value
for rpc_release cannot be read and the check for the value
will result in a null return, causing all artifacts to be
disabled.

Also, when deploying RPC-O master using strict mode for
artifacts, the upstream elasticsearch repo is used instead
of the apt artifacts repo. This patch implements the correct
overrides to resolve this issue.

Issue: [RE-1267](https://rpc-openstack.atlassian.net/browse/RE-1267)

Issue: [RO-3624](https://rpc-openstack.atlassian.net/browse/RO-3624)

### DIFF
--- a/gating/pre_merge_test/run_deploy.sh
+++ b/gating/pre_merge_test/run_deploy.sh
@@ -28,6 +28,15 @@ elif [[ ${RE_JOB_IMAGE} =~ loose_artifacts$ ]]; then
   DEBIAN_FRONTEND=noninteractive apt-get -y upgrade
 fi
 
+# In order to get the value for rpc_release, we need to use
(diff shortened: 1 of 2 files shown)


odyssey4me:
recheck_cit_xenial_swift_deploy gating venv download problem

odyssey4me:
The artifact test has passed all but the container builds, which are failing due to https://rpc-openstack.atlassian.net/browse/RO-3626 - so merging this patch.

PR #2782, opened 2018-01-16
Add tasks to revert artifacts if they're disabled

prometheanfire:
Presently if artifacts are enabled there's no way to automate their
removal. This change adds tasks to backup files we modify and to
restore the backs (original) files should a deployer chose to disable
the artifacts.


cloudnull:
Sorry for the delay but https://github.com/rcbops/rpc-openstack/pull/2720 has been rebased with the needed fixes to address the moved var file(s)

prometheanfire:
so it worked :D

PR #2784, opened 2018-01-17
RLM-1136 Lookup RPC_APT_ARTIFACT_MODE env var

antonym:
AIO looks up the environment variable but doing a
non-AIO deploy does not.

Issue: [RLM-1136](https://rpc-openstack.atlassian.net/browse/RLM-1136)

### DIFF
--- a/group_vars/all/apt.yml
+++ b/group_vars/all/apt.yml
@@ -134,4 +134,4 @@ hwraid_apt_keys: "{{ rpco_apt_gpg_keys }}"
 #  being deployed.
 # This setting may be overridden by overriding it in the
 # /etc/openstack_deploy/user_*.yml files.
-rpco_apt_artifacts_mode: "strict"
+rpco_apt_artifacts_mode: "{{ lookup('env', 'RPC_APT_ARTIFACT_MODE') | default('strict', true) }}"


group_vars/all/apt.yml
```diff
@@ -134,4 +134,4 @@ hwraid_apt_keys: "{{ rpco_apt_gpg_keys }}"
 #  being deployed.
 # This setting may be overridden by overriding it in the
 # /etc/openstack_deploy/user_*.yml files.
-rpco_apt_artifacts_mode: "strict"
+rpco_apt_artifacts_mode: "{{ lookup('env', 'RPC_APT_ARTIFACT_MODE')| default('RPC_APT_ARTIFACT_MODE', 'strict') }}"
```
odyssey4me: I'm not sure this is going to work as expected. Did you mean to do this?

  rpco_apt_artifacts_mode: "{{ lookup('env', 'RPC_APT_ARTIFACT_MODE') | default('strict', true) }}"

Even then, this may cause problems later because this env var may not be set when executing playbooks on day two or three just using 'openstack-ansible ...'.

It may be better to use:

  ansible_local['rpc_openstack']['rpc_artifacts']['apt_artifact_mode'] | default('strict')

antonym: K, tested that and it works fine, updated PR.

PR #2785, opened 2018-01-17
RE-1280 Update rpc-artifacts submodule

odyssey4me:
Issue: [RE-1280](https://rpc-openstack.atlassian.net/browse/RE-1280)


odyssey4me:
As this update cannot be exercised through any of these tests, merging the PR.

PR #2790, opened 2018-01-19
RO-3654 Update OSA SHA for apt-cacher-ng fix

odyssey4me:
With the repo cache enabled, while using container artifacts,
the apt-cacher-ng log folder is not setup after the initial
bind mount. As a result apt-cacher-ng fails to start. This
issue is fixed in https://review.openstack.org/532581.

As there have been no python package SHA bumps in newton, this
patch brings the OSA SHA forward to include the upstream patch.


odyssey4me:
recheck_cit_artifact

group_vars/all/apt.yml
```diff
@@ -39,6 +39,12 @@ rpco_gpg_key_id: 52AA252F #SET IN STATIC (to force key verification per release)
 lxc_container_cache_files_from_host:
   - "/etc/apt/sources.list.d/{{ rpco_mirror_apt_filename }}.list"
 
+# We won't be using the repo package cache, as we'll have a full mirror.
```
cloudnull: We are in fact using the repo cache, even with our full mirror. Without the repo cache every host and container will connect to the rpc-repo which makes deploying painfully slow especially in remote cdc locations. IMO the best solution would be to move the SHA forward to pull in the fix in or make the missing directory accordingly.

odyssey4me: Pulling the SHA forward requires an artifact rebuild which negates the soak test for the current RC. This change implements the minimal change needed which does not require an artifact rebuild.

odyssey4me: Actually, scratch that. I see that there's been no further python SHA bumps in OSA newton so I'll adjust this to bring the OSA SHA forward.

PR #2795, opened 2018-01-22
RO-3729 Source functions correctly in post_deploy

odyssey4me:
The functions currently fail to source due to the wrong path.

Issue: [RO-3729](https://rpc-openstack.atlassian.net/browse/RO-3729)


odyssey4me:
The ``CIT/artifact`` job will fire after https://github.com/rcbops/rpc-gating/pull/607 merges and a recheck for it is issued.

PR #2802, opened 2018-01-23
RO-3316 Enable container images for MNAIO tests

odyssey4me:
As discussed in [1] the container images should now
be used in build tests, rather than disabled. The
setting for the MNAIO tests was missed in [1] so
this patch resolves that.

[1] https://github.com/rcbops/rpc-openstack/pull/2797

(cherry picked from commit 94fee0467a75b8fedb75ae591994d2915f802990)

Issue: [RO-3316](https://rpc-openstack.atlassian.net/browse/RO-3316)


odyssey4me:
As the PR tests aren't touching any of these changes, this is being merged ahead of the test completion.

PR #2805, opened 2018-01-25
Notify users & protect the rpcm var file

cloudnull:
While we expect users to largely ignore this file in deployments,
especially given the fact that it's only used it when artifacts are
enabled, we have the ability to protect the file and notify operators
that this file should not be written to. This change sets the mode of
the user variable files we put into place to 0444 which will, by
default, make the file read only. This change also adds a comment to the
top of the file notifying users that any change written to the rpcm
variable file will be undone should the playbooks be executed.

Issue: [RO-3440](https://rpc-openstack.atlassian.net/browse/RO-3440)


prometheanfire:
recheck_cit_artifact

PR #2812, opened 2018-01-31
RE-1304 Add workaround for nova-novncproxy fail

hughsaunders:
For all newton releases up to r14.8.0 when executing the os-nova-install.yml
playbook the ``nova-novncproxy`` and ``nova-spicehtml5proxy`` services will
fail. The workaround to resolve this issue is to restart the services.

This patch adds a release note to communicate the known issue. The fix will
be brought into the r14.9.0 release as part of the upstream dependency
updates.

(cherry picked from commit e57b236fc064513b59d774fe8ecbdc947ef48dfb)

Issue: [RE-1304](https://rpc-openstack.atlassian.net/browse/RE-1304)

### DIFF
--- a/releasenotes/notes/nova-novncproxy-restart-6aa641c97867e6ef.yaml
+++ b/releasenotes/notes/nova-novncproxy-restart-6aa641c97867e6ef.yaml
@@ -1,0 +1,17 @@
+---
+issues:
+  - |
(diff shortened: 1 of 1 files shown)


hughsaunders:
This was done in #2812

hughsaunders:
This was done in #2811

PR #2817, opened 2018-02-01
RLM-322 Change openstack_hosts role to rpc one

Frank-ZhangXin:
Link the rpc openstack_hosts role which is removed period mark from
templates/openstack-host-hostfile-setup.sh.j2

### DIFF
--- a/ansible-role-requirements.yml
+++ b/ansible-role-requirements.yml
@@ -16,4 +16,4 @@
 - name: openstack_hosts
   scm: git
   src: https://github.com/rcbops/openstack-ansible-openstack_hosts
-  version: ce249ec4386d013d76b617f009a93a408f76619e 
+  version: 235ec6f0fc5570e7c657b243bcf4208cc2c7739c
--- a/ansible-role-requirements.yml
+++ b/ansible-role-requirements.yml
@@ -10,3 +10,7 @@
   scm: git
   src: https://github.com/ceph/ansible-ceph-osd.git
   version: 434a13f4d3ee3f60c410052d2670d5dfea034bf8
+- name: openstack_hosts
+  scm: git
+  src: https://github.com/rcbops/openstack-ansible-openstack_hosts
+  version: ce249ec4386d013d76b617f009a93a408f76619e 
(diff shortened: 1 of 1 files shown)


ansible-role-requirements.yml
```diff
@@ -10,3 +10,7 @@
   scm: git
   src: https://github.com/ceph/ansible-ceph-osd.git
   version: 434a13f4d3ee3f60c410052d2670d5dfea034bf8
+- name: openstack_hosts
```
antonym: Let's add a comment here detailing why we're forking the repo and include the JIRA number as well so we can reference this change easily in the future.

antonym: Other than that, it looks good, points to the https://github.com/rcbops/openstack-ansible-openstack_hosts/tree/leapupgrade-version which is the change on top of the newton code.

antonym:
recheck_all

ansible-role-requirements.yml
Frank-ZhangXin: Yeah, gonna add comment to explain the separated role.

antonym:

Frank-ZhangXin:
@antonym comment is added to the role requirement.

ansible-role-requirements.yml
```diff
@@ -10,3 +10,10 @@
   scm: git
   src: https://github.com/ceph/ansible-ceph-osd.git
   version: 434a13f4d3ee3f60c410052d2670d5dfea034bf8
+# Created for JIRA RLM-322
+# Frank: this is a forked ansible role for removing
+# hardcoded period mark from hostname template.
+- name: openstack_hosts
+  scm: git
+  src: https://github.com/rcbops/openstack-ansible-openstack_hosts
+  version: ce249ec4386d013d76b617f009a93a408f76619e
```
antonym: Merged in PR on openstack-ansible-openstack_hosts, updated stable/newton version is now: 235ec6f0fc5570e7c657b243bcf4208cc2c7739c

Frank-ZhangXin: cool, update is on the way.

PR #2821, opened 2018-02-05
RE-1309 Ensure update_dependencies script works

odyssey4me:
Currently if the update_dependencies script is run from a
directory which is not /opt/rpc-openstack, the get-rpc_release.py
script returns an empty result. This causes the version comparison
between the RC branch the the current development head to always
fail, so the version is never incremented.

This patch ensures that the RPC_RELEASE value is always properly
returned, and also ensures that the update_dependencies script is
easily tested on a test VM. Various updates are made to the script
to make it easier to re-run and to make it more robust.

The extended explanation for skipping the version bump is removed
due to bashate being incapable of understanding that the text using
the word "for" does not mean it's a bash for loop.

Adjustments are also made to other gating scripts which now fail due
to the new verification being done.

Issue: [RE-1309](https://rpc-openstack.atlassian.net/browse/RE-1309)

### DIFF
--- a/gating/update_dependencies/run
+++ b/gating/update_dependencies/run
@@ -1,11 +1,40 @@
-#!/bin/bash -xe
+#!/bin/bash
 
+# Set options to ensure easier debugging
+# and failure on any problems encountered
+set -euxo pipefail
+
+# Find the absolute path to the directory this script
+# is executed from.
 export GATING_PATH="$(readlink -f $(dirname ${0}))"
-source "${GATING_PATH}/../../scripts/functions.sh"
+
+# Set the base RPC-O directory which the functions use
+export BASE_DIR="$(readlink -f ${GATING_PATH}/../..)"
+
+# As get-rpc_release.py requires a yaml import, make
+# sure that it is installed and available.
+if ! python -c 'import yaml' &>/dev/null; then
+  echo "Python YAML library not available. Installing it."
+  apt-get install -y python-yaml
+fi
+
+# Source the functions
+source "${BASE_DIR}/scripts/functions.sh"
+
+# We need pip on the host in order to install the required
+# SemVer library, so make sure it is installed and available.
+if ! pip --version &>/dev/null; then
+  echo "Pip not available. Installing it."
+  apt-get install -y python-pip
+fi
 
 # Install the semver library to manipulate the
 # rpc_release version strings
-pip install semver==2.7.9
+semver_version="2.7.9"
+if [[ "$(pip --disable-pip-version-check freeze | grep semver)" != "semver==${semver_version}" ]]; then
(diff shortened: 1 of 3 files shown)


gating/update_dependencies/run
```diff
@@ -1,11 +1,47 @@
...
+# sure that it is installed and available.
+if ! python -c 'import yaml' &>/dev/null; then
+  echo "Python YAML library not available. Installing it."
+  apt-get install -y python-yaml
+fi
+
+# Source the functions
+source "${BASE_DIR}/scripts/functions.sh"
+
+# Validate that RPC_RELEASE is set and has a value
```
git-harry: Is there are reason this only needs to be validated for the dependency update? Maybe it would be better to have this validation in the functions file. I think other scripts should be able to expect that if sourcing the file didn't fail, things are set up correctly. Validating in that file also means that all jobs/deployments benefit from this check.

gating/update_dependencies/run
```diff
@@ -1,11 +1,47 @@
...
+  echo "Pip not available. Installing it."
+  apt-get install -y python-pip
+fi
 
 # Install the semver library to manipulate the
 # rpc_release version strings
-pip install semver==2.7.9
+if ! python -c 'import semver' &>/dev/null; then
+  echo "Python SemVer library not available. Installing it."
+  pip install semver==2.7.9
```
git-harry: If this requires a specific version the test that it's installed is too loose.

gating/update_dependencies/run
```diff
@@ -52,27 +100,23 @@ release_data_file="${WORKSPACE}/rc-release-data.yml"
...
 # old method
 elif git cat-file -e ${old_file_to_fetch} 2>/dev/null; then
   git show ${old_file_to_fetch} > ${release_data_file}
   export RC_BRANCH_VERSION=$(awk '/^rpc_release/{print $2}' ${release_data_file} | tr -d '"')
 
 else
-  echo "RC branch ${rc_branch} not found, skipping rpc_release version bump.
-If there is no RC branch then the mainline branch is considered unreleased
-and therefore the rpc_release value is left alone. It is still important
-for the dependencies to be updated regularly though, so that part continues
```
git-harry: If you move `for` to the previous line it will pass the test.

odyssey4me: Great, thanks!

gating/update_dependencies/run
odyssey4me: Good call - updated to be more specific.

gating/update_dependencies/run
odyssey4me: Good call - moved this to the functions.

git-harry:

odyssey4me:

PR #2824, opened 2018-02-05
RO-3937 Add ironic/nova to EOL repo list

odyssey4me:
Both ironic and nova have deleted the stable/newton branch
and need to now make use of the EOL tags. When building
the containers without the python artifacts we use developer
mode for the role execution, so we need to set the correct
var to ensure the EOL tag is used.

Issue: [RO-3937](https://rpc-openstack.atlassian.net/browse/RO-3937)

### DIFF
--- a/scripts/artifacts-building/containers/build-process.sh
+++ b/scripts/artifacts-building/containers/build-process.sh
@@ -131,10 +131,10 @@ if ! python_artifacts_available; then
     # which no longer exists.
     # We implement the override here to cater for PR tests
     # only.
-    eol_repo_list="cinder glance heat horizon keystone magnum_dashboard"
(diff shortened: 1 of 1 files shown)


scripts/artifacts-building/containers/build-process.sh
```diff
@@ -131,10 +131,10 @@ if ! python_artifacts_available; then
     # which no longer exists.
     # We implement the override here to cater for PR tests
     # only.
-    eol_repo_list="cinder glance heat horizon keystone magnum_dashboard"
+    eol_repo_list="cinder glance heat ironic horizon keystone magnum_dashboard"
```
git-harry: The list appears to be alphabetical, so this is in the wrong place.

odyssey4me: Ah, good catch - fixing.

PR #2826, opened 2018-02-05
RO-3856 Remove container artifact usage

odyssey4me:
Trying to use pre-built container images with the current
deployment system has turned out to be a bad idea. It has
raised a constant game of whack-a-mole with missing log
file folders [1][2][3][4][5] and other yet-to-find obscure
errors due to the way that OSA was designed to be deployed
(everything done at runtime) versus the desired RPC-O
implementation (pre-build as much as possible).

While we go back to the drawing board and rather implement
the artifact-based deployment method upstream, it'd be
better for us to cut out this technical debt so that we
have better deployment and upgrade success while we revisit
this approach upstream for a more sustainable solution.

This patch removes the artifacted container image usage
for deployments, but still allows the usage of apt and
python artifacts. The patch still maintains the promise of
what we deploy today is the same as what we deploy tomorrow
based on the fact that a container image was always just a
pre-built set of bits coming from the apt & python artifacts.

[1] RO-3654
[2] RO-3856
[3] RLM-1094
[4] RLM-324
[5] RO-3522

Issue: [RO-3856](https://rpc-openstack.atlassian.net/browse/RO-3856)


major:
@odyssey4me I'm going to need this split into at least 15 separate patches... :trollface:

odyssey4me:
@major https://media.giphy.com/media/yJpIlhh6ZTqHm/giphy.gif

odyssey4me:
Updated to clean up the last remnants of ``container_artifact_enabled`` from the README and ``playbooks/configure-apt-sources.yml``.

PR #2830, opened 2018-02-06
RO-3856 Remove container artifact usage (newton)

odyssey4me:
Trying to use pre-built container images with the current
deployment system has turned out to be a bad idea. It has
raised a constant game of whack-a-mole with missing log
file folders [1][2][3][4][5] and other yet-to-find obscure
errors due to the way that OSA was designed to be deployed
(everything done at runtime) versus the desired RPC-O
implementation (pre-build as much as possible).

While we go back to the drawing board and rather implement
the artifact-based deployment method upstream, it'd be
better for us to cut out this technical debt so that we
have better deployment and upgrade success while we revisit
this approach upstream for a more sustainable solution.

This patch removes the artifacted container image usage
for deployments, but still allows the usage of apt and
python artifacts. The patch still maintains the promise of
what we deploy today is the same as what we deploy tomorrow
based on the fact that a container image was always just a
pre-built set of bits coming from the apt & python artifacts.

Given that all files removed are in-tree, and never copied
into user-space, there is no need for us to add any clean-up
of user-space files. For an upgrade from any existing newton
version the current version, container artifacts are not used
anyway (they're only used for greenfield deployments). As
such there are no upgrade concerns relating to the removal
of this code. This will, in fact, make the greenfield
deployment more reliable as it will no longer suffer strange
and inconsistent errors relating to the bind mounting of
the host log folder over the container's var/log folder.

[1] RO-3654
[2] RO-3856
[3] RLM-1094
[4] RLM-324
[5] RO-3522

Issue: [RO-3856](https://rpc-openstack.atlassian.net/browse/RO-3856)


odyssey4me:
recheck_cit_all

odyssey4me:
recheck_cit_all

odyssey4me:
recheck_issue_link

odyssey4me:
recheck_cit_artifact now that https://github.com/rcbops/rpc-gating/pull/637 has merged
recheck_cit_xenial_swift_deploy as it seems it was a victim of an apt mirror failure

odyssey4me:
recheck_cit_artifact now that rcbops/rpc-gating#637 has merged
recheck_cit_xenial_swift_deploy as it seems it was a victim of an apt mirror failure

PR #2839, opened 2018-02-08
RO-3952 Revert elasticsearch SHA update

git-harry:
The update of the elasticsearch role SHA introduced a breaking change.
This reverts the SHA to its previous value while a fix is prepared to
move the SHA on again.

Reverts the change introduced by https://github.com/rcbops/rpc-openstack/pull/2815

Issue: [RO-3952](https://rpc-openstack.atlassian.net/browse/RO-3952)

PR #2841, opened 2018-02-08
Add elasticsearch_major_version

d34dh0r53:
The updated ansible role from elastic.co does checking on the
es_major_version, we need to set it to the correct major version in
order to pull from the appropriate repository.

Fixes: RO-3952

Issue: [RO-3952](https://rpc-openstack.atlassian.net/browse/RO-3952)


prometheanfire:
Do we use the `elasticsearch_*` variables other than to set `es_*` variables?  I'm just trying to determine if we need the additional layer of indirection or not.

d34dh0r53:
@prometheanfire we do, so I just followed the pattern. It's probably not necessary and I may clean that up in the future.

odyssey4me:
recheck_cit_xenial_no_artifacts_swift_deploy now that https://github.com/rcbops/rpc-openstack/pull/2839 has merged

PR #2842, opened 2018-02-08
RO-3973 Override MariaDB repo configuration

odyssey4me:
MariaDB released new packages yesterday which breaks MariaDB
clustering for multi-node deployments. This breaks non-artifacted
builds which use MariaDB 10.1.

As a workaround until an upstream fix is implemented, we
override the defaults set in OSA to ensure that we use the
last known good version of MariaDB.

As these settings are unnecessary when using apt artifacts,
the overrides are removed if apt artifacts are enabled so
that only the apt artifacts repo is used.

Issue: [RO-3973](https://rpc-openstack.atlassian.net/browse/RO-3973)

### DIFF
--- a/etc/openstack_deploy/group_vars/all/mariadb.yml
+++ b/etc/openstack_deploy/group_vars/all/mariadb.yml
@@ -1,0 +1,25 @@
+---
+# Copyright 2017, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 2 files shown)


odyssey4me:
The master branch functional test will not pass until https://github.com/rcbops/rpc-openstack/pull/2841 and/or https://github.com/rcbops/rpc-openstack/pull/2839 merges. The second is probably best to merge as it takes us back to a known good state, then the other can be rebased and also update the SHA along with it. Once one of those has merged, this can be rechecked to validate that it's not breaking anything.

odyssey4me:
recheck_cit_xenial_no_artifacts_swift_deploy now that https://github.com/rcbops/rpc-openstack/pull/2839 has merged

PR #2843, opened 2018-02-08
RO-3857 Add support for MNAIO log collection

git-harry:
This change replaces the rsync commands used to gather logs and
configuration files on an AIO with a playbook that can be used on both
AIOs and MNAIOs. For an AIO the playbook targets localhost and for an
MNAIO the playbook uses the MNAIO inventory host file created by
openstack-ansible-ops.

Issue: [RO-3857](https://rpc-openstack.atlassian.net/browse/RO-3857)

### DIFF
--- a/gating/pre_merge_test/collect_logs.yml
+++ b/gating/pre_merge_test/collect_logs.yml
@@ -1,0 +1,102 @@
+- hosts: "{{ target_hosts }}"
+  gather_facts: false
+  tasks:
+    - name: Ensure artefacts directories exists
+      file:
+        path: "{{ item }}"
+        state: directory
+      with_items:
+        - "{{ artifacts_dir }}/{{ inventory_hostname }}/host"
+        - "{{ artifacts_dir }}/{{ inventory_hostname }}/containers"
+      delegate_to: "localhost"
+
+    - name: Grab host data
+      command: >
+               rsync
+               --archive
+               --compress
+               --verbose
+               --relative
+               --rsh 'ssh -o StrictHostKeyChecking=no'
+               --ignore-missing-args
+               --safe-links
+               --no-perms
+               --no-owner
+               --no-group
+               {{ inventory_hostname }}:{{ item }}
+               {{ artifacts_dir }}/{{ inventory_hostname }}/host
+      with_items:
+        - "/openstack/log"
+        - "/etc"
+        - "/var/./log"
+      delegate_to: "localhost"
+      ignore_errors: true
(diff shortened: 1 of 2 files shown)


odyssey4me:
recheck_cit_artifact

gating/pre_merge_test/collect_logs.yml
```diff
@@ -0,0 +1,99 @@
...
+      with_nested:
+        - "{{ rootfilesystems.results }}"
+        -
+          - "etc"
+          - "var/log"
+      delegate_to: "localhost"
+      tags:
+        - skip_ansible_lint
+  vars:
+    artifacts_dir: "/tmp/artifacts"
```
odyssey4me: As per the changes in https://github.com/rcbops/rpc-gating/pull/638, this is a hard-coded dir, rather than the actual hook dir. Can we change this to rather pull to the defined hook dir, with a default to this in case that is not defined?

gating/pre_merge_test/collect_logs.yml
```diff
@@ -0,0 +1,99 @@
...
+      register: results_files
+
+    - name: Copy tempest results to RE_HOOK_RESULT_DIR
+      copy:
+       src: "{{ item.path }}"
+       dest: "{{ result_dir }}/"
+      with_items: "{{ results_files.files }}"
+      when: results_files.matched > 0
+  vars:
+    result_dir: "/tmp/result"
```
odyssey4me: This should be ``RE_HOOK_RESULT_DIR`` per the ``RE for Projects`` page.

git-harry: This is a default, take a look at gating/pre_merge_test/post_deploy.sh to see that it is being passed in by the script.

gating/pre_merge_test/collect_logs.yml
git-harry: This is a default, take a look at gating/pre_merge_test/post_deploy.sh to see that it is being passed in by the script.

gating/pre_merge_test/collect_logs.yml
odyssey4me: Ah, I see. That could be removed if this was the following, and it would be a bit less confusing I think:

``{{ lookup('env', 'RE_HOOK_RESULT_DIR') | default('/tmp/result', true) }}``

git-harry:

git-harry:
recheck_release

PR #2849, opened 2018-02-12
RE-36 Update freeze and thaw process

mattt416:
This commit updates gating/pre_merge_test/post_deploy.sh to include
the full gating-specified image name and scenario in the name used to
create the instance snapshot. This is necessary to differentiate the
different artifact modes and whether the scenario is swift, ironic, etc.

Additionally, we update gating/thaw/thaw.yml to use lxc-autostart to
containers so that the start order defined in each container's config
is honoured.

Issue: [RE-36](https://rpc-openstack.atlassian.net/browse/RE-36)


odyssey4me:
@mattt416 Is this still WIP?

mattt416:
@major @odyssey4me WIP removed, but pushed a small additional change.  I'll need to backport this one to newton when it merges.  Thanks!

PR #2867, opened 2018-03-02
(ASC-144) Source MNAIO_SSH from an exported file

johnduarte:
This commit sources the `MNAIO_SSH` variable in the `run` script
from a file created by the `run_deploy_mnaio.sh`. This mechanism is
used to make the variable available to the parent shell in order to
pass it to the `run_system_tests.sh` script as expected.

Because the variable is created by a sub-shell to the one used by
the `run` command, exporting it does not make it available to the
parent shell. To get around this, the variable is stored to an
external file and sourced after the `run_deploy_mnaio.sh` script has
returned control to the parent.

Issue: [ASC-144](https://rpc-openstack.atlassian.net/browse/ASC-144)

### DIFF
--- a/gating/pre_merge_test/run
+++ b/gating/pre_merge_test/run
@@ -13,6 +13,7 @@ echo "+-------------------- ENV VARS --------------------+"
 ## Vars ----------------------------------------------------------------------
 
 export DEPLOY_AIO="true"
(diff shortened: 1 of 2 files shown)


mattt416:

mattt416:
recheck_cit_release
recheck_cit_artifact

johnduarte:
recheck_issue_link

PR #2873, opened 2018-03-06
update instructions for passing defcore

prometheanfire:
Issue: [asc-144](https://rpc-openstack.atlassian.net/browse/asc-144)


prometheanfire:
1 review required, 1 found

PR #2878, opened 2018-03-12
Asc 219 manage exits

prometheanfire:
The following commits are in this PR

* Fix spelling errors in run_system_tests.sh script

* update run_system_tests.sh to record/use exit code
Record an exit code if the previous command fails.
Use that exit code to exit at the very end so collection can work.

* update run_system_tests.sh for formatting

Issue: ASC-219

Issue: [RO-3972](https://rpc-openstack.atlassian.net/browse/RO-3972)

### DIFF
--- a/gating/pre_merge_test/run_system_tests.sh
+++ b/gating/pre_merge_test/run_system_tests.sh
@@ -21,8 +21,7 @@ export SYS_INVENTORY="/opt/openstack-ansible/playbooks/inventory"
 
 ## Main ----------------------------------------------------------------------
 
-
-# 1. Clone test repo into working directory.
+# 1. Clone test repository into working directory.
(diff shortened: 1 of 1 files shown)


gating/pre_merge_test/run_system_tests.sh
```diff
@@ -21,28 +21,38 @@ export SYS_INVENTORY="/opt/openstack-ansible/playbooks/inventory"
 
 ## Main ----------------------------------------------------------------------
 
-
-# 1. Clone test repo into working directory.
+# 1. Clone test repository into working directory.
```
johnduarte: `-e` needs to be set here for this section to fail hard if an error is encountered prior to step # 2.

PR #2881, opened 2018-03-13
(RI-213) Run tempest tests on MNAIO

johnduarte:
This commit runs the tempest playbook on MNAIO deployments after the
deployment has concluded.

Issue: [RI-213](https://rpc-openstack.atlassian.net/browse/RI-213)

### DIFF
--- a/gating/pre_merge_test/run
+++ b/gating/pre_merge_test/run
@@ -33,6 +33,13 @@ if [[ $RE_JOB_ACTION == "tox-test" ]]; then
 elif [[ $RE_JOB_IMAGE =~ .*mnaio.* ]]; then
   bash -c "$(readlink -f $(dirname ${0})/run_deploy_mnaio.sh)"
   source "${MNAIO_VAR_FILE}"
+  if [[ $RE_JOB_ACTION == "deploy" ]]; then
+    # run tempest tests on MNAIO for deploy action
+    ${MNAIO_SSH} <<EOS
+      cd /opt/openstack-ansible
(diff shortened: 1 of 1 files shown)


gating/pre_merge_test/run
```diff
@@ -33,6 +33,11 @@ if [[ $RE_JOB_ACTION == "tox-test" ]]; then
 elif [[ $RE_JOB_IMAGE =~ .*mnaio.* ]]; then
   bash -c "$(readlink -f $(dirname ${0})/run_deploy_mnaio.sh)"
   source "${MNAIO_VAR_FILE}"
+  # run tempest tests on MNAIO after deploy
+  ${MNAIO_SSH} <<EOS
+    cd /opt/openstack-ansible
+    openstack-ansible playbooks/os-tempest-install.yml
```
odyssey4me: This will only install it, not run it - unless you have ``tempest_run: yes`` set somewhere? To make it also run you can add ``-e tempest_run=yes`` to this command.

johnduarte:
Thanks @odyssey4me . I have updated the command to include the `-e tempest_run=yes` setting.

odyssey4me:
@johnduarte One thing I've just thought of is that this will execute tempest for *all* MNAIO builds. I thought you wanted to have tempest for the ``deploy`` action, but execute other things for the ``system`` action? Are you happy to have tempest execute with the system tests too?

johnduarte:
Thanks @odyssey4me . I have updated to only execute on the `deploy` action. Cheers.

PR #2895, opened 2018-03-21
Fix for get-rpc_release.py ignoring RPC_PRODUCT_RELEASE

shannonmitchell:
The scripts/get-rpc_release.py used by the deployment to allow
for what looks like branchless deploments was ignoring the
RPC_PRODUCT_RELEASE.  It looks like it argparse pulls that
environment variable to use for a release_series variable.
A default of 'queens' was being set as well.  The EnvDefault
class being used by that argument only uses the environment
variable if no default is set.  This change is to allow any
setting of RPC_PRODUCT_RELEASE to override or set the default.

Issue: RO-4110

Issue: [RO-4110](https://rpc-openstack.atlassian.net/browse/RO-4110)

### DIFF
--- a/scripts/get-rpc_release.py
+++ b/scripts/get-rpc_release.py
@@ -22,9 +22,8 @@ import yaml
 # Sourced from: https://stackoverflow.com/a/10551190
 class EnvDefault(argparse.Action):
     def __init__(self, envvar, required=True, default=None, **kwargs):
-        if not default and envvar:
-            if envvar in os.environ:
-                default = os.environ[envvar]
+        if envvar in os.environ:
+            default = os.environ[envvar]
         if required and default:
             required = False
         super(EnvDefault, self).__init__(default=default, required=required,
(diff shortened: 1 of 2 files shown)


scripts/get-rpc_release.py
```diff
@@ -22,7 +22,7 @@
 # Sourced from: https://stackoverflow.com/a/10551190
 class EnvDefault(argparse.Action):
     def __init__(self, envvar, required=True, default=None, **kwargs):
-        if not default and envvar:
+        if envvar:
```
coreywright: i suggest just dropping this check because `envvar` should always be set to something [1] significant [2].

this would make the order of preference:
1. user-provided argument
1. user-provided environment variable
1. dev-provided `default`

[1] it doesn't have a default arg value
[2] otherwise why are we using the EnvDefault class for the argparse argument action

shannonmitchell:

shannonmitchell:

coreywright:
@odyssey4me i reviewed it from a python perspective, i just told @shannonmitchell  that it needed someone with domain experience (ie rpc-openstack) to review it (to make sure that it's doing as intended within the context of rpc-openstack).

PR #2897, opened 2018-03-22
[master] RO-3976 Remove mariadb.yml from group_vars

cjloader:
MariaDB was fixed upstream so we neeed to delete the variable file.

Issue: [RO-3976](https://rpc-openstack.atlassian.net/browse/RO-3976)


prometheanfire:
are we consuming the fix though (didn't think we had time to update our sha).

PR #2907, opened 2018-03-29
INTAKERPC-132 Remove artifacting and references to rpc-repo

odyssey4me:
Given that the artifacting code path is now totally untested,
and the artifacts are no longer being built either, it is best
to remove the code entirely so as not to provide any false
promises to the consumers of pike releases.

This patch borrows from [1] and [2] and some other subsequent
patches to the queens implementation.

[1] e250dc4f27011e1b534e4a5d6a09b7a13c8bc559
[2] 0aeffb2704608beb59ca3224fc58b2b521c1aa21


odyssey4me:
Marking don't merge for now until the next release is done to ensure that we have a long period of testing to find any issues caused by this.

PR #2913, opened 2018-04-02
DNM: allow ansible role overrides

prometheanfire:
Issue: [RO-4125](https://rpc-openstack.atlassian.net/browse/RO-4125)

### DIFF
--- a/ansible-role-overrides.yml
+++ b/ansible-role-overrides.yml
@@ -1,0 +1,5 @@
+---
+- name: pip_install
+  scm: git
+  src: https://git.openstack.org/openstack/openstack-ansible-pip_install
+  version: 78e615c712771e33c1a7436e05bc91644318ece1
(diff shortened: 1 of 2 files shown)


prometheanfire:
if this fails, we could try this sha instead 78e615c712771e33c1a7436e05bc91644318ece1

https://github.com/openstack/openstack-ansible-pip_install/commit/78e615c712771e33c1a7436e05bc91644318ece1

odyssey4me:
FYI @cloudnull specifically implemented Pike+ in such a way that we can not implement downstream overrides in order to force fixes to go upstream and for any and all focus for development to be as upstream as possible. Implementing something like this allows us to do what we should not be doing and put our focus downstream again.

prometheanfire:
yep, this was a DNM patch for a reason :P (testing).  It looks like we can close this anyway now that upstream merged.

PR #2915, opened 2018-04-05
RO-4147 Use forked repo_build role to fix PBR build issue

odyssey4me:
A new version of PBR was released which causes some services, like
gnocchi, to break in new and exciting ways. This change installs PBR
within the repo_build role which will ensure all of the correct tools
are available to the repo build process and are installed using known
constraints.


odyssey4me:
recheck_cit_artifact It would appear that the rackspace Ubuntu mirror flaked out mid-test for the apt artifact build.

odyssey4me:
The IAD rackspace ubuntu mirrors are out of sync, which is why the artifact builds are failing. I'm in #linux-patching via slack hoping to get that sorted out.

odyssey4me:
recheck_cit_artifact The mirrors are updated now.

odyssey4me:
recheck_cit_artifact

PR #2919, opened 2018-04-06
RE-1399 Run release update when RC branch exists

mattt416:
The master-rc branch was deleted today, causing the dep_update PR job
to fail. This commit updates gating/update_dependencies/run by setting
RC_BRANCH_VERSION to an empty string and then conditionally running
release-update.py when RC_BRANCH_VERSION is not an empty string.

Issue: [RE-1399](https://rpc-openstack.atlassian.net/browse/RE-1399)

### DIFF
--- a/gating/update_dependencies/run
+++ b/gating/update_dependencies/run
@@ -101,6 +101,7 @@ else
 If there is no RC branch then the mainline branch is considered unreleased and
 therefore the rpc_release value is left alone. It is still important for the
 dependencies to be updated regularly though, so that part continues to be done."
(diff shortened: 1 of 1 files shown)


git-harry:
Setting it to an empty string looks fine to me, I think it's preferable to a string representing a null value. I agree with @odyssey4me that it should be set in the else block at around line 100 and that the conditional introduced by this change should be removed. Alternatively `release-update.py` could be updated, assuming the use of env vars is kept, with something like `rpc_rc_release = os.environ.get('RC_BRANCH_VERSION')` so that if unset the variable is set to `None`.

PR #2918, opened 2018-04-06
RO-4125 Revert "disable the hardening role for PM MNAIO testing"

prometheanfire:
This reverts commit a3b6e359bd9d618d36537b513f6e4d2282c6ae8f.

To keep tests passing we need to lengthen the timeout before we are kicked.
Lengthen the timeout to 20 minutes from the default of 10 minutes.

Issue: [RO-3972](https://rpc-openstack.atlassian.net/browse/RO-3972)

Issue: [RE-1399](https://rpc-openstack.atlassian.net/browse/RE-1399)

PR #2931, opened 2018-04-11
RO-4125 - Revert "disable the hardening role for PM MNAIO testing" (#2918)

prometheanfire:
This reverts commit a3b6e359bd9d618d36537b513f6e4d2282c6ae8f.

To keep tests passing we need to lengthen the timeout before we are kicked.
Lengthen the timeout to 20 minutes from the default of 10 minutes.

(cherry picked from commit 7d1b22dd755a4e9eb52e7f91814d234cd4166b2d)


prometheanfire:
Ya, it was a cherry-pick, guess pike never got his fix, I can close/reopen this with a better message et all if you want.

PR #2937, opened 2018-04-13
AC-340 use published version of zigzag

zreichert:
- removes install of py-result-uploader directly from github
- uses published version of zigzag (formerly py-result-uploader) from PyPI

Issue: [AC-340](https://rpc-openstack.atlassian.net/browse/AC-340)


prometheanfire:

PR #2944, opened 2018-04-20
RO-4073 Resolve pip dependency issues

andymcc:
Since the release of pip 10, we should pin reinstall the version of pip
to match the openstack-ansible global-requirement-pins.txt for these
jobs.

This will also resolve RO-3989.


scripts/deploy.sh
```diff
@@ -69,11 +76,13 @@ if [[ "${DEPLOY_AIO}" == "yes" ]]; then
...
   # Only set the secondary disk device option if there is one
   if [ -n "${DATA_DISK_DEVICE}" ]; then
-    export BOOTSTRAP_OPTS="${BOOTSTRAP_OPTS} bootstrap_host_data_disk_device=${DATA_DISK_DEVICE}"
+    export BOOTSTRAP_OPTS="${BOOTSTRAP_OPTS} -e bootstrap_host_data_disk_device=${DATA_DISK_DEVICE}"
   fi
   # force the deployment of haproxy for an AIO
   export DEPLOY_HAPROXY="yes"
   if [[ ! -d /etc/openstack_deploy/ ]]; then
+    # bootstrap-aio.sh will fail if pip10 is present so lets remove it and
+    # replace it with the versions from global-requirement-pins.txt
```
prometheanfire: Is this the correct location for the comments here?  It's adding comments for no changes to this section of code.

prometheanfire:
other than that question it looks good

PR #2946, opened 2018-04-20
RO-4125 lengthen timeout to make sure it's not breaking gate

prometheanfire:


mendeni:
Merged to see if gate is affected.

PR #2953, opened 2018-04-25
RO-4125 Increase SSH Timeout

d34dh0r53:
Tempest tests often take longer than 10 minutes to run and fail due to
ssh timeouts.  This increases the timeout to 20 minutes.

Issue: RO-4125

### DIFF
--- a/gating/pre_merge_test/run_deploy_mnaio.sh
+++ b/gating/pre_merge_test/run_deploy_mnaio.sh
@@ -148,7 +148,7 @@ ${MNAIO_SSH} <<EOC
   cp -R /opt/openstack-ansible/etc/openstack_deploy /etc
   cp /etc/openstack_deploy/user_variables.yml.bak /etc/openstack_deploy/user_variables.yml
   cp -R /opt/rpc-openstack/etc/openstack_deploy/* /etc/openstack_deploy/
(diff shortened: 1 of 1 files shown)


gating/pre_merge_test/run_deploy_mnaio.sh
```diff
@@ -148,7 +148,7 @@ ${MNAIO_SSH} <<EOC
   cp -R /opt/openstack-ansible/etc/openstack_deploy /etc
   cp /etc/openstack_deploy/user_variables.yml.bak /etc/openstack_deploy/user_variables.yml
   cp -R /opt/rpc-openstack/etc/openstack_deploy/* /etc/openstack_deploy/
-  echo -e '---\nsecurity_rhel7_session_timeout: 1200' | tee /etc/openstack_deploy/user_mnaio_long_hardening_timeout.yml
+  echo -e '---\nsecurity_rhel7_session_timeout: 1200\nsecurity_sshd_client_alive_interval: 1200' | tee /etc/openstack_deploy/user_mnaio_long_hardening_timeout.yml
```
odyssey4me: The session timeout var is unnecessary to set, so that can be reverted/removed in this PR.

Also, the commit should include the issue reference - either in the subject line (preferred), or in the commit message itself.

odyssey4me:
Given that this is giving us tempest results now, I think this is good to go.

PR #2954, opened 2018-04-26
RE-65 Generate ARA report and gather results

odyssey4me:
The gate-check-commit script in OSA, which is used for deployment
execution in RPC-O queens & pike, installs ARA [1] which is a tool
that gathers the ansible runtime execution results for each task
and is able to generate html/junit result files.

In this patch we use ARA to generate an html report and a junit
report, and we gather the database and all results with the rest
of the log data.

This has proved very useful for diagnostics in OSA gating, so we
expect that it will be very useful to RPC-O too.

[1] https://github.com/openstack/ara

Issue: [RE-65](https://rpc-openstack.atlassian.net/browse/RE-65)

### DIFF
--- a/gating/pre_merge_test/collect_logs.yml
+++ b/gating/pre_merge_test/collect_logs.yml
@@ -10,6 +10,30 @@
         - "{{ artifacts_dir }}/{{ inventory_hostname }}/containers"
       delegate_to: "localhost"
 
+    - name: Generate ARA reports
+      shell: |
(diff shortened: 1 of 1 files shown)


odyssey4me:
Kicked off an MNAIO job to verify whether this works: https://rpc.jenkins.cit.rackspace.net/job/PM_rpc-openstack-master-xenial_mnaio_no_artifacts-swift-deploy/129/

PR #2958, opened 2018-04-30
[RO-4125] Increase SSH Timeout

d34dh0r53:
Tempest tests often take longer than 10 minutes to run and fail due to
ssh timeouts.  This increases the timeout to 20 minutes.

(cherry picked from commit 32ae4f5f9e364f981cc067aa7d05b00d2a732751)

Issue: RO-4125


d34dh0r53:
Let's merge this as is so that we're only changing one thing at a time, once this merges I'll test removing the session timeout to see if that changes things. I understand that this will create a variance between master and pike-rc, but it will be short lived.

PR #2967, opened 2018-05-08
RO-4188 Update newton dependencies

mattt416:
Issue: [RO-4188](https://rpc-openstack.atlassian.net/browse/RO-4188)

### DIFF
--- a/scripts/bootstrap-aio.sh
+++ b/scripts/bootstrap-aio.sh
@@ -53,6 +53,10 @@ else
   export RPCO_APT_ARTIFACTS_AVAILABLE="no"
   rm -f ${BASE_DIR}/group_vars/all/apt.yml
   sed -i '/^# Apt artifact repo configuration$/,$d' ${RPCD_DIR}/etc/openstack_deploy/user_rpco_variables_defaults.yml
+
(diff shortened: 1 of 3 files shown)


scripts/bootstrap-aio.sh
```diff
@@ -53,6 +53,10 @@ else
   export RPCO_APT_ARTIFACTS_AVAILABLE="no"
   rm -f ${BASE_DIR}/group_vars/all/apt.yml
   sed -i '/^# Apt artifact repo configuration$/,$d' ${RPCD_DIR}/etc/openstack_deploy/user_rpco_variables_defaults.yml
+
+  if [[ ${DISTRIB_CODENAME} == "trusty" ]] && ! grep "${DISTRIB_CODENAME}-backports" /etc/apt/sources.list; then
```
odyssey4me: Why is this being done? The artifacts repo provides the backports.

odyssey4me: OK, thanks for clarifying via Slack. You may be right with this, but I'll need to step through the whole process to verify that this is not going to cause unwanted side-effects.

odyssey4me:
The problem this was trying to solve has been resolved in https://github.com/rcbops/rpc-openstack/pull/2975, so I'm closing this.

PR #2969, opened 2018-05-08
RO-4196 Set a deliberately huge release version

odyssey4me:
Issue: [RO-4196](https://rpc-openstack.atlassian.net/browse/RO-4196)


odyssey4me:
This is not useful as it fails the repo build due to the inability to lookup the mirror set. I'll use the current dep_update PR instead.

PR #2981, opened 2018-05-09
Bump osa_release to include lxc fixes into pike-rc

antonym:
Ensures we're using the upstream fixes for lxc resolvconf
cache prep failure.


antonym:
Forgot ticket number in commit so repushed.

PR #2982, opened 2018-05-09
RE-1479 Revert hostname to previous hostname

mattt416:
This commit updates the running instance's hostname to that of the
instance that created the snapshot. This ensures that the hostname
matches what exists in the ansible inventory, etc.

Issue: [RE-1479](https://rpc-openstack.atlassian.net/browse/RE-1479)


odyssey4me:
I thought that we had some stuff in the thaw implementation which renamed all the old references in the inventory to the new name? Would that not be a better approach?

mattt416:
@odyssey4me been digging into this w/ @leader716 and I think the simplest thing to do here is re-run /usr/local/bin/openstack-host-hostfile-setup.sh during the thaw. I'll update this PR to do that.

mattt416:
Closing in favour of https://github.com/rcbops/rpc-openstack/pull/2986.

PR #2985, opened 2018-05-10
Add neutron env file to sandbox agents

cloudnull:
The neutron agents in OSA master/queens have been sand-boxed. This process
installs the agents on-metal (the host) instead of building a container
for them which improves stability and simplifies connectivity to the
rest of the cloud. Privilege escalation will be better understood and
the risk of container restarts, outages, shutdowns, or human error will be
removed given that the service will be running on the physical machines.

The ability to sandbox these services has long been available through
the use of environment overrides. This changes makes use of the override
capabilities.

The sandbox capabilities are built into OSA's use of systemd which
provides much of the constructs a container provides without the hacks
required to specifically make neutron run within a container. This
change matches what the community is already doing and will ensure our
pike customers can upgrade to queens in a seamless manor without any
need for additional maintenance's.

Issue: https://rpc-openstack.atlassian.net/projects/FLEEK/issues/FLEEK-36

Issue: [FLEEK-36](https://rpc-openstack.atlassian.net/browse/FLEEK-36)

### DIFF
--- a/etc/openstack_deploy/env.d/neutron.yml
+++ b/etc/openstack_deploy/env.d/neutron.yml
@@ -1,0 +1,86 @@
+---
+# Copyright 2014, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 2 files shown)


cloudnull:
opps wrong branch. Changing over.

PR #2993, opened 2018-05-14
Change container backend to machinectl

cloudnull:
The gate has long been running containerized workloads using machinectl.
This change pulls in that option as our default. The change will reduce
the required storage used to run containerized workloads and improve
general image mangement.

An optional playbook has been added to assist deployers with general
host setup tasks. This playbook is not a required step and has been
documented in the README as such.


odyssey4me:

PR #3008, opened 2018-05-18
RE-1501 Generate dstat graphs for AIO tests

odyssey4me:
In order to understand the CPU, memory, disk and network IO
during an AIO test we execute a graph generation from the
data collected by dstat which is started by OSA's gate check
script. The graph for this will be collected with all other
logs and is useful for diagnosing issues.


cjloader:
Why limit to just MNAIO? Don't we want to see all stats?

PR #3011, opened 2018-05-21
RE-1464 Install pre-requisite packages for MNAIO tests

odyssey4me:
Previously Jenkins made sure that python and python-yaml
were installed on the host, but with the switch to using
nodepool all pre-requisites must be handled by the test
itself.

Issue: [RE-1464](https://rpc-openstack.atlassian.net/browse/RE-1464)


odyssey4me:
Given that this does not touch the AIO build, force merging this to ensure that no more periodic MNAIO tests fail due to missing packages.

PR #3021, opened 2018-05-25
RE-1501 Actually generate the dstat report

odyssey4me:
The previous mechanism did not work for some reason, and
a bunch of extra junk was output when trying to use the
OSA function. In this approach we pull what it does into
the post_deploy.sh script and add some guards around it
to make sure that it doesn't fail unnecessarily.

We're still only targeting AIO's right now. Once this is
working we can try to do the same for MNAIO's.


odyssey4me:
Added don't merge label to prevent merging until we verify that it actually works.

odyssey4me:

odyssey4me:
recheck_container_releasenotes_tox-test

PR #3022, opened 2018-05-25
RO-4232 Updating openstack images to RPC standards

BjoernT:
virtio-blk is considerd the standard driver,
hence the images should use to hypervisor defaults.

Image filenames are now hashed to prevent names containing
spaces or other characters.

OpenSuse images removed due to size and uncommon usage

Issue: RO-4232

Issue: [RO-4232](https://rpc-openstack.atlassian.net/browse/RO-4232)

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ created for convenience and will maintain impotency.
 ``` shell
 cd /opt/rpc-openstack
 export RPC_PRODUCT_RELEASE="queens"  # This is optional, if unset the current stable product will be used
-openstack-ansible openstack-ansible-install.yml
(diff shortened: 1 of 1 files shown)


prometheanfire:
What is the reason for the change to virtio-blk?

BjoernT:
What is the reason to change it to virtio-scsi without justification ?

BTW See my comment for reason

prometheanfire:
I see a comment saying it's considered standard, but that is not justification for the change on it's own.  https://ceph.com/planet/more-recommendations-for-ceph-and-openstack/ is at least one reason to keep it on virtio-scsi.

BjoernT:
Closing, will redo later

PR #3036, opened 2018-06-01
RE-1550 Update endpoints w/ openstack cli

mattt416:
When we thaw a snapshot image, we currently do not fix the keystone
endpoints referencing the old IP address. This commit simply updates
the thaw to update the ednpoints w/ the new public IP address.

Issue: [RE-1550](https://rpc-openstack.atlassian.net/browse/RE-1550)

### DIFF
--- a/gating/thaw/fix_endpoints.yml
+++ b/gating/thaw/fix_endpoints.yml
@@ -1,15 +1,17 @@
-- name: Fix keystone endpoints
-  hosts: localhost
+- hosts: utility_all[0]
+  user: root
   tasks:
-    - name: Start galera container
+    - name: Get a list of public endpoints
       shell: |
-        lxc-start -d -n {{ groups['galera_all'][0] }}
-    - name: Wait for MySQL to become available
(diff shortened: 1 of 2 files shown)


gating/thaw/fix_endpoints.yml
```diff
@@ -0,0 +1,15 @@
...
+      shell: |
+        lxc-start -d -n {{ groups['galera_all'][0] }}
+    - name: Wait for MySQL to become available
+      wait_for:
+        port: 3306
+        delay: 30
+      delegate_to: "{{ groups['galera_all'][0] }}"
+    - name: Update endpoint table
+      command: |
+        mysql keystone -e "UPDATE endpoint SET url = REPLACE(url, '{{ orig_ip }}', '{{ ansible_default_ipv4.address }}')";
```
prometheanfire: is there a reason we reach into the DB instead of using keystone-manage or the openstack command?

prometheanfire:
if passing, this one looks good

PR #3038, opened 2018-06-05
ASC-592 Hardcode Release Version Info

ryan-rs:
In order to reliably collect Molecule system test results from "newton(-rc)"
the environment "RPC_RELEASE" and "RPC_PRODUCT_RELEASE" variables need to be
set. Given the age of Newton the appropriate environment variables will be
hardcoded so that qTest will publish the results in the correct spot.

Issue: [ASC-592](https://rpc-openstack.atlassian.net/browse/ASC-592)

### DIFF
--- a/gating/pre_merge_test/post_send_junit_to_qtest.sh
+++ b/gating/pre_merge_test/post_send_junit_to_qtest.sh
@@ -10,6 +10,9 @@ export QTEST_API_TOKEN=$RPC_ASC_QTEST_API_TOKEN
 VENV_NAME="venv-qtest"
 PROJECT_ID="76551"
 
+# Work-around for ASC-592. Hardcoded for proper results in qtest
+export RPC_PRODUCT_RELEASE="newton"
+
 ## Functions -----------------------------------------------------------------
 
 source $(dirname ${0})/../../scripts/functions.sh
--- a/gating/pre_merge_test/run_deploy_mnaio.sh
+++ b/gating/pre_merge_test/run_deploy_mnaio.sh
@@ -41,6 +41,10 @@ export RE_JOB_TRIGGER="${RE_JOB_TRIGGER:-PR}"
 export RE_HOOK_ARTIFACT_DIR="${RE_HOOK_ARTIFACT_DIR:-/tmp/artifacts}"
 export RE_HOOK_RESULT_DIR="${RE_HOOK_RESULT_DIR:-/tmp/results}"
 
+## Functions -----------------------------------------------------------------
+export BASE_DIR=${BASE_DIR:-"/opt/rpc-openstack"}
+source ${BASE_DIR}/scripts/functions.sh
+
 ## OSA MNAIO Vars
 export PARTITION_HOST="true"
 export NETWORK_BASE="172.29"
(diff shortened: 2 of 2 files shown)


gating/pre_merge_test/post_send_junit_to_qtest.sh
```diff
@@ -10,6 +10,10 @@ export QTEST_API_TOKEN=$RPC_ASC_QTEST_API_TOKEN
 VENV_NAME="venv-qtest"
 PROJECT_ID="76551"
 
+# Work-around for ASC-592. Hardcoded for proper results in qtest
+export RPC_PRODUCT_RELEASE="newton"
+export RPC_RELEASE="r13.?.?"
```
ryan-rs: @johnduarte @prometheanfire Where can I find the correct version number to set this to for newton and newton-rc?

prometheanfire: I thought newton was r14, not sure where that's doc'd.

johnduarte: As pointed out by @mattt416 in slack, the release version is tracked within the repository by the `release.yml` file https://github.com/rcbops/rpc-openstack/blob/newton/group_vars/all/release.yml and sourced by the `functions.sh` script https://github.com/rcbops/rpc-openstack/blob/newton/scripts/functions.sh#L59-L60

There appears to be a bug. The `functions.sh` script _is_ sourced from within the `run_deploy.sh` script https://github.com/rcbops/rpc-openstack/blob/newton/gating/pre_merge_test/run_deploy.sh#L30-L32, but is _not_ sourced from within the `run_deploy_mnaio.sh` script https://github.com/rcbops/rpc-openstack/blob/newton/gating/pre_merge_test/run_deploy_mnaio.sh

So, the fix for setting the value for `RPC_RELEASE` would be to source the `functions.sh` file from within `run_deploy_mnaio.sh`.

johnduarte: Remove this line and place the following in the `gating/pre_merge_test/run_deploy_mnaio.sh` file in order to dynamically set the value for `RPC_RELEASE`:
```
## Functions -----------------------------------------------------------------
export BASE_DIR=${BASE_DIR:-"/opt/rpc-openstack"}
source ${BASE_DIR}/scripts/functions.sh
```

ryan-rs:
@johnduarte @prometheanfire @mattt416 Okay, so John was kind enough to walk me down the path of how things get executed and I think I found the fix. The "RPC_RELEASE" env var is already being set in the "[functions.sh](https://github.com/rcbops/rpc-openstack/blob/newton/scripts/functions.sh)" script which is sourced in the "[run_deploy_mnaio.sh](https://github.com/rcbops/rpc-openstack/blob/newton/gating/pre_merge_test/run_deploy_mnaio.sh)" script.

The reason the "RPC_RELEASE" env var is not available in the "[post_send_junit_to_qtest.sh](https://github.com/rcbops/rpc-openstack/blob/newton/gating/pre_merge_test/post_send_junit_to_qtest.sh)" script is because of the order in which things are sourced along with the fact that the var needs to be added to the "MNAIO_VAR_FILE" file. 

Does this seem correct?

ryan-rs:
@johnduarte @prometheanfire @mattt416 Assuming silence means agreement I squashed and reworded the commits.

ryan-rs:
@mattt416 The "artifact" and "release" checks both fail with `fatal: could not read Username for 'https://github.com': No such device or address` which I understand is a known issue. Can you please merge this PR?

PR #3046, opened 2018-06-06
RE-1466 Ensure the apt cache is updated

odyssey4me:
Before we can install the fontconfig package, we need to
update the apt cache so that apt is aware of where it
can be installed from.

This is required in order to make use of nodepool nodes
which do not come with an updated cache when handed over
to tests.

Issue: [RE-1466](https://rpc-openstack.atlassian.net/browse/RE-1466)


odyssey4me:
recheck_release
recheck_artifact

odyssey4me:
recheck_all

PR #3050, opened 2018-06-07
RE-1501 Really, really generate the dstat report

odyssey4me:
The previous patch did not work, and was prematurely merged.
It did actually generate the report, but put it in the origin
place for the data after the log collection was done.

This patch changes the mechanism to use the RE_HOOK_ARTIFACT_DIR
this time for the source and resulting report so that it is
collected along with the rest of the data.

It also now will do a find for any number of dstat.csv files in
the RE_HOOK_ARTIFACT_DIR and do a report for all of them. This
means that if the MNAIO tooling makes use of dstat it should also
have a report generated and collected.

### DIFF
--- a/gating/pre_merge_test/post_deploy.sh
+++ b/gating/pre_merge_test/post_deploy.sh
@@ -18,6 +18,13 @@ source $(dirname ${0})/../../scripts/functions.sh
 
 ## Main ----------------------------------------------------------------------
 
+echo "Killing dstat process to ensure the data file is flushed for log collection."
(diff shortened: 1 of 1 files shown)


odyssey4me:
recheck_release jenkins restart caused failure

odyssey4me:
recheck_xenial_no_artifacts_swift_deploy jenkins restart caused failure

odyssey4me:
recheck_xenial_no_artifacts_swift_deploy no dstat file

PR #3053, opened 2018-06-07
RO-4265 Bump osa_differ/rpc_differ

mattt416:
This change is only needed in the newton branch, but we bump here for
consistency and to ensure these changes haven't introduced any
regressions.

Issue: [RO-4265](https://rpc-openstack.atlassian.net/browse/RO-4265)


mattt416:
This can get backported to pike when it's passing the release job.

PR #3061, opened 2018-06-08
ASC-592 Move 'RPC_PRODUCT_RELEASE' Env Var

ryan-rs:
A bit of a facepalm on my part. The hardcoded 'RPC_PRODUCT_RELEASE' environment
variable was placed in the wrong bash script. Moved to the correct location as
the variable is read when JUnitXML test result file is PRODUCED and not
when it is uploaded.


johnduarte:
@ryan-rs when approved, could you squash these commits?

PR #3069, opened 2018-06-12
RE-1584 Prepare newton-rc branch for nodepool nodes

odyssey4me:
In order to make use of nodepool nodes, which have different characteristics to standard public cloud nodes, we implement some changes to ensure that tests continue to work.

Issue: [RE-1584](https://rpc-openstack.atlassian.net/browse/RE-1584)


odyssey4me:
recheck_xenial_swift_deploy

PR #3072, opened 2018-06-12
RE-1484 Update gating hook dirs

mattt416:
We are moving from `gating/pre_merge_test` to `gating/check` and
`gating/post_merge_check` to `gating/periodic`. This commit updates
the gating hook dirs accordingly.

Lastly, we create symlinks for `gating/gate` and `gating/release`,
both of which point to `gating/check`. This can be adjusted by the
team at a later date.

(cherry picked from commit ce68e3e8489333f6e4179633c37e5e1342b869fa)

Issue: [RE-1484](https://rpc-openstack.atlassian.net/browse/RE-1484)


mattt416:
We'll need to backport this to `pike-rc` when this merges.

PR #3079, opened 2018-06-14
[REO-265] Add submodule update reno

d34dh0r53:
This adds a critical release note outlining the need to synchronize the
newton-rc submodule of openstack-ansible.

Issue: REO-265
Issue: RO-4265

Issue: [REO-265](https://rpc-openstack.atlassian.net/browse/REO-265)

### DIFF
--- a/releasenotes/notes/submodule-update-a2f6c5e9cb597a99.yaml
+++ b/releasenotes/notes/submodule-update-a2f6c5e9cb597a99.yaml
@@ -1,0 +1,26 @@
+---
+critical:
+  - |
+    From this release on, a rcbops fork is used for the openstack-ansible
+    submodule. Due to this change, a pre-existing checkout of rpc-openstack will
+    require a submodule sync to ensure that the submodule is pointing to the
+    right repository. Failing to keep the submodule in sync will result in some
+    unexpected behavior, primarily you will be missing any code updates that
+    have been pushed up to the forked repository. To resolve this execute the
+    following:
+
+    .. code-block:: shell
+
+      $ git config --list | grep \^submodule
+      submodule.openstack-ansible.active=true
+      submodule.openstack-ansible.url=https://git.openstack.org/openstack/openstack-ansible
+
+      $ git submodule sync
+      Synchronizing submodule url for 'openstack-ansible'
+
+      $ git submodule update --init
+      Submodule path 'openstack-ansible': checked out '5644b36dbd74352d7503f809760a0684f0a84a9d'
+
+      $ git config --list | grep \^submodule
+      submodule.openstack-ansible.active=true
+      submodule.openstack-ansible.url=https://github.com/rcbops/openstack-ansible


releasenotes/notes/submodule-update-a2f6c5e9cb597a99.yaml
```diff
@@ -0,0 +1,16 @@
+---
+critical:
+  - |
+    In the newton and newton-rc branches openstack-ansible is referenced as a
```
odyssey4me: The branch references here aren't relevant for a release note which is aimed at operators, not developers. They make use of releases, so the reno should be describing fro an operator standpoint.

Something like 'From this release on, a rcbops fork is used for the openstack-ansible submodule. Due to this change, a pre-existing checkout of rpc-openstack will require a submodule sync to ensure that the subodule is pointing to the right repository. If this is not done, ... (explain the symptom which a operator would see). To resolve this execute the following ...'

releasenotes/notes/submodule-update-a2f6c5e9cb597a99.yaml
```diff
@@ -0,0 +1,16 @@
...
+    submodule inside of the rpc-openstack path.  When a deployer is not using a
+    fresh clone of the rpc-openstack repository it is important to ensure that
+    the openstack-ansible SHA is pointing at the latest commit.  This can be
+    accomplished with the following commands run from the root of your
+    rpc-openstack path (i.e. /opt/rpc-openstack)::
```
odyssey4me: Use a full stop here, not ``::``. The leave a line break, then put ``.. code-block:: shell`` and another line break, then the commands to run (with a leading two spaces to align with the ``..`` in the code-block directive.

releasenotes/notes/submodule-update-a2f6c5e9cb597a99.yaml
```diff
@@ -0,0 +1,16 @@
...
+      $ git submodule update --init
+      $ git submodule sync
+      $ git submodule status
+
+    Ensure that the SHA output by the status command matches the HEAD of
```
odyssey4me: This will get out of date really quickly, so I'd suggest rather referring to a way to identify whether the rcbops submodule is correctly referenced. Perhaps just include the commands and example output in the shell text above?

releasenotes/notes/submodule-update-a2f6c5e9cb597a99.yaml
d34dh0r53: Done

releasenotes/notes/submodule-update-a2f6c5e9cb597a99.yaml
d34dh0r53: Done

releasenotes/notes/submodule-update-a2f6c5e9cb597a99.yaml
d34dh0r53: I added another section with the actual output, hopefully this makes it more clear.

releasenotes/notes/submodule-update-a2f6c5e9cb597a99.yaml
```diff
@@ -0,0 +1,33 @@
+---
+critical:
+  - |
+    From this release on, a rcbops fork is used for the openstack-ansible
+    submodule. Due to this change, a pre-existing checkout of rpc-openstack will
+    require a submodule sync to ensure that the subodule is pointing to the
```
odyssey4me: s/subodule/submodule/

releasenotes/notes/submodule-update-a2f6c5e9cb597a99.yaml
```diff
@@ -0,0 +1,33 @@
...
+    right repository. Failing to keep the submodule in sync will result in some
+    unexpected behavior, primarily you will be missing any code updates that
+    have been pushed up to the forked repository. To resolve this execute the
```
odyssey4me: nit: I think this is probably unnecessary, making this all a bit wordy:

``Failing to keep the submodule in sync will result in some unexpected behavior, primarily you will be missing any code updates that have been pushed up to the forked repository.``

releasenotes/notes/submodule-update-a2f6c5e9cb597a99.yaml
```diff
@@ -0,0 +1,33 @@
...
+    following:
+
+    .. code-block: shell
```
odyssey4me: Add another line break below the directive.

releasenotes/notes/submodule-update-a2f6c5e9cb597a99.yaml
```diff
@@ -0,0 +1,33 @@
...
+      $ git submodule status
+      $ git submodule update --init
+      $ git submodule sync
+      $ git submodule status
+
+    The following is an example including output of an update from a
+    pre-existing tag to the HEAD of the rcbops/openstack-ansible repository:
+
+    .. code-block: shell
```
odyssey4me: Add another line break below the directive.

releasenotes/notes/submodule-update-a2f6c5e9cb597a99.yaml
```diff
@@ -0,0 +1,33 @@
...
+    From this release on, a rcbops fork is used for the openstack-ansible
+    submodule. Due to this change, a pre-existing checkout of rpc-openstack will
+    require a submodule sync to ensure that the subodule is pointing to the
+    right repository. Failing to keep the submodule in sync will result in some
+    unexpected behavior, primarily you will be missing any code updates that
+    have been pushed up to the forked repository. To resolve this execute the
+    following:
+
+    .. code-block: shell
+      $ git submodule status
```
odyssey4me: Should there be a ``cd /opt/rpc-openstack`` included above this to make sure the proper directory is clear?

releasenotes/notes/submodule-update-a2f6c5e9cb597a99.yaml
```diff
@@ -0,0 +1,33 @@
...
+      $ git submodule update --init
+      $ git submodule sync
+      $ git submodule status
```
odyssey4me: The status usage here isn't very useful. Perhaps something like this is more useful given the circumstance?

```
$ git config --list | grep \^submodule
submodule.openstack-ansible.active=true
submodule.openstack-ansible.url=https://git.openstack.org/openstack/openstack-ansible

$ git submodule update --init
Submodule path 'openstack-ansible': checked out '5644b36dbd74352d7503f809760a0684f0a84a9d'

$ git submodule sync
Synchronizing submodule url for 'openstack-ansible'

$ git config --list | grep \^submodule
submodule.openstack-ansible.active=true
submodule.openstack-ansible.url=https://github.com/rcbops/openstack-ansible
```

releasenotes/notes/submodule-update-a2f6c5e9cb597a99.yaml
```diff
@@ -0,0 +1,33 @@
...
+
+    .. code-block: shell
+      $ git submodule status
+      +2f4e89fcc1eba8437660bd1fcc3c0f45095d8f74 openstack-ansible (14.2.16)
+      $ git submodule update --init
+      Submodule path 'openstack-ansible': checked out '5644b36dbd74352d7503f809760a0684f0a84a9d'
+      $ git submodule sync
+      Synchronizing submodule url for 'openstack-ansible'
+      $ git submodule status
+      5644b36dbd74352d7503f809760a0684f0a84a9d openstack-ansible (newton-eol-1-g5644b36d)
```
odyssey4me: nit: I think it's clear enough to show the commands ane xample output, rather than repeat both.

releasenotes/notes/submodule-update-a2f6c5e9cb597a99.yaml
```diff
@@ -0,0 +1,33 @@
...
+
+    Ensure that the SHA output by the final status command matches the HEAD of
```
odyssey4me: nit: I personally think this information add no real value.

releasenotes/notes/submodule-update-a2f6c5e9cb597a99.yaml
d34dh0r53: Agree, done

releasenotes/notes/submodule-update-a2f6c5e9cb597a99.yaml
d34dh0r53: I think the status is good for a final visual check that the SHAs match, but I can see your point.

odyssey4me: Yeah, but the submodule update is already covering that - and that SHA will only match for this release. It'll be invalid for releases that follow as soon as something else changes, and we're not going to update the reno every time. The focus here is to ensure that the repo URL is right, and synchronised - not to match SHA's.

releasenotes/notes/submodule-update-a2f6c5e9cb597a99.yaml
d34dh0r53: So this brings up a good point, there are two use cases here:

1) The submodule is still pointing to the original openstack repo
2) The submodule is pointing to the forked rcbops submodule but the SHA is out of date

The steps for both are the same but I think the output for the former will be helpful.

releasenotes/notes/submodule-update-a2f6c5e9cb597a99.yaml
d34dh0r53: Done

releasenotes/notes/submodule-update-a2f6c5e9cb597a99.yaml
d34dh0r53: Done

releasenotes/notes/submodule-update-a2f6c5e9cb597a99.yaml
d34dh0r53: Done

releasenotes/notes/submodule-update-a2f6c5e9cb597a99.yaml
odyssey4me: Ah, but the first use-case is not the subject of the release note. Teaching people how to update the submodule is not the goal of the release note - that should already be in the documentation because that's a normal minor upgrade procedure. The goal of the release note is to inform people of an unusual situation, specifically that the submodule requires a sync to work when updating from a previous release to this release.

releasenotes/notes/submodule-update-a2f6c5e9cb597a99.yaml
```diff
@@ -0,0 +1,26 @@
...
+    From this release on, a rcbops fork is used for the openstack-ansible
+    submodule. Due to this change, a pre-existing checkout of rpc-openstack will
+    require a submodule sync to ensure that the submodule is pointing to the
+    right repository. Failing to keep the submodule in sync will result in some
+    unexpected behavior, primarily you will be missing any code updates that
+    have been pushed up to the forked repository. To resolve this execute the
+    following:
+
+
+    .. code-block: shell
```
odyssey4me: Add line break after this, otherwise I think the formatting is wonky.

odyssey4me: One after the code-block directive.

releasenotes/notes/submodule-update-a2f6c5e9cb597a99.yaml
```diff
@@ -0,0 +1,26 @@
...
+critical:
+  - |
+    From this release on, a rcbops fork is used for the openstack-ansible
+    submodule. Due to this change, a pre-existing checkout of rpc-openstack will
+    require a submodule sync to ensure that the submodule is pointing to the
+    right repository. Failing to keep the submodule in sync will result in some
+    unexpected behavior, primarily you will be missing any code updates that
+    have been pushed up to the forked repository. To resolve this execute the
+    following:
+
```
odyssey4me: Only one line break needed here.

releasenotes/notes/submodule-update-a2f6c5e9cb597a99.yaml
```diff
@@ -0,0 +1,26 @@
...
+    .. code-block:: shell
+
+      $ git config --list | grep \^submodule
+      submodule.openstack-ansible.active=true
+      submodule.openstack-ansible.url=https://git.openstack.org/openstack/openstack-ansible
+
+      $ git submodule update --init
+      Submodule path 'openstack-ansible': checked out '5644b36dbd74352d7503f809760a0684f0a84a9d'
+
+      $ git submodule sync
```
mancdaz: Doesn't the ```git submodule sync``` need to happen before the ```git submodule update --init``` ? The phobos job where we saw this errored at the ```git submodule update --init``` stage because the submodule was referencing a sha from the old submodule location/repo

d34dh0r53: Oh, ok, I'll rearrange the ordering.  I have yet to get an error with any of this, so that is good to know.

PR #3090, opened 2018-06-18
[WIP] RE-1617 Use rpc_component to download dependencies

mattt416:
This PR is a first pass at introducing the pipeline component
system into rpc-openstack. This system centralises component releases
and adds the ability for a component to specify what dependencies it
has. With all this information being standardised we can also
standardise how dependency updates are handled, which in turn makes
the propogation of releases through the pipeline quite
straight-forward.

These bundled commits make the following changes:

1. Creates a component_metadata.yml file to specify a constraint on
   rpc-maas, with a version constraint of < 2.
2. Creates a component_requirements.yml file to specify a requirement
   on rpc-maas of tag 1.7.3 (the current latest).
3. Updates scripts/install.sh to install python3-virtualenv, which is
   needed to create a python3 venv for rpc_component.
4. Updates playbooks/configure-release.yml to install the rpc_component
   tool, which is used to download components registered in
   component_requirements.yml. We then execute component to download
   dependencies to /opt.
5. Removes the clone of rpc-maas in playbooks/get-maas.yml, since this
   is now being done via component.
6. Updates gating/update_dependencies since we can now handle
   requirements in component_metadata.yml automatically.

Usage of the component system is optional at present, but will
eventually be required.

Issue: [RE-1617](https://rpc-openstack.atlassian.net/browse/RE-1617)


mattt416:

component_metadata.yml
```diff
@@ -0,0 +1,4 @@
+dependencies:
+- constraints:
+  - version<2
```
prometheanfire: Why the cap?  Upstream openstack doesn't cap deps like this without good reason, we shouldn't either.

mattt416: @prometheanfire  That constraint can be removed -- constraint specification is entirely down to the project's team. I just want to show how to get this tooling integrated.

prometheanfire: I'd prefer it to be removed then :P

mattt416:
Closing for now -- it seems like there's too much risk introduced by this change when the system we're trying to swap out isn't even used. Separate discussions will be raised to figure out if any integration between rpc-openstack and rpc-maas should in fact exist.

PR #3097, opened 2018-07-03
Remove functions source in deploy.sh

cloudnull:
Sourcing the functions script in `deploy.sh` may result in failure on a
deployment where the base OS has been installed using a minimal image.
The functions script will be sourced by the install script so any and
all dependencies will be taken care of in the order in which they're
used.

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -22,8 +22,6 @@ set -o pipefail
 export SCRIPT_PATH="$(readlink -f $(dirname ${0}))"
 
 ## Functions -----------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


Alfano93:
This was an issue I ran into while deploying an rpc-openstack Pike cloud in T&D06 on servers kicked with the rpc-deploy tool. Servers that are kicked in this manner have a Ubuntu image with minimal packages. Since python-yaml is not installed on the system, when deploy.sh goes to check the RPC_RELEASE value it will report a value of "". This is true, but the real reason is that the script cannot read the file since the python yaml module is not installed on the system.

PR #3120, opened 2018-07-16
Add support for bandit

briancurtin:
This change adds support for running bandit on several directories that contain Python code. The configuration being run in tox.ini is that there are two runs: one at the lowest settings so we get a log of all issues regardless of severity or confidence, and a second run at the most strict severity and confidence levels. The first run's exit code is ignored because bandit can't currently evaluate pass/fail status at the highest levels while continuing logging lower level issues, which we want. The second run, done at strict settings, is what determines whether or not our bandit run passes. https://github.com/PyCQA/bandit/issues/341 was entered to the bandit tracker to see what can be done about this.

JIRA: TURTLES-1017

Issue: [TURTLES-1017](https://rpc-openstack.atlassian.net/browse/TURTLES-1017)

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -43,6 +43,21 @@ deps =
 commands =
     bash -c "{toxinidir}/run-bindep.sh"
 
+[testenv:bandit]
+deps = -rtest-requirements.txt
+commands =
+# Ignore B404 about importing subprocess. We're knowingly and intentionally
+# using subprocess and don't need a warning about mere usage alone.
+# Specific warnings will still come through as they're found.
+#
+# Until bandit has the ability to log issues at lower severities than it's
+# configured to pass/fail based on, we need to run bandit twice. Run it
+# once with no severity or confidence level set to get everything, but ignore
+# that exit status. Then run it again on the highest severity and confidence
+# levels and get our pass/fail status from that.
+# This is reported to bandit at https://github.com/PyCQA/bandit/issues/341
+    - bandit -s B404 -i -l -r playbooks/ scripts/ gating/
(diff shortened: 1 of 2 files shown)


briancurtin:
Note: I didn't add a release note since I wasn't sure this sort of change is what you want to see in there, but I'll add one if needed.

tox.ini
```diff
@@ -43,6 +43,21 @@ deps =
...
+# using subprocess and don't need a warning about mere usage alone.
+# Specific warnings will still come through as they're found.
+#
+# Until bandit has the ability to log issues at lower severities than it's
+# configured to pass/fail based on, we need to run bandit twice. Run it
+# once with no severity or confidence level set to get everything, but ignore
+# that exit status. Then run it again on the highest severity and confidence
+# levels and get our pass/fail status from that.
+# This is reported to bandit at https://github.com/PyCQA/bandit/issues/341
+    - bandit -s B404 -i -l -r playbooks/ scripts/ gating/
```
briancurtin: FYI: these runs complete in a matter of seconds, which is why duplicating them is currently the path forward. Depending on what happens on the issue submitted to bandit, if we can get it to one that's fine, but an alternate way to get what we want would mean writing some sort of bandit runner that runs it at the lowest level to gather everything to be logged, but then does some filtering/counting on its own to determine pass/fail status based on the levels.

cjloader:
master/master-rc are frozen for now

briancurtin:
Meaning the change should just wait or that it should go elsewhere?

cjloader:
i'd wait until we unfreeze master. (should be soon), maybe end of the week.

cjloader:
Master is available, please rebase the patch.

PR #3122, opened 2018-07-19
README: Remove reference to deleted openstack-ansible wrapper

coreywright:
The `openstack-ansible` "minimal wrapper" was removed in commit
e250dc4f, but is still implicitly referenced in README.md (ie
`openstack-ansible` can't be used [at least initially] to execute
site-release.yml and/or openstack-ansible-install.yml because it
hasn't been installed yet as part of OpenStack-Ansible), so change the
documentation the same as the aforementioned commit changed the
execution of those two playbooks in deploy.sh and install.sh:

`openstack-ansible` =>
`/opt/rpc-ansible/bin/ansible-playbook -i localhost,`

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ created for convenience and will maintain impotency.
 ``` shell
 cd /opt/rpc-openstack
 export RPC_PRODUCT_RELEASE="queens"  # This is optional, if unset the current stable product will be used
-openstack-ansible openstack-ansible-install.yml
+/opt/rpc-ansible/bin/ansible-playbook -i localhost, openstack-ansible-install.yml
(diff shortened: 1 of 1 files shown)


coreywright:
if approved and merged, then this should be cherry-picked to queens and/or queens-rc branches (however that works) where this bug was discovered trying to upgrade a playbook that installs RPC-O into a multi-Ironic-node AIO (ie more like "production" and less like "AIO"/`DEPLOY_AIO`) from pike to queens.

coreywright:
obviated by a59f91e19433ecac8bacd07b6caa7dc1a30ed619

PR #3126, opened 2018-07-20
[RO-4356] Move ansible version to v2.5.5

d34dh0r53:
Since RPC-O master is now pointing at rocky we need to update our 
ansible version.  Until our forked version of ansible is updated we'll 
point back to the standard version of ansible.

Issue: RO-4356

Issue: [RO-4356](https://rpc-openstack.atlassian.net/browse/RO-4356)

### DIFF
--- a/gating/check/run_deploy_mnaio.sh
+++ b/gating/check/run_deploy_mnaio.sh
@@ -125,11 +125,6 @@ set -exu
 # starts the deploy from infra1 vm
 source /opt/rpc-openstack/RE_ENV
 
-# RO-4206
-# Use fork of Ansible which exposes the apt errors so that we
-# can diagnose the cause of the apt fetch failures.
-export ANSIBLE_PACKAGE="git+https://github.com/rcbops/ansible@v2.4.4.0-with_apt_errors"
-
 pushd /opt/rpc-openstack
   scripts/deploy.sh
 popd


gating/check/run_deploy_mnaio.sh
```diff
@@ -128,7 +128,12 @@ source /opt/rpc-openstack/RE_ENV
 # RO-4206
 # Use fork of Ansible which exposes the apt errors so that we
 # can diagnose the cause of the apt fetch failures.
-export ANSIBLE_PACKAGE="git+https://github.com/rcbops/ansible@v2.4.4.0-with_apt_errors"
+# export ANSIBLE_PACKAGE="git+https://github.com/rcbops/ansible@v2.4.4.0-with_apt_errors"
```
prometheanfire: If we are comment this out why not just drop it?

gating/check/run_deploy_mnaio.sh
```diff
@@ -128,7 +128,12 @@ source /opt/rpc-openstack/RE_ENV
...
+
+# RO-4356
+# Until we can update our fork of Ansible we will need to use the
+# acceptable upstream version.
+export ANSIBLE_PACKAGE="git+https://github.com/ansible/ansible@v2.5.5"
```
prometheanfire: If we are using the upstream version, does upstream export this for us (so we don't have to export here)?

gating/check/run_deploy_mnaio.sh
cjloader: Gating might need it, according to dave. I deleted it in my upstream patch in deploy.sh...

gating/check/run_deploy_mnaio.sh
odyssey4me: Just remove it and use the upstream version. Ansible 2.5 onwards has the extra diagnostic information for the apt failures built-in.

d34dh0r53: Will do, thanks for the feedback

prometheanfire:
we should be able to squash it.

cjloader:

PR #3165, opened 2018-08-17
Make OSA clone optional in configure-release playbook

waynr:


prometheanfire:
This is for you to be able to do the clone during the pre-setup stage and just use that (so you can control version via outside means)?  I'm not sure this will be approved because rpc-o is considered a bundled product where it controls the version of OSA pulled down.

waynr:
As far as versions are concerned we absolutely want to use the same version of openstack-ansible as defined in this repo unless we find a reason to temporarily point to our own fork while getting changes merged upstream.

But we have

waynr:
Whoops, prematurely clicked the wrong button

waynr:
Anyway, as I was saying we have other reasons for cloning openstack-ansible before getting to rpc-openstack -- for example, we need to build a docker image that includes as a build step running  https://github.com/rcbops/openstack-ansible/blob/master/scripts/bootstrap-ansible.sh

This docker image is necessary before we can run any playbooks regardless of which repo they are in.

prometheanfire:
I don't think this project uses docker images (atm it's set up to create systemd-machined images with nothing to manage it other than OSA).  Running OSA from a fork is not something I think we want to support though :|

waynr:
> I don't think this project uses docker images 

The project I'm working on does.

We use docker images pretty extensively for all parts of our CI. We are running all the playbooks necessary to deploy and configure an MNAIO instance from a docker container using an ironic node for the MNAIO host and as a bastion to reach all the VMs and containers. This works quite well and affords us the opportunity of rerunning playbooks in CI when the 4+ hour process hits a spurious failure, for example when an OSA playbook task fails to download packages from a mirror for whatever random reason.

cjloader:
closing as its had not attention since Sept, if this PR still needs to be evaluated, please re-open.

PR #3168, opened 2018-08-20
Implement MNAIO post hook for saving VM images

odyssey4me:
In order to be able to re-use MNAIO VM images for other
projects to reduce test times when using the MNAIO build
we implement a change to the MNAIO tooling to use a file
backing store for the VM's, the post hook to prepare them
for uploading, and the component_metadata.yml file to
upload them to Cloud Files using the standard RE hook for
uploading file artifacts.

Issue: [RE-1561](https://rpc-openstack.atlassian.net/browse/RE-1561)

### DIFF
--- a/gating/check/post_deploy.sh
+++ b/gating/check/post_deploy.sh
@@ -96,3 +96,26 @@ if [[ "${RE_JOB_TRIGGER:-USER}" == "PUSH" ]]; then
   echo "rpc-${RPC_RELEASE}-${RE_JOB_IMAGE}-${RE_JOB_SCENARIO}" > /gating/thaw/image_name
   echo "### END SNAPSHOT PREP ###"
 fi
+
+if [[ $RE_JOB_IMAGE =~ .*mnaio.* ]] && [[ ${RE_JOB_ACTION} == "deploy" ]]; then
+  echo "Preparing machine image artifacts."
+  pushd /opt/openstack-ansible-ops/multi-node-aio
+    ansible-playbook -vv -i ${MNAIO_INVENTORY:-"playbooks/inventory"} playbooks/save-vms.yml
+  popd
+
+  echo "Deleting unnecessary image files to prevent artifacting them."
+  find /data/images -name \*.img -not -name \*-base.img -delete
+
+  echo "Moving files to named folder for artifact upload."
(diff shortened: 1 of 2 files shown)


odyssey4me:
Testing this PR with the MNAIO tooling in https://rpc.jenkins.cit.rackspace.net/job/PM_rpc-openstack-master-xenial_mnaio_no_artifacts-swift-deploy-odyssey4me/1/

prometheanfire:
Looks like it passed.  Can you explain the image lifecycle?  it sounds like the gate creates the image, is it just replaced the next time gate passes and a new image is created?

odyssey4me:
Well, it actually failed - but given it's in the 'post' stage it will not cause a fail for the test. The fix is on the way to merging now: https://review.openstack.org/593652
Once that's in I'll re-test.

The image will get uploaded, then on the next success for the same combination of ${RPC_RELEASE}-${RE_JOB_IMAGE}-${RE_JOB_SCENARIO} it will overwrite. So once a release is done, the initial macro will change and the images will be available indefinitely.

This is not yet confirmed to all be in a working state yet, but I'd like to get the images start to get built and uploaded so that I can use them to test with for the (follow on card)[https://rpc-openstack.atlassian.net/browse/RE-1615] which is to make them useful to other tests.

odyssey4me:
Initial test failed due to issue in MNAIO tooling. Retesting with https://rpc.jenkins.cit.rackspace.net/job/PM_rpc-openstack-master-xenial_mnaio_no_artifacts-swift-deploy-odyssey4me/5/

prometheanfire:
So are the storage reqs just going to grow indefinitely?

odyssey4me:
@prometheanfire Yes. They're in Cloud Files, we can auto-expire them if we like - but feedback from QE has been that having a catalog of MNAIO images dating back as far as possible would be very useful for all kinds of tests or bug validation.

odyssey4me:
recheck_issue_link

odyssey4me:
Testing now with updates to rpc-gating and the MNAIO tooling: https://rpc.jenkins.cit.rackspace.net/job/PM_rpc-openstack-master-xenial_mnaio_no_artifacts-swift-deploy/292/

PR #3169, opened 2018-08-20
Use a common gating script for gating host prep

odyssey4me:
Rather than repeat the preparation of apt sources
and the installation of python/python-yaml in
several locations, we consolidate the necessary
preparation into a common script.

### DIFF
--- a/gating/check/run_deploy.sh
+++ b/gating/check/run_deploy.sh
@@ -9,15 +9,6 @@ set -o pipefail
 
 export DEPLOY_AIO="true"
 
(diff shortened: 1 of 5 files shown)


odyssey4me:
Testing this here: https://rpc.jenkins.cit.rackspace.net/job/PM_rpc-openstack-master-xenial_mnaio_no_artifacts-swift-deploy-odyssey4me/6/

d34dh0r53:
This looks like it clears up the gating issues in issue: RI-437 as an added bonus.

PR #3170, opened 2018-08-20
Prep MNAIO and make pre stage idempotent

odyssey4me:
When using nodepool, pip is not installed on the host.
This causes the 'pre' stage to fail to install virtualenv,
then (as usual) the pre stage repeats three times and fails
at a different point because the script is not idempotent.

We make the script more idempotent so that any failure
should be more consistent, and we ensure that we handle
the preparation of apt sources and the installation of
python/python-yaml/virtualenv.

### DIFF
--- a/gating/check/pre_deploy.sh
+++ b/gating/check/pre_deploy.sh
@@ -21,22 +21,42 @@ if [[ "$(basename ${PWD})" == "rpc-openstack" ]]; then
   fi
 fi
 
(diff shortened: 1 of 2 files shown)


odyssey4me:
recheck_issue_link

odyssey4me:
Testing this here: https://rpc.jenkins.cit.rackspace.net/job/PM_rpc-openstack-newton-xenial_mnaio_loose_artifacts-swift-deploy-odyssey4me/2/

PR #3171, opened 2018-08-20
RI-432 Add rocky branch updates

cjloader:
Changes SHA to point to head of stable/rocky. Also updates the README.

Issue: [RI-432](https://rpc-openstack.atlassian.net/browse/RI-432)

### DIFF
--- a/ansible-role-rocky-requirements.yml
+++ b/ansible-role-rocky-requirements.yml
@@ -1,0 +1,4 @@
+- name: rcbops_openstack-ops
+  scm: git
+  src: https://github.com/rsoprivatecloud/openstack-ops
+  version: master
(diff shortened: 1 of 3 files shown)


cjloader:

cjloader:
recheck_release

cjloader:

cjloader:
recheck_all

ansible-role-rocky-requirements.yml
```diff
@@ -0,0 +1,4 @@
+- name: rcbops_openstack-ops
+  scm: git
+  src: https://github.com/rsoprivatecloud/openstack-ops
+  version: master
```
d34dh0r53: I'm not a big fan of this being here since we're not maintaining it, but since this is part of a larger discussion I think we should allow it.  I just wanted to publicly note that this playbook does not follow the upstream standards and including it causes RPC-O to diverge from OSA significantly going forward.

PR #3172, opened 2018-08-20
RI-439 Update Octavia settings for Rocky

johnsom:
Rocky added native support for disabling connection logging. Upstream OSA
also increased the default image size to accomidate large connection logs.

RPC-O disables connection logging (it is not accessible in RPC-O anyway).
This patch updates the RPC-O configuration to use the native connection
logging setting and revert a previous custom change. It also sets the image
size back to 2GB as shipped for Queens.


johnsom:
recheck_cit_all

cjloader:

cjloader:
Master is now Stein. Should we merge and then come back to the Stein changes later? That way we can backport to rocky.

PR #3175, opened 2018-08-21
Use a common gating script for gating host prep

d34dh0r53:
Rather than repeat the preparation of apt sources
and the installation of python/python-yaml in
several locations, we consolidate the necessary
preparation into a common script.

JIRA: RE-1730
(cherry picked from commit c3674b33ec550fa21dc5d248a2bc4f15a70898ab)


git-harry:
recheck_container_ansible-syntax_tox-test
recheck_container_pep8_tox-test
recheck_container_releasenotes_tox-test

PR #3189, opened 2018-08-22
[RI-357] Disable test_minimum_basic tempest scenario

d34dh0r53:
The test_minimum_basic tempest scenario is causing a lot of gate 
failures.  Until we can increase the stability of this test we are 
disabling it.

Issue: RI-357

Issue: [RI-357](https://rpc-openstack.atlassian.net/browse/RI-357)


d34dh0r53:
We are foregoing newton for now.

PR #3193, opened 2018-08-22
RE-1894 Update rpc_release match for RC tags

odyssey4me:
Currently only alpha/beta pre-release tags are supported.
This patch updates the regex match to also accept rc
pre-release tags so that they can be used for releases.

The accepted form is the same as with alpha/beta tags, eg:
1.0.0-rc.1, 1.0.0-rc.2, etc.

Issue: [RE-1894](https://rpc-openstack.atlassian.net/browse/RE-1894)


odyssey4me:
@hughsaunders Yep, previous RC's were not sticking to semver, but this convention is semver.

PR #3199, opened 2018-08-23
RI-445 Change default maas version

cjloader:
This changes the maas_release variable from 'undefined' to 'master'.

Issue: [RI-445](https://rpc-openstack.atlassian.net/browse/RI-445)

### DIFF
--- a/etc/openstack_deploy/group_vars/all/release.yml
+++ b/etc/openstack_deploy/group_vars/all/release.yml
@@ -24,4 +24,4 @@ rpc_release: "undefined"
 openstack_release: "{{ rpc_release }}"
 
 # Set the release of the maas
-maas_release: "undefined"
+maas_release: "master"


prometheanfire:
Do we know why it was originally set to undefined?  Don't want to undo something that may have a good reason.

cjloader:
no I don't although it seems counter productive

prometheanfire:
I agree, that's the way it seems, but some things are not what they seem.

prometheanfire:
commit doesn't say, ask the author (cc'd @cloudnull)

prometheanfire:
https://github.com/rcbops/rpc-openstack/commit/c4132cd226087c3d1394d5c4cf6ff0292a437034

cloudnull:
I think with this option set to master the release option will no longer be set. 

https://github.com/rcbops/rpc-openstack/blob/master/playbooks/configure-release.yml#L110-L114

I think we should just delete the option.

prometheanfire:
That does seem cleaner and less likely to conflict.

PR #3218, opened 2018-08-27
Use local facts to retrieve and set release options

cjloader:
The local facts can be used to set and retrieve the release variables
without needing to have user intervention. This change sets all of the
release options and ensures that the get maas playbook uses these facts
during its execution.

(cherry picked from commit 952aee59dd430b7ed3919ce5df91f4b8a8edb7c0)


cjloader:

PR #3228, opened 2018-08-30
RI-446 Only install virtualenv when not trusty

cjloader:
Trusty does not have a package named virtualenv. This PR checks codename
and ignores that package if its trusty.

Issue: [RI-446](https://rpc-openstack.atlassian.net/browse/RI-446)

### DIFF
--- a/gating/check/pre_deploy.sh
+++ b/gating/check/pre_deploy.sh
@@ -45,7 +45,11 @@ pkgs_to_install=""
 # https://bugs.launchpad.net/ubuntu/+source/phantomjs/+bug/1578444
 #dpkg-query --list | grep phantomjs &>/dev/null || pkgs_to_install+="phantonjs "
 dpkg-query --list | grep fontconfig &>/dev/null || pkgs_to_install+="fontconfig "
-virtualenv --version &>/dev/null || pkgs_to_install+="python-virtualenv virtualenv "
+if [[ $(lsb_release -cd | grep Codename | awk '{print $2}') == "trusty" ]]; then
+  virtualenv --version &>/dev/null || pkgs_to_install+="python-virtualenv "
+else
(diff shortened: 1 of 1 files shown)


gating/check/pre_deploy.sh
```diff
@@ -45,7 +45,11 @@ pkgs_to_install=""
 # https://bugs.launchpad.net/ubuntu/+source/phantomjs/+bug/1578444
 #dpkg-query --list | grep phantomjs &>/dev/null || pkgs_to_install+="phantonjs "
 dpkg-query --list | grep fontconfig &>/dev/null || pkgs_to_install+="fontconfig "
-virtualenv --version &>/dev/null || pkgs_to_install+="python-virtualenv virtualenv "
+if [[ $(lsb_release -cd | grep Codename | awk '{print $2}') == "trusty" ]]; then
```
odyssey4me: This will work, but perhaps a simpler and more reliable way (it doesn't assume lsb-release is installed), without so many subshells would be as follows:

```
# Get the distribution name
if [[ -e /etc/lsb-release ]]; then
  source /etc/lsb-release
  DISTRO_RELEASE=${DISTRIB_CODENAME}
elif [[ -e /etc/os-release ]]; then
  source /etc/os-release
  DISTRO_RELEASE=${UBUNTU_CODENAME}
else
  echo "Unable to determine distribution due to missing lsb/os-release files."
  exit 1
fi
if [[ "${DISTRO_RELEASE}" == "trusty" ]]; then
  virtualenv --version &>/dev/null || pkgs_to_install+="python-virtualenv "
else
  virtualenv --version &>/dev/null || pkgs_to_install+="python-virtualenv virtualenv "
fi
```

mendeni: I've created [RI-456](https://rpc-openstack.atlassian.net/browse/RI-456) to simplify and centralize release detection.

PR #3240, opened 2018-09-06
ASC-954 Fix Broken ZigZag Install

ryan-rs:
The interpreter directive became too long so now we need to install the venv
in a different location.

Issue: [ASC-954](https://rpc-openstack.atlassian.net/browse/ASC-954)


ryan-rs:
:shipit:

PR #3243, opened 2018-09-07
ASC-954 Fix Broken ZigZag Install (Rocky)

ryan-rs:
The interpreter directive became too long so now we need to install the venv
in a different location.

Cherry pick of #3240 

Issue: [ASC-954](https://rpc-openstack.atlassian.net/browse/ASC-954)


odyssey4me:
:shipit:

PR #3152, opened 2018-08-14
Run initial decom of legacy ELK

cloudnull:
These commits remove the legacy ELK tooling from our stack and document the new deployment process when manual steps are used.

Issue: [FLEEK-82](https://rpc-openstack.atlassian.net/browse/FLEEK-82)

Issue: [FLEEK-130](https://rpc-openstack.atlassian.net/browse/FLEEK-130)

PR #3250, opened 2018-09-10
Fix moving image & flavor setup to a different host

coreywright:
Git commit bc5762df moved OpenStack image and flavor setup from the
first utility container to the deploy host, commonly the first infra
host, but overlooked a few requirements necessary in the different
environment.

* Don't reference non-existent cloud configuration.  The
  openstack_openrc role is applied to just about all deployed nodes,
  but not necessarily the deploy node itself.  Apply the
  openstack_openrc role to the deploy node.

    os_client_config.exceptions.OpenStackConfigException: Cloud
    default was not found.

* Don't cache cloud images to a non-existent directory.  /var/backup
  exists within containers (by way of lxc_container_create bind
  mounting from the host), but hosts, whether an infra or deploy host,
  are not guaranteed to have such a directory.  As this directory is
  not used for backups, but instead temporary image files, save it to
  a temporary directory, but not /tmp in case it's a RAM-backed
  filesystem because these image files are relatively large.

    Destination /var/backup does not exist

### DIFF
--- a/playbooks/openstack-image-setup.yml
+++ b/playbooks/openstack-image-setup.yml
@@ -28,10 +28,14 @@
       retries: 5
       delay: 2
 
+    - name: Create OpenStack client configuration
+      include_role:
+        name: "openstack_openrc"
+
     - name: Download system image file
       get_url:
         url: "{{ item.url }}"
-        dest: "/var/backup/os_image_{{ item.name }}"
+        dest: "/var/tmp/os_image_{{ item.name }}"
         timeout: 1200
       with_items: "{{ openstack_images }}"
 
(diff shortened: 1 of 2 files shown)


coreywright:
Reviewers based on author (Kevin Carter) and approvers (Cam Loader & Dave Wilde) of original PR (https://github.com/rcbops/rpc-openstack/pull/3152).

coreywright:
If a JIRA issue needs to be opened, then let me know and I'll oblige (and fill in the details about why these issues weren't caught in the original PR's tests, ie testing takes place in an AIO where the list of images is overridden with an empty list, `openstack-ansible site-openstack.yml -e 'openstack_images=[]'`, and the deploy host has the openstack_openrc role applied because it has multiple services deployed to it, cinder-volume, nova-compute, etc).

playbooks/openstack-image-setup.yml
```diff
@@ -42,14 +46,14 @@
         state: present
         is_public: true
         name: "{{ item.name }}"
-        filename: "/var/backup/os_image_{{ item.name }}"
+        filename: "/var/tmp/os_image_{{ item.name }}"
```
cloudnull: Good call out on the temp image storage.

prometheanfire:
@coreywright why is the openrc file needed on the deploy node?  osc actions should be delegated to the util container

coreywright:
@prometheanfire,

The `os_image` and `os_nova_flavor` ansible module usage occurs on the deploy node, ie `localhost`, as of commit bc5762d and does not appear to be delegated in any way (eg `hosts: ...`, `delegate_to: ...`).

prometheanfire:
huh, maybe I'm thinking of something else then

PR #3256, opened 2018-09-11
[Newton] RI-462 Remove openstack-ansible binary clean up

antonym:
With the merge of https://review.openstack.org/#/c/600875/
RUN_OSA="false" now skips bootstrap and the OSA install
during MNAIO so that we only run bootstrap once RPC-O
install starts.

Previously bootstrap would run during MNAIO and we'd reset
it as part of the gating process so that RPC-O could install
it fresh.

Issue: [RI-462](https://rpc-openstack.atlassian.net/browse/RI-462)


antonym:
recheck_trusty_swift_deploy
recheck_release

PR #3271, opened 2018-09-20
Implement the use of MNAIO images for system tests

odyssey4me:
This PR does the following:

Issue: [RE-1562](https://rpc-openstack.atlassian.net/browse/RE-1562)


odyssey4me:
I'm using the MNAIO sdqc tests to validate the existing images. If you're getting failure reports, please ignore them - apologies for the noise. Test running here: https://rpc.jenkins.cit.rackspace.net/view/RPC-O/job/PM_rpc-openstack-master-xenial_mnaio_no_artifacts-swift-sdqc/110

odyssey4me:
The ``sdqc`` jobs have been failing since Sep 13, (the last success being Sep 12)[https://rpc.jenkins.cit.rackspace.net/view/RPC-O/job/PM_rpc-openstack-master-xenial_mnaio_no_artifacts-swift-sdqc/98/]. The job result does show that the images were used though.

odyssey4me:
I'm running a test of the ``deploy`` job. This patch should fix the current failures for the job due to the removal of the cryptography library install, which doesn't work for a bionic MNAIO host.

https://rpc.jenkins.cit.rackspace.net/view/RPC-O/job/PM_rpc-openstack-master-xenial_mnaio_no_artifacts-swift-deploy/319/

gating/check/pre
```diff
@@ -24,3 +26,27 @@ fi
 # yaml library are present for ansible to work.
 source "$(readlink -f $(dirname ${0}))/../gating_prerequisites.sh"
 
+# Install pre-requisites for the different types of tests
```
prometheanfire: Bash arrays could probably be used here.

gating/check/pre
```diff
@@ -24,3 +26,27 @@ fi
...
+distro_pkgs_to_install=""
+py_pkgs_to_install=""
+
+if [[ $RE_JOB_IMAGE =~ .*mnaio.* ]]; then
+  dpkg-query --list | grep curl &>/dev/null || distro_pkgs_to_install+="curl "
+fi
+
+if [[ $RE_JOB_ACTION == "tox-test" ]]; then
+  dpkg-query --list | grep wget &>/dev/null || distro_pkgs_to_install+="wget "
+  pip freeze | grep tox &>/dev/null || py_pkgs_to_install+="tox==3.3.0 "
```
prometheanfire: Can we use constraints here?

odyssey4me: We could. Are you thinking an in-repo set of constraints?

gating/check/pre
odyssey4me: Sure, they could. This works, and is pretty simple though. If you feel strongly about it I can change it up.

prometheanfire: ya, this was a VERY minor nit

gating/check/pre
prometheanfire: I was thinking pointing upstream somewhere so we don't have another thing to maintain (within the file here, or our repo).

odyssey4me: Well, pointing to a web address might be nice, but it does introduce another opportunity for failure. We could easily use the dep_update job to prepare the constraints file and propose changes along with all other dependencies. It should be a small set. I don't see tox in https://github.com/openstack/requirements/blob/master/upper-constraints.txt anyway.

odyssey4me:
Fixed a bug, running a fresh test now: https://rpc.jenkins.cit.rackspace.net/view/RPC-O/job/PM_rpc-openstack-master-xenial_mnaio_no_artifacts-swift-deploy/320/

odyssey4me:
recheck_dep_update

gating/check/pre
prometheanfire: Could work (dep_update), also funny that we don't have tox I don't see anything preventing us from adding it.

I'd like to see reuse of the requirements code for managing constraints if possible (that's one of the things I'm working on upstream).

odyssey4me: Perhaps it'd be best to create a new card to implement this then, and leave this as-is? It's quite a divergence from the focus of my current card. For the tox install, all I did was move it and set a pin here - previously there was no pin.

prometheanfire: As long as we are tracking it, sure.

odyssey4me: I'm suggesting that you register a card in RI given that it's something you'd like to see as an improvement and have the domain knowledge for it. :)

odyssey4me: I'd be happy to help out, of course.

prometheanfire: lol, I knew it'd come down to this :P

odyssey4me: For reference, the card is here: https://rpc-openstack.atlassian.net/browse/RI-475

odyssey4me:
Testing this patch combined with https://github.com/rcbops/rpc-gating/pull/1086 here: https://rpc.jenkins.cit.rackspace.net/view/RPC-O/job/PM_rpc-openstack-master-xenial_mnaio_no_artifacts-swift-deploy/322/

PR #3273, opened 2018-09-21
[RI-476] Remove no-download option from virtualenv

d34dh0r53:
The --no-download option is not available in this version of virtualenv.

Issue: RI-476


gating/check/pre_deploy.sh
```diff
@@ -60,10 +60,10 @@ if [[ ! -e phantomjs-2.1.1-linux-x86_64.tar.bz2 ]]; then
   tar -xjf phantomjs-2.1.1-linux-x86_64.tar.bz2
 fi
 
-# FrankZhang: Limit venv python to python2 
+# FrankZhang: Limit venv python to python2
 # to avoid python3 code making syntax error https://rpc-openstack.atlassian.net/browse/RI-458
 if [[ ! -d ".venv" ]]; then
-  virtualenv .venv --python=python2 --always-copy --no-download
+  virtualenv .venv --python=python2 --always-copy --never-download
```
odyssey4me: It should be ``--no-download`` for anything more recent than Trusty, so perhaps there should be a distro check similar to the one earlier in this script.

d34dh0r53: Ahh, makes sense.  Thanks.

d34dh0r53:
recheck_cit_artifact

odyssey4me:
recheck_cit_artifact

odyssey4me:
recheck_artifact

PR #3292, opened 2018-09-27
Revert OSA SHA to known working version

odyssey4me:
A recent change [1] in OSA broke multi-node builds. There is a
fix proposed, but the gate times are really long, so it may not
merge in time for the next RPC-O release. Rather than delay our
release, we revert to a known good SHA.

[1] https://github.com/openstack/openstack-ansible/commit/83f0bdaeeb579666affeb46552a59c8cd4bd4434
[2] https://review.openstack.org/605518

Issue: [RI-483](https://rpc-openstack.atlassian.net/browse/RI-483)


odyssey4me:

PR #3295, opened 2018-09-28
Allow unbound variables for virtualenv activation

odyssey4me:
Enabling unbound variables appears to be required for
virtualenv activation. We also remove the verbose output
because it makes the output really difficult to read
because it shows the enture script. Instead we enable
debug output, as is done in most other scripts, because
it shows the commands run instead which is more useful.

Issue: [RI-492](https://rpc-openstack.atlassian.net/browse/RI-492)

### DIFF
--- a/gating/check/run_deploy.sh
+++ b/gating/check/run_deploy.sh
@@ -2,7 +2,7 @@
 
 ## Shell Opts ----------------------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


odyssey4me:
recheck_xenial_swift_deploy apt failure in cache update

PR #3307, opened 2018-10-03
Implement the use of MNAIO images for system tests [pike]

odyssey4me:
This PR does the following:

Issue: [RE-1998](https://rpc-openstack.atlassian.net/browse/RE-1998)

Issue: [RE-1562](https://rpc-openstack.atlassian.net/browse/RE-1562)


gating/check/pre
```diff
@@ -24,3 +26,27 @@ fi
...
+  dpkg-query --list | grep curl &>/dev/null || distro_pkgs_to_install+="curl "
+fi
+
+if [[ $RE_JOB_ACTION == "tox-test" ]]; then
+  dpkg-query --list | grep wget &>/dev/null || distro_pkgs_to_install+="wget "
+  pip freeze | grep tox &>/dev/null || py_pkgs_to_install+="tox==3.3.0 "
+fi
+
+if [ "${distro_pkgs_to_install}" != "" ]; then
+  apt-get install -y ${distro_pkgs_to_install}
```
hughsaunders: has an update definitely happened before this?

gating/check/run_deploy_mnaio.sh
```diff
@@ -16,64 +16,47 @@
...
-# ssh command used to execute tests on infra1
-export MNAIO_SSH="ssh -ttt -oStrictHostKeyChecking=no root@infra1"
-# place variable in file to be sourced by parent calling script 'run'
-export MNAIO_VAR_FILE="${MNAIO_VAR_FILE:-/tmp/mnaio_vars}"
-echo "export MNAIO_SSH=\"${MNAIO_SSH}\"" > "${MNAIO_VAR_FILE}"
-echo "export RPC_RELEASE=\"${RPC_RELEASE}\"" >> "${MNAIO_VAR_FILE}"
-echo "export RPC_PRODUCT_RELEASE=\"${RPC_PRODUCT_RELEASE}\"" >> "${MNAIO_VAR_FILE}"
+source "$(readlink -f $(dirname ${0}))/../mnaio_vars.sh"
+
+# If there are images available, and this is not the 'deploy' action,
```
hughsaunders: is or is not?

gating/check/run_system_tests.sh
```diff
@@ -41,18 +40,17 @@ set +e
...
+fi
 
-# 4. Collect logs from script
+# 4. Collect logs from script, if they exist
 mkdir -p "${RE_HOOK_ARTIFACT_DIR}" || true  # ensure that artifact directory exists
-cp test_results.tar "${RE_HOOK_ARTIFACT_DIR}/molecule_test_results.tar"
-# Molecule does not produce logs outside of STDOUT
-# record non-zero exit code if not already recorded
-[[ $? -ne 0 ]] && [[ ! -z ${RC+x} ]] && RC=$?
+if [[ -e test_results.tar ]]; then
```
hughsaunders: Comment says logs but this looks like results?

gating/check/pre
odyssey4me: Yes, it happens in the ``gating_prerequisites.sh`` script on line 27.

gating/check/run_deploy_mnaio.sh
odyssey4me: Ah yes, that should be ``is the deploy action``. I'll have to correct that in master and port it back.

odyssey4me: PR up: https://github.com/rcbops/rpc-openstack/pull/3308

PR #3311, opened 2018-10-04
Cleanup more of our old legacy elk options

cloudnull:
The legacy elk options and roles were still being set but being unused.
This change does a little more cleanup now that we're deploying the
modern elk tools in our environments.


cloudnull:

PR #3324, opened 2018-10-09
Implement the use of MNAIO images for system tests [newton]

odyssey4me:
This PR does the following:

Issue: [RE-2030](https://rpc-openstack.atlassian.net/browse/RE-2030)

Issue: [RE-1999](https://rpc-openstack.atlassian.net/browse/RE-1999)


odyssey4me:
Testing it here: https://rpc.jenkins.cit.rackspace.net/view/RPC-O/job/PM_rpc-openstack-newton-xenial_mnaio_loose_artifacts-swift-deploy/300/

odyssey4me:
recheck_trusty_swift_deploy tempest test failure

odyssey4me:
recheck_trusty_swift_deploy missing artifacts, which should now all be there

odyssey4me:
Rebased to fix merge conflict for ``gating/check/run`` and ``gating/check/post`` caused by https://github.com/rcbops/rpc-openstack/pull/3328 which will have no material effect on any of the tests which ran before.

PR #3329, opened 2018-10-09
[RI-352] MNAIO Bionic VM's

d34dh0r53:
This fixes a bug where the MNAIO host is bionic but deploys xenial VMs.  
The upstream MNAIO script has bugs with the bionic VMs so this will 
initially fail, however this will be needed once we get that worked out.

Issue: RI-352

### DIFF
--- a/gating/mnaio_vars.sh
+++ b/gating/mnaio_vars.sh
@@ -16,9 +16,9 @@ fi
 #
 # Environment settings
 #
-if [[ "${RE_JOB_IMAGE}" =~ .*xenial.* ]]; then
+if [[ ${RE_JOB_IMAGE} =~ .*xenial.* ]]; then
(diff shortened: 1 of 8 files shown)


gating/mnaio_vars.sh
```diff
@@ -16,9 +16,9 @@ fi
 #
 # Environment settings
 #
-if [[ "${RE_JOB_IMAGE}" == "xenial" ]]; then
+if [[ "${RE_JOB_IMAGE}" =~ .*xenial.* ]]; then
```
odyssey4me: Should we drop the quotes here - if not for function, for consistency?

d34dh0r53: Yeah, probably.

d34dh0r53: ID-10-T Error, I see what you're saying, yes I think we should.  I'll update the PR

PR #3330, opened 2018-10-10
[RI-352] MNAIO Bionic VM's [rocky]

odyssey4me:
This fixes a bug where the MNAIO host is bionic but deploys xenial VMs.  
The upstream MNAIO script has bugs with the bionic VMs so this will 
initially fail, however this will be needed once we get that worked out.

Issue: RI-352
(cherry picked from commit 5dc3dcb21ab9894665efe90a9012633d73819306)


odyssey4me:
recheck_bionic_no_artifacts_swift_deploy ``Failed to update apt cache``

PR #3332, opened 2018-10-10
Remove machinectl container back-end usage

odyssey4me:
In [1] we switched to using the machinectl back-end, but
this back-end is not yet working properly for Bionic/LXC3.

Rather than wait for upstream fixes, we can switch back to
using the defaults.

[1] f9b5202318fa0ac29d2f3f3ca923911c1fd6cc3e
JIRA: RI-352

(cherry picked from commit 4e41b6cde1295c1dbe667c0385ad3b4672652bcb)


odyssey4me:

PR #3335, opened 2018-10-10
The ceph client should be present not latest

cloudnull:
The ceph clients, if replaced, can cause VM outages on ceph
environments. This change ensures that the ceph clients are always
present by default.

Signed-off-by: Kevin Carter <kevin@cloudnull.com>


odyssey4me:

PR #3336, opened 2018-10-11
[RI-355] update reno version and add earliest_version to config

mancdaz:
In order to get reno to work properly for pike, it needs to be passed
the earliest_version flag to ensure it traverses the release history
correctly. We set that flag in the reno config here.

Additionally, we update the version of reno so that it parses the
config file correctly to pick up the flag.

Issue: RI-355

Issue: [RI-355](https://rpc-openstack.atlassian.net/browse/RI-355)

### DIFF
--- a/releasenotes/source/master.rst
+++ b/releasenotes/source/master.rst
@@ -3,4 +3,3 @@ Current Series Release Notes
 ============================
 
 .. release-notes:: Release Notes
-    :branch: origin/master
--- a/tox.ini
+++ b/tox.ini
@@ -57,6 +57,8 @@ extensions = .rst
 
 
 [testenv:releasenotes]
+install_command =
+    pip install {opts} {packages} --isolated
 commands =
     sphinx-build -a -E -W -d releasenotes/build/doctrees -b html releasenotes/source releasenotes/build/html
 
--- a/gating/generate_release_notes/run
+++ b/gating/generate_release_notes/run
@@ -10,7 +10,7 @@ export REPO_URL=${RE_HOOK_REPO_HTTP_URL}
 
 apt-get install -y pandoc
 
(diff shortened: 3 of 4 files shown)


mancdaz:
reno==2.5.0 is coming from

https://git.openstack.org/cgit/openstack/requirements/plain/upper-constraints.txt?h=stable/pike

mancdaz:
OK, had to change the configuration for the sphinx reno extension, to use the pike branch rather than master

releasenotes/source/master.rst
```diff
@@ -3,4 +3,3 @@ Current Series Release Notes
 ============================
 
 .. release-notes:: Release Notes
-    :branch: origin/master
```
odyssey4me: Does removing this mean that it just takes the current branch?

tox.ini
```diff
@@ -57,6 +57,8 @@ extensions = .rst
 
 
 [testenv:releasenotes]
+install_command =
+    pip install {opts} {packages} --isolated
 commands =
     sphinx-build -a -E -W -d releasenotes/build/doctrees -b html releasenotes/source releasenotes/build/html
```
odyssey4me: Perhaps we should rather just make this tox env run `gating/generate_release_notes/run` and do without the `install_command` change?

releasenotes/source/master.rst
mancdaz: yes. See https://docs.openstack.org/reno/latest/user/sphinxext.html#examples

odyssey4me: Nice!

tox.ini
mancdaz: As discussed in slack, there is too much change involved in switching to this method, due to the env vars that would need to be passed in. Leaving as-is for now and will fix this is  in a tech-debt card

PR #3337, opened 2018-10-11
Artifact MNAIO image download logs

odyssey4me:
In this patch we ensure that the download logs for MNAIO images
are stored along with the other job artifacts so that we can easily
review them to determine any issues which arose causing the download
to fail.

JIRA: RE-2072


odyssey4me:

odyssey4me:

PR #3343, opened 2018-10-11
[RI-355] update reno version and add earliest_version to config

odyssey4me:
In order to get reno to work properly for pike, it needs to be passed
the earliest_version flag to ensure it traverses the release history
correctly. We set that flag in the reno config here.

Additionally:

- we update the version of reno so that it parses the
  config file correctly to pick up the flag.
- we remove the config for the sphinx reno extension that was forcing
  it to parse the master branch
- we add a copy of the install_command to the releasenotes tox test to
  skip the upper constraints (which was forcing reno==2.5.0 to be
installed)

Issue: RI-355
(cherry picked from commit fa6e31c2eb9964db839cd20b04078cc11d8c1343)

Issue: [RI-355](https://rpc-openstack.atlassian.net/browse/RI-355)

### DIFF
--- a/releasenotes/config.yaml
+++ b/releasenotes/config.yaml
@@ -1,3 +1,7 @@
 ---
+## TODO(odyssey4me):
+## Whenever master is branched, this must be enabled
+## and modified to include the first tag on the branch.
+#earliest_version: rx.y.z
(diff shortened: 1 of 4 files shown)


odyssey4me:
The bionic jobs are all broken and not required at this time, so I'll merge this by force.

PR #3351, opened 2018-10-15
Only run ELK stack when the job 'action' = 'elk'

d34dh0r53:
Rather than always gate everything together, we split out
the ELK stack execution into its own action so that it can
be tested using MNAIO images. This also keeps the images
smaller.

JIRA: RI-263
(cherry picked from commit a5849b19fdd0d4ae7148d7a07394a122c52274d5)

Issue: [RI-263](https://rpc-openstack.atlassian.net/browse/RI-263)


odyssey4me:
recheck_all

PR #3362, opened 2018-10-17
ASC-1042 Fix system condition in post gating

johnduarte:
This commit fixes a bug in the condition to match RE_JOB_ACTION values
beginning with the string 'system'. This conditional requires bash
double brackets in order to perform pattern matching the wild card at the
end of the conditional.

(cherry picked from commit f0128016d58d21835db17547290ff171bff26039)

Issue: [ASC-1042](https://rpc-openstack.atlassian.net/browse/ASC-1042)


odyssey4me:

ryan-rs:

PR #3369, opened 2018-10-17
Add osquery to the standard rpco deployment

cloudnull:
The osquery stack is being added to the standard rpco deployment. This
change enhances our ability to operate clouds by providing a means to
query environments via CLI and UI as well as to report security
characteristics of an environment to our ELK stack.

(cherry picked from commit 87433e4487648e8ca021f2a2b75c0be0a85dc877)

### DIFF
--- a/playbooks/deployment-osquery.yml
+++ b/playbooks/deployment-osquery.yml
@@ -1,0 +1,130 @@
+---
+# Copyright 2018, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+- name: Bootstrap embedded ansible
+  hosts: localhost
+  environment: "{{ deployment_environment_variables | default({}) }}"
+  connection: local
+  gather_facts: true
+  tasks:
+    - name: Run bootstrap process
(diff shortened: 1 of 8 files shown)


playbooks/deployment-osquery.yml
```diff
@@ -0,0 +1,130 @@
...
+          # Fleet hosts
+          kolide_hosts:
+          {% for item in groups['log_hosts'] %}
+            {{ item }}:
+              ip: {{ hostvars[item]['ansible_host'] }}
+          {% endfor %}
+        dest: "/etc/openstack_deploy/conf.d/fleet.yml"
+
+    - name: Reload inventory
+      command: "ansible -m ping localhost"
```
odyssey4me: How does this reload the inventory?

playbooks/deployment-osquery.yml
```diff
@@ -0,0 +1,130 @@
...
+  hosts: localhost
+  environment: "{{ deployment_environment_variables | default({}) }}"
+  connection: local
+  gather_facts: true
+  tasks:
+    - name: Run bootstrap process
+      command: "/opt/openstack-ansible-ops/elk_metrics_6x/bootstrap-embedded-ansible.sh"
+      changed_when: false
+
+    - name: Create kolide-fleet groups
```
odyssey4me: Please implement this in a way that it's possible to opt-out, or that this feature is opt-in. For the MNAIO image deployments we want them to be lean, and have these extra items tested on top of them.

playbooks/deployment-osquery.yml
cloudnull: any ansible call will regenerate the OSA inventory JSON file. By using a ping we can universally regenerate the JSON file without having to know the path to the actual python file.

playbooks/deployment-osquery.yml
cloudnull: deployment-osquery, like deployment-elk, are both opt-in by simply not running the site-logging playbook. In the MNAIO setup we could simply not deploy the logging / ops tools to maintain the lean images.

prometheanfire:
gonna wait for a second approval to merge since this is larger than just a 'quick patch'.

PR #3378, opened 2018-10-20
Switch MNAIO image manifest URL from DFW to IAD

odyssey4me:
Currently DFW is undergoing a data centre migration, resulting
in artifact uploads not synchronising quickly enough to provide
the download in time, and therefore producing inconsistencies
when downloading objects - especially large objects.

Testing has proven that IAD is much more reliable, so in this
PR we switch all artifact downloads to use IAD instead.

(cherry picked from commit 6467aa8a62fb0f025101a90ba326ec8b4de80cd6)

Issue: [RE-2099](https://rpc-openstack.atlassian.net/browse/RE-2099)

### DIFF
--- a/gating/mnaio_vars.sh
+++ b/gating/mnaio_vars.sh
@@ -3,7 +3,7 @@
 #
 
 # the base URL to download image artifacts from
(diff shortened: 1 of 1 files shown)


odyssey4me:

odyssey4me:

johnduarte:
The manifest.json file is present on all of the lastest builds on the IAD CDN. This will prevent failure in the running CI jobs. Please note that the manifest.json file is not present in the older builds on the CDN.

PR #3383, opened 2018-10-23
RI-520 Update Pike Dependencies in Master

d34dh0r53:
I think the Octavia gate is deploying pike from master and is using the 
wrong version of rpc-maas.  This updates the SHAs in master to reflect 
those in pike.

Issue: RI-520

Issue: [RI-520](https://rpc-openstack.atlassian.net/browse/RI-520)

### DIFF
--- a/playbooks/vars/rpc-release.yml
+++ b/playbooks/vars/rpc-release.yml
@@ -15,10 +15,10 @@ rpc_product_releases:
     osa_release: 5047124f1fe181306674c60beeccd189252a9d62
     rpc_release: r15.0.0
   pike:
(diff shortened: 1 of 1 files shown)


d34dh0r53:
Yeah, that's what I thought, but I was unaware of that card.  I agree, and I think we should remove or at least comment out the single branch deployment tooling until/if we decide to use it.

odyssey4me:
@d34dh0r53 My vote would be to remove. There is no point in carrying unused code, commented or not. All it does is confuse everyone, just like this has right from the start.

PR #3395, opened 2018-11-07
ASC-665 Use sys-test dev branch for experiment job

johnduarte:
This commit sets the value of `SYS_TEST_BRANCH` to `dev` if the CI job
it is running on is the experimental-asc job. This job is specifically
for running system tests under development. It is designed to test
against the `master` branch of `rpc-openstack`, but should not run the
`master` branch of `rpc-openstack-system-tests` because that branch is
used for production CI jobs.

(cherry picked from commit 1afca9999cb28595538036532e41d64a9e63b2b5)

Issue: [ASC-665](https://rpc-openstack.atlassian.net/browse/ASC-665)


johnduarte:
recheck_xenial_swift_deploy

PR #3391, opened 2018-11-07
RI-560 Adjust to upstream changes

coreywright:
* RI-560 Remove installation of shade
* Track upstream change in embedded Ansible location

Issue: https://rpc-openstack.atlassian.net/browse/RI-560

PR #3397, opened 2018-11-09
[RI-538 and RI-561] ELK Fixes

d34dh0r53:
This makes the ELK HAproxy services conditional based on whether or not 
the proper inventory exists, if not those services are not enabled in 
HAProxy.

JIRA: RI-538

The elk-deployment.yml file was renamed to deployment-elk.yml but the 
references to elk-deployment.yml were not adjusted.  This rectifies 
that.

JIRA: RI-561

### DIFF
--- a/gating/check/run_elk_tests.sh
+++ b/gating/check/run_elk_tests.sh
@@ -32,5 +32,5 @@ source "$(readlink -f $(dirname ${0}))/../mnaio_vars.sh"
 
 ${MNAIO_SSH} <<EOS
   cd /opt/rpc-openstack
-  openstack-ansible playbooks/elk-deployment.yml
+  openstack-ansible playbooks/site-logging.yml
 EOS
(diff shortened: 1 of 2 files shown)


d34dh0r53:
This is broken until #3391 merges.

coreywright:
with #3391 merged, let's rekick failed checks:

gating/check/run_elk_tests.sh
```diff
@@ -32,5 +32,5 @@ source "$(readlink -f $(dirname ${0}))/../mnaio_vars.sh"
 
 ${MNAIO_SSH} <<EOS
   cd /opt/rpc-openstack
-  openstack-ansible playbooks/elk-deployment.yml
+  openstack-ansible playbooks/site-logging.yml
```
Frank-ZhangXin: renaming 'elk-deployment.yml', so shouldn't we use 'deployment-elk.yml' here?

d34dh0r53: `site-logging.yml` takes care of some housekeeping and then calls `deployment-elk.yml`

PR #3403, opened 2018-11-14
Update RPC-O Octavia for Stein

johnsom:
The haproxy template overrides have been upstreamed in Stein and
are no longer required in RPC-Openstack. This patch removes those
RPC-O customizations.


johnsom:

PR #3411, opened 2018-11-16
Update RPC-O Octavia for disabled members

johnsom:
Rocky Octavia changed how administratively disabled members are
handled in the amphora. They are now included, but marked in the
"disabled" state. This improves health reporting.
This patch makes sure those "disabled" members are rendered to
the haproxy configuration file.


prometheanfire:
also, did this hit master first?

johnsom:
Yes, it will include all members.
It is not on master as we removed this jinja template override in RPC master as it is no longer necessary.

prometheanfire:
sgtm

PR #3426, opened 2018-11-29
Remove legacy ELK env.d file and ensure new version is used [queens]

odyssey4me:
The legacy env.d file for elk is incomplete. It also puts the
elk containers into the log_hosts group, resulting in unused
ELK containers and haproxy back-ends being configured without
the associated services.

This patch does the following:

1. Removes the legacy ELK env.d file.
2. Ensures that the ELK deployment playbook copies over the
   new elk env.d file from the openstack-ansible-ops repo.
3. Removes the duplicated README information informing the
   deployer how to deploy the new ELK stack, replacing it
   with the use of the playbook to remove the element of
   human error. This was done because the instructions were
   missing the addition of the new host groups into the
   inventory and the copy of the new env.d file.

Also, when building an environment, there should be no situation
where the resulting build has a haproxy back-end that is down.
Recently some implementations were merged which did this,
causing the use of snapshot AIO images to fail due to the
failed back-ends in haproxy.

To prevent this happening again, we ensure that part of the
AIO PR/PM test is to do the same check and to fail if there
is a back-end down.


d34dh0r53:
:shipit:

odyssey4me:
@d34dh0r53 Love the enthusiasm, but the gate job won't work until all PR tests are complete and have passed. :)

d34dh0r53:
Ohh, I thought it queued :/ my bad

odyssey4me:
recheck_all

d34dh0r53:
:shipit:

PR #3429, opened 2018-12-04
Remove legacy ELK env.d file and ensure new version is used [master]

odyssey4me:
The legacy env.d file for elk is incomplete. It also puts the
elk containers into the log_hosts group, resulting in unused
ELK containers and haproxy back-ends being configured without
the associated services.

This patch does the following:

Removes the legacy ELK env.d file.
Ensures that the ELK deployment playbook copies over the
new elk env.d file from the openstack-ansible-ops repo.
Removes the duplicated README information informing the
deployer how to deploy the new ELK stack, replacing it
with the use of the playbook to remove the element of
human error. This was done because the instructions were
missing the addition of the new host groups into the
inventory and the copy of the new env.d file.
Also, when building an environment, there should be no situation
where the resulting build has a haproxy back-end that is down.
Recently some implementations were merged which did this,
causing the use of snapshot AIO images to fail due to the
failed back-ends in haproxy.

To prevent this happening again, we ensure that part of the
AIO PR/PM test is to do the same check and to fail if there
is a back-end down.


odyssey4me:
The failure here will go away once 
https://review.openstack.org/622473 merges.

mendeni:

mendeni:

d34dh0r53:

odyssey4me:
With thanks to https://github.com/rcbops/rpc-openstack/pull/3357/files#diff-4b2a015225363d3a6e1fa1663aad98ea the logging services are now being deployed by force for the AIO. I would argue that this is way over the top and that this should be opt-in, and not deployed on the AIO due to resource constraints. I've registered https://rpc-openstack.atlassian.net/browse/RI-632 to express that.

cjloader:
recheck_all

rackernexus:

odyssey4me:
recheck_whisperer

PR #3446, opened 2019-01-14
Replace grafana repo with new official one

odyssey4me:
As of 7 January 2019, packagecloud is no longer used
as a repo for grafana. This updates the URL and key
so that the apt artifact repo can complete the mirror
process using the new repo.

ref: https://grafana.com/blog/2019/01/05/moving-to-packages.grafana.com/

PR #3452, opened 2019-01-16
Remove bad Ceph repos and add new Grafana repo

cjloader:
The upstream ceph-* role repositories have been deleted, causing the ansible bootstrap to fail. Given that we use RPC-Ceph now we do not need these any more, so we remove them from the Ansible role requirements.

Also, as of 7 January 2019, packagecloud is no longer used as a repo for grafana. This adds a new mirror for it and stops updates to the old mirror. We also include the new mirror in the snapshots and remove the old one.

Previously, we were using the @evrardjp fork of the aptly repo to deploy and configure aptly. To ensure that we are in control of the role repo, we have forked it to rcbops and updated the configuration to use our fork.

ref: https://grafana.com/blog/2019/01/05/moving-to-packages.grafana.com/

Issue: [RI-641](https://rpc-openstack.atlassian.net/browse/RI-641)


odyssey4me:
I think that the release job failing is just because it's trying to pull the roles from master, so I think the failure can be ignored. The artifact build failure should be resolved by 
https://github.com/rcbops/rpc-openstack/pull/3446

cjloader:
@odyssey4me added a release note

odyssey4me:
Once https://github.com/rcbops/osa_differ/pull/23 is merged, we'll need to release osa-differ, publish it to pypi, then we can add a constraint bump here to finalize this PR.

evrardjp:
Does this mean I can remove my repo now, @cjloader?

prometheanfire:
I imagine there are still releases that depend on the repo.

odyssey4me:
@evrardjp Yes, you can now remove your fork of the aptly role.

evrardjp:
@odyssey4me that was my understanding too that it was only used in for artifacts and master branch, so nice to see a confirmation.

Good I will archive it, so that it is not deleted. I will just add a deprecation message into top of master branch, removing all files. We'll see how it goes quite quickly :p

odyssey4me:
@evrardjp Yeah, we have an exact copy of your repo ``rax`` branch here: https://github.com/rcbops/ansible-role-aptly/tree/rax

That said, we're just using the ``master`` branch instead now that the necessary parts of the changes you made are in the upstream role.

PR #3457, opened 2019-01-21
ASC-967 qtest publish on staging

zreichert:
- publishes to a specific project if action is system_staging

Issue: [ASC-967](https://rpc-openstack.atlassian.net/browse/ASC-967)


zreichert:
@johnduarte @ryan-rs Updated

PR #3460, opened 2019-01-22
Use upstream OSA checkout method for repos

antonym:
Utilizes method from OSA get-ansible-role-requirements.yml
to checkout repos instead of using ansible-galaxy.


antonym:

PR #3470, opened 2019-01-31
Update flavor and image setup for newer ansible and images

cloudnull:
This change adds newer images and sets the playbook up to only
deploy images for a given virt-type, which will speed up the
general gate times and test cycles as only the images using the
supported virt type will be enabled.

The deploy scripts have been updated to default to qemu (the
current default) when deploying an AIO. When the playbooks are
executed outside of the gate environment the default will be
KVM, which is the openstack default.

Signed-off-by: Kevin Carter <kevin@cloudnull.com>

### DIFF
--- a/playbooks/openstack-flavor-setup.yml
+++ b/playbooks/openstack-flavor-setup.yml
@@ -22,18 +22,22 @@
       include_role:
         name: "openstack_openrc"
 
-    - name: Create flavors of nova VMs
-      os_nova_flavor:
-        endpoint_type: internal
-        cloud: default
-        state: present
-        name: "{{ item.name }}"
-        ram: "{{ item.ram }}"
-        vcpus: "{{ item.vcpus }}"
-        disk: "{{ item.disk }}"
-        swap: "{{ item.swap }}"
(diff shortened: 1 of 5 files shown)


antonym:
```
2019-01-30T22:25:36-0600 Variable files: "-e @/etc/openstack_deploy/user_aio_variables.yml -e @/etc/openstack_deploy/user_rpco_secrets.yml -e @/etc/openstack_deploy/user_secrets.yml -e @/etc/openstack_deploy/user_variables.yml "
2019-01-30T22:25:38-0600 ERROR! 'args' is not a valid attribute for a Play
2019-01-30T22:25:38-0600 
2019-01-30T22:25:38-0600 The error appears to have been in '/var/lib/jenkins/workspace/PR_rpc-openstack-master-xenial_no_artifacts-swift-deploy/rpc-openstack/playbooks/site-openstack.yml': line 16, column 3, but may
2019-01-30T22:25:38-0600 be elsewhere in the file depending on the exact syntax problem.
```

prometheanfire:
recheck_all

I think master is unblocked now

PR #3493, opened 2019-02-25
ASC-978 Update "rpc-zigzag" Version

ryan-rs:
Update the script to pin to the 1.0 series for "rpc-zigzag". Also, add a
variable for the "rpc-zigzag" PyPI package at the top so it can be easily
updated in the future without hunting around in the script.


johnduarte:
@ryan-rs can you articulate why this PR should not be merged?

ryan-rs:

PR #3504, opened 2019-02-27
Enable snapshot image creation for PUSH and TIMER [master]

odyssey4me:
To ensure that the snapshot image creation happens more regularly,
we ensure that the preparation for it happens for both a PUSH and
a TIMER trigger.

This will ensure that the images are updated on a daily basis,
not just when a new patch merges.

Issue: [RE-2278](https://rpc-openstack.atlassian.net/browse/RE-2278)

### DIFF
--- a/gating/check/post_deploy.sh
+++ b/gating/check/post_deploy.sh
@@ -66,9 +66,10 @@ else
 fi
 echo "#### END DSTAT CHART GENERATION ###"
 
-# Only enable snapshot when triggered by a commit push.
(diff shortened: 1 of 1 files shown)


odyssey4me:
recheck_all

odyssey4me:
Until https://github.com/rcbops/rpc-openstack/pull/3509 merges, the syntax/lint tests will fail.

odyssey4me:
recheck_container_ansible-syntax_tox-test

PR #3510, opened 2019-02-27
RI-668: Update thaw scripts to rebuild container networking

leader716:
Issue: [RI-668](https://rpc-openstack.atlassian.net/browse/RI-668)

### DIFF
--- a/gating/thaw/run
+++ b/gating/thaw/run
@@ -33,9 +33,13 @@ cd /opt/rpc-openstack/openstack-ansible/playbooks/
 # IP fact being up to date.
 export ANSIBLE_GATHERING=implicit
 openstack-ansible -v /opt/rpc-openstack/gating/thaw/thaw.yml
+
 openstack-ansible -t haproxy_server-config haproxy-install.yml
 
 lxc-autostart --all
+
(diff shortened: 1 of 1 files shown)


gating/thaw/run
```diff
@@ -33,6 +33,9 @@ cd /opt/rpc-openstack/openstack-ansible/playbooks/
 # IP fact being up to date.
 export ANSIBLE_GATHERING=implicit
 openstack-ansible -v /opt/rpc-openstack/gating/thaw/thaw.yml
+# This should rebuild the networking for the containers and ensure that they are all up
+openstack-ansible -t lxc_container_create-networks lxc-containers-create.yml
```
odyssey4me: I honestly have no idea whether this is the best thing to do, but I do think that this might be better suited to happening after ``lxc-autostart --all`` below, because that's the thing that ensures all the containers are actually running, I think. I will defer to @d34dh0r53 or @cloudnull for this - they know the underlying LXC/networking bits better than I do.

leader716: That makes sense, let me move it and resubmit the PR.

odyssey4me:
recheck_trusty_swift_deploy

PR #3512, opened 2019-02-28
Ri 669: Update snapshot thaw scripts to work with rocky

leader716:


gating/thaw/run
```diff
@@ -34,7 +34,7 @@ export ANSIBLE_STDOUT_CALLBACK=debug
 # IP fact being up to date.
 export ANSIBLE_GATHERING=implicit
 openstack-ansible -v /opt/rpc-openstack/gating/thaw/thaw.yml
-openstack-ansible -t haproxy_server-config haproxy-install.yml
+openstack-ansible -t haproxy-server haproxy-install.yml
```
odyssey4me: Perhaps it's better just to remove the tag and let the whole role run? That way if tags are changed again, it doesn't matter.

odyssey4me:
recheck_all

PR #3519, opened 2019-03-19
Updating rcbops/openstack-ansible submodule sha

shannonmitchell:
This is to include the linux-image-extra name change
from https://github.com/rcbops/openstack-ansible/pull/2
in the rpc-openstack's openstack-ansible submodule.


shannonmitchell:
Canceling.  Looks like these tests use an old version, so we might need to fine tune.

PR #3525, opened 2020-03-05
Fixing newton openssh issue breaking repo jobs

shannonmitchell:
The artifact builds in the late stages are failing during
an aio build with rpc-openstack/newton on xenial.  This is
due to the lxc cache prep failing to install the
openssh-server package due to missing random/urandom devices.


shannonmitchell:
Although more issues need taken care of involving trusty, this takes care of the artifact job cache/openssh issues when tested.

